PR #114, opened 2018-10-12
[WIP] Queuing - this time for real

tcbyrd:
After some discussion with @bkeepers and @gimenete we decided to go the route of dynamically creating queues for each installation ID. To accomplish this, @gimenete gave me a quick idea for this `getForInstallation` method:

```js
getForInstallation(installationId) {
  const name = `installation-${installationId}`
  if (queues[name]) return queues[name]
  const queue = new Queue(name, redisConnection)
  queue.on('complete', () => {
    queue.close()
    delete queues[name]
  })
  queues[name] = queue
  return queue
}
```
I ran with that idea and modified the `worker.start()` method to spin up a small message bus that receives the job from `Subscription.syncRepository` and creates the queue from there.  A huge benefit here is we can actually use Bull's rate limiting features to guarantee each installation is only processing one job at a time on an interval we specify (currently set to 5000ms, but we can probably reuse the `LIMITER_PER_INSTALLATION` variable we setup before)

TODOs:
- [ ] Setup a properly authenticated route from the web page to initiate the sync job
- [ ] Update the UI to allow for syncing a specific repository (see [current DVCS UI](https://marketplace-cdn.atlassian.com/files/images/com.atlassian.jira.plugins.jira-bitbucket-connector-plugin/89b25236-c47e-4006-89c2-2800e5fab4d5.png) for inspiration)
- [ ] Figure out how to handle the existing queues (maybe a separate worker script that processes what's left?)
- [ ] Reimplement #112 to use these dynamic queues
- [ ] Write tests
- [ ] Anything else?

### DIFF
--- a/lib/frontend/app.js
+++ b/lib/frontend/app.js
@@ -80,7 +80,7 @@ module.exports = (appTokenGenerator) => {
 
   app.get('/jira/configuration', verifyJiraMiddleware, getJiraConfiguration)
   app.delete('/jira/configuration', verifyJiraMiddleware, deleteJiraConfiguration)
-  app.get('/jira/sync', retrySync)
+  app.get('/jira/sync/:owner/:repo', retrySync)
 
   app.get('/', async (req, res, next) => {
     const { data: info } = (await res.locals.client.apps.get({}))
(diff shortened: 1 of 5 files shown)


tcbyrd:
Closing in favor of #118

PR #118, opened 2018-10-15
Short lived jobs

gimenete:
This PR solves some problems:

- Improves queueing by doing only one page of results at a time and then putting a new job at the end of the queue if there are more pages. Closes https://github.com/github/ecosystem-integrations/issues/194
- Sends the `migrate` event correctly to event, not assuming that the commits queue is the last one and not assuming when one repo is done, everything is done.
- Parses PR bodies. Closes https://github.com/integrations/jira/issues/120 and https://github.com/integrations/jira/issues/119
- Adds more logging. See https://github.com/github/ecosystem-integrations/issues/193

### DIFF
--- a/lib/frontend/retry-sync.js
+++ b/lib/frontend/retry-sync.js
@@ -1,13 +1,24 @@
-const { Subscription } = require('../models')
+const { Installation, Subscription } = require('../models')
+const jwt = require('atlassian-jwt')
 
 module.exports = async (req, res, next) => {
-  if (!req.session.jiraHost) {
-    return next(new Error('Unauthorized'))
+  const { jiraHost, installationId, token } = req.body
+
+  const installation = await Installation.getForHost(jiraHost)
+
+  if (!installation) {
+    next(new Error('Not Found'))
+  } else {
+    try {
+      jwt.decode(token, installation.sharedSecret)
+
+      const subscription = await Subscription.getSingleInstallation(jiraHost, installationId)
+
+      await Subscription.findOrStartSync(subscription)
+
+      return res.sendStatus(202)
+    } catch (error) {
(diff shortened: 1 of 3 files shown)


tcbyrd:
@gimenete Take a look at the changes I added in 3ab5c87. We don't really need to retry a bunch of times, so instead I just threw a `try/catch` around the function that gets each content type. If it fails to get `per_page: 100`, it falls back to trying `per_page: 50` for each job. Since we're doing one page per job, this feels like a pretty simple solution. If it fails to get `per_page: 50`, it's likely there's some other error that can't be resolved by retrying anyway.

I'm getting some inconsistent results, though. Try syncing the `integrations` org against your test instance. About 2/3 of the time when I run the sync, it finishes all the pages but never goes from "active" to "complete" because a couple of the repos still show pending. The other 1/3 of the time it completes the sync with no issues, so I think something is off about how we're updating `repoSyncState`. I've looked at it a few ways but couldn't put my finger on what's causing it.

tcbyrd:
To follow up, 5720e40 fixes the issue with some jobs never signaling they were complete. We were running into race conditions where the `subscription` object we were passing around between jobs had stale data, so the jobs would complete, but the `repoSyncState` object would be in an older state with jobs in pending status. The update ensures we get the most current state from the database before pushing another update.

I also noticed a couple of situations where we weren't sending `avatarUrl` everywhere, so a fresh sync wouldn't display an avatar for a commit that was previously pushed.

tcbyrd:
@gimenete I added a few things to how the jobs are getting re-triggered. Mainly just verifying it's coming from an admin of the Jira instance by vaildating the JWT and adding csrf protection like our other POST routes. I verified everything looks good in staging. Feel free to deploy this prod.

PR #127, opened 2018-10-17
Open public url from within iframe

tcbyrd:
This allows the iframe to remain a heroku URL while ensuring that when we send users to a GitHub specific page, it will show the official GitHub URL.


tcbyrd:
:shipit: 

Live now on `jira.github.com`.

PR #131, opened 2018-10-22
Single queue

gimenete:
Refactor queues to have one single queue.

The single queue checks the status of the sync by getting the `repoSyncState` from the db. So now the `repository` object is not stored in the job. So in order to get that info I've decided to put the repo info also in the `repoSyncState` in order to avoid additional requests in each job execution. That's the only change in the format of the job data and `repoSyncState` data.

Other than that it just checks what's pending and processes it. Having only one queue has this benefits.

- Only one job at a time makes requests to a given installation.
- Which means we can update the concurrency to whatever level we want without having to worry about rate limits
- An installation with lots of repos doesn't fill the queues.
- No race conditions updating the `repoSyncState`
- We can delay a job if we reach the rate limit of an installation by exactly the time we need after doing the next request

TODO:

- [x] Error handling for rate limiting and graphql timeouts
- [x] Testing migration of old jobs to new ones. Specially the `repository` object inside `repoSyncState`
- [x] Make each processor return edges and jiraPayload and not pass a jiraClient to them.
- [x] Make the `repository` object stored in the db as small as possible


lib/sync/discovery.js
```diff
@@ -16,21 +21,16 @@ module.exports.discovery = (app, queues) => {
       return
     }
 
-    // Store an empty object per repository to indicate that the sync for that repo is pending
+    // Store the repository object to prevent doing an additional query in each job
+    // Also, with an object per repository we can calculate which repos are synched or not
     const repos = repositories.reduce((obj, repo) => {
-      obj[repo.id] = {}
+      obj[repo.node_id] = { repository: getRepositorySummary(repo) }
```
tcbyrd: Since we're storing the database ID as part of the summary, and because we can use the global `node_id` to look up a repository in GitHub, I changed this.

lib/sync/installation.js
```diff
@@ -34,8 +35,11 @@ module.exports.processInstallation = (app, queues) => {
     if (edges.length > 0) {
       // there's more data to get
       subscription.set(`repoSyncState.repos.${repositoryId}.${getCursorKey(task)}`, edges[edges.length - 1].cursor)
+      await subscription.save()
```
tcbyrd: I was noticing it processing jobs for the same cursor twice without this.

lib/sync/installation.js
```diff
@@ -34,8 +35,11 @@ module.exports.processInstallation = (app, queues) => {
...
+
       const { removeOnComplete, removeOnFail } = job.opts
-      queues.installation.add(job.data, { removeOnComplete, removeOnFail })
+      const delay = Number(process.env.LIMITER_PER_INSTALLATION) || 1000
```
tcbyrd: Thought we'd reuse the `LIMITER_PER_INSTALLATION` here since it helps us not hit abuse limits as quickly. Basically all jobs will be added to the queue with this default delay, and we get the added benefit of it continuing to work the same if we add more workers. @gimenete any thoughts on this approach?

lib/sync/installation.js
```diff
@@ -49,12 +53,28 @@ module.exports.processInstallation = (app, queues) => {
...
+
+    // Record elapsed time for each GraphQL request
+    github.hook.before('request', () => {
+      github.timeStart = Date.now()
+    })
+
+    github.hook.after('request', () => {
+      const elapsed = Date.now() - github.timeStart
+      app.log.debug(`GitHub Request time for ${installationId}: (${elapsed}ms)`)
+    })
```
tcbyrd: Just a little enhancement so we can see how long each GraphQL request takes.

lib/sync/discovery.js
gimenete: Hmm the problem is that old jobs (the ones that are migrated) need that id to get the repository summary https://github.com/integrations/jira/pull/131/files#diff-9311ea9c09f3ca1ea3c71393232b21feR97

lib/sync/installation.js
gimenete: I don't think hitting the rate limiting quickly is bad per se. If a sync requires 6000 API requests, I prefer to do 5000 API requests as quickly as possible, hit the rate limit, wait for the next reset and make 1000 API requests as fast as possible.

By delaying always jobs we guarantee that we don't hit the rate limit but we end up doing a lot less processing that we could do, I think.

Adding new workers should not be a problem since we now only have one job per installation. So either that job is being processed or not, but it's one job maximum per installation at the same time.

tcbyrd: The problem I'm trying to avoid with this is abuse rate limits, which have no direct correlation to the 5000 API requests per hour. In my testing against https://github.com/fusion-qa/, it was consistently triggering abuse detection within 2-3 minutes of the start of the job, and then we have to back off for at least 60 seconds when that is triggered. This allows us to smooth that out and not wait ~60 seconds every few minutes because we're hitting abuse limits. I've yet to ever see us come near the 5000/hr limit, but it's much easier to hit the abuse limit with this app due to the cost of he GraphQL queries.

gimenete: Sounds good 👍

lib/sync/discovery.js
tcbyrd: Good point. I'll change it back. We can always calculate the `node_id` if we need it anyway.

tcbyrd:
:shipit: !!

Looking good. There's not a lot of initial sync activity at the moment, but I tested a few new installs on prod and they went through cleanly.

Thanks for all the help here @gimenete!

gimenete:
@mbd-dbc-dk yes. This is in production now. It would be great if you could test to sync now with the new code. Thanks!

mbd-dbc-dk:
@gimenete Sounds good. I have just configured the Jira system at github to include all repos, and asked the github plugin at jira to "retry". It has changed from Complete to Pending, now active.

This was started approximately 17:05 UTC+1. jira end: dbcjira.atlassian.net, github end, DBCDK (organisation).

I will look at it again tomorrow morning. Don't hesitate to ping me, if I forget.

gimenete:
@mbd-dbc-dk can you check the status of the sync? Thanks!

mbd-dbc-dk:
Yes, sorry, I checked it earlier (about 12 UTC+1), and it was in COMPLETE, and it works as expected. Forgot to write here (been a busy day).

We were unable to go from ACTIVE to COMPLETE before, so this is major improvement for us, and AFAICT, everything now works as expected and intended.

Thanks for the great work!

Kind regards

Mads

gimenete:
That's great to hear @mbd-dbc-dk 

Thanks!

jgeorge-gc:
How long should it take to sync around 300 repos? Is the sync time a function of repos, commits, or jira tickets?

I was directed to this PR by github support because sync is taking over 5 days.

mbd-dbc-dk:
We did 369 in less than 17 days. Abort your current. Set to sync a single repo. When it is done change to all. Sync.

Good luck.

Mads

17. dec. 2018 22.08 skrev Joseph George <notifications@github.com>:

How long should it take to sync around 300 repos? Is the sync time a function of repos, commits, or jira tickets?

I was directed to this PR by github support because sync is taking over 5 days.

PR #132, opened 2018-10-23
Check for previous installations from jira site

Dahaden:
On subsequent install requests from Jira, the request will be signed by the previously agreed upon shared secret.

### DIFF
--- a/lib/jira/install.js
+++ b/lib/jira/install.js
@@ -1,28 +1,43 @@
 const { Installation } = require('../models')
 const getAxiosInstance = require('../jira/client/axios')
+const jwt = require('./util/jwt')
 
 module.exports = async (req, res) => {
   req.log('Received installation payload.')
   const { baseUrl: host, clientKey, sharedSecret } = req.body
-  res.sendStatus(204)
+  let successfulRes = false;
+
+  try {
+    const hostData = await Installation.getForClientKey(clientKey)
+    if (jwt.hasValidJwt(hostData.sharedSecret, host, req, res)) {
+      res.sendStatus(204)
+      successfulRes = true
+    }
+  } catch (err) {
+    // Client key has not been seen before. Safe to install.
+    res.sendStatus(204)
+    successfulRes = true
+  }
   res.on('finish', async () => {
-    const instance = getAxiosInstance(null, {}, host, sharedSecret)
-    try {
-      // Validate the clientKey and the sharedSecret
-      // We have to do this after responding with 204 because
-      // otherwise we get an error "The add-on is not enabled"
-      // no matter what we send
-      const result = await instance.get(`/rest/devinfo/0.10/existsByProperties?fakeProperty=1`)
-      if (result.status === 200) {
-        req.log('App installed on Jira. Adding secrets.')
-        await Installation.install({ host, clientKey, sharedSecret })
-      } else {
-        req.log('App not installed. Verification failed.')
-      }
-    } catch (err) {
-      if (err.response.status === 401) {
-        req.log('Jira does not recognize this installation. Deleting it.')
-        await Installation.uninstall({ clientKey })
+    if (successfulRes) {
+      const instance = getAxiosInstance(null, {}, host, sharedSecret)
+      try {
+        // Validate the clientKey and the sharedSecret
+        // We have to do this after responding with 204 because
+        // otherwise we get an error "The add-on is not enabled"
+        // no matter what we send
+        const result = await instance.get(`/rest/devinfo/0.10/existsByProperties?fakeProperty=1`)
+        if (result.status === 200) {
+          req.log('App installed on Jira. Adding secrets.')
+          await Installation.install({ host, clientKey, sharedSecret })
+        } else {
+          req.log('App not installed. Verification failed.')
+        }
+      } catch (err) {
+        if (err.response.status === 401) {
+          req.log('Jira does not recognize this installation. Deleting it.')
+          await Installation.uninstall({ clientKey })
+        }
       }
     }
   })


lib/jira/install.js
```diff
@@ -1,28 +1,43 @@
...
-        req.log('App installed on Jira. Adding secrets.')
-        await Installation.install({ host, clientKey, sharedSecret })
-      } else {
-        req.log('App not installed. Verification failed.')
-      }
-    } catch (err) {
-      if (err.response.status === 401) {
-        req.log('Jira does not recognize this installation. Deleting it.')
-        await Installation.uninstall({ clientKey })
+    if (successfulRes) {
```
Dahaden: I believe I could also use `res.statusCode === 204`, but I couldnt start this up to try it out :/

lib/jira/install.js
```diff
@@ -1,28 +1,43 @@
 const { Installation } = require('../models')
 const getAxiosInstance = require('../jira/client/axios')
+const jwt = require('./util/jwt')
 
 module.exports = async (req, res) => {
   req.log('Received installation payload.')
   const { baseUrl: host, clientKey, sharedSecret } = req.body
-  res.sendStatus(204)
+  let successfulRes = false;
```
gimenete: Can we rename it to `requiresNewInstallation` or similar?

lib/jira/install.js
```diff
@@ -1,28 +1,43 @@
...
+
+  try {
+    const hostData = await Installation.getForClientKey(clientKey)
+    if (jwt.hasValidJwt(hostData.sharedSecret, host, req, res)) {
+      res.sendStatus(204)
+      successfulRes = true
```
gimenete: This should be `false`, right? If the installation already exists and is valid, we don't want to store a new one. We could even `return` here.

lib/jira/install.js
gimenete: In order to simplify the code indentation and the diff, can you do `if (!successfulRes) return`?

lib/jira/install.js
Dahaden: Unfortunately some information can change in the installation flow.
If the app is uninstalled from a site, then reinstalled, the JWT will be signed with the old sharedSecret but a new sharedSecret will be supplied.

Also there is a chance that the baseUrl can change (site rename) and a new installation is sent to update the app

lib/jira/install.js
Dahaden: Since we also need to update if its not the first installation, maybe I should change this to `requiresInstallionUpdate`?

lib/jira/install.js
Dahaden: Done :)

Dahaden:
@gimenete Thanks heaps for the prompt review :)

Dahaden:
I pushed my lint fixes into my last commit but I may have broken Travis :( No idea how to restart it

gimenete:
Hi again @Dahaden 

Sorry for the delayed response. I'm trying to understand where the current implementation fails in the edge case this PR is solving, and see if we can simplify the solution. In the current flow we check against `existsByProperties` that the `sharedSecret` is ok, which is enough to check that the request is legitimate, and only in that case we do `Installation.install` that internally does a `findOrCreate`. Inside that method we have:

```js
    if (!created) {
      await installation.update({
        sharedSecret: payload.sharedSecret,
        enabled: true
      })
    }
```

Maybe we just need to add the `clientKey` in that `update`?

Dahaden:
Hey sorry for the delay.

The client key will never update, but the sharedSecret and baseUrl / host url can.

So currently, it looks to me that someone could spoof a Jira site and update the sharedSecret.

1. Legitimate Jira sends install payload to Github-app,
1. Github-app accepts and saves data,
1. Spoof Jira sends new payload with same clientKey as legitimate Jira,
    - host doesnt seem to follow any whitelist
1. Github-app accepts and overwrites sharedSecret
    - Verification of host and sharedSecret only uses the incoming host and sharedSecret even though these cannot be trusted as valid sources

What I am trying to achieve in this PR, is to use the JWT in subsequent installations (where we can find a matching clientKey in the db) to prevent malicious apps from potentially gaining access data to Github via this app :)

I hope that this helps!

gimenete:
@Dahaden thanks for your explanation.

What I'm trying to find is the minimal changes to do what you are suggesting. As far as I understand we could simply change the `lib/models/installation.js` file like it is now in this PR:

```js
    if (!created) {
      await installation.update({
        sharedSecret: payload.sharedSecret,
        jiraHost: payload.host, // this is new
        enabled: true
      })
    }
```

With only that change we'll have:

1. Legitimate Jira sends install payload to Github-app,
1. Github-app accepts and saves data,
1. Spoof Jira sends new payload with same clientKey as legitimate Jira,
    - new shared secret is used to call the JIRA endpoint
    - the JIRA endpoint does not return successfully
    - we haven't updated anything in the db

If the call is legit we'll have:

1. Legitimate Jira sends install payload to Github-app,
1. Github-app accepts and saves data,
1. Legitimate Jira sends new payload with same clientKey
    - new shared secret is used to call the JIRA endpoint
    - the JIRA endpoint returns successfully
1. Github-app accepts and overwrites sharedSecret

So I think that could be the only change to make to the code, right?

Dahaden:
In theory we could do this but:
1. Currently in the install request, we are using the jiraHost that is provided by the incoming request, and
2. If the Jira site has legitimately changed its URL, then we will need to use the jiraHost from the incoming request to ensure that the host and the sharedSecret work together

rachellerathbone:
Hey @Dahaden. I'm from a team at Atlassian that is currently in the process of migrating the ownership over to us from GitHub. As a part of this, we're currently going through the backlog of issues and prs to see what can still be addressed and what should be closed.

As this PR is now 2.5 years old so we will likely close this unless you're interested in updating your changes and resolving conflicts. Please let me know if this is something you're interested in doing. Otherwise, I'll close this out at the end of the week.

Dahaden:
Hey @rachellerathbone I noticed that https://github.com/integrations/jira/blob/master/lib/jira/install.js#L17 still just returns 204 without checking secrets.
It literally just replaces the `sharedSecret` provided in the request without checking anything.
I tried explaining this 2.5 years ago and am happy to try again.

Else you can reach out to the Atlassian Ecosystem team who might be able to chat to you about this

rachellerathbone:
Hey @Dahaden. Thanks for your prompt response. 

We're happy to review your PR and get it merged in if it resolves this issue but would first require you to update your changes as one of the files you edited, `lib/jira/install.js`, no longer exists.

KAbakumovAtl:
Hi @Dahaden, thank you for contributing. We recently revisited authentication of our Connect App events endpoints. 

Recently Atlassian Connect changed the way install and uninstall callbacks are authenticated. (More details here https://developer.atlassian.com/cloud/jira/platform/understanding-jwt-for-connect-apps/#verifying-a-asymmetric-jwt-token-for-install-callbacks)  

**TLDR**: These 2 events now use asymmetrically signed JWT tokens. Jira don't sign them with a shared secret anymore, but it signs it with its own private key. When receiving such token we download public key from Atlassian CDN to verify the signature and then check other token parameters. You can find the implementation here: https://github.com/atlassian/github-for-jira/blob/97213b4a4c9486bd2f912f46e5fbbd8d79b1146b/src/jira/util/jwt.ts#L200

 I am going to close this PR hence these changes are not needed anymore

PR #133, opened 2018-10-24
Tracking unique project keys

tcbyrd:
This allows us to track unique project identifiers so we know how many projects are being used, and allows for cross referencing any identifiers the integration has seen that may not exist on Jira (e.g. we might see `UTF-8`, but that's likely not a valid Jira project).

/cc @annafil


annafil:
🙌yay!!

PR #145, opened 2018-11-15
Handle pushes in background job

tcbyrd:
The app currently does not close the connection from GitHub when push payloads come in that have more than 10 commits in them. Since GitHub webhooks close after 10 seconds and we have to make an API call for each commit to get the files data, this is causing some issues like memory leaks and errors even when data is being processed and sent to Jira correctly.

This PR helps with this by putting all push events in a queue. Since even a small payload of 3-4 commits can take a bit of time to process, I thought it would be better if we just always use the worker since it's available to us. Additionally, to help limit the amount of commits we need to process, I chose to parse the smart commits before adding the payload to the queue. This way if there are 100 commits, but only 1 of them has an issue key, that payload doesn't have to make 100 calls to the GitHub API.

Left to do:
- [x] Restructure tests for jira payload processing. Currently this is only testing that the github payload was added to the queue

Any thoughts on this implementation? I tried to avoid having to use the queue, but basically the goal is that we respond to the webhook payload as quickly as possible so it closes the connection properly and continues processing in the background.


tcbyrd:
Ok tests have been refactored and it's feeling pretty good. Will deploy to staging and do some real world testing on this tomorrow.

lib/github/push.js
```diff
@@ -1,19 +1,25 @@
...
-  await transformPush(context.payload, context.github, async ({ data: jiraPayload, commands }) => {
-    await jiraClient.devinfo.repository.update(jiraPayload)
-
-    // Don't run Jira commands
-    // await util.runJiraCommands(commands)
+module.exports = async (context, jiraClient) => {
+  // This keeps the context object the same shape, but
+  // filters out any commits that don't have issue keys
+  // so we don't have to process them.
+  context.payload.commits = context.payload.commits.map(commit => {
```
gimenete: In general I would prefer to create a new variable and not mutate the context to avoid any side effects if the context is used somewhere else.

```
const newContext = {
  ...context,
  payload: {
    ...context.payload,
    commits: context.payload.commits.map(...)
  }
}
```

tcbyrd:
@gimenete Yeh I was being a lazy programmer, good callout! 😆Since it was only sending the payload to the queue anyway, instead of copying the whole context object (which has that enormous `context.github` object), I just created a smaller `payload` object that only contains the data we need to process.

tcbyrd:
Looking good in the before/after:
image

All metrics point to this working just as we'd hoped. Pushes are still handled quickly if they only have a few commits, but I've already seen a few pushes come through with dozens of commits and the webhook was responded to quick enough that it's not hitting any network timeouts (the flat lines shown above). Will have a better picture tomorrow if it fully addressed the memory leak.

Issue #159, opened 2018-12-14
Jira bot duplicating ticket links


returnlytom:
Similar to #141, the Jira bot is duplicating links to tickets. Please see this gif for a reproduction:

![link_inject](https://user-images.githubusercontent.com/22224829/50033902-d51d8f00-003d-11e9-9095-c4c27207f568.gif)

tcbyrd:
This is an interesting issue. I think the best way to solve for it is we just don't unfurl to the same syntax we use within the unfurled link.

What would you think if we instead unfurled to something like this: [jira/TES-123](https://example.com/browse/TES-123). This follows the way github unfurls links (e.g. https://github.com/integrations/slack/issues/1). Done this way it's also visually more clear that the link will take you to Jira.

tcbyrd:
Check out #161 for a proposed solution.

returnlytom:
@tcbyrd Love it! Nice work, thanks for the prompt turnaround :)

returnlytom:
Closed by #165

PR #160, opened 2018-12-17
Fix: ignore Markdown links to Jira tickets in smart-commit regex [Issue #159]

tomholford:
This is to address #159. Note that this is not a complete PR yet, just a failing test for future TDD :)

@tcbyrd What would be the best approach to resolve this? One naive approach I can think of would be adding a negative lookahead or lookbehind for the `atlassian.net` URL to the `issueKeysRegex` in `smart-commit.js`. But this seems brittle and hacky :) Curious to hear your thoughts!

### DIFF
--- a/test/unit/smart-commit.test.js
+++ b/test/unit/smart-commit.test.js
@@ -253,6 +253,16 @@ describe('Smart commit parsing', () => {
         issueKeys: ['JRA-090']
       })
     })
+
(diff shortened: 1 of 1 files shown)


tcbyrd:
Deployed a solution for this in #165 

Docs here: https://github.com/integrations/jira#link-issues-in-github

PR #163, opened 2019-01-07
Updates to address sync getting stuck due to invalid issue keys

tcbyrd:
The regex we were using to parse Issue Keys is getting false positives against commit messages and PR titles that have an underscore in them.

**Example:**
`JSP_JIR-123` currently matches, but `JSP_JIR` is not a valid identifier according to the devinfo API. I looked through the logs from sync jobs that were getting this error and none of them looked like valid Jira issue keys.

To verify, I tried creating an issue key with underscores, but Jira will not allow this:

This PR prevents the integration from matching on those keys. In addition, it will no longer throw when it encounters a bad request. Instead, it will just log an error and continue on to the next page of results. This will allow the sync to continue while allowing for further diagnostics.

### DIFF
--- a/lib/sync/installation.js
+++ b/lib/sync/installation.js
@@ -128,9 +128,8 @@ module.exports.processInstallation = (app, queues) => {
         } catch (err) {
           if (err.response && err.response.status === 400) {
             const errorMessages = err.response && err.response.data && err.response.data.errorMessages
-            app.log(`JIRA Returned ${err.response.status}. Error messages: ${JSON.stringify(errorMessages)}. Payload sent=${jiraPayload}`)
+            app.log.error(`JIRA Returned ${err.response.status}. Error messages: ${JSON.stringify(errorMessages)}. Payload sent=${jiraPayload}`)
           }
-          throw err
         }
       }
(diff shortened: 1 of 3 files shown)


tcbyrd:
Deployed. Ran a sync on an example repo with a bad commit before the deploy to verify failure state, then ran the sync again after deploying and the sync completed successfully.

fooware:
@tcbyrd  Our JIRA integration has finally started working today (hasn't been working since we switched a month ago). We have taskbranches with names like "taskbranch/ABC-123_2" after rebasing, could that underscore have triggered this problem?

tcbyrd:
@fooware That syntax is ok. What was causing issues was underscores before the dash (so for example if you instead had `2_ABC-123`, it would try to find a Jira project with key `2_ABC`). Your installation was failing because of a deleted user that authored an old PR (fixed by #164). Since this a public repo I won't go into more details, but if you'd like to know specifically which PR or what user it was related to, let me know and I'll contact you privately.

fooware:
Thank you @tcbyrd for the details. Very happy to finally have the integration working again and can leave it at that.

PR #164, opened 2019-01-08
Handle deleted users that were authors of a PR

tcbyrd:
In the REST API, we return the [Ghost User](https://github.com/ghost) when a user that's an author of a PR deletes their account. In GraphQL we handle this by returning `null`, which is causing syncs of Pull Requests to fail since we expect properties to be returned on the `avatar` object. The Devinfo API [requires at least an author name](https://github.com/integrations/jira/blob/053cc76af05ea045d964b7fd1634473534aa9bb1/docs/jira-dev-info-0.10-swagger.yaml#L395-L398) to be sent over with the PR object, so this will just return the ghost user info when GraphQL returns `null` for author.

### DIFF
--- a/lib/sync/transforms/pull-request.js
+++ b/lib/sync/transforms/pull-request.js
@@ -15,7 +15,7 @@ function mapStatus ({ state, merged }) {
 }
 
 function getAuthor (author) {
-  //if author is null, return the ghost user
+  // If author is null, return the ghost user
   if (!author) {
     return {
       avatar: 'https://github.com/ghost.png',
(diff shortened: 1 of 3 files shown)


tcbyrd:
Tested on staging. Ghost user is properly showing up both in an initial sync and if the user has been deleted when we get a `pull_request.edited` event. The latter did not require any code changes since the webhook event sends the ghost user object in place of the original user.

image

tcbyrd:
🚢 Working on prod.

PR #166, opened 2019-01-09
[WIP] Adding datadog live process monitoring

tcbyrd:
We're currently only sending Heroku host level information to Datadog. This updates the datadog agent config to use [Live Process monitoring](https://docs.datadoghq.com/graphing/infrastructure/process/?tab=linuxwindows) as well. I'm not clear if this feature is enabled on our DD infrastructure, so going to deploy this to staging first to see if it shows up.

### DIFF
--- a/datadog/conf.d/datadog.yaml
+++ b/datadog/conf.d/datadog.yaml
@@ -1,0 +1,2 @@
+process_config:
+  enabled: "true"


tcbyrd:
Not seeing this in Datadog yet on staging. Live process is an Enterprise feature, so need to check if it needs to be enabled.

PR #167, opened 2019-01-15
Staff API routes

tcbyrd:
Since the main config page is only available on a Jira instance, we currently have limited ability to see what's going on with a user's installation. This adds some API routes to allow us to complete or undo a migration, as well as see the current state of an in progress migration for better troubleshooting.

For security purposes, these routes can only be accessed by a token that's been authorized to access the GitHub Org, which means the token has to be authorized through GitHub's 2FA protected IdP. Additionally, the token has to belong to someone with `write` access to our internal repo. This protects against unauthorized employees and ensures that if an employee leaves GitHub, their token will no longer work.


lib/api/index.js
```diff
@@ -0,0 +1,202 @@
...
+  const { installationId } = req.params
+  const { jiraHost } = req.body
+
+  const installation = await Installation.getForHost(jiraHost)
+
+  if (!installation) {
+    res.sendStatus(404)
+  } else {
+    try {
+      const subscription = await Subscription.getSingleInstallation(jiraHost, installationId)
```
gimenete: Could we get the installationId from the installation object and not from the query params?

gimenete:
I've pushed a commit adding tests (we can still improve them a little bit) and I've left a comment.

lib/api/index.js
tcbyrd: Hah, I had to double check this because of the poor naming choices that have plagued us.

So the installation object doesn't have the github installation ID. In this case `Installation` is the installation of the integration in Jira, so all we have available is the `jiraHost` and `clientKey`. One `jiraHost` can have many `Subscriptions` to various GitHub Installation IDs (different orgs or users). So the parameter here is a way to say "I want the status of the sync from GitHub Installation 12345 on Jira Host tcbyrd.atlassian.net", because you could also have the same GitHub Installation ID on multiple instances.

So TL;DR, we need a way to provide the github installation ID as a value, and the path param makes it feel "RESTful" to me. Thoughts?

gimenete: Ok. Makes sense.

But then the `const installation = await Installation.getForHost(jiraHost)` is not needed in this endpoint, right? Initially I thought it was to get the installationId, but I see that it's a different type of installation (jira vs github installation).

tcbyrd: Ah, yeh. Good point. I pulled this code over from another route that checks for the Installation and sends a `404` if there isn't one, but it would make sense here to send that if no Subscription is found for the installation ID. I'll update it!

PR #171, opened 2019-02-12
Delete old queues

gimenete:
The jobs in these queues were migrated and processed in the `installation` queue and no new jobs are stored in there. So we don't need to keep them in the code.


gimenete:
@tcbyrd Cool. Thanks for checking that!

PR #173, opened 2019-02-13
Find issue keys in branch names when opening PRs

gimenete:
Have into account also the branch name since GitHub sometimes humanizes the PR title based on the branch name by default and the issue key is lost that way. So, using the branch name should keep that issue key reference.

cc https://github.com/github/ecosystem-primitives/issues/145

### DIFF
--- a/lib/transforms/pull-request.js
+++ b/lib/transforms/pull-request.js
@@ -18,7 +18,7 @@ function mapStatus ({ state, merged }) {
 module.exports = (payload, author) => {
   // eslint-disable-next-line camelcase
   const { pull_request, repository } = payload
-  const { issueKeys } = parseSmartCommit(pull_request.title)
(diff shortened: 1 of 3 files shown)


tcbyrd:
Nice that it was such an easy fix. Looks like it's working perfectly!

image
image

ACyphus:
@gimenete To confirm, how does this impact older branches with issue keys in their name. Will those add pull requests that are already created to the Jira issue?

For example, I opened a branch 1 month ago named `TS-3` and 2 weeks ago opened a pull request that didn't include the issue key in the title. With this change, will my pull request now appear in my Jira issue?

gimenete:
@ACyphus sorry but no, you need to trigger a pull request event (close, open, edit title/body) in that PR.

ACyphus:
Got it. Thanks @gimenete

PR #177, opened 2019-03-12
Feature: Handle payloads where issue keys are removed

tcbyrd:
On Pull Requests, a user can freely edit the title to contain issue keys ore remove them. Right now if you update the Issue Key, the Jira API will automatically move the PR from one issue to another. But if you remove all issue keys, we stop processing the payload since it doesn't contain smart commit data. This PR adds a feature that will delete the Pull Request from Jira if you remove the issue key from the title by parsing the `payload.changes` hash, which contains the title that the PR was changed _from_. We don't need to know the issue keys to delete it from Jira. We just tell the Jira API to delete the reference to the Pull Request.

### DIFF
--- a/lib/github/pull-request.js
+++ b/lib/github/pull-request.js
@@ -1,11 +1,19 @@
 const { Project } = require('../models')
 const transformPullRequest = require('../transforms/pull-request')
 const reduceProjectKeys = require('../jira/util/reduce-project-keys')
+const parseSmartCommit = require('../transforms/smart-commit')
 
 module.exports = async (context, jiraClient, util) => {
   const author = await context.github.users.getForUser({ username: context.payload.pull_request.user.login })
   const { data: jiraPayload } = transformPullRequest(context.payload, author.data)
   const { pull_request: pullRequest } = context.payload
+
+  if (!jiraPayload && context.payload.changes) {
+    const hasIssueKeys = !!parseSmartCommit(context.payload.changes.title.from)
+    if (hasIssueKeys) {
+      return jiraClient.devinfo.pullRequest.delete(context.payload.repository.id, pullRequest.number)
+    }
+  }
 
   const linkifiedBody = await util.unfurl(pullRequest.body)
   if (linkifiedBody) {
--- a/lib/github/pull-request.js
+++ b/lib/github/pull-request.js
@@ -8,7 +8,7 @@ module.exports = async (context, jiraClient, util) => {
   const { data: jiraPayload } = transformPullRequest(context.payload, author.data)
   const { pull_request: pullRequest } = context.payload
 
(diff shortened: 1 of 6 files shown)


lib/github/pull-request.js
```diff
@@ -1,12 +1,20 @@
...
 const reduceProjectKeys = require('../jira/util/reduce-project-keys')
+const parseSmartCommit = require('../transforms/smart-commit')
 
 module.exports = async (context, jiraClient, util) => {
   const author = await context.github.users.getForUser({ username: context.payload.pull_request.user.login })
   const { data: jiraPayload } = transformPullRequest(context.payload, author.data)
   const { pull_request: pullRequest } = context.payload
 
+  if (!jiraPayload && context.payload.changes) {
+    const hasIssueKeys = !!parseSmartCommit(context.payload.changes.title.from)
```
gimenete: Is it `title` always an object. Can it be null or undefined? I would check first in order to not crash if it's not there.

We are now parsing the title and the branch name looking for issue keys. We should check the branch name as well.

lib/github/pull-request.js
```diff
@@ -1,12 +1,20 @@
...
+    if (hasIssueKeys) {
+      return jiraClient.devinfo.pullRequest.delete(context.payload.repository.id, pullRequest.number)
```
gimenete: I guess we don't need to specify which issue keys were deleted, right?

lib/github/pull-request.js
tcbyrd: Good thinking! The only downside is the reference would stay if you remove the issue key from the title on a branch with issue keys, but that's probably an acceptable workflow. We already remove the references when a branch is deleted anyway. Will update this.

lib/github/pull-request.js
tcbyrd: Nope! It stores the reference in Jira as [`/repository/{repositoryId}/{entityType}/{entityId}`](https://github.com/integrations/jira/blob/master/docs/jira-dev-info-0.10-swagger.yaml#L325), so all we need to do is tell the Jira API that reference is gone and it removes it from the issue.

lib/github/pull-request.js
tcbyrd: > Is it title always an object. Can it be null or undefined?

I think it cannot be null. We're not supposed to allow blank titles, but I do wonder if there can be other objects off the end of `context.payload.changes`.

tcbyrd: Added a test case here, but basically the branch issue is already handled since `jiraPayload` will exist if a branch has an issue key.

`context.payload.changes` can have `body` instead of title, so I added check for `changes.title` to handle that case.

tcbyrd:
🚢 Tested everything out and it's working as expected!

PR #184, opened 2019-04-04
Add access for Admins of github too

tcbyrd:
Adds Staff API access for GitHub staff that have `ADMIN` access in addition to `WRITE`.

CC https://github.com/github/ecosystem-primitives/issues/251

### DIFF
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -16,7 +16,6 @@ async function getInstallation (client, subscription) {
 }
 
 function validViewerPermission (viewer) {
-  console.log(viewer)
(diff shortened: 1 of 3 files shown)


lib/api/index.js
```diff
@@ -15,6 +15,17 @@ async function getInstallation (client, subscription) {
   }
 }
 
+function validViewerPermission (viewer) {
+  console.log(viewer)
```
gimenete: A leftover here ;-)

tcbyrd: Woops!

tcbyrd:
🚢 📖 💥

PR #187, opened 2019-04-16
Add Full resync option

tcbyrd:
This adds an option through the UI and the staff API to allow a full resync. The way the sync works today, it will not rescan your entire installation if for example you add or removed repositories while the sync is in progress. This works fine if you add or remove a repository from the installation after the sync is complete, but a few users have realized they installed very large repositories and then removed them while the sync was in progress. This left the `repoSyncState` in a state where it is trying to sync repositories that it can no longer access. If this happens, the user needs a way to restart the sync from the beginning so we can clear out the state of any repositories that the sync can no longer access.

image

image

This also changes the API call to `repoSyncState` to require you to pass the `jiraHost` parameter. Before it was just finding the first subscription for a given github installation, but this is not always the subscription since you could have multiple jira instances wired up to one Org on GitHub.

cc @izuzak

### DIFF
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -37,19 +37,18 @@ You can check your sync status in the integration settings:
 | ACTIVE   | The sync has started and is still in progress. No information will be displayed in Jira. |
 | COMPLETE | The sync has finished. Information will be displayed in Jira. |
 
-The sync should take a maximum of ~5 hours. If your sync status is still `ACTIVE` after 5 hours:
+The time it takes to complete the sync will depend on the size of your installation. Since the sync scans commits for every repository in your installation, be mindful that selecting "All Repositories" will perform a scan of every repository in your account, including forks. If you have repositories with hundreds of thousands of forks (e.g. a fork of the Linux repo), the scan might take several hours to complete.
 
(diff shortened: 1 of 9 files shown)


izuzak:
Thanks so much for adding this, @tcbyrd ❤️ 

Should we also update the public troubleshooting documentation to mention when this should be used? I'm thinking this section:

https://github.com/integrations/jira/blob/master/SUPPORT.md#sync-status-not-reaching-complete

which currently says:

> If after another 5 hours your status is still not COMPLETE, try selecting just the repositories that Jira needs access to:
> 
> 1. Open the GitHub Apps settings in GitHub.
> 1. Click Configure on Jira.
> 1. In Repository access, select only the repositories that Jira needs access to
> 1. Click Save
>
> This will automatically start a new sync.

tcbyrd:
Good call @izuzak! I updated the docs in 101261d to reflect how I'm thinking about when to need a re-sync.

tcbyrd:
Added a second modal to give some information about what Sync Status means:
image

tcbyrd:
🚢

gimenete:
Sorry that I didn't review this PR 🙏 

Great work 👍

Issue #189, opened 2019-04-18
commit title / pull request title [#ABC-123] do not link correctly


ayadav:
> When using the DVCS Connector, a commit title / pull request title [#ABC-123] Test JIRA was being correctly linked to the ticket ABC-123 in JIRA. This doesn't work anymore with the new GitHub for Jira app. I have to use [ABC-123] Test JIRA (please notice the missing hash sign).

Related to https://github.com/integrations/jira/issues/188

The root cause seems to be code in lib/transforms/smart-commit.js. It considers [#ABC-123] to be a command and excludes it from list of ISSUE KEYS such. 

The basic command line syntax for a Smart Commit message is:

`<ignored text> <ISSUE_KEY> <ignored text> #<COMMAND> <optional COMMAND_ARGUMENTS>
`

Which should not match a message like below (but it does in current implementation).
`[#ABC-123] done something...
`

matisnape:
We use naming convention `#ABC-123` and now with this new integration we lost all history in Jira. 
Please, don't treat `#ABC-123` as comments 🙏

movieglu:
The same has happened to us.  For example, we had a Github commit comment such as "Implemented #MG-1234 to fix bug".

However the new Smart Commit rules no longer recognises Jira ticket numbers that are followed by a #. It only works if there is no #.

This appears to be a regression in the functionality. Unfortunately there is no easy way for us to search and replace 5+ years with of commits so that they appear on Jira tickets.

jnraine:
@adamniedzielski @movieglu Hello! :wave:

Some good news: going forward, commits and pull request titles that reference issue keys like `[#JRA-123]` or `#JRA-123]` will now be associated to those issues in Jira. If you have historical commits you'd like associated with issues, [trigger a full sync](https://github.com/integrations/jira/blob/master/SUPPORT.md#sync-status-definitions) to pull those into Jira as well.

If you still experience trouble with this, please do re-open this and I'll look into it. 🏄

PR #190, opened 2019-04-18
#189 - Update commandRegex to be more restrictive

ayadav:
A proper fix for #189 would be to check full command syntax but short term fix can be to just restrict the commandRegx to not match the general issues keys. 

New regex enforces either "end of string" or "one or more spaces" to be present after the  `#<COMMAND>` to ensure it does not match something like `[#ABC-123]` but match only `#ABC`


jnraine:
Thanks for encouraging us to fix this and offer up a quick fix! I ended up going with the full Smart Commits parsing in #213 so closing this one out.

PR #199, opened 2019-06-10
Migration to add `jiraClientKey` to subscriptions table

tcbyrd:
This will update the database so we can key off the Jira `clientKey` value instead of `baseUrl`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,10 @@ node_js:
   - "10.14.1"
 notifications:
   disabled: true
+
+# Setup Postgres database for testing
+services:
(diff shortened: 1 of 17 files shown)


tcbyrd:
Made some additional updates in 22993ed to help ensure the subscription is only setup after extracting the clientKey from a valid JWT. Will play around with this in staging a bit before promoting it.

IAmHughes:
> Made some additional updates in 22993ed to help ensure the subscription is only setup after extracting the clientKey from a valid JWT. Will play around with this in staging a bit before promoting it.

Looks great!

PR #203, opened 2019-06-14
Add build badge for TravisCI

IAmHughes:
This PR adds the build badge for TravisCI to the `README` doc! 🚀

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,5 +1,7 @@
 ## Contributing
 
+[![Build Status](https://travis-ci.com/integrations/jira.svg?branch=master)](https://travis-ci.com/integrations/jira)
+
 [code-of-conduct]: CODE_OF_CONDUCT.md
 [fork]: /fork
 [license]: LICENSE
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,7 +1,5 @@
 ## Contributing
 
-[![Build Status](https://travis-ci.com/integrations/jira.svg?branch=master)](https://travis-ci.com/integrations/jira)
-
 [code-of-conduct]: CODE_OF_CONDUCT.md
 [fork]: /fork
 [license]: LICENSE
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 # GitHub.com + Jira Software integration
 
+| Build |
+|-------|
+| [![Build Status](https://travis-ci.com/integrations/jira.svg?branch=master)](https://travis-ci.com/integrations/jira) |
+
 ## About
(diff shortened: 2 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -1,5 +1,7 @@
 ## Contributing
 
+[![Build Status](https://travis-ci.com/integrations/jira.svg?branch=master)](https://travis-ci.com/integrations/jira)
```
swinton: I think this should go on the README.md?

IAmHughes: So @tcbyrd @wilhelmklopp and @gimenete were saying we put them on the `CONTRIBUTING` instead because knowing build statuses (and hopefully codecov for Jira once we add that) are more of a CONTRIBUTOR thing than something everyone needs.

I'm not opposed to changing this and changing it in the Slack GitHub App as well to be in the README though. Thoughts?

tcbyrd: Yeh I think since it's not a library, the badge is more relevant in the `CONTRIBUTING` file. Since that's what we're doing in Slack, let's do it here as well.

swinton: > Yeh I think since it's not a library, the badge is more relevant in the `CONTRIBUTING` file. 

I dunno... There are lots of _non-library_ repos that share build status in their `README`. [VS Code](https://github.com/microsoft/vscode), and [Atom](https://github.com/atom/atom) are examples. Build status is a strong signal of a project's health, and can provide confidence to users, both new and existing, and those users can also end up becoming contributors.

I appreciate this is a minor detail, but I think it's worth calling out the health of our project in a more public way.

Thoughts?

IAmHughes: Well now I'm on the side of putting it in the `README` again 😆. Which means I need to also PR that change in the Slack integration if we decide to go that route. Open to thoughts too 👍

tcbyrd: If y'all like it in README, put it in README. It's definitely more common and helps signal "here's what's working". I could also see an argument for a status page style link in the README that pulls from some yet-to-be-created status page.

IAmHughes: Alrighty I'll commit a change to the README instead ❤️ 

> README that pulls from some yet-to-be-created status page.

This would be great, and we can use the badges for now and then maybe replace (or add) to them with this status page one day if we go that route 🎉

IAmHughes:
Per [this conversation](https://github.com/integrations/jira/pull/203#discussion_r295533567), moved this over to the `README` instead. I also made it into a table as we would hopefully add things like Codecov and Greenkeeper in the future.

Ready for review!

PR #208, opened 2019-07-19
Sync more data from branches

tcbyrd:
It looks like there's a lot going on in this PR, but a lot of it is just beefing up our tests to make sure we're catching edge cases like this. The core of what I'm proposing here is in the `mapBranch` function:

https://github.com/integrations/jira/blob/5dd752f5dd270887bb084b70ab0c7438333c211e/lib/sync/transforms/branch.js#L4-L39

The goal of this PR is to map more data to the DevInfo API in Jira during the initial sync. Currently, if the branch doesn't have an issue key in it, we stop processing that branch. What users expect is for any commit with an issue key to also make that branch show up in the DevInfo panel. This is how it works for new data, but during the initial sync we were filtering out branches if the name of the branch didn't look like it needed to sync.

The function above does this by looking at the data we get back from the `getBranches` query, which includes the branch name and the latest 50 commits for that branch. If it finds issue keys anywhere in that branch, it will now sync the branch and PR to Jira. Additionally, if the PR has an issue key in the title, the branch for that PR will also be included.

I'm also adding one little fix that's been bothering me: 

https://github.com/integrations/jira/blob/5dd752f5dd270887bb084b70ab0c7438333c211e/lib/sync/commits.js#L6-L19

This will ensure a full historical commit sync is done from the `default ref`, which __may or may not__ be called `master`. This way if someone is using a default branch called something like `dev`, it will sync everything from that branch instead of `master`.

### DIFF
--- a/lib/sync/queries.js
+++ b/lib/sync/queries.js
@@ -35,9 +35,9 @@ module.exports = {
     }
   }`,
 
-  getCommits: `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String) {
+  getCommits: `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String, $default_ref: String!) {
     repository(owner: $owner, name: $repo){
-      ref(qualifiedName: "master") {
+      ref(qualifiedName: $default_ref) {
         target {
           ... on Commit {
             history(first: $per_page, after: $cursor) {
@@ -111,6 +111,13 @@ module.exports = {
         }
       }
     }
-  }
-  `
+  }`,
+
+  getDefaultRef: `query ($owner: String!, $repo: String!) {
+    repository(owner: $owner, name: $repo){
+        defaultBranchRef {
+          name
+        }
+    }      
+  }`
 }
--- a/lib/sync/queries.js
+++ b/lib/sync/queries.js
@@ -114,7 +114,7 @@ module.exports = {
   }`,
 
   getDefaultRef: `query ($owner: String!, $repo: String!) {
-    repository(owner: $owner, name: $repo){
+    repository(owner: $owner, name: $repo) {
         defaultBranchRef {
           name
         }
--- a/lib/sync/transforms/branch.js
+++ b/lib/sync/transforms/branch.js
@@ -1,25 +1,54 @@
 const parseSmartCommit = require('../../transforms/smart-commit')
 const { getJiraId } = require('../../jira/util/id')
 
+/**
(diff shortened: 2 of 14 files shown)


lib/sync/queries.js
```diff
@@ -35,9 +35,9 @@ module.exports = {
     }
   }`,
 
-  getCommits: `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String) {
+  getCommits: `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String, $default_ref: String!) {
     repository(owner: $owner, name: $repo){
-      ref(qualifiedName: "master") {
+      ref(qualifiedName: $default_ref) {
```
scepticulous: That's awesome. ❤️

lib/sync/transforms/branch.js
```diff
@@ -1,25 +1,54 @@
...
   }
 
-  const allKeys = issueKeys.filter(Boolean)
+  if (!allKeys) {
+    // If we get here, no issue keys were found anywhere they might be found
+    return
+  }
+
+  // If we do have some keys, let's flatten the array and filter outo any undefined values
+  const filteredKeys = allKeys.filter(Boolean)
```
scepticulous: I really like the incremental implementation, that makes it very clear what is happening.
We *could* reduce the amount of conditionals by leaning into the array behavior a little bit.
Example:
```javascript
const allKeys = [].concat(branchKeys).concat(pullRequestKeys).concat(commitKeys)
const filteredKeys = allKeys.filter(boolean)

// or something like
const allKeys = [branchKeys, pullRequestKeys, commitKeys]
const filteredKeys = allKeys.filter(Boolean).flat() 

// The idea being that we dont need to care about the if checks since filter will do that for us:
> a.concat([]).concat([1,2,3]).concat(null).concat(undefined)
[ 1, 2, 3, null, undefined ]
> a.concat([]).concat([1,2,3]).concat(null).concat(undefined).filter(Boolean)
[ 1, 2, 3 ]
```

tcbyrd: Yeh I had actually intended to refactor this when I finished writing the tests, but I was getting tired. Implemented in 86a92becf76eb740de231c807db31418985447c3!

tcbyrd:
Looking good. Image shows branches syncing when PR title has a key or commits in the branch have keys. The branch `TES-8` also shows up in that relevant issue. Here it shows up in `TES-15` because there are commits in it related to `TES-15`.

image

PR #211, opened 2019-07-31
New feature: Subscription management

tcbyrd:
This branch has been under development for a bit to make sure it passes security audit, so it's fairly large, but @IAmHughes and I have been pairing on it to make sure it does what we want. The TL;DR is it allows an admin of an Organization to view/delete subscriptions to any Jira instance connected to the Org. Since there's no view in the GitHub UI that can show an Org admin what instances are connected, we're going to be requesting a permissions update in the Jira integration itself so we can provide this feature. Here's a quick walkthrough of how that works:

1) Admins will get an e-mail asking to update the permissions. If they don't see this e-mail or don't accept from there, when going to the "Manage Organizations" page, they'll now see a button to "Update" the permission for the installation:

image

2) This will take them directly to the permission update page where they can accept the new Organization membership permission. Once accepted, **if the user is an admin of the Organization**, they will now have an option to "Manage" that installation:

image

3) Which takes them to a screen that lists all subscriptions and gives them an option to delete subscriptions they no longer want:

image

If the user is **not** an admin of the Organization, this option will be disabled:

image

I've got this installed on staging and everything seems to be working as expected.

cc @scepticulous @jmilas @philipbremer for 👀

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,3 +4,16 @@ node_js:
   - "10.14.1"
 notifications:
   disabled: true
+env:
+  global:
+    - NODE_ENV=test
+install:
+  - npm install
+script:
+  - sequelize db:create --env test
+  - sequelize db:migrate --env test
+  - npm test
+# Setup Postgres database for testing
+services:
+  - postgresql
+  - redis-server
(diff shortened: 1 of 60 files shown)


tcbyrd:
Going to merge this in so the docs are live when the permission request e-mail goes out, then will promote staging to prod.

Issue #188, opened 2019-04-16
Document exact patterns for branches, commits and pull requests


adamniedzielski:
Hey 👋I'm not sure if that's the right side of the integration to post this issue. Feel free to close or forward to the Atlassian side.

We are trying to migrate from the DVCS Connector to the new GitHub for Jira app, prompted by an email titled "Action required: Switch to the new GitHub for Jira integration". We're following this article - https://community.atlassian.com/t5/Jira-articles/Everything-you-need-to-know-about-migrating-from-the-DVCS/ba-p/945910

The article mentions that:

> So commits, branches, and pull requests are all being synced as they were and smart commits / workflow triggers work the same way they did previously.

I tried to find out in both GitHub and Atlassian documentation what "the same way they did" precisely means, but didn't manage to find it within 1 hour. Namely, I'm interested what exactly the accepted format is, that I have to put in my branches, commits in pull requests for the integration to work.

Some examples:
1. When using the DVCS Connector a branch name `feature/lief-12155-test-jira` was being correctly linked to the ticket `LIEF-12155` in JIRA. This doesn't work anymore with the new GitHub for Jira app. I have to use `feature/LIEF-12155-test-jira` (please note the capitalised project name).
2. When using the DVCS Connector, a commit title / pull request title `[#LIEF-12155] Test JIRA` was being correctly linked to the ticket `LIEF-12155` in JIRA. This doesn't work anymore with the new GitHub for Jira app. I have to use `[LIEF-12155] Test JIRA` (please notice the missing hash sign).

Based on these two cases I believe that it's beneficial to document the exact patterns that are used to match branches, commits and pull requests to JIRA tickets. Moreover, it would also be helpful to point out the breaking changes in the migration guide.

matisnape:
We use naming convention `#ABC-123` and now with this new integration we lost all history in Jira. 
Please, don't treat `#ABC-123` as comments 🙏

jnraine:
@adamniedzielski Hello! Thank you for the bug report. I was able to reproduce the bug parsing `[#LIEF-12155]` and have a fix in the pipe (#213).

However, I'm having trouble with the other case:

> When using the DVCS Connector a branch name feature/lief-12155-test-jira was being correctly linked to the ticket LIEF-12155 in JIRA. This doesn't work anymore with the new GitHub for Jira app. I have to use feature/LIEF-12155-test-jira (please note the capitalised project name)

[...]

PR #213, opened 2019-08-01
Improve Smart Commits support

jnraine:
**The diff is big! Most of it is `package-lock.json`. If you know how to avoid such a change, let me know.**

This improves issues key extraction from Smart Commits strings, requested in #188, #189. Here are some examples of the cases now supported:

```javascript
[#JRA-123]      // Associate to JRA-123
#JRA-123 #close // Associate to JRA-123 and trigger close
```

At first glance, both requests seem like minor bug fixes. My initial intent was to modify the existing code but it proved more difficult than anticipated. After a spike and a conversation with @scepticulous, I decided to take a different approach.

## Why the different approach?

The previous code split a line into two chunks: everything before the command and the command. For example, `JRA-123 #close all done` was split into `JRA-123` and `close all done`. We only looked for issue keys in the first half.

This approach makes supporting `#JRA-123 #close all done` difficult because it splits into `<empty string>` and `JRA-123 #close all done`.

## How it works

Instead of a single pass, parsing [Smart Commits](https://confluence.atlassian.com/fisheye/using-smart-commits-960155400.html) now happens in two steps:

### 1. Tokenize

```
JRA-123 #close all done
```

becomes

```javascript
[
  {type: "issueKey", value: "JRA-123"},
  {type: "transition", value: "close"},
  {type: "comment", value: "all done"},
]
```

### 2. Extract issue keys

Once tokenized, it is converted into a hash that can be sent to the [JIRA developer information endpoint](https://developer.atlassian.com/cloud/jira/software/rest/#api-group-Development-Information).

```javascript
[
  {type: "issueKey", value: "JRA-123"},
  {type: "transition", value: "close"},
  {type: "comment", value: "all done"},
]
```

becomes

```javascript
{issueKeys: ["JRA-123"]}
```

Prior to this commit, we also extracted the commands but because we didn't use them, I've decided to remove support. If we ever need it, we can use the implementation in https://github.com/integrations/jira/pull/213/commits/3f1c8a979b10dd6bd427dfefad85f5e4b58a137e or write a [grammar file](https://nearley.js.org/docs/tokenizers).

## Risks

**This is a less familiar approach.**

You need to understand how [Moo](https://github.com/no-context/moo) works to make changes. If there is an unexpected character in the source, the tokenizer will raise a syntax error.

However, I think it's well worth it. Our previous implementation wasn't at parity with what DVCS offered and wasn't easy to modify to get there. The two-step approach also makes the code easier to extend and debug. If an issue key isn't appearing, it is easy to test drive a fix at the tokenizer without needing to consider the transform code.

**This is a rewrite.**

Since learning that JIRA handles the command parsing for us, I'm less concerned about this change. The issue key extraction is fairly unambiguous and is well-tested. 

~~This replaces the existing Smart Commits parsing code. It's possible that previously supported cases no longer work. To mitigate this, I've expanded test coverage to include every case [documented by Atlassian](https://confluence.atlassian.com/fisheye/using-smart-commits-960155400.html) and those mentioned by users of the integration. I've also completed some manual testing. However, I'd like to complete further testing before shipping.~~

~~One option is to use [scientist](https://github.com/trello/scientist) to deploy the new implementation in tandem. There will be mismatches—the new implementation supports more cases—but having visibility into how things are different seems worthwhile.~~


lib/transforms/smart-commit.js
```diff
@@ -1,100 +1,95 @@
-// RDVCS Connector regexp provided by Atlassian
```
jnraine: Same file name, entirely different contents. If you're not already, switch to a "split diff" for a better review experience.

test/unit/smart-commit.test.js
```diff
@@ -253,6 +254,16 @@ describe('Smart commit parsing', () => {
         issueKeys: ['JRA-090']
       })
     })
+
+    it('uppercases issue keys', () => {
+      const text = 'jra-090'
```
tcbyrd: I've gone back and forth on this, but in general it's always seemed best to ignore lowercases so we eliminate a whole class of false positives. Branch names especially are almost always lower case and have dashes in them. I can't recall any feedback where people requested to be able to lower case issue keys, and all of the entities in the official QA repositories are capitalized: https://github.com/fusion-qa

jnraine: Your point about false positives makes sense and the [Atlassian docs](https://confluence.atlassian.com/bitbucket/use-smart-commits-298979931.html) don't mention supporting lower case keys either.

The reason I added this was because of https://github.com/integrations/jira/issues/188. A user reported that DVCS was able to pull lower case tickets out of branch names:

> When using the DVCS Connector a branch name feature/lief-12155-test-jira was being correctly linked to the ticket LIEF-12155 in JIRA. This doesn't work anymore with the new GitHub for Jira app. I have to use feature/LIEF-12155-test-jira (please note the capitalised project name).
> – https://github.com/integrations/jira/issues/188

I hadn't considered it until now but perhaps DVCS made an exception for branches, allowing lower case issue keys. Do you think that would be a better approach?

In the meantime, I'll try to setup DVCS on my developer Atlassian account to confirm the behaviour is scope to branches only.

jnraine: After playing with DVCS, I found that it also does not support lower case issue keys, except when they appear at the beginning of a branch name.

```bash
feature/ws-1-something-good # doesn't associate to WS-1
ws-1-something-good         # does associate to WS-1
```

scepticulous: I agree with not supporting lower-case identifiers. All caps is the de facto standard for JIRA and within the JIRA UX this is how it is always presented and to my knowledge, everybody does it like that. 
Since it makes the solution more resilient to only support uppercase, I suggest doing so until somebody actually asks for lower case support.

jnraine: :+1: That seems reasonable. I'll drop support for lower case identifiers.

I'll open an issue to capture the special case for branches starting with lower case identifiers in the DVCS Connector to explore adding later.

package.json
```diff
@@ -27,6 +27,7 @@
     "express-sslify": "^1.2.0",
     "helmet": "^3.13.0",
     "markdown": "^0.5.0",
+    "moo": "^0.5.0",
```
tcbyrd: Do you have experience with this? Seems like a great  solution, but I'm always cautious about bringing in new packages that are `0.*`.

test/unit/smart-commit-tokenizer.test.js
```diff
@@ -0,0 +1,191 @@
...
+    })
+
+    it('extracts issue keys embedded in branch names', () => {
+      const tokens = tokenize('feature/JRA-123-my-feature')
+
+      expect(valuesForType(tokens, 'issueKey')).toEqual(['JRA-123'])
+    })
+
+    it('extracts issue keys prefixed with a hash', () => {
+      const tokens = tokenize('#JRA-123 [#JRA-456]')
```
tcbyrd: 👍

package.json
jnraine: Yeah, I'd feel better if it were 1.0 too.

I haven't used it before but am comfortable using it because it's [been around for years](https://www.npmjs.com/package/moo?activeTab=versions), [sees regular maintenance](https://github.com/no-context/moo/commits/master), and is a dependency of [nearley](https://nearley.js.org/#projects-using-nearley), a popular JavaScript  parser toolkit.

If you think further due diligence is necessary or that it's too great a risk, I'm all ears.

tcbyrd: No objections. Let's give it a shot!

jnraine:
@tcbyrd tested against `fusion-qa/over-20k` and `fusion-qa/lots-of-branches` and everything was 👌 . It's getting close to EOD so I'm going to hold off deploying this.

When do expect https://github.com/integrations/jira/pull/222 to be merged? It would be good to have exception tracking before this one goes out. 😅

tcbyrd:
> When do expect #222 to be merged?

I've got it in a good enough state now. I'll promote it to prod first thing tomorrow so we can merge.

jnraine:
Waiting to deploy this until other Sentry issues are resolve. 🕐

PR #222, opened 2019-08-06
Better error handling and reporting

tcbyrd:
Our errors in Sentry were pretty difficult to grok and not full of the right context to troubleshoot errors. This pulls in the newest Sentry client, which overwrites the one included in Probot, and adds a new Error handler module that is placed inside our various middleware functions. There's definitely room to improve things further and add additional logging output, but this will get our middleware wired up so the context has the metadata we need to better debug issues.

To add detailed error messages that show up in Sentry, you can log an error using the Bunyan streams this sets up.

e.g.

In github middlware (Probot):

```js
app.log.error('github error')
```

In the Jira client middleware:

```js
logger.error('jira error')
```

In the frontend app (Express routes);

```js
req.log.error('Express error')
```

There is also a `/boom` route that will send a test error to Sentry if everything is setup.


lib/frontend/app.js
```diff
@@ -108,6 +109,23 @@ module.exports = (appTokenGenerator) => {
...
 
+    Sentry.withScope(async scope => {
+      scope.setTag('jiraHost', req.session.jiraHost || req.query.xdm_e)
+
+      if (req.session.githubToken) {
+        try {
+          const { data: { login } } = (await res.locals.github.users.get())
+          Sentry.setUser({ login })
+        } catch (_) {
+          console.log('unable to get user info')
```
jnraine: Is this something that should be logged using the probot logger, [similar to this](https://github.com/integrations/jira/blob/6239bf0176c1bfbf2a65548a4a073c455a9c8cc5/lib/jira/client/axios.js#L4)?

lib/github/middleware.js
```diff
@@ -1,9 +1,17 @@
...
 const getJiraUtil = require('../jira/util')
+const { Sentry } = require('../error-handler')
 
 module.exports = function middleware (callback) {
   return async (context) => {
+    Sentry.setContext('GitHub Payload', {
+      event: context.name,
+      action: context.payload.action,
+      id: context.id,
+      nwo: (context.payload.repository) ? context.repo() : undefined
```
jnraine: What does `nwo` stand for?

lib/frontend/app.js
tcbyrd: Any preferences? I wrote it this way to explicitly swallow the error inside the `withScope` block. My thought here is we just want to attempt to get the user info if we can, but otherwise we don't want the rest of this block to fail.

lib/github/middleware.js
tcbyrd: It's "Name With Owner", which is a general abbreviation I see used when we want to log the repo name with it's owner. We could probably just call it `repo` though. It returns an object that has both the repo name and owner login, so while it has more data than just the repo, maybe it's less confusing than calling it `nwo`, especially since the function is `.repo()`.

jnraine: So `integrations/jira` instead of `jira`?

I like `repo` for the reasons you've outlined. 💯

lib/frontend/app.js
jnraine: I like how you've added the user when it is available. It works as it stands. If you know what could go wrong, it may help to match on the error before swallowing or including the error in the log message.

In regards to _where_ to log, I lean toward using a logger over printing directly to STDOUT with `console.log`. It gives us a few extra details, which may prove useful.

I do have a follow-up: does everything printed to STDOUT make it to Loggly?

tcbyrd: > does everything printed to STDOUT make it to Loggly?

Yep. And we choose what goes to STDOUT based on the LOG_LEVEL env variable.

> If you know what could go wrong

Yeh that's part of the issue. It's just a catchall for any kind of problem getting data from the GitHub API. If it were a critical path I'd want to rescue various errors, but since this is just about decorating the error, I think we're ok here just moving on.

I do like the idea of using the real application logger. Let me see if I can get that in scope here.

lib/github/middleware.js
tcbyrd: > So integrations/jira instead of jira?

Yeh exactly. Since a repo name by itself isn't enough data.

I updated to `repo` in 83fd55fe83ec1ceb271f778bf6a978e4be74bc87

PR #225, opened 2019-08-07
Add indexes for subscriptions transactions

tcbyrd:
This adds two indexes for the Subscriptions table. One for `githubInstallationId` and `jiraHost`, and another for just `gitHubInstallationId`.

### DIFF
--- a/db/migrations/20190807230019-add-indexes.js
+++ b/db/migrations/20190807230019-add-indexes.js
@@ -4,7 +4,7 @@ module.exports = {
   up: async (queryInterface, Sequelize) => {
     await queryInterface.addIndex('Subscriptions', {
       fields: ['gitHubInstallationId', 'jiraHost'],
-      name: 'index-on-installation-and-jira-host'
+      name: 'Subscriptions_gitHubInstallationId_jiraHost_idx'
     })
 
     await queryInterface.addIndex('Subscriptions', {
--- a/db/migrations/20190807230019-add-indexes.js
+++ b/db/migrations/20190807230019-add-indexes.js
@@ -4,17 +4,23 @@ module.exports = {
   up: async (queryInterface, Sequelize) => {
     await queryInterface.addIndex('Subscriptions', {
       fields: ['gitHubInstallationId', 'jiraHost'],
-      name: 'index-on-installation-and-jirahost'
+      name: 'index-on-installation-and-jira-host'
     })
 
     await queryInterface.addIndex('Subscriptions', {
-      fields: ['gitHubInstallationId'],
-      name: 'index-on-github-installation-id'
+      fields: ['jiraHost'],
+      name: 'index-on-jira-host'
+    })
+
+    await queryInterface.addIndex('Subscriptions', {
+      fields: ['jiraClientKey'],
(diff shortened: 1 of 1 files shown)


db/migrations/20190807230019-add-indexes.js
```diff
@@ -0,0 +1,20 @@
...
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addIndex('Subscriptions', {
+      fields: ['gitHubInstallationId', 'jiraHost'],
+      name: 'index-on-installation-and-jirahost'
+    })
+
+    await queryInterface.addIndex('Subscriptions', {
+      fields: ['gitHubInstallationId'],
```
jnraine: If I understand correctly, the composite index from line 5 will support querying against _only_ `gitHubInstallationId`.

```sql
SELECT * FROM "Subscriptions" WHERE "gitHubInstallationId" = 123; -- covered by index
SELECT * FROM "Subscriptions" WHERE "gitHubInstallationId" = 123 AND jiraHost = "github.jira.com"; -- covered by index
SELECT * FROM "Subscriptions" WHERE jiraHost = "github.jira.com"; -- NOT covered by index
```

Do we ever query using only the `jiraHost`? If so, what do you think about adding that as a separate index?

tcbyrd: > Do we ever query using only the jiraHost? If so, what do you think about adding that as a separate index?

Yeh working on that now. We also need one for the jira client key. Multi-column indexes always make my head hurt, so thanks!

db/migrations/20190807230019-add-indexes.js
```diff
@@ -0,0 +1,26 @@
+'use strict'
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addIndex('Subscriptions', {
+      fields: ['gitHubInstallationId', 'jiraHost'],
+      name: 'index-on-installation-and-jira-host'
```
jnraine: I'm new to Postgres so take this with a grain of salt. After some Googling, I found some [naming conventions](https://stackoverflow.com/questions/4107915/postgresql-default-constraint-names/4108266#4108266) that seem to match internal Postgres indexes (`SELECT * FROM pg_indexes`).

What do you think of this name?

```suggestion
      name: 'Subscriptions_gitHubInstallationId_jiraHost_idx'
```

tcbyrd: Works for me! I was doing a mental conversion between how we define them in `github/github` and what that might look like in JS, but making them look more like Postgres is probably better.

db/migrations/20190807230019-add-indexes.js
```diff
@@ -0,0 +1,26 @@
+'use strict'
```
jnraine: Minor: I just noticed the generator adds this while making my own. Since we're using Node 8.x, we're always running in strict mode so [we don't need this](https://stackoverflow.com/questions/31685262/not-recommended-to-use-use-strict-in-es6).

That said, it doesn't hurt.

PR #230, opened 2019-08-09
Improve installation verification

jnraine:
This aims to fix [`Cannot destructure property jiraHost of 'undefined' or 'null'.`](https://sentry.io/organizations/github-integrations/issues/1143967447)

The [installation flow](https://developer.atlassian.com/cloud/jira/platform/authentication-for-apps/) for an [Jira Cloud app](https://developer.atlassian.com/cloud/jira/platform/integrating-with-jira-cloud/) (like this codebase) has two steps:

1. **Install**. Atlassian makes an unauthenticated request to us, providing a shared secret. We store the secret in the database. The secret can't be used until step two.
2. **Enable**. Atlassian makes an authenticated request using the shared secret. We then make an authenticated request to their API. If we are successful, we mark the installation as enabled.

We have more than [30 installations](https://data.heroku.com/dataclips/xrrkxmxsgsgzgpujqbkbrstlbrjf) that have repository subscriptions but aren't enabled. We don't know for certain what causes this. My current hypothesis is that a race condition causes our verification in step two to never happen.

We send the request status _before_ adding the `finish` callback:

https://github.com/integrations/jira/blob/c903c49eed1a7bfce5e5e2ffece71158b1736f22/lib/jira/enable.js#L7-L8

## Solution

1. **Fix the (potential) bug.** This removes the possible race condition by adding the callback before sending a response.

2. **Add endpoint to fix bad installations.** This adds `POST /api/jira/<id>/verify`, accessible to GitHub staff only, to trigger verification for a given Jira installation ID. This should allow us to fix installations that are still valid and clean up those that aren't.

### DIFF
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -4,6 +4,9 @@ const app = express()
 const bodyParser = require('body-parser').urlencoded({ extended: false })
 const octokit = require('@octokit/rest')()
 
+const { Installation } = require('../models')
+const verifyInstallation = require('../jira/verify-installation')
+
 async function getInstallation (client, subscription) {
   const id = subscription.gitHubInstallationId
   try {
@@ -180,6 +183,32 @@ app.post('/:installationId/sync', bodyParser, async (req, res) => {
   } catch (err) {
     console.log(err)
     return res.sendStatus(401)
+  }
+})
+
+app.post('/jira/:installationId/verify', bodyParser, async (req, response) => {
+  const { installationId } = req.params
+  const installation = await Installation.findById(installationId)
+
+  const respondWith = function (message) {
+    const data = {
+      message: message,
+      installation: { enabled: installation.enabled, id: installation.id, jiraHost: installation.jiraHost }
+    }
+
+    return response.send(JSON.stringify(data))
+  }
+
+  if (installation.enabled) {
+    respondWith('Installation already enabled')
+  } else {
+    await verifyInstallation(installation, req.log)()
+
+    if (installation.enabled) {
+      respondWith('Verification successful')
+    } else {
+      respondWith('Verification failed')
+    }
   }
 })
 
--- a/lib/jira/verify-installation.js
+++ b/lib/jira/verify-installation.js
@@ -1,0 +1,28 @@
+const getAxiosInstance = require('../jira/client/axios')
+const { Sentry } = require('../error-handler')
+
+module.exports = function (installation, log) {
+  return async () => {
+    const instance = getAxiosInstance(null, {}, installation.jiraHost, installation.sharedSecret)
+
+    try {
+      const result = await instance.get(`/rest/devinfo/0.10/existsByProperties?fakeProperty=1`)
+      if (result.status === 200) {
+        log('App enabled on Jira')
+        installation.enable()
+      } else {
+        const message = `Unable to verify Jira installation: ${installation.jiraHost} responded with ${result.status}`
+        log(message)
+        Sentry.captureMessage(message)
+      }
+    } catch (err) {
+      if (err.response && err.response.status === 401) {
+        log('Jira does not recognize this installation. Deleting it.')
+        installation.destroy()
+      } else {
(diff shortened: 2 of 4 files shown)


lib/api/index.js
```diff
@@ -183,6 +186,32 @@ app.post('/:installationId/sync', bodyParser, async (req, res) => {
...
+
+    return response.send(JSON.stringify(data))
+  }
+
+  if (installation.enabled) {
+    respondWith('Installation already enabled')
+  } else {
+    await verifyInstallation(installation, req.log)()
+
+    if (installation.enabled) {
```
scepticulous: Am I right, that the idea here is, that the function `respondWith` closes (as in closure) over the mutable `installation` and that `verifyInstallation` mutates `installation`?
If so, I am a little afraid of this shared mutable state.

I personally would have loved to have `verifyInstallation` to return a new object instead of mutating its argument.

lib/jira/verify-installation.js
```diff
@@ -0,0 +1,28 @@
+const getAxiosInstance = require('../jira/client/axios')
+const { Sentry } = require('../error-handler')
+
+module.exports = function (installation, log) {
+  return async () => {
```
scepticulous: I am not sure why we are returning a function here? 
Would be great if you could share your thoughts about this.

lib/api/index.js
jnraine: Your understand of this code is bang on.

I like the idea of returning a mutated copy but I'm unsure if it fits this context. `installation` is a Sequelize model that behaves like ActiveRecord: to change the row, you update the object's attribute then persist that in the database. We want to set `enabled=true`. To avoid mutation, we would need to do something like this:

```javascript
const verifyInstallation = async function (installation) {
  const installationCopy = installation.copy() // Hand-waving -- not sure if Sequelize supports this
  // ...verify the install
  await installationCopy.update({enabled: true})
  return installationCopy
}
```

I think we could make this work but I'm hesitant to go against the conventions of Sequelize.

**Disclaimer**: I'm new to Node and Sequelize so I may be missing something obvious. If so, please tell me! 😄

lib/jira/verify-installation.js
jnraine: Yeah, it's a bit weird.

We use it in two places. [In the first](https://github.com/integrations/jira/pull/230/files#diff-766fc9b709027b4794f61f1326ab6a21R205), we immediately call the returned function so the outer function serves no purpose. [In the second](https://github.com/integrations/jira/pull/230/files#diff-a7e9a0e95ff84e5221ba1ddbd31f60ddR9), we need to return a callback function that has `installation` and `log` available to it. Because it is called _later_, I've used a function that returns a function to ensure `installation` and `log` within the functions closure. 

I hope that made sense. It was difficult to describe with text. If you know a way around this, I'd happily give you sparkles! ✨

lib/api/index.js
scepticulous: Got your point. What about just returning the mutated installation from that function so that the code is more obvious. Especially the second (nested) `if (installation.enabled)`.
Even though technically it is the same object, the API would be cleaner and would allow potential future changes, even though we stick with the current semantics for now.
I agree however that we stick with conventions (internally). But I am also not a Sequelize expert ✌️

lib/jira/verify-installation.js
scepticulous: Makes sense 👍

jnraine:
@scepticulous Just saw your comment about tests. I have _not_ added tests for this yet. I spent more time that I'd like to admit trying to get tests for the existing `installation` endpoint. I got stuck trying to figure out how to ensure my assertions ran _after_ the response "finish" callbacks. Ultimately, I decided to get this out ahead of the tests to plug the errors.

I have the tests stashed and plan to come back to them later this week.

lib/api/index.js
jnraine: I need to refresh more—just saw this comment. I've made a note of this and will make the change along with the tests.

PR #233, opened 2019-08-09
Don't show generic error page in development

jnraine:
When an error occurs within a request, we display a generic error page to the user. This is useful for production but hides problems in development: instead of seeing a stacktrace or error, you see an Octocat.

62794609-60dea280-ba89-11e9-9ed2-efe96b827111

Cute but not helpful.

This pull request disables the generic error handler middleware when `NODE_ENV=development`. This also adds `NODE_ENV=development` to `.env.example` so folks get this behaviour for free.


lib/frontend/app.js
```diff
@@ -126,12 +118,28 @@ module.exports = (appTokenGenerator) => {
       }
       Sentry.captureException(err)
```
tcbyrd: This needs `next(err)` so the error falls through to the next handler
```suggestion
       Sentry.captureException(err)
       next(err)
```

lib/frontend/app.js
```diff
@@ -126,12 +118,28 @@ module.exports = (appTokenGenerator) => {
...
-      .render('github-error.hbs', {
-        info,
-        title: `${err.message} - GitHub + Jira integration`
-      })
-  })
+  }
+
+  const catchErrors = async (err, req, res, next) => {
+    if (process.env.NODE_ENV === 'development') {
+      return next()
```
tcbyrd: Same here. If we pass `err` into the context, the browser will show the error.

```suggestion
      return next(err)
```

PR #234, opened 2019-08-09
Add script/server

jnraine:
Mirroring the monolith, you can now start the app with a single command:

```
script/server
```

script-server

This sets up a tunnel using ngrok or localtunnel, starts a web server, and starts a worker. Both the web and worker will restart upon file changes.

In order to handle the tunnel, I’ve added `TUNNEL_PORT` and `TUNNEL_SUBDOMAIN` to `.env.example`. Once set, these values will be used to open a tunnel using a consistent subdomain.

While ngrok provides the best experience, I have included support for localtunnel because it is free. Even if you can’t buy an ngrok license, you can setup the app locally.

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,0 +1,3 @@
+tunnel: script/tunnel
+web: nodemon --exec "npm start"
+worker: nodemon --exec "npm run worker"
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -1,3 +1,3 @@
 tunnel: script/tunnel
-web: nodemon --exec "npm start"
-worker: nodemon --exec "npm run worker"
+web: npm run dev
+worker: npm run dev:worker
--- a/script/tunnel
+++ b/script/tunnel
@@ -14,9 +14,9 @@ elif [ -z ${TUNNEL_SUBDOMAIN} ]; then
   exit 1
 fi
 
-if is_installed ngrok1; then
+if is_installed ngrok; then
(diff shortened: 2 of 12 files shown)


Procfile.dev
```diff
@@ -0,0 +1,3 @@
+tunnel: script/tunnel
+web: nodemon --exec "npm start"
+worker: nodemon --exec "npm run worker"
```
tcbyrd: Just a note that I had to `npm i -g nodemon` in order for this to work, which I didn't need before due to how `npm run test` resolve binaries. May want to include it in `script/bootstrap` or you could just do `./node_modules/.bin/nodemon`.

jnraine: Good find. I've tweaked it to use commands from `package.json` so it should just work now.

script/tunnel
```diff
@@ -0,0 +1,27 @@
...
+fi
+
+if is_installed ngrok1; then
+  exec ngrok http "${TUNNEL_PORT}" --subdomain="${TUNNEL_SUBDOMAIN}" --log=stdout
+elif is_installed lt1; then
+  exec lt --port "${TUNNEL_PORT}" --subdomain "${TUNNEL_SUBDOMAIN}" --print-requests
+else
+  echo_red "No tunnel utility found!"
+  echo "To install, run this:"
+  echo_bold "  brew install ngrok # requires license"
```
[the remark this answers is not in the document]
jnraine: Fixed! Turns out the localtunnel one was wrong too. 😆

PR #236, opened 2019-08-13
Set Sentry environment to NODE_ENV

jnraine:
This will allow us to send production and staging to a single project on Sentry.

### DIFF
--- a/lib/error-handler.js
+++ b/lib/error-handler.js
@@ -4,9 +4,8 @@ const Integrations = require('@sentry/integrations')
 
 Sentry.init({
   dsn: process.env.SENTRY_DSN,
-  release: process.env.HEROKU_SLUG_COMMIT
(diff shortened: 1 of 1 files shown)


lib/error-handler.js
```diff
@@ -4,9 +4,8 @@ const Integrations = require('@sentry/integrations')
 
 Sentry.init({
   dsn: process.env.SENTRY_DSN,
-  release: process.env.HEROKU_SLUG_COMMIT
-    ? process.env.HEROKU_SLUG_COMMIT
-    : undefined,
+  environment: process.env.NODE_ENV,
+  release: process.env.HEROKU_SLUG_COMMIT,
```
jnraine: I tweaked this because I believe `null` results in the same behaviour from Sentry.

PR #237, opened 2019-08-13
Fix webhook signature mismatch

jnraine:
This fixes an error that causes webhooks delivered by GItHub to be ignored due to an apparent signature mismatch.

Since adding error tracking last week, we’ve lost 2,400 webhooks.

This mirrors the approach taken to fix the same issue in Slack: https://github.com/integrations/slack/pull/839.


jnraine:
Hmm, test failures. Removing review requests until I have that sorted.

PR #239, opened 2019-08-15
Stop sending error to Sentry multiple times

jnraine:
When an exception was raised, it was sent to Sentry three or more times. This stops that. It also improves the metadata that comes with each error and stops metadata from leaking between exceptions.

Sentry used to be configured within `lib/error-handler.js` where we called `Sentry.init` and then exported `Sentry`. Anywhere we needed to use Sentry to add context, send errors, or add middleware, we would require `lib/error-handler.js`.

However, `Sentry.init` enables error delivery _globally_ and each time it is called, it appeared to cause errors to be delivered multiple times. We also called `Sentry.captureException` from a function that was called when an exception was raised, which sent the error to Sentry one extra time.

_(^^ These two assertions ought to be double-checked as I put this together quickly at the end of a long day.)_

While working on this, I found code that was calling `Sentry.configureScope` instead of `withScope`. The former is intended for metadata that applies to all exceptions but was being called with request-specific metadata. This metadata then leaked across into other exceptions.

This can be tested by [setting SENTRY_DSN](https://forum.sentry.io/t/where-can-i-find-my-dsn/4877/4) locally and then hitting the test endpoints:

- `POST /boom`
- `GET /boom`
- `POST /jira/boom`
- `GET /jira/boom`

For example, you can use curl:

```
curl -v -X POST --data "foo=bar" https://jnraine-jira.ngrok.io/jira/boom\?xdm_e\=foo.com
```

This doesn’t have automated test coverage. Without the time pressure of reducing our error volume, I would have added some. I recommend we test this manually and get it out as it will reduce error volume by 2-3x times and improve the quality of our exception data. If volume remains an issue, it is relatively easy to filter out unwanted excpetions using `Sentry.configureScope`.


lib/frontend/app.js
```diff
@@ -105,27 +103,30 @@ module.exports = (appTokenGenerator) => {
...
   })
 
-  const sentryMiddleware = async (err, req, res, next) => {
+  const addSentryContext = async (err, req, res, next) => {
     Sentry.withScope(async scope => {
-      scope.setTag('jiraHost', req.session.jiraHost || req.query.xdm_e)
-
-      if (req.session.githubToken) {
-        try {
-          const { data: { login } } = (await res.locals.github.users.get())
```
jnraine: Removing this because, if I understand it correctly, it would make an HTTP request to GitHub for each exception raised.

lib/frontend/app.js
```diff
@@ -144,7 +145,6 @@ module.exports = (appTokenGenerator) => {
     })
   }
 
-  app.use(sentryMiddleware)
```
jnraine: Moved this to before `Sentry.Handlers.errorHandler()`. If it comes after, it isn't guaranteed to have run by the time the error is sent to Sentry.

lib/jira/client/axios.js
```diff
@@ -43,14 +42,6 @@ function getErrorMiddleware (id, installationId) {
       params: error.config.fields
     }, `Jira request: ${error.request.method} ${error.request.path} - ${status} ${statusText}`)
 
-    Sentry.configureScope(scope => {
```
jnraine: This is the bit that set context globally. Once this ran, any exception that came after it was tagged with this context.

bin/worker
```diff
@@ -5,6 +5,14 @@ if (process.env.NEWRELIC_KEY) {
   require('newrelic')
 }
 
+const Sentry = require('@sentry/node')
+
+Sentry.init({
```
jnraine: I've copy/paste this for now to ensure it isn't called twice. This can't go in `lib/worker/index.js` because we require that from within the web process.

I like the idea of extracting this out in the future.

lib/frontend/app.js
```diff
@@ -105,27 +103,30 @@ module.exports = (appTokenGenerator) => {
...
-        }
+      const jiraHost = (req && req.session && req.session.jiraHost) || (req && req.query && req.query.xdm_e)
+      if (jiraHost) {
+        scope.setTag('jiraHost', jiraHost)
       }
 
       if (req.body) {
         Sentry.setContext('Body', req.body)
       }
-      Sentry.captureException(err)
```
jnraine: This is the bit that takes an exception we are in the process of sending to Sentry and tells Sentry to send it again.

.vscode/launch.json
```diff
@@ -0,0 +1,14 @@
+{
```
jnraine: Oops, accidentally added. Will remove before deploy.

PR #240, opened 2019-08-16
Add STALLED sync status

jnraine:
When an active sync has not been updated in the last fifteen minutes, assume something went wrong and mark it as staled. We make a best effort to mark syncs that have failed due to an error as “FAILED” but there are cases where the error handler doesn’t run.

image

While we work to resolve these issues, the customer shouldn’t need to wait 48 hours before knowing something isn’t working as it should. This helps avoid that.

### DIFF
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -36,6 +36,8 @@ You can check your sync status in the integration settings:
 | PENDING  | The sync has not started.  |
 | ACTIVE   | The sync has started and is still in progress. No information will be displayed in Jira. |
 | COMPLETE | The sync has finished. Information will be displayed in Jira. |
+| STALLED  | The sync has not finished but has stopped being updated. Partial information may appear in Jira. |
+| FAILED   | The sync hit an error and stopped without completing. Partial information may appear in Jira. |
 
 The time it takes to complete the sync will depend on the size of your installation. Since the sync scans commits for every repository in your installation, be mindful that selecting "All Repositories" will perform a scan of every repository in your account, including forks. If you have repositories with hundreds of thousands of forks (e.g. a fork of the Linux repo), the scan might take several hours to complete.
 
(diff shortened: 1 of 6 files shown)


izuzak:
Another idea: how do you feel about exposing the updated_at time to end-users? I think that would reduce their confusion around syncs that really are still ACTIVE. So, if a user checks the status and they see ACTIVE, and they check again and again see ACTIVE, and again and again -- they might start to wonder if things are still progressing as expected. If we told them "yeah, the sync last updated 1 minute ago", then that would give them confidence that things are still progressing and they wouldn't need to reach out to Support to ask what's going on.

What do you think? 💭 This definitely doesn't need to happen as a part of this pull request, just wanted to share the idea here since it has a similar note 🎶

jnraine:
@izuzak Good thought about the rate limiting. I checked and didn't find _any_ instances of us delaying due to an exceeded rate limit on [Loggly](https://githubint.loggly.com/search#terms="Delaying%20job%20for"&from=2019-08-09T23:11:29.079Z&until=2019-08-16T23:11:29.079Z&source_group=). It goes back two weeks. It's possible that a customer could it as STALLED due to rate limiting. I could increase it to an hour to eliminate that possibility. I'm leaning towards keeping it at 15 minutes to give people faster feedback. What do you think?

Great idea about the "Last updated" column. I've added it. 🏄

jnraine:
> Do we have any way of figuring out the update frequency of regular syncs?

@scepticulous the `updatedAt` on active syncs is at most one minute old for currently active syncs. This excludes any stalled syncs (no updates for days). I'll send you a link with this data if you're curious to dive in deeper.

jnraine:
@izuzak Right! I forgot to add the additional instructions to SUPPORT.md. I've done so now. In regards to the rate limit, we can theoretically delay the sync for as much as one hour. In practice, we haven't historically hit those limits. Still, since it is possible, I've opened a ticket to differential between ACTIVE and "waiting for rate limit to reset": https://github.com/integrations/jira/issues/243

jnraine:
Here is the final product:

image

PR #242, opened 2019-08-17
Add additional context to exceptions during jobs

jnraine:
Even though we see errors as they happen, we don't always have the context we need to diagnose and correct the issues. For example, we often don't know **what** was being done to trigger the error or **who** did it.

We have some information for the installation queue but nothing for pull requests. Further, because Sentry doesn't support Bull out-of-the-box, our existing strategy for capturing metadata may cause data to leak between jobs.

This improves on the existing code in a few ways:

1. Errors from all queues will now included job metadata, including job ID to tie it back the logs.
2. Errors from all queues will be searchable by `jiraHost` or `queue`.
3. Errors while updating Jira will included additional request and response data.
4. Error state will be isolated to each job.
5. Additional error metadata can be easily added by any job using `job.sentry.addExtra(...)`

[Click here](https://sentry.io/organizations/github-integrations/issues/1164659823/events/bb0928886e634d8784ccdd06be5ba547/?project=1240714&statsPeriod=14d) to see an error with the new metadata. 

Here's a look at the new tags:

Screen Shot 2019-08-19 at 2 58 48 PM

## Details

This adds a middleware function that wraps job handlers with a Sentry hub. This is a variation on the example from [this issue](https://github.com/getsentry/sentry-javascript/issues/2171#issuecomment-516410622).

Now, jobs can access their own Sentry context using `job.sentry`:

```javascript
// Within the job
job.sentry.setExtra(“some info”, {foo: “bar”})
job.sentry.setTag(“jiraHost”, “whatever”)
```

I’ve added this middleware to all three queues so we should get information any job that has a problem, including the `queue`,`jiraHost`, and more.

---

1: We use multiple Express apps (Probot events, staff API, Jira configuration) and run background jobs using [Bull](https://optimalbits.github.io/bull/). Each needs to be monitored separately.

### DIFF
--- a/lib/sync/installation.js
+++ b/lib/sync/installation.js
@@ -1,5 +1,3 @@
-const Sentry = require('@sentry/node')
-
 const { Subscription } = require('../models')
 const getJiraClient = require('../jira/client')
 const { getRepositorySummary } = require('./jobs')
@@ -80,9 +78,7 @@ module.exports.processInstallation = (app, queues) => {
   return async function (job) {
     const { installationId, jiraHost } = job.data
 
-    Sentry.setExtra('jobData', job.data)
-    Sentry.setTag('jiraHost', jiraHost)
-    Sentry.setUser({ gitHubInstallationId: installationId, jiraHost })
+    job.sentry.setUser({ gitHubInstallationId: installationId, jiraHost })
 
     app.log(`Starting job for installationId=${installationId}`)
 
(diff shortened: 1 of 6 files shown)


scepticulous:
@jnraine I did a quick review and the code changes seem good to me.
Would you mind sharing a little bit more `why`(Problem with the current approach) and the difference in the result (before, after)?
Having this additional context in the PR body will make it easier to understand the code change when doing the review.

jnraine:
@scepticulous Done! I'm in the process of fixing tests up now.

jnraine:
Deployed as part of https://github.com/integrations/jira/pull/244. Merging in separately to keep the history.

PR #246, opened 2019-08-20
Prevent transitions during initial sync

tcbyrd:
Closes https://github.com/integrations/jira/issues/192

This will ensure transitions are not called during initial sync. I've done this by changing the Jira client to allow a second parameter, which defaults to `false` (the original value), while passing `true` into the one call site where we send all payloads to a Jira instance during the initial sync jobs.

### DIFF
--- a/lib/jira/client/index.js
+++ b/lib/jira/client/index.js
@@ -107,8 +107,8 @@ module.exports = async (id, installationId, jiraHost) => {
         delete: (repositoryId) => instance.delete(`/rest/devinfo/0.10/repository/${repositoryId}`, {
           fields: { _updateSequenceId: Date.now() }
         }),
-        update: (data, preventTransitions = false) => instance.post('/rest/devinfo/0.10/bulk', {
-          preventTransitions,
+        update: (data, options) => instance.post('/rest/devinfo/0.10/bulk', {
+          preventTransitions: (options && options.preventTransitions) || false,
           repositories: [data],
           properties: {
             installationId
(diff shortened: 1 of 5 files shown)


lib/jira/client/index.js
```diff
@@ -107,8 +107,8 @@ module.exports = async (id, installationId, jiraHost) => {
         delete: (repositoryId) => instance.delete(`/rest/devinfo/0.10/repository/${repositoryId}`, {
           fields: { _updateSequenceId: Date.now() }
         }),
-        update: (data) => instance.post('/rest/devinfo/0.10/bulk', {
-          preventTransitions: false,
+        update: (data, preventTransitions = false) => instance.post('/rest/devinfo/0.10/bulk', {
```
jnraine: What do you think about using a keyed value to improve clarity when calling the method?

```suggestion
        update: (data, { preventTransitions }) => instance.post('/rest/devinfo/0.10/bulk', {
```

The calling code is longer but might be worth the extra characters.

```javascript
jiraClient.devinfo.repository.update(jiraPayload, true)
// becomes
jiraClient.devinfo.repository.update(jiraPayload, { preventTransitions: true })
```

tcbyrd: Yeh definitely like the API better that way. Take a look at https://github.com/integrations/jira/pull/246/commits/4c1b6294cce3595bf59d9dcef1c2ace68343c659 for how I implemented it. Just throwing `{ preventTransitions }` doesn't work out because of how destructuring works, so instead I opted for just passing an options param.

tcbyrd:
:shipit: 

Confirmed initial sync did not call transitions that already happened.

PR #250, opened 2019-08-23
Deduplicate Issue Keys in Jira client

srinjoym:
This PR deduplicates the `issueKeys` referenced from commits and branches before a bulk update to Jira is performed in the Jira `index.js`.

As part of this work, we also now check and throw a custom error if the number of `issueKeys` referenced by the commit or branch is above the limit of 400 defined by the [Jira API](https://developer.atlassian.com/cloud/jira/software/rest/#api-rest-devinfo-0-10-bulk-post).


lib/jira/client/index.js
```diff
@@ -107,13 +130,24 @@ module.exports = async (id, installationId, jiraHost) => {
...
-          preventTransitions: (options && options.preventTransitions) || false,
-          repositories: [data],
-          properties: {
-            installationId
+        update: (data, options) => {
+          // Deduplicate issue keys in commits and branches before request
+          if ("commits" in data) data.commits = dedupIssueKeys(data.commits)
+          if ("branches" in data) data.branches = dedupIssueKeys(data.branches)
+
+          if (issueKeySet(data.commits).length <= 400 && issueKeySet(data.branches).length <= 400) {
```
jnraine: Screen Shot 2019-08-23 at 4 00 14 PM

From [Atlassian API docs](https://developer.atlassian.com/cloud/jira/software/rest/#api-rest-devinfo-0-10-bulk-post). I believe the issue key limit per commit or branch is 100 and the commits array limit is 400.

---

We take the issues keys from every commit, put them in a set, then convert the set to an array. If that array contains for than 400 items, we bail out. Will this fail in cases when it should succeed?

For example, say we have 10 commits that each reference 40 unique issue keys. Each commit is under the issue key limit but when we add them together, there are 400 unique issue keys. Instead, does it make sense to get the max issue key count?

```javascript
const withinLimit = (resources) => {
  const issueKeyCounts = resource.map((resource) =>  resource.issueKeys.length)
  const maxIssueKeyCount = Math.max(...issueKeyCounts)

  return maxIssueKeyCount <= issueKeyLimit
}
```

lib/jira/client/index.js
```diff
@@ -107,13 +132,24 @@ module.exports = async (id, installationId, jiraHost) => {
...
         }),
-        update: (data, options) => instance.post('/rest/devinfo/0.10/bulk', {
-          preventTransitions: (options && options.preventTransitions) || false,
-          repositories: [data],
-          properties: {
-            installationId
+        update: (data, options) => {
+          // Deduplicate issue keys in commits and branches before request
+          if ('commits' in data) data.commits = dedupIssueKeys(data.commits)
+          if ('branches' in data) data.branches = dedupIssueKeys(data.branches)
```
jnraine: Is it possible for `branches[0].lastCommit.issueKeys` to exceed the issue key limit?

lib/jira/client/index.js
```diff
@@ -107,13 +132,24 @@ module.exports = async (id, installationId, jiraHost) => {
...
+          if (issueKeySet(data.commits).length <= 100 && issueKeySet(data.branches).length <= 100) {
+            instance.post('/rest/devinfo/0.10/bulk', {
+              preventTransitions: (options && options.preventTransitions) || false,
+              repositories: [data],
+              properties: {
+                installationId
+              }
+            })
+          } else {
+            // We have more than 400 issue keys so raise an exception
```
jnraine:
```suggestion
            // We have more than 100 issue keys so raise an exception
```

.vscode/launch.json
```diff
@@ -0,0 +1,20 @@
+{
```
jnraine: Did you mean to commit this change?

lib/jira/client/index.js
srinjoym: Great catch! Just added a check for that here: https://github.com/integrations/jira/blob/df5547f6ab8075425b28386e241cec6a48e8b7e7/lib/jira/client/index.js#L140-L144

.vscode/launch.json
srinjoym: my bad 😆 will merge this in after the VSCode PR so the diff is fixed

lib/jira/client/index.js
srinjoym: Another great catch! Just updated the logic here:https://github.com/integrations/jira/blob/ec19cc2d644e28e32a7eb5fa6b106562eeb065dd/lib/jira/client/index.js#L22-L27

lib/jira/client/index.js
```diff
@@ -107,13 +132,31 @@ module.exports = async (id, installationId, jiraHost) => {
...
-          properties: {
-            installationId
+        update: (data, options) => {
+          // Deduplicate issue keys in commits and branches before request
+          if ('commits' in data) data.commits = dedupIssueKeys(data.commits)
+          if ('branches' in data) {
+            data.branches = dedupIssueKeys(data.branches)
+            data.branches.forEach(branch => {
+              if ('lastCommit' in branch) {
+                dedupIssueKeys([branch.lastCommit])
```
jnraine: One other recommendation: sometime function that mutates can be surprising. Is it possible to refactor this code so the dedup and the mutation are separate and more obvious?

```
const dedup = (array) => { [...new Set(array)] }

const dedupRepository = (repositoryObj) => {
  // ...
  if ('commits' in data) {
    data.commits.forEach((commit) => {
      commit.issueKeys = dedup(commit.issueKeys)
    })
  }  
}

// later
dedupRepository(repositoryObj)
```

`dedupRepository` still mutates the object but the mutation is front and centre.

lib/jira/client/index.js
```diff
@@ -2,6 +2,30 @@ const { Installation } = require('../../models')
...
+}
+
+/**
+ * Returns if the max length of the issue
+ * key field is within the limit
+ */
+const withinIssueKeyLimit = (resources) => {
+  if (resources == null) return []
+
+  let issueKeyCounts = resources.map((resource) => resource.issueKeys.length)
```
jnraine: Minor: can this be `const`?

lib/sync/installation.js
```diff
@@ -146,7 +146,11 @@ module.exports.processInstallation = (app, queues) => {
             })
           }
 
-          throw new Error(`Failed Jira API request: ${err.response.status}`)
+          if (err.response && err.response.status) {
+            throw new Error(`Failed Jira API request: ${err.response.status}`)
```
scepticulous: I wonder if we should log the original exception here. Otherwise, we would only keep the status and all error details are inaccessible for debugging.

scepticulous:
> @scepticulous I held off on adding tests here because it looks like we don't have any form of testing for the Jira client yet in the project. [...]

Makes sense and your plan sounds great! 🎉

PR #258, opened 2019-08-29
Truncate issue keys and show warning to user

srinjoym:
This PR limits the issue keys referenced from a commit or branch to the first 100. In #250 we started deduplicating the issue keys and throwing an error if there were more than 100. Instead of throwing an error, now we truncate the list of issue keys and display a warning to the user after the sync is complete.

image

### DIFF
--- a/db/migrations/20190828145758-add-sync-warning-column-on-subscriptions.js
+++ b/db/migrations/20190828145758-add-sync-warning-column-on-subscriptions.js
@@ -1,0 +1,12 @@
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn('Subscriptions', 'syncWarning', {
+      type: Sequelize.STRING,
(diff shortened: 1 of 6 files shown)


srinjoym:
@scepticulous Totally agree that this could be jarring for users, here's the latest version building off @jnraine's suggestion to use the '*' character

image

@jnraine Just refactored all of the issue-key related functions so that it's hopefully clearer what each function does. The call from `update` is now just either `dedupIssueKeys` or `truncateIssueKeys` which makes it much more readable

ACyphus:
I would love to get a mention of this message in the [`SUPPORT.md`](https://github.com/integrations/jira/blob/master/SUPPORT.md) file. I don't want to block this pull request so it can be in a different pull request if necessary.

jnraine:
Ah, sorry — I broke the style when resolving the merge conflict!

PR #259, opened 2019-08-29
Remove Jira client “fields” path interpolation

jnraine:
**Note: this change is in preparation for the Axios 0.19 update.**

Prior to this commit, there was an interceptor that added something similar to [Express.js route parameters] to Axios.

Looking at the [PR that introduced it], it was introduced to add debug logging. However, there is a problem: as of version 0.19, [Axios strips out unexpected key/value pairs] passed to the request methods like `get` and `post`.

This commit removes that interceptor and replaces the code with vanilla Axios code, interpolating values into the path or using `params`. This changes the debug logs, though

```
Jira request: GET /rest/api/latest/issue/:issue_id?fields=summary - 200  (id=47084f00-c934-11e9-9c6b-1691d649534a, installation=1339471)
{
  baseURL: “https://jnraine.atlassian.net”,
  fields: “summary”,
  issue_id: “WS-1”
}

-- becomes 👇

Jira request: GET /rest/api/latest/issue/WS-1?fields=summary - 200  (id=47084f00-c934-11e9-9c6b-1691d649534a, installation=1339471)
```

It’s more terse but I think it’s still useful. The only information we lose is the Jira host. Since these logs are development only, you likely already know what Jira host you’re using.

As a bonus, there are no more `eslint-disable-next-line camelcase` exceptions in this file. 🤗

[Express.js route parameters]: https://expressjs.com/en/guide/routing.html#route-parameters
[PR that introduced it]: https://github.com/integrations/jira/pull/12
[Axios strips out unexpected key/value pairs]: https://github.com/axios/axios/compare/v0.18.0...v0.19.0#diff-a3c2b8d114f9b727107446e1d3fd9847

### DIFF
--- a/lib/jira/client/axios.js
+++ b/lib/jira/client/axios.js
@@ -1,5 +1,6 @@
 const axios = require('axios')
 const jwt = require('atlassian-jwt')
+const querystring = require('querystring')
 const url = require('url')
 
 const { logger } = require('probot/lib/logger')
@@ -7,38 +8,37 @@ const { logger } = require('probot/lib/logger')
 const instance = process.env.INSTANCE_NAME
 const iss = 'com.github.integration' + (instance ? `.${instance}` : '')
 
+/*
+ * Create an Atlassian request from an Axios config.
+ * Compatible with jwt.createQueryStringHash.
+ */
+jwt.reqFromAxiosConfig = (config) => {
+  const parsedUrl = url.parse(config.url)
+  const parsedQuery = querystring.parse(parsedUrl.query)
+
+  return {
+    method: config.method,
+    originalUrl: parsedUrl.pathname,
+    query: Object.assign(parsedQuery, config.params)
+  }
+}
+
 function getAuthMiddleware (secret) {
   return (config) => {
-    const { query, pathname } = url.parse(config.url, true)
+    const qsh = jwt.createQueryStringHash(jwt.reqFromAxiosConfig(config))
+    const jwtToken = jwt.encode({ ...getExpirationInSeconds(), iss, qsh }, secret)
 
-    const jwtToken = jwt.encode({
-      ...getExpirationInSeconds(),
-      iss,
-      qsh: jwt.createQueryStringHash({
-        method: config.method,
-        originalUrl: pathname,
-        query
-      })
-    }, secret)
+    config.headers['Authorization'] = `JWT ${jwtToken}`
 
-    return {
-      ...config,
-      headers: {
-        ...config.headers,
-        Authorization: `JWT ${jwtToken}`
-      }
-    }
+    return config
   }
 }
 
 function getErrorMiddleware (id, installationId) {
   return (error) => {
-    const { status, statusText } = error.response || {}
-    logger.debug({
-      id,
-      installation: installationId,
-      params: error.config.fields
-    }, `Jira request: ${error.request.method} ${error.request.path} - ${status} ${statusText}`)
+    if (error.response) {
+      logResponse(id, installationId, error.response)
+    }
 
     return Promise.reject(error)
   }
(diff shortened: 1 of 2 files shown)


lib/jira/client/axios.js
```diff
@@ -1,86 +1,61 @@
...
+ * Compatible with jwt.createQueryStringHash.
+ */
+jwt.reqFromAxiosConfig = (config) => {
+  const parsedUrl = url.parse(config.url)
+  const parsedQuery = querystring.parse(parsedUrl.query)
+
+  return {
+    method: config.method,
+    originalUrl: parsedUrl.pathname,
+    query: Object.assign(parsedQuery, config.params)
```
srinjoym: I think we were passing `parsedUrl.query` directly in this object for the `query` field when constructing the token before this change. Might be missing context, but is there any reason to change the payload here for the Axios upgrade? Not sure if this could break the token verification if `config.params` adds some extra fields to the `query` which the verification doesn't account for.

I love the idea of breaking this out into a method but maybe we can keep the parameters for the payload the same as before?

lib/jira/client/axios.js
```diff
@@ -1,86 +1,61 @@
...
-    }
-
-    config.fields['baseUrl'] = config.baseURL
+const logResponse = function (id, installationId, response) {
+  const { status, statusText } = response
+  const { method, path } = response.request
 
-    return {
-      ...config,
-      originalUrl: config.url,
```
srinjoym: Transferring @scepticulous's comment from the last PR about why we're removing this; Looks like `UrlMiddleware` was adding an `originalUrl` and `url` field to the request config. Do we not need these anymore with the axios upgrade?

lib/jira/client/index.js
```diff
@@ -62,62 +59,33 @@ module.exports = async (id, installationId, jiraHost) => {
...
-          }
-        }),
-        // eslint-disable-next-line camelcase
-        addForIssue: (issue_id, payload) => instance.post(`/rest/api/latest/issue/:issue_id/comment`, payload, {
-          fields: {
-            issue_id
-          }
-        })
+        getForIssue: (issueId) => instance.get(`/rest/api/latest/issue/${qs.escape(issueId)}/comment`),
+        addForIssue: (issueId, payload) => instance.post(`/rest/api/latest/issue/:${qs.escape(issueId)}/comment`, payload)
```
srinjoym:
```suggestion
        addForIssue: (issueId, payload) => instance.post(`/rest/api/latest/issue/${qs.escape(issueId)}/comment`, payload)
```

lib/jira/client/axios.js
```diff
@@ -1,86 +1,61 @@
...
-
-    return {
-      ...config,
-      headers: {
-        ...config.headers,
-        Authorization: `JWT ${jwtToken}`
-      }
-    }
+    const qsh = jwt.createQueryStringHash(jwt.reqFromAxiosConfig(config))
+    const jwtToken = jwt.encode({ ...getExpirationInSeconds(), iss, qsh }, secret)
```
scepticulous: I suggest using more explicit names instead of abbreviations. looking up `qsh` forced me to read the line above and I still don't know what `iss` is 😆 .
I know, I am lazy, but I would love to hear your thoughts.

`Edit` ... found it at the top of the file. Not sure if it is worth assuming jwt API knowledge just to have the object property shorthand syntax... 🤔

lib/jira/client/index.js
```diff
@@ -62,62 +59,33 @@ module.exports = async (id, installationId, jiraHost) => {
...
-            id: transition_id
-          }
-        }, {
-          fields: {
-            issue_id
-          }
-        })
+        getForIssue: (issueId) => instance.get(`/rest/api/latest/issue/${qs.escape(issueId)}/transitions`),
+        updateForIssue: (issueId, transitionId) => {
+          return instance.post(`/rest/api/latest/issue/${qs.escape(issueId)}/transitions`, { transition: { id: transitionId } })
```
scepticulous: I am not sure I like the idea of performing string interpolation and sanitization at the same time. 
IMO we should either escape before interpolating or (better IMO) let the `instance` object receive an additional parameter that contains the strings to interpolate. This might be a lot of work and maybe it's not worth it...

It seems to be the common approach to have some kind of format-string and arguments, then sanitize the arguments and interpolate them. Just thinking out loud here.

lib/jira/client/axios.js
```diff
@@ -1,86 +1,61 @@
...
 const instance = process.env.INSTANCE_NAME
 const iss = 'com.github.integration' + (instance ? `.${instance}` : '')
 
+/*
+ * Create an Atlassian request from an Axios config.
+ * Compatible with jwt.createQueryStringHash.
+ */
+jwt.reqFromAxiosConfig = (config) => {
+  const parsedUrl = url.parse(config.url)
+  const parsedQuery = querystring.parse(parsedUrl.query)
```
scepticulous: Do we need to parse the query string in the first place? If not, I would opt for backward compatibility and just not doing it at all.
If we need to, 👍

jnraine:
I delayed merging this because it turned out [the Axios upgrade didn't need to happen](https://github.com/integrations/jira/pull/255#issuecomment-526353949). I've since found a reason that generic paths, like `/rest/api/latest/issue/:issue_id`, are useful: it reduces the `path` tag cardinality when sending metrics to Datadog. See https://github.com/integrations/jira/pull/270 for more details.

We'll probably need to upgrade Axios at some point and when that time comes, we'll need to get rid of `fields`. Until then, I'm closing this.

PR #261, opened 2019-08-30
Add ability to uninstall Jira installation via stafftools

jnraine:
We recently discovered a number of Jira installations that had lost the ability to authenticate to the Jira API. This renders them quite useless. Each time we process a webhook for one of these installations, an error is thrown.

A investigation found more than 150 such instances. While it’s not clear how they ended up in that state, it is likely prudent to clean them up. This commit adds the ability to do so using `bin/stafftools`.

```
$ bin/stafftools help jira-uninstall
Usage:
  stafftools jira-uninstall <client_key>

Options:
  [--force]                # Force uninstall for an authorized Jira installation
  [--token=TOKEN]          # Personal access token (defaults to GITHUB_TOKEN)
  [--host=HOST]            # Jira integration host (useful to test against staging or development)
                           # Default: https://jira.github.com
  [--debug], [--no-debug]  # Output full request and response (hepful when an error occurs)

Description:
  Trigger the same uninstall process that occurs when a customer uninstalls the GitHub for Jira Atlassian Marketplace app.

  Note: this does not uninstall the app from a customer's Jira instance, it only removes the data from our side.
```

To protect against accidentally uninstalling a working installation, the endpoint checks whether or not we can talk to the Jira instance. If we can, the uninstall is aborted. For cases where you still need to uninstall, run the command again with `--force`.

### DIFF
--- a/bin/stafftools
+++ b/bin/stafftools
@@ -67,6 +67,30 @@ class JiraStaffTools < Thor
     output_json(response.body)
   end
 
+  desc "jira-uninstall <client_key>", "Uninstall the Jira instance"
+  long_desc <<~DESC
+    Trigger the same uninstall process that occurs when a customer
+    uninstalls the GitHub for Jira Atlassian Marketplace app.
+    
+    Note: this does not uninstall the app from a customer's Jira
+          instance, it only removes the data from our side.
+  DESC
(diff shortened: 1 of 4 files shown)


jnraine:
@scepticulous Yep, tests are a great idea. Once we reduce overall error rates, I'd like to take some time build up testing approaches for the app, including stafftools.

PR #264, opened 2019-09-04
Ignore webhooks from known baddie

jnraine:
There is a Jira installation with a repository that sees 100 push events per hour. Each push event contains 1,000 commits. Right now, this causes an enormous number of GitHub API requests, quickly resulting in rate limit errors.

To stop these errors and keep the Jira integration from wasting time processing these requests, I’ve hard coded the repo ID and ignored it.

**This is a temporary measure.** We’ll reach out to the customer and resolve it ASAP.

### DIFF
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -22,6 +22,16 @@ const withSentry = function (callback) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


jnraine:
The customer is still pushing on the same cadence so I've merged this.

PR #266, opened 2019-09-05
Reduce page limit on branch query

srinjoym:
Reduces the initial page size for the branch query to 20. We have a hunch that the subsequent page sizes do not actually get tried when the request times out since the page limit of 50 does not exceed the node limit ("MAX_NODE_LIMIT_EXCEEDED").

I'll follow up on this PR with more info about the error handling for smaller page sizes after I get a chance to debug later today / tomorrow.

https://github.com/integrations/jira/blob/3a6a26dd2c89d73a028cab7b710e88687ea5a020/lib/sync/installation.js#L114

### DIFF
--- a/lib/sync/queries.js
+++ b/lib/sync/queries.js
@@ -86,7 +86,7 @@ module.exports = {
                   name
                 }
                 authoredDate
-                history(first: $per_page) {
+                history(first: 50) {
                   nodes {
                     message
                     oid
--- a/lib/sync/installation.js
+++ b/lib/sync/installation.js
@@ -111,7 +111,7 @@ module.exports.processInstallation = (app, queues) => {
     }
 
     const execute = async () => {
-      for (const perPage of [50, 10, 5, 1]) {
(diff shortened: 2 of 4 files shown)


lib/sync/queries.js
```diff
@@ -86,7 +86,7 @@ module.exports = {
                   name
                 }
                 authoredDate
-                history(first: $per_page) {
+                history(first: 50) {
```
jnraine: Minor: how about pulling this value out into a variable so it's more clear what it represents?

```suggestion
                history(first: numberOfCommitsPerBranch) {
```

srinjoym: 👋 Good idea! Not sure how to do this without passing a variable into the query which would make this into a bigger change because we'd have to pass this variable in every spot we make this query. I think we can keep this hardcoded for now (I've seen dotcom queries like this as well) and then revisit the issue in the follow-up PR when debugging the page size retries?

PR #267, opened 2019-09-06
Supercharge Octokit request errors

jnraine:
When a request from Octokit to the GitHub API fails, this wraps the error, adding additional context and ensuring the grouping is more meaningful.

```
HttpError: {“message":"API rate limit exceeded for installation ID 1339471.","documentation_url":"https://developer.github.com/v3/#rate-limiting"}
```

becomes

```
OctokitError: PATCH /repos/:owner/:repo/issues/:number responded with 403
```

The response from GitHub—`API rate limit exceeded…`—is available in full under the `response.body` metadata. The GitHub request ID can be found under `response.headers` metadata.

Errors are also now grouped by method, path, and response code. This should ensure better grouping in Sentry.

### DIFF
--- a/lib/frontend/github-client-middleware.js
+++ b/lib/frontend/github-client-middleware.js
@@ -20,7 +20,6 @@ module.exports = (getAppToken) => {
     })
 
     res.locals.client = appClient
-
     res.locals.isAdmin = isAdminFunction
 
     next()
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -1,6 +1,8 @@
 const Sentry = require('@sentry/node')
 
 const AxiosErrorEventDecorator = require('../models/axios-error-event-decorator')
+const OctokitError = require('../models/octokit-error')
+const SentryScopeProxy = require('../models/sentry-scope-proxy')
 const { Subscription } = require('../models')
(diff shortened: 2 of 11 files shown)


jnraine:
> Is there no place we can add the OctokitError.wrapRequestErrors call so we only call it once?

@jules2689 I know, right?

Because we rely on Probot to instantiate the Octokit client for us on each request or job, I couldn't find a great place do it just once. Some things I explored:

- **[Octokit plugins](https://octokit.github.io/rest.js/#plugins).** In theory, this would be a great plugin. However, plugins must be added _before_ the client is instantiated. Because this happens [deep within Probot](https://github.com/probot/probot/blob/91cf5b0d367c56cc31e548425c025373f5f19b49/src/github/index.ts#L15), plugins are out. (This is further complicated because we use a version of Octokit that doesn't support plugins.)
- **Extracting a helper function to be used across all jobs.** This would reduce the number of places we do the same thing but I decided against it because I don't think it would make the code better, just more hidden. There is at least one object dying to be extracted from the jobs. Once we do that, the sharing becomes much easier.

jules2689:
Sounds like you made the right call then @jnraine ❤️ ✨ Great job adding all this contextual debuggability.

PR #268, opened 2019-09-07
Add Datadog histogram for GitHub requests

jnraine:
When a request is made to the GitHub API using Octokit, a timing histogram is sent to localhost:8125 over UDP. The value of the metric is the elapsed time of the request. Because we’re using a histogram, Datadog will turn this into six different metrics: `count`, `avg`, `median`, `95percentile`, `max`, `min`. For more information on Datadog histograms, see https://docs.datadoghq.com/developers/metrics/histograms/.

This metric will help us measure our interactions with the GitHub API to address timeout issues, monitor usage, and optimize GraphQL queries.

To test this out locally:

1. Start the server
2. In a new tab run `script/datadog-spy`
3. Trigger a sync

You should see metrics printed in the second tab.

Note: Because we don’t have a Datadog agent running on production, _these won't go anywhere yet._ That’s okay! This acts as the first step to adding meaning metrics.

### DIFF
--- a/lib/config/enhance-octokit.js
+++ b/lib/config/enhance-octokit.js
@@ -1,0 +1,29 @@
+const OctokitError = require('../models/octokit-error')
+const statsd = require('./statsd')
+
+const instrumentRequests = (octokit, log) => {
+  octokit.hook.wrap('request', async (request, options) => {
+    const requestStart = Date.now()
+    try {
+      const response = await request(options)
+      return response
+    } finally {
+      const elapsed = Date.now() - requestStart
+      statsd.histogram('github-request', elapsed, { path: options.url })
+      log.debug(`GitHub request time: ${elapsed}ms`)
+    }
+  })
+}
+
+/*
+ * Customize an Octokit instance behavior.
+ *
+ * This acts like an Octokit plugin but works on Octokit instances.
+ * (Because Probot instantiates the Octokit client for us, we can't use plugins.)
+ */
+module.exports = (octokit, log) => {
+  OctokitError.wrapRequestErrors(octokit)
+  instrumentRequests(octokit, log)
+
+  return octokit
+}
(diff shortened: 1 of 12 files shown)


lib/config/statsd.js
```diff
@@ -0,0 +1,9 @@
+const StatsD = require('hot-shots')
+
+module.exports = new StatsD({
+  prefix: 'jira-integration.',
+  globalTags: {
+    env: process.env.NODE_ENV || 'unknown',
```
srinjoym: What happens if this defaults to 'unknown'? Would this throw an exception then? Might be worth wrapping this in a `try` or something like that since we don't want to throw exceptions on logging

lib/config/enhance-octokit.js
```diff
@@ -0,0 +1,29 @@
...
+const statsd = require('./statsd')
+
+const instrumentRequests = (octokit, log) => {
+  octokit.hook.wrap('request', async (request, options) => {
+    const requestStart = Date.now()
+    try {
+      const response = await request(options)
+      return response
+    } finally {
+      const elapsed = Date.now() - requestStart
```
srinjoym: Can we add a unit test for this to ensure the request timing is getting sent to Datadog? Maybe as a start, we can just make sure `statds.histogram` gets called?

lib/config/statsd.js
jnraine: If `process.env.NODE_ENV ` isn't set, it will send `env:unknown` to Datadog. No exceptions will be raised.

lib/config/enhance-octokit.js
jnraine: Phew, several hours later, we now have [`toHaveSentMetrics`](https://github.com/integrations/jira/pull/268/files#diff-f65ac65e4d6e45df7914e615b009d4c2). This matcher should help us as we add metrics in the future.

I've added a basic test to ensure the histogram is sent as expected.

PR #272, opened 2019-09-12
Reduce cardinality of path tag

jnraine:
Yesterday, we connected production to Datadog and started seeing metrics! 🎉 Unfortunately, we quickly realized that there were far too many values being sent for the `path` tag. 😞 

image

To avoid this, I've modified the Jira Axios client to always use Express.js-like route parameters to interpolate user data into paths and to remove URL query parameters. This should reduce the total number of `path` values we send to Datadog and save us 💰.

✨ Props to @jules2689 for [warning of this problem](https://github.com/integrations/jira/pull/269#pullrequestreview-286245560) before it happened!

### DIFF
--- a/lib/config/enhance-octokit.js
+++ b/lib/config/enhance-octokit.js
@@ -1,5 +1,6 @@
 const OctokitError = require('../models/octokit-error')
 const statsd = require('./statsd')
+const { extractPath } = require('../jira/client/axios')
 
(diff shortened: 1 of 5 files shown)


jules2689:
Looks great. Sorry I didn't take a look sooner (I did, but apparently forgot to press "send" on my approve 🙃)

PR #274, opened 2019-09-16
Swallow repo NOT_FOUND error during Initial Sync

srinjoym:
**Problem:** If a user deletes a repo after starting a long-running sync the entire sync will fail.

**Context:** The repo sync state column is populated at the beginning of the sync in the discovery queue. If the sync runs over multiple days, a user can delete a repo that was initially in the sync state column. This causes the entire sync to fail since the sync job hits a `NOT_FOUND` error from GitHub when it tries to process that repo's sync state.

**Fix:** If the repo no longer exists, we should just be able to swallow the error and not fail the sync.

### DIFF
--- a/lib/sync/installation.js
+++ b/lib/sync/installation.js
@@ -114,7 +114,7 @@ module.exports.processInstallation = (app, queues) => {
         try {
           return await pagedProcessor(perPage)
         } catch (err) {
-          processJobGitHubError(err)
+          handleGitHubError(err)
         }
       }
       throw new Error(`Error processing GraphQL query: installationId=${installationId}, repositoryId=${repositoryId}, task=${task}`)
--- a/lib/sync/installation.js
+++ b/lib/sync/installation.js
@@ -103,13 +103,10 @@ module.exports.processInstallation = (app, queues) => {
     }
 
     const handleGitHubError = (err) => {
-      const ignoredErrors = [
-        'MAX_NODE_LIMIT_EXCEEDED',
-        'NOT_FOUND'
-      ]
-      const isIgnoredError = ignoredErrors.find(ignoredError => String(err).includes(ignoredError))
+      const ignoredErrorTypes = ['MAX_NODE_LIMIT_EXCEEDED']
+      const isIgnoredError = err.errors.filter(error => ignoredErrorTypes.includes(error.type)).length
 
-      if (isIgnoredError) {
+      if (!isIgnoredError) {
         throw (err)
       }
     }
@@ -174,6 +171,12 @@ module.exports.processInstallation = (app, queues) => {
         app.log(`Abuse detection triggered. Retrying in 60 seconds: installationId=${installationId}, repositoryId=${repositoryId}, task=${task}`)
         const { removeOnComplete, removeOnFail } = job.opts
         queues.installation.add(job.data, { delay: 60000, removeOnComplete, removeOnFail })
+        return
+      }
+      // Checks if parsed error type is NOT_FOUND: https://github.com/octokit/graphql.js/tree/master#errors
+      const isNotFoundError = err.errors.filter(error => error.type === 'NOT_FOUND').length
+      if (isNotFoundError) {
+        app.log(`Repository deleted after discovery, skipping initial sync: installationId=${installationId}, repositoryId=${repositoryId}, task=${task}`)
         return
       }
       throw err
--- a/lib/sync/installation.js
+++ b/lib/sync/installation.js
@@ -104,9 +104,9 @@ module.exports.processInstallation = (app, queues) => {
 
     const handleGitHubError = (err) => {
       const ignoredErrorTypes = ['MAX_NODE_LIMIT_EXCEEDED']
-      const isIgnoredError = err.errors.filter(error => ignoredErrorTypes.includes(error.type)).length
+      const notIgnoredError = err.errors.filter(error => !ignoredErrorTypes.includes(error.type)).length
 
-      if (!isIgnoredError) {
+      if (notIgnoredError) {
         throw (err)
       }
     }
--- a/lib/sync/installation.js
+++ b/lib/sync/installation.js
@@ -102,7 +102,7 @@ module.exports.processInstallation = (app, queues) => {
       return processor(github, repository, cursor, perPage)
     }
 
-    const processJobGitHubError = (err) => {
+    const handleGitHubError = (err) => {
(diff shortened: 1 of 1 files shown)


lib/sync/installation.js
```diff
@@ -102,14 +102,19 @@ module.exports.processInstallation = (app, queues) => {
       return processor(github, repository, cursor, perPage)
     }
 
+    const processJobGitHubError = (err) => {
+      if (!String(err).includes('MAX_NODE_LIMIT_EXCEEDED') ||
```
jnraine: This code behaves the way we want but is dense<sup>1</sup>. If I were to explain this to someone, I'd say it something like this: "swallow ignored errors, otherwise throw"

Can we make the code say that too? Here's an example refactoring:

```javascript
const processJobGitHubError = (err) => {
  const ignoredErrors = [
    'MAX_NODE_LIMIT_EXCEEDED',
    'NOT_FOUND'
  ]

  const isIgnoredError = ignoredErrors.find(ignoredError => String(err).includes(ignoredError))

  if (!isIgnoredError) {
    throw err
  }
}
```

---

1: "Dense" is ambiguous. Let me clarify what I mean. When a single statement/expression contains more than one "thing", the density goes up. Lines became hard to understand at 2-3 things. For this `if` statement, I count 3 things: how to check if an error is ignored, what errors are ignored, what we do when an error is ignored. The `||` and two `!` make it denser yet.

lib/sync/installation.js
```diff
@@ -102,14 +102,19 @@ module.exports.processInstallation = (app, queues) => {
...
+
     const execute = async () => {
       for (const perPage of [20, 10, 5, 1]) {
         try {
           return await pagedProcessor(perPage)
         } catch (err) {
-          if (!String(err).includes('MAX_NODE_LIMIT_EXCEEDED')) {
-            throw err
-          }
+          processJobGitHubError(err)
```
jnraine:
```suggestion
          handleGitHubError(err)
```

Two reasons:

1. This function is defined and only used within the context of a job so `Job` seems superfluous to me.
2. `handle` feels closer to what we're doing here: take an error, decide how handle it (throw or ignore). I use _feel_ intentionally. This is a matter of opinion and if you like `process` better, please do keep it! 😄

lib/sync/installation.js
```diff
@@ -102,14 +102,19 @@ module.exports.processInstallation = (app, queues) => {
       return processor(github, repository, cursor, perPage)
     }
 
+    const processJobGitHubError = (err) => {
+      if (!String(err).includes('MAX_NODE_LIMIT_EXCEEDED') ||
+          !String(err).includes('NOT_FOUND')) {
```
jnraine: I would **love** to see references to why these magic strings matter.

I've confirmed that both are errors sent by the GitHub API so a note about them or a link to docs about them would be 💯 .

lib/sync/installation.js
```diff
@@ -102,14 +102,24 @@ module.exports.processInstallation = (app, queues) => {
...
     }
 
+    const handleGitHubError = (err) => {
+      const ignoredErrors = [
+        'MAX_NODE_LIMIT_EXCEEDED',
+        'NOT_FOUND'
+      ]
+      const isIgnoredError = ignoredErrors.find(ignoredError => String(err).includes(ignoredError))
+
+      if (isIgnoredError) {
```
jules2689: Excuse me if I'm reading this wrong, but aren't we inverting the logic?

Before it was `!err.include?(MAX_EXCEEDED)` which says `isNotIgnoredError` `err.include?(MAX_EXCEEDED) || err.include?(NOT_FOUND)` which is `isIgnoredError`.

It's a bit confusing if the way it's written is intended because it's `if ignore this error then throw the error`

srinjoym: @jules2689 Great point, I just refactored this PR in general because the check was in the wrong place but I still tried to clean up the wording. The latest version filters through all of the errors and checks if any of them has a "non-ignored" error type, and then throws the error. 

https://github.com/integrations/jira/blob/a56843186404eb1f552d155a0643fe89b0dd5b28/lib/sync/installation.js#L105-L112

lib/sync/installation.js
```diff
@@ -102,14 +102,21 @@ module.exports.processInstallation = (app, queues) => {
       return processor(github, repository, cursor, perPage)
     }
 
+    const handleGitHubError = (err) => {
+      const ignoredErrorTypes = ['MAX_NODE_LIMIT_EXCEEDED']
+      const notIgnoredError = err.errors.filter(error => !ignoredErrorTypes.includes(error.type)).length
```
jules2689: I'm not gonna block, but I wanted to point out one thing to keep in mind 😄 

You're using a `!` inside the filter and a `not` in the var name... all the nots make it more difficult to parse. I get what you're doing here, and it's limited scope, but be careful in a larger setting ✨

srinjoym: 👋 Thanks for bringing this up @jules2689 and @jnraine! Totally understand the confusion here and I can see how the naming is not ideal. 

Since there can be multiple errors on the response (at least there are on the Axios return object) I filter the errors to check if we have **any** error type that we shouldn't be ignoring. I figured we don't want to swallow the error if there is a different error with a non-ignored error type. If I invert the variable then we'd swallow the error regardless of any of the other error types.

I'm building off this PR and this section of the codebase to catch timeout errors in another upcoming PR so I'll revisit it again to see if there's a better way to name this 😆

lib/sync/installation.js
```diff
@@ -166,6 +173,12 @@ module.exports.processInstallation = (app, queues) => {
         queues.installation.add(job.data, { delay: 60000, removeOnComplete, removeOnFail })
         return
       }
+      // Checks if parsed error type is NOT_FOUND: https://github.com/octokit/graphql.js/tree/master#errors
+      const isNotFoundError = err.errors.filter(error => error.type === 'NOT_FOUND').length
+      if (isNotFoundError) {
+        app.log(`Repository deleted after discovery, skipping initial sync: installationId=${installationId}, repositoryId=${repositoryId}, task=${task}`)
```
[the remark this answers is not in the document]
srinjoym: Thank you so much for the detailed comments here! 🎉  I really like the idea of passing in metadata as the first argument 😄 Also totally agree that we should be using an explicit log level, I just updated this to use `app.log.info` ✨ 

For this `app.log` statement I mirrored what we were doing in the other `if-then` blocks above this line to keep them consistent. I'll follow up with another PR to change all of these calls to pass the metadata first after we merge this in.

PR #275, opened 2019-09-17
WIP Instrumenting Failed and Errored jobs

zmoazeni:
This is mega-rough, but submitting a WIP draft to iterate on.

This is currently breaking the test suite. It also doesn't have any automated tests. May need to restructure the code in order to allow this to be easily tested in a handler-agnostic way.

### DIFF
--- a/lib/worker/index.js
+++ b/lib/worker/index.js
@@ -15,6 +15,8 @@ const { Subscription } = require('../models')
 const AxiosErrorEventDecorator = require('../models/axios-error-event-decorator')
 const SentryScopeProxy = require('../models/sentry-scope-proxy')
 
+const statsd = require('../config/statsd')
+
 // Setup queues
 const queues = {
   discovery: new Queue('Content discovery', REDIS_URL),
@@ -39,6 +41,7 @@ Object.keys(queues).forEach(name => {
 
     Sentry.setTag('queue', name)
     Sentry.captureException(err)
+    statsd.increment('job-error')
   })
 
   queue.on('failed', async (job, err) => {
@@ -46,6 +49,7 @@ Object.keys(queues).forEach(name => {
 
     const subscription = await Subscription.getSingleInstallation(job.data.jiraHost, job.data.installationId)
     await subscription.update({ syncStatus: 'FAILED' })
+    statsd.increment('job-failed')
   })
 })
 


lib/worker/index.js
```diff
@@ -39,13 +41,15 @@ Object.keys(queues).forEach(name => {
...
     Sentry.captureException(err)
+    statsd.increment('job-error')
   })
 
   queue.on('failed', async (job, err) => {
     app.log.error(`Error occurred while processing job id=${job.id} on queue name=${name}`)
 
     const subscription = await Subscription.getSingleInstallation(job.data.jiraHost, job.data.installationId)
     await subscription.update({ syncStatus: 'FAILED' })
+    statsd.increment('job-failed')
```
zmoazeni: I'm also wondering if there are any appropriate tags to pass across. But I'm not sure yet if there are.

lib/worker/index.js
```diff
@@ -39,13 +41,15 @@ Object.keys(queues).forEach(name => {
 
     Sentry.setTag('queue', name)
     Sentry.captureException(err)
+    statsd.increment('job-error')
```
jnraine: Metric names is always a bit up in the air but for the first few in this app, I've followed this pattern: `<prefix>.<area/concern>`. This has led to metric names like:

- `jira_integration.github_requests.95th`
- `jira_integration.jira_requests.95th`

In this same vein, what do you think about creating a `jobs` namespace and adding states beneath that?

- `jira_integration.jobs.started`
- `jira_integration.jobs.completed`
- `jira_integration.jobs.errored`
- etc

zmoazeni:
> One stretch goal would be to use histogram instead of increment. In addition to counts, this would capture timing data for each job.

Yeah totally. I was planning on doing that in a separate effort. One metric for histogram/timing information for all jobs. A different metric for the number of errors and failures we encounter.

But I have no problem storing the number of failures/errors as a histogram. That'll give us both timing for just failures/errors as well as counts.

jnraine:
>  I was planning on doing that in a separate effort.

👍 Shipping the counts now and the histograms later sounds good.

My one recommendation would be to use the same count key as the histogram:

```javascript
statsd.increment('jira_integration.jobs.started.count')
# then later
statsd.histogram('jira_integration.jobs.started')
```

While this isn't required, it ensures the metric name doesn't change when the histogram work is done.

You may have already considered recommendation. If so, carry on! 😄

zmoazeni:
> You may have already considered recommendation.

Oh I didn't! Good idea!

zmoazeni:
This is all set as of https://github.com/integrations/jira/pull/312

PR #276, opened 2019-09-18
Fix non-default branch commit history sync + Batch commit payloads to Jira

srinjoym:
Fixes https://github.com/github/integrations-jira-internal/issues/29

**Problem:** We only process one commit for each non-default branch instead of processing all 50. The key issue is in this line: https://github.com/integrations/jira/blob/0cb6270d7c7dea71c0a188013ddcbbadaa370452/lib/sync/transforms/branch.js#L94-L96

The `[variable]` syntax destructures the array returned and only takes the first commit.

The Jira API also limits us to sending 400 commits per request during the initial sync. As a result, we have to batch the commits and send multiple requests to Jira.

**Solution:** This PR removes the destructure and processes all 50 commits on non-default branches: https://github.com/integrations/jira/blob/95cc82e9efbffe9b77d46428028659b079c97603/lib/sync/transforms/branch.js#L94-L96

 In the Jira client we now automatically batch requests with 400 commit chunks: https://github.com/integrations/jira/blob/95cc82e9efbffe9b77d46428028659b079c97603/lib/jira/client/index.js#L144-L147

This will likely need to be updated in another future PR to support batching across other fields that are limited.


lib/jira/client/index.js
```diff
@@ -126,16 +126,20 @@ module.exports = async (jiraHost, gitHubInstallationId, logger) => {
...
           }
 
-          await instance.post('/rest/devinfo/0.10/bulk', {
-            preventTransitions: (options && options.preventTransitions) || false,
-            repositories: [data],
-            properties: {
-              installationId: gitHubInstallationId
-            }
-          })
+          if (data.commits && data.commits.length >= 400) {
```
jules2689: I would recommend putting the `400` in a constant somewhere, otherwise it's a magic comment with only this PR as reference as to what it means :)

lib/jira/client/index.js
```diff
@@ -144,6 +148,36 @@ module.exports = async (jiraHost, gitHubInstallationId, logger) => {
...
+  }
+
+  const batchedUpdates = []
+  commitChunks.forEach((commitChunk) => {
+    const repositoryObj = {
+      ...data,
+      commits: commitChunk
+    }
+
+    batchedUpdates.push(instance.post('/rest/devinfo/0.10/bulk', {
```
jules2689: the instance.post is near identical to the one we see above. The only difference is `respositoryObj` instead of `data`. I would be concerned these could drift. 

One question is ... when would we not have `data.commits`? Can we otherwise make these the same path somehow?

lib/jira/client/index.js
```diff
@@ -144,6 +148,36 @@ module.exports = async (jiraHost, gitHubInstallationId, logger) => {
   return client
 }
 
+/**
+ * Splits commits in data payload into chunks of 400 and makes separate requests
+ * to avoid Jira API limit
+ */
+const batchedCommitRequest = async (data, options, instance, installationId) => {
```
jnraine:
```suggestion
const batchedBulkUpdate = async (data, options, instance, installationId) => {
```

After reading this code, I can't help but think the original name is under selling the utility of this method.

lib/jira/client/index.js
```diff
@@ -152,28 +142,27 @@ module.exports = async (jiraHost, gitHubInstallationId, logger) => {
  * Splits commits in data payload into chunks of 400 and makes separate requests
  * to avoid Jira API limit
  */
-const batchedCommitRequest = async (data, options, instance, installationId) => {
-  const filteredCommits = dedupCommits(data)
+const batchedBulkUpdate = async (data, options, instance, installationId) => {
+  const dedupedCommits = data.commits ? dedupCommits(data.commits) : []
```
jnraine: What do you think about pushing this conditional into `dedupCommits`? It would mean `dedupCommits` accepts either an empty array or a null value. I favour that only because it pushes some complexity out of this function, which already does a fair bit. Within `dedupCommits`, this case can be handled elegantly:

```javascript
const dedupCommits = (commits) => {
  return (commits || []).filter((obj, pos, arr) => {
    // ...
  })
}
```

This is a matter of preference so I'd encourage you to pick whichever approach you find most readable and maintainable (even if it differs from my suggestion).

lib/jira/client/index.js
```diff
@@ -152,28 +142,27 @@ module.exports = async (jiraHost, gitHubInstallationId, logger) => {
...
 
-  const commitChunks = []
-  while (filteredCommits.length) {
-    commitChunks.push(filteredCommits.splice(0, 400))
+  // Initialize with an empty chunk of commits so we still process the request if there are no commits in the payload
+  const commitChunks = [[]]
+  while (dedupedCommits.length) {
+    commitChunks.push(dedupedCommits.splice(0, 400))
   }
```
jnraine: Since we always start with an empty commit chunk, does that mean any bulk update that includes even a single commit will make at least two requests to the Jira API? 

If that's the case, this may be a good fit for a `do ... while` block, which will always run at least once.

```javascript
const commitChunks = []
do {
  commitChunks.push(dedupedCommits.splice(0, 400))
} while (dedupedCommits.length)
```

This should handle the "no commits" case without adding an empty array in scenarios where commits are present.

lib/jira/client/index.js
```diff
@@ -152,28 +142,27 @@ module.exports = async (jiraHost, gitHubInstallationId, logger) => {
...
+  const batchedUpdates = commitChunks.map((commitChunk) => {
+    if (commitChunk.length) {
+      data.commits = commitChunk
     }
 
-    batchedUpdates.push(instance.post('/rest/devinfo/0.10/bulk', {
+    return instance.post('/rest/devinfo/0.10/bulk', {
       preventTransitions: (options && options.preventTransitions) || false,
-      repositories: [repositoryObj],
+      repositories: [data],
```
jnraine: In it's current state, we'll resend any data outside of the commits when we exceed 400 commits. That raises two questions in my mind:

1. Is it problematic to send certain data twice? This one we need to answer before shipping this. I suspect it isn't a problem but it's worth checking.
2. Is there some data we can remove from subsequent requests? For example, on subsequent requests, can we drop `pullRequests` and `branches`, if present? This one we can punt to a later PR.

lib/sync/installation.js
```diff
@@ -107,9 +107,9 @@ module.exports.processInstallation = (app, queues) => {
 
     const handleGitHubError = (err) => {
       const ignoredErrorTypes = ['MAX_NODE_LIMIT_EXCEEDED']
-      const notIgnoredError = err.errors.filter(error => !ignoredErrorTypes.includes(error.type)).length
+      const isIgnoredError = err.errors && err.errors.filter(error => ignoredErrorTypes.includes(error.type)).length
```
jnraine: I like that we're asserting positives. Easier on my brain. 😄 To your earlier point, does this change cause us to swallow not ignored errors when they are accomanied by an ignored one?

For example, if `err.errors == [{type: "MAX_NODE_LIMIT_EXCEEDED"}, {type: "Some other error"}]`.

srinjoym: Good point! I had originally updated this in an attempt to fix the CI failures, just reverted this change for this PR so we don't change the functionality here.

lib/jira/client/index.js
srinjoym: The description for the field in the docs is:

> List of commits to update in this repository. Must not contain duplicate entity IDs. Maximum number of commits is 400

I'm reasonably confident that we won't run into any issues since it only "updates" the list of commits when sending multiple requests as long as we use the same repository `id` for each request. I think it's a good idea to drop `pullRequests` and `branches` in a follow up (possibly in addition to adding support for batching other entities)

lib/jira/client/index.js
srinjoym: Fantastic idea ✨ Just implemented this and it's super clean

PR #278, opened 2019-09-20
Clean up failed and completed jobs on start

jnraine:
The background job library we use doesn’t consistently remove completed jobs from Redis. On production, this led to unbounded memory growth.

While the [Bull reference guide] doesn’t specifically mention the need to manual clean old data, the library prov ides [`Queue#clean`]. Prior to this commit, we manually cleaned the queues by running `bin/clean-old-jobs`. ~To ensure we clean on a regular basis and to minimze chores, **old jobs are now cleaned on start up.**~ To decouple clean up from restarting the workers, old jobs will be cleaned up using the [Heroku Scheduler](https://devcenter.heroku.com/articles/scheduler). (Hat tip: @jules2689)

Cleaning has also been limited to 50,000 jobs at a time. Doing too many at once can overload Redis. The specific number is a guess based on past experience: cleaning 500,000 jobs was too much for Redis.

[Bull reference guide]: https://optimalbits.github.io/bull/
[`Queue#clean`]: https://optimalbits.github.io/bull/

### DIFF
--- a/lib/worker/index.js
+++ b/lib/worker/index.js
@@ -82,17 +82,21 @@ module.exports = {
     queues.discovery.process(5, sentryMiddleware(discovery(app, queues)))
     queues.installation.process(Number(CONCURRENT_WORKERS), sentryMiddleware(processInstallation(app, queues)))
     queues.push.process(Number(CONCURRENT_WORKERS), sentryMiddleware(limiterPerInstallation(processPush(app))))
+
     app.log(`Worker process started with ${CONCURRENT_WORKERS} CONCURRENT WORKERS`)
   },
 
   async clean () {
+    const gracePeriod = 10000
+    const limit = 50000
+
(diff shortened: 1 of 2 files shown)


zmoazeni:
Also def a hat-tip to @jules2689's suggestion on just doing this regularly. Solid suggestion!

jnraine:
@zmoazeni ah, `removeOnComplete` is perfect. It’s smaller than fun-sized but would be an easy starting PR. Would you be up to adding that? 

From what I’ve seen, memory use will likely stay below 50 MB now that we’re regularly cleaning out old jobs. Once we add `removeOnComplete`, it should stay even lower.

zmoazeni:
Yeah man. Totally not necessary for this PR.

Sounds like a great fun sized issue! I'll spin up one and assign it to myself.

PR #279, opened 2019-09-23
Send sync status metrics to Datadog

jnraine:
Every n minutes<sup>1</sup>, send over the count of sync statuses to Datdog. In other words, we want this query available as a Datadog metric:

```
psql> SELECT "syncStatus", COUNT(*) FROM "Subscriptions" GROUP BY "syncStatus";
 syncStatus | count
------------+-------
            | 5
 PENDING    | 1
 COMPLETE   | 100
 ACTIVE     | 10
 FAILED     | 5
(5 rows)
```

This adds `bin/send-sync-status-metrics` that runs the above query and sends it along to Datadog under the `jira_integration.syncs` metric.

```
jira_integration.syncs:5|g|#status:
jira_integration.syncs:1|g|#status:PENDING
jira_integration.syncs:100|g|#status:COMPLETE
jira_integration.syncs:10|g|#status:ACTIVE
jira_integration.syncs:5|g|#status:FAILED
```

This will allow us to track syncs over time and proactively track the health of syncs. My hope is that this can focus our attention on driving the total number of failed syncs to zero.

---

1: Once deployed, I plan to add a [Heroku scheduler](https://devcenter.heroku.com/articles/scheduler) job that sends the metrics every ten minutes. This is the most frequent interval possible with that add-on. I'm working to install the [Advanced Scheduler](https://elements.heroku.com/addons/advanced-scheduler) add-on, which will allow this metric to be collected every minute.

### DIFF
--- a/bin/send-sync-status-metrics
+++ b/bin/send-sync-status-metrics
@@ -1,0 +1,29 @@
+#!/usr/bin/env node
+require('dotenv').config()
+
+const Sentry = require('@sentry/node')
+
+Sentry.init({
+  dsn: process.env.SENTRY_DSN,
+  environment: process.env.SENTRY_ENV || process.env.NODE_ENV,
+  release: process.env.HEROKU_SLUG_COMMIT
+})
+
+const statsd = require('../lib/config/statsd')
+const { Subscription } = require('../lib/models')
+
+Subscription.syncStatusCounts().then((results) => {
+  results.forEach((row) => {
+    statsd.gauge('syncs', row.count, { status: row.syncStatus })
+  })
+
+  // Close client, triggering flush of metrics, then exit.
+  // (Without exit, process hangs. Without close, metric is never sent.)
+  statsd.close(() => {
+    process.exit()
+  })
+}).catch((error) => {
+  console.log('An error occurred while sending sync status metrics:', error)
+  Sentry.captureException(error)
+  process.exit(1)
+})
--- a/lib/models/subscription.js
+++ b/lib/models/subscription.js
@@ -109,6 +109,16 @@ module.exports = class Subscription extends Sequelize.Model {
     return queues.installation.add({ installationId, jiraHost })
   }
 
+  /*
+   * Returns array with sync status counts. [ { syncStatus: 'COMPLETED', count: 123 }, ...]
+   */
+  static async syncStatusCounts () {
+    const syncStatusCountQuery = 'SELECT "syncStatus", COUNT(*) FROM "Subscriptions" GROUP BY "syncStatus"'
+    const [ results ] = await this.sequelize.query(syncStatusCountQuery)
+
+    return results
+  }
+
   async uninstall () {
     return this.destroy()
   }
--- a/test/unit/config/enhance-octokit.test.js
+++ b/test/unit/config/enhance-octokit.test.js
@@ -26,7 +26,7 @@ describe(enhanceOctokit, () => {
         }).toHaveSentMetrics({
           name: 'jira-integration.github-request',
           type: 'h',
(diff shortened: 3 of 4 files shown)


bin/send-sync-status-metrics
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env node
+require('dotenv').config()
+
+const Sentry = require('@sentry/node')
+
+Sentry.init({
```
zmoazeni: I see this init copy/pasted in two different places. Not a blocker for this effort, but I think it'd be great for us to abstract this out.

Perhaps something like:

```node
const { Sentry, initializeSentry } = require("sentry"); # our file
initializeSentry();
```

That way the other areas only need to change to:

```node
const { Sentry } = require("sentry");
```

_I don't feel that attached to that suggestion if there's a better node-ism._

lib/models/subscription.js
```diff
@@ -109,6 +109,16 @@ module.exports = class Subscription extends Sequelize.Model {
     return queues.installation.add({ installationId, jiraHost })
   }
 
+  /*
+   * Returns array with sync status counts. [ { syncStatus: 'COMPLETED', count: 123 }, ...]
+   */
+  static async syncStatusCounts () {
+    const syncStatusCountQuery = 'SELECT "syncStatus", COUNT(*) FROM "Subscriptions" GROUP BY "syncStatus"'
+    const [ results ] = await this.sequelize.query(syncStatusCountQuery)
```
zmoazeni: Hrm, would it be better to do something like this?

```suggestion
    return this.sequelize.query(syncStatusCountQuery)
```

The only consumer right now uses it as a promise with the `then()` syntax (which is great!) but I don't think we need to `await` here do we?

jnraine:
> By the way, I've noticed you have posted these metrics in other PRs:
> 
> > jira_integration.syncs:5|g|#status:
> 
> I think that's great! How do you print that out? I'd like to do the same in my PRs.

`script/datadog-spy` listens for statsd metrics and prints them to the console. It's a quick way to test that things are working as expected.

There's one quirk: because [`hot-shots`](https://www.npmjs.com/package/hot-shots) doesn't end messages with `\n`, everything gets printed to one line. 🤦‍♂ If you figure out a way to inject a newline between metrics, I'd be in your debt!

bin/send-sync-status-metrics
jnraine: This bugged me too. [Previously](https://github.com/integrations/jira/pull/239), we did something similar but it led to initializing the client multiple times, causing duplicate errors.

Your suggested approach makes initialization more obvious and seems more likely to be used appropriately! I'll make a follow-up PR for that.

lib/models/subscription.js
jnraine: There's some subtly here.

`sequelize.query` returns an array containing two elements. We only want the first one. In order to only return that value, we need to await the query, extract the first element and return it.

I wasn't sure how to make it more obvious that there was multiple items in the array. When I made it explicit, the style checker yelled at me. 😄 

image

If you have a more obvious syntax, I'm eager for it!

zmoazeni: Ahh I see now. All good then!

zmoazeni:
> There's one quirk: because hot-shots doesn't end messages with \n, everything gets printed to one line. 🤦‍♂ If you figure out a way to inject a newline between metrics, I'd be in your debt!

I just ran into this from another scenario. I found https://stackoverflow.com/a/24574563 and created a bash script in my PATH:

```bash
#!/bin/bash

# Found from https://stackoverflow.com/a/24574563
# Convert "\n" literals into new lines

awk '{gsub(/\\n/,"\n")}1'
```

I run it like `some_command_that_outputs_literals | literal_to_newline`

PR #282, opened 2019-10-02
Instrument response only when present

jnraine:
When an error is thrown while making a request with to the Jira API, there’s a good chance the error has a response attached to it. However, that’s not always the case.

Because the instrumentation relies on the response property, we now only call `instrumentRequest` when it is present.

This issue was raised because we saw a number of `Cannot read property 'config' of undefined` errors.

It’s not immediately obvious to me how to add an automated test to trigger this case. If you can think of one, please let me know.

### DIFF
--- a/lib/jira/client/axios.js
+++ b/lib/jira/client/axios.js
@@ -127,7 +127,12 @@ const instrumentRequest = (response) => {
 }
 
 const instrumentFailedRequest = (error) => {
-  instrumentRequest(error.response)
+  if (error.response) {
+    instrumentRequest(error.response)
+  } else {
+    console.log('Error during Axios request has no response property:', error)
+  }
+
   return Promise.reject(error)
 }
 


jnraine:
> What's an example of when would it not have a response?

I wish I knew. It's occurred about 100 times so far but the specific cause isn't clear. With some spelunking in the Axios codebase, we could find the answer but because of how few times it is happening, I'm happy to plug the hole and 🚢 .

jules2689:
I like the statsd counter. Alternative is a sentry "exception" report that doesn't raise.

jnraine:
Thanks for the suggestions, @zmoazeni @jules2689!

I've opted to go with a log message. It is infrequent enough that it won't create log noise and I like that it can tell us what kind of error doesn't have the response object.

Since this happens only when an error occurs, it will end up in Sentry too. Once we see a few log entries, we can go find the error on Sentry and learn more about why it happens.

PR #283, opened 2019-10-02
Only mark sync as failed during sync

jnraine:
Customers were reporting syncs switching from `COMPLETE` to `FAILED` without triggering a sync themselves. The underlying problem was with how and when we marked syncs as failed.
    
Syncs occur in the background on the `installation` queue.  When a sync background job throws an error, a global failure handler is called that updates `subscription.syncStatus` to `FAILED`. However, this global failure handler runs when _other_ jobs fail too.
    
This led to a sync being marked as `FAILED` when any job for that customer failed.
    
This commit avoids that by moving the `subscription.syncStatus` update out of the global failure handler and into the sync job.

### DIFF
--- a/lib/sync/installation.js
+++ b/lib/sync/installation.js
@@ -81,7 +81,10 @@ module.exports.processInstallation = (app, queues) => {
     const github = await getEnhancedGitHub(installationId)
 
     const nextTask = getNextTask(subscription)
-    if (!nextTask) return
+    if (!nextTask) {
+      await subscription.update({ syncStatus: 'COMPLETE' })
+      return
+    }
 
     await subscription.update({ syncStatus: 'ACTIVE' })
 
--- a/lib/sync/installation.js
+++ b/lib/sync/installation.js
@@ -182,6 +182,8 @@ module.exports.processInstallation = (app, queues) => {
         await updateJobStatus(jiraClient, job, edgesLeft, task, repositoryId)
         return
       }
+
+      await subscription.update({ syncStatus: 'FAILED' })
       throw err
     }
   }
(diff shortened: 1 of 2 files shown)


lib/sync/installation.js
```diff
@@ -81,7 +81,10 @@ module.exports.processInstallation = (app, queues) => {
     const github = await getEnhancedGitHub(installationId)
 
     const nextTask = getNextTask(subscription)
-    if (!nextTask) return
+    if (!nextTask) {
+      await subscription.update({ syncStatus: 'COMPLETE' })
```
jules2689: I don't understand where this came from. It didn't seem to be in the worker/index.js... so why do we need it now?

jnraine: This is a new behaviour intended to resolve the following pain point:

1. Customer installs Jira integration
2. Initial sync completes all work but fails while updating sync to COMPLETE
3. We fix bug and retrigger sync to correct problem

Prior to this change, the sync would remain FAILED even after a sync was retriggered. This ensures that if a sync runs and there is no remaining tasks to complete, we mark it as COMPLETE.

jnraine:
> Do we have tests we can add / change? I've one comment otherwise.

I'm sad to say this file has no tests. Adding it for this specific case is perhaps feasible though. I'll look into it tomorrow.

lib/sync/installation.js
zmoazeni: Sounds like a fairly uncommon edge case. I would recommend adding a log statement here so that we can see how often we're running into it. Or if/when we're doing any sleuthing for a particular user. It's not much, but a little visibility.

jules2689: We could also ping datadog so we can see this over time. Nice edge case!

jnraine: This case happens when the customer triggers a "normal" sync or, more likely, _we_ trigger the sync on their behalf.

PR #285, opened 2019-10-02
Fix sync status metrics bug

jnraine:
Added in #279, sync status metrics were intended to give greater insight into the overall health of the app. However, due to the way `bin/send-sync-status-metrics` was run using Heroku Scheduler, the metrics never made it to Datadog. [1]

This should remedy the problem by moving the metric sending into a background job. Because we already send metrics from background jobs to Datadog, I’m confident that the metrics will arrive. Heroku Scheduler still plays an important part but is now only repsonsible to enqueue the job.

Note: Bull (the background job library) supports what it calls [“repeated jobs”](https://github.com/OptimalBits/bull#repeated-jobs). However, I had difficulty getting these to work locally.

1: While I haven’t uncovered the cause, my hunch is that the Heroku kills the dyno before the Datadog agent sends the buffered metrics to Datadog.

### DIFF
--- a/bin/send-sync-status-metrics
+++ b/bin/send-sync-status-metrics
@@ -1,29 +1,10 @@
 #!/usr/bin/env node
 require('dotenv').config()
 
-const Sentry = require('@sentry/node')
-
-Sentry.init({
-  dsn: process.env.SENTRY_DSN,
-  environment: process.env.SENTRY_ENV || process.env.NODE_ENV,
-  release: process.env.HEROKU_SLUG_COMMIT
-})
-
-const statsd = require('../lib/config/statsd')
-const { Subscription } = require('../lib/models')
-
-Subscription.syncStatusCounts().then((results) => {
-  results.forEach((row) => {
-    statsd.gauge('syncs', row.count, { status: row.syncStatus })
+const { queues } = require('../lib/worker')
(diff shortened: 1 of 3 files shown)


jules2689:
Interesting. I've never had to use a bg job for metrics before because I thought statsd did everything over UDP/async. Is this not the case in the JS lib?

jnraine:
@jules2689 Metrics are sent over UDP in this app as well. The reason I've moved this into a background job isn't because it takes a long time but rather because metrics sent from the script never made it to Datadog.

I haven't confirmed but here is my understanding of the problem.

1. Every ten minutes, Heroku Scheduler spins up a dyno to run `bin/send-sync-status-metrics`
2. During dyno boot, Datadog agent starts up.
3. `bin/send-sync-status-metrics` is executed and sends metrics via UDP
4. The Datadog agent receives the metrics and stores them in a buffer, to be sent to Datadog at a later interval
5. The script exits and the dyno is killed, along with the Datadog agent.

So, the metrics never make it to Datadog. Instead, this now enqueues a job that runs on one of our background worker dynos, which live for longer than a few seconds.

It's also possible that some other thing is to blame here. When I start a dyno and run the script myself, the metrics _still_ don't make it to Datadog. 🤷‍♂

PR #286, opened 2019-10-03
Add `bin/stafftools errors`

jnraine:
This opens the browser to Sentry’s discover tab with errors for a given Jira host. Useful when investigating a customer problem.

In order to use, you must have access to the production Sentry instance.

(This codifies a practice I've been using and turns a multiple step process into single command. Since you're probably using `bin/stafftools` to diagnose a customer problem, this makes the hop to Sentry more smooth.)

### DIFF
--- a/bin/stafftools
+++ b/bin/stafftools
@@ -4,6 +4,8 @@ require 'cgi'
 require 'open3'
 require 'io/console'
 require 'json'
+require 'shellwords'
+require 'uri'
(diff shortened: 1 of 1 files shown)


jnraine:
Lightning fast review after hours! ✨🥇

jules2689:
That's what happens in a hotel when you're traveling lol :P

PR #288, opened 2019-10-05
Exit 0 when successful

jnraine:
This script was exiting non-zero when everything was fine. This fixes that.


jnraine:
@jules2689 Yep, this'll remedy that. 👌

PR #289, opened 2019-10-08
Add support for unicode line separator to SmartCommitTokenizer

jnraine:
A string ending with a [line separator], a whitespace unicode character (everyone’s favourite kind), would result in a syntax error.

This fixes that by matching tokenizing the `workLogComment` using a “not whitespace” regex pattern. This pattern matches any character that isn’t a carriage return, newline, tab, or space. You know, the normal whitespace stuff.

This results in the line separator being included as a work log comment but because we’re currently only extracting the issue keys from the tokenized smart commit, it will have no ill effect.

[line separator]: https://www.fileformat.info/info/unicode/char/2028/index.htm

### DIFF
--- a/lib/smart-commit-tokenizer.js
+++ b/lib/smart-commit-tokenizer.js
@@ -10,6 +10,7 @@ module.exports = function () {
   const newline = {match: /\n/, lineBreaks: true, next: 'main'}
   const time = {match: /(?<= )#time(?= )/, push: 'workLog'}
   const whitespace = /[ \t]+/
+  const notWhitespace = /[^\r\n\t ]+/
 
   return moo.states({
     main: {
@@ -24,7 +25,7 @@ module.exports = function () {
     transition: {
       time,
       transition,
-      comment: /[^\r\n\t ]+/,
+      comment: notWhitespace,
       whitespace,
       carriageReturn,
       newline
@@ -36,7 +37,7 @@ module.exports = function () {
       hours: {match: /[\d.]+?h(?!\B)/, value: text => text.slice(0, -1)},
       minutes: {match: /[\d.]+?m(?!\B)/, value: text => text.slice(0, -1)},
       whitespace,
-      workLogComment: /.+/,
+      workLogComment: notWhitespace,
       carriageReturn,
       newline
     }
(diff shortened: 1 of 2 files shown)


lib/smart-commit-tokenizer.js
```diff
@@ -10,6 +10,7 @@ module.exports = function () {
   const newline = {match: /\n/, lineBreaks: true, next: 'main'}
   const time = {match: /(?<= )#time(?= )/, push: 'workLog'}
   const whitespace = /[ \t]+/
+  const notWhitespace = /[^\r\n\t ]+/
```
jules2689: Is there any reason we can't use `\s` to match any whitespace?

jnraine: Yes...sort of. There's some history here.

1. The `whitespace` pattern above must exclude newline and carriage return characters because they are used to delimit commands and must be matched separately. In order to accomplish this, I listed whitespace characters explicitly. My list didn't include unicode characters.
2. The `notWhitespace` pattern needed to match any characters that `whitespace` didn't. This meant it needed to match Unicode whitespace characters, which led to the pattern you see above. More details on this here: https://github.com/integrations/jira/pull/213/commits/c8ea4ecbb3ffac4210cb16436c62fc79cd670f9e

If characters didn't match at least one of those patterns, it was possible to trigger a syntax error, which we saw here.

Taking a fresh look at this code today, I realized a better, albeit brain-twisting, solution: match using a negated character set that includes non-whitespace (`\S`) and `\r\n`. This precisely excludes `\n\r` while including other unicode whitespace characters.

New code pushed!

jules2689: Ah this is great context. Thank you!
It might be nice to include this as a comment above this const so the context is inline 😄

jnraine: Already done! Do you think more detail is warranted? https://github.com/integrations/jira/pull/289/files#diff-742bcc471ec3bbc2c448c13a42e19feaR12

jules2689: nope! That's great. Thank you ❤️

PR #294, opened 2019-11-13
Bump the sentry client version and configure Sentry Var

tebriel:
app.json tells heroku what the var is and that we desire it.

Closes: https://github.com/github/ecosystem-primitives/issues/459


jules2689:
This PR can be rebased on master to get the `build` CI build, or we can rely on travis and override the merge PR button :)

tebriel:
@jules2689 we have to merge master in every time before deploying because heroku doesn't do it!

PR #295, opened 2019-11-13
Add actions build script

jules2689:
Adds GitHub Actions instead of Travis.
It seems to be about 15-20 seconds faster too!


jules2689:
> Should we be concerned about this: https://github.com/integrations/jira/pull/295/checks?check_run_id=301423012#step:6:274

I don't think so. It's a new relic agent that's failing. As long as the npm packages are being installed, it should be good. Also, it was failing on travis too :)

image

PR #297, opened 2019-11-25
Lint Jira Integration

tebriel:
Bring the Jira Integration up to our Slack linting standards.

Currently the Jira project uses [Standard](https://standardjs.com/), but Slack (and all of dotcom) use [ESLint](https://eslint.org/) with the airbnb styleguide and some other rules turned on. This is probably too much work for a single PR, but it's giving me exposure into the project, which is generally how I learn. We can throw this away/part it out if we decide it's too much.


tebriel:
Will revisit this idea later.

PR #301, opened 2019-12-06
Use Throng for Probot Concurrency

tebriel:
We sometimes have huge spikes in request queuing, we only have 5 workers, that may be part of the cause, let's use throng [like we did in slack](https://github.com/integrations/slack-private/pull/29) to add more workers on the same nodes to better utilize our resources.

## Env Vars

Heroku will calculate `WEB_CONCURRENCY` for us based on Dyno Ram Size / `WEB_MEMORY` == `WEB_CONCURRENCY`. We can also set it manually (which we will do at first to ensure a no-change rollout at first, see below).

## Preflight Checklist

- [x] Do we have enough redis connections available
- [x] Do we have enough postgres connections available -- **NO** Decrease Web worker autoscale to 5 max (see comment below)
- [x] Set the `WEB_CONCURRENCY` env var in heroku to `1` before deploying (this ensures we don't spin up more than 1 worker to start with)
- [x]  Set the `WEB_MEMORY` env var in heroku to 768 before deploying (this ensures that we have enough memory once we remove the `WEB_CONCURRENCY` variable.

### DIFF
--- a/db/config.json
+++ b/db/config.json
@@ -19,6 +19,9 @@
     "ssl": true,
     "dialectOptions": {
       "ssl": true
+    },
+    "pool": {
+      "max": 5
     }
   }
 }
--- a/lib/run.js
+++ b/lib/run.js
@@ -1,6 +1,7 @@
 #!/usr/bin/env node
 
 require('dotenv').config()
+const throng = require('throng')
 
 if (process.env.NEWRELIC_KEY) {
   require('newrelic') // eslint-disable-line global-require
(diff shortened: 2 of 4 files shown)


tebriel:
Redis is a premium-3 like slack, we have enough available connections.

tebriel:
Postgres is a standard-0, which has 120 max connections. The pool size is 5, this means we can support 24 instances (between web and workers).

## Queue Workers

We currently have 3 dynos hard-configured (no autoscaling), on 1x dynos (can only support 1 instance per dyno) this means at most we'll use 15 database connections.

## Web Workers

We support up to 10 Performance-M Dynos which have 2.5Gb of ram. At 768mb of RAM per instance, this is a max of 3 workers per dyno.
10 dynos * 3 workers * 5 pool size == 150 connections. We will need to decrease the max scaling of the dynos.

## Total Budget

Workers: 5 * 3 == 15 connections (budget now at 105 connections)
Web: 105 / 5 == 21 instances / 3 per dyno == 7 dynos max.

## New Dyno settings

Worker: Stays the same at 3
Web: Decrease max autoscaling from 10 to 5 (gives us buffer for postgres follower, ad-hoc queries, deploys, etc)

PR #302, opened 2019-12-09
Don't raise an exception if there's no jira client

tebriel:
Exceptions are expensive in nodejs (recovering, capturing the callback, submitting it to sentry, etc). We've seen this exception 55k times, and generally see it 1.1-1.3k times per day: https://sentry.io/organizations/github-integrations/issues/1201310569/events/013891d2c99b4797b1ba91139276cae8/

> Cannot read property 'jiraHost' of null

This is because if a user disables the integration, the `Installation.getForHost` function returns `null`, and then we try to use it anyways.

The simple solution is: If there isn't an installation, exit, we can't process it anyways, so no reason to raise an exception about it.

Since this is a configuration option (the user has configured the system to disable the plugin but also it still receives webhooks), I don't think that sending a metric is warranted, as the number of requests that we skipped due to users disabling the installation would not be actionable (as far as I can tell).

### DIFF
--- a/test/unit/jira/client/index.test.js
+++ b/test/unit/jira/client/index.test.js
@@ -13,7 +13,7 @@ describe('Test getting a jira client', () => {
     await installation.enable()
   })
 
-  test('Instllation exists', async () => {
+  test('Installation exists', async () => {
     const client = await getJiraClient(BASE_URL, 1, {})
     expect(client).toMatchSnapshot()
   })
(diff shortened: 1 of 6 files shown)


test/unit/jira/client/index.test.js
```diff
@@ -0,0 +1,32 @@
...
+  beforeEach(async () => {
+    const installation = await Installation.install({
+      host: BASE_URL,
+      sharedSecret: 'shared-secret',
+      clientKey: 'client-key'
+    })
+    await installation.enable()
+  })
+
+  test('Instllation exists', async () => {
```
zmoazeni:
```suggestion
  test('Installation exists', async () => {
```

PR #303, opened 2019-12-09
Upgrade Bull to major redis cli improvements

tebriel:
Upgrading from [Bull 3.4.7](https://github.com/OptimalBits/bull/blob/develop/CHANGELOG.md#v347) to [Bull 3.12.1](https://github.com/OptimalBits/bull/blob/develop/CHANGELOG.md#v3121)

biggest changes to note are from [Bull 3.5.3](https://github.com/OptimalBits/bull/blob/develop/CHANGELOG.md#v353) where ioredis was upgraded from 3.x to [4.x](https://github.com/luin/ioredis/blob/master/Changelog.md). Which has numerous performance, security, and stability improvements

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1994,17 +1994,32 @@
       "integrity": "sha1-Jw8HbFpywC9bZaR9+Uxf46J4iS8="
     },
     "bull": {
-      "version": "3.4.7",
-      "resolved": "https://registry.npmjs.org/bull/-/bull-3.4.7.tgz",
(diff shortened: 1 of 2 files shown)


tebriel:
@jules2689 ioredis is only used by bull (which makes our life easier). Bull has been on the newer version of ioredis for a while, so the problems with any major bumps there should be fine. We're also on the latest redis instance version in heroku, so I don't think we should expect any conflicts there either!

PR #312, opened 2020-02-04
More statsd sync metrics

zmoazeni:
This is in an effort to aid https://github.com/github/integrations-jira-internal/issues/17 which is more metrics on Jira to help us diagnose issues early. And to see positive trends when we push improvements.

This implements:

- How long does a complete sync take (per repo, per org)? (mean, 95th, max)
- How long does each job type take?(histogram)
- How often does each job fail?
- Aging jobs. How many have stayed PENDING for more than 24 hours?

### DIFF
--- a/lib/models/subscription.js
+++ b/lib/models/subscription.js
@@ -1,5 +1,4 @@
 const Sequelize = require('sequelize')
-const moment = require('moment')
 
 module.exports = class Subscription extends Sequelize.Model {
   static init (sequelize, DataTypes) {
@@ -121,12 +120,15 @@ module.exports = class Subscription extends Sequelize.Model {
   }
 
   static async agedSyncCounts () {
-    const ageThreshold = moment().subtract(7, 'd').format('YYYY-MM-DD')
     const [ results ] = await this.sequelize.query(
-      `SELECT "syncStatus", COUNT(*) FROM "Subscriptions" WHERE "updatedAt" < :ageThreshold AND ("syncStatus" IS NULL OR "syncStatus" in ('ACTIVE', 'PENDING')) GROUP BY "syncStatus"`,
-      {
-        replacements: { ageThreshold }
-      }
+      `
+        SELECT "syncStatus", COUNT(*)
+        FROM   "Subscriptions"
+        WHERE  "updatedAt" < NOW() - INTERVAL '7 days'
+              AND ( "syncStatus" IS NULL
+                      OR "syncStatus" IN ( 'ACTIVE', 'PENDING' ) )
+        GROUP  BY "syncStatus"
+      `
     )
     return results
   }
--- a/lib/models/subscription.js
+++ b/lib/models/subscription.js
@@ -120,8 +120,8 @@ module.exports = class Subscription extends Sequelize.Model {
     return results
   }
 
-  static async agedSyncCounts() {
-    const ageThreshold = moment().subtract(7,'d').format('YYYY-MM-DD')
+  static async agedSyncCounts () {
+    const ageThreshold = moment().subtract(7, 'd').format('YYYY-MM-DD')
     const [ results ] = await this.sequelize.query(
       `SELECT "syncStatus", COUNT(*) FROM "Subscriptions" WHERE "updatedAt" < :ageThreshold AND ("syncStatus" IS NULL OR "syncStatus" in ('ACTIVE', 'PENDING')) GROUP BY "syncStatus"`,
       {
--- a/lib/models/subscription.js
+++ b/lib/models/subscription.js
@@ -1,4 +1,5 @@
 const Sequelize = require('sequelize')
+const moment = require('moment')
 
 module.exports = class Subscription extends Sequelize.Model {
   static init (sequelize, DataTypes) {
@@ -116,6 +117,17 @@ module.exports = class Subscription extends Sequelize.Model {
     const syncStatusCountQuery = 'SELECT "syncStatus", COUNT(*) FROM "Subscriptions" GROUP BY "syncStatus"'
     const [ results ] = await this.sequelize.query(syncStatusCountQuery)
 
+    return results
+  }
+
+  static async agedSyncCounts() {
+    const ageThreshold = moment().subtract(7,'d').format('YYYY-MM-DD')
+    const [ results ] = await this.sequelize.query(
+      `SELECT "syncStatus", COUNT(*) FROM "Subscriptions" WHERE "updatedAt" < :ageThreshold AND ("syncStatus" IS NULL OR "syncStatus" in ('ACTIVE', 'PENDING')) GROUP BY "syncStatus"`,
+      {
+        replacements: { ageThreshold }
+      }
+    )
     return results
   }
 
--- a/lib/sync/discovery.js
+++ b/lib/sync/discovery.js
@@ -10,6 +10,7 @@ const jobOpts = {
 
 module.exports.discovery = (app, queues) => {
   return async function discoverContent (job) {
+    const startTime = new Date();
     const { jiraHost, installationId } = job.data
     const github = await app.auth(installationId)
     enhanceOctokit(github, app.log)
@@ -34,6 +35,6 @@ module.exports.discovery = (app, queues) => {
     await subscription.save()
 
     // Create job
-    queues.installation.add({ installationId, jiraHost }, jobOpts)
+    queues.installation.add({ installationId, jiraHost, startTime }, jobOpts)
   }
 }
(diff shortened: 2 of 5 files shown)


lib/models/subscription.js
```diff
@@ -119,6 +120,17 @@ module.exports = class Subscription extends Sequelize.Model {
     return results
   }
 
+  static async agedSyncCounts () {
+    const ageThreshold = moment().subtract(7, 'd').format('YYYY-MM-DD')
```
zmoazeni: I just chose 7 days as a threshold for aged jobs. We can definitely tweak this.

lib/sync/discovery.js
```diff
@@ -34,6 +35,6 @@ module.exports.discovery = (app, queues) => {
     await subscription.save()
 
     // Create job
-    queues.installation.add({ installationId, jiraHost }, jobOpts)
+    queues.installation.add({ installationId, jiraHost, startTime }, jobOpts)
```
zmoazeni: We'll only pass the `startTime` when we come through the discovery.

lib/sync/installation.js
```diff
@@ -45,7 +46,15 @@ module.exports.processInstallation = (app, queues) => {
       // no more data (last page was processed of this job type)
       if (!getNextTask(subscription)) {
         subscription.set('syncStatus', 'COMPLETE')
-        app.log(`Sync status for installationId=${installationId} is complete`)
+        var message = `Sync status for installationId=${installationId} is complete`
+        if (job.data.startTime !== undefined) {
+          const endTime = new Date()
+          const timeDiff = endTime - Date.parse(job.data.startTime)
+          message = `${message} startTime=${job.data.startTime} endTime=${endTime.toJSON()} diff=${timeDiff}`
+          statsd.histogram('full_sync', timeDiff)
```
zmoazeni: I'm completely open to better metric names than `full_sync`

lib/worker/index.js
```diff
@@ -15,6 +16,12 @@ const { CONCURRENT_WORKERS = 1 } = process.env
 const AxiosErrorEventDecorator = require('../models/axios-error-event-decorator')
 const SentryScopeProxy = require('../models/sentry-scope-proxy')
 
+function measureElapsedTime (startTime, tags) {
+  const endTime = new Date()
+  var timeDiff = endTime - startTime
+  statsd.histogram('job_duration', timeDiff, tags)
```
zmoazeni: These histograms will give us several nice pieces of data for completed and failed jobs. Including counts, timing, and percentiles.

lib/worker/metrics-job.js
```diff
@@ -7,4 +7,9 @@ module.exports = async (job) => {
   syncStatusCounts.forEach((row) => {
     statsd.gauge('syncs', row.count, { status: row.syncStatus })
   })
+
+  const agedSyncCounts = await Subscription.agedSyncCounts()
```
tebriel: should we `Promise.all()` these two `async/await` functions so that we're not synchronously waiting for a then b?

lib/worker/index.js
```diff
@@ -29,10 +36,17 @@ Object.keys(queues).forEach(name => {
...
   })
 
   queue.on('completed', function (job, jobPromise) {
     app.log.info(`Job completed name=${name} id=${job.id}`)
+    measureElapsedTime(job.meta_time_start, { queue: name, status: 'completed' })
+  })
+
+  queue.on('failed', async (job, err) => {
+    app.log.error(`Error occurred while processing job id=${job.id} on queue name=${name}`)
+    measureElapsedTime(job.meta_time_start, { queue: name, status: 'failed' })
```
tebriel: Should we `measureElapsedTime` for the `error` as well?

lib/worker/index.js
tebriel: Not necessary, but I find it very helpful to indicate the time unit in the metric name. Something like `job_duration_sec` or `job_duration_ms` to make dashboarding easier.

lib/sync/installation.js
tebriel: I think it's serviceable. Do you mind putting a comment above the `statsd.histogram` call that says what `full_sync` measures? That way when someone looks at it in datadog and then goes and searches the code, they'll find something like (I dunno if the below phrasing is correct or well worded):
```js
// full_sync measures the duration from start to finish of a complete scan and sync of github issues translated to tickets
```

lib/models/subscription.js
tebriel: If we change the threshold, will the metric reported be invalid for previous times? Do we need to add the threshold to the metric name so that it doesn't suddenly shift the dashboards/monitors in an unexpected way?

lib/models/subscription.js
```diff
@@ -119,6 +120,17 @@ module.exports = class Subscription extends Sequelize.Model {
...
+    const [ results ] = await this.sequelize.query(
+      `SELECT "syncStatus", COUNT(*) FROM "Subscriptions" WHERE "updatedAt" < :ageThreshold AND ("syncStatus" IS NULL OR "syncStatus" in ('ACTIVE', 'PENDING')) GROUP BY "syncStatus"`,
```
tebriel: I think we could save a little work in node by moving the date calculation to the sql (then we also don't need to require moment.js here)
```js
const [ results ] = await this.sequelize.query(
    `SELECT "syncStatus", 
           COUNT(*) 
    FROM   "Subscriptions" 
    WHERE  "updatedAt" < NOW - INTERVAL '7 days'
           AND ( "syncStatus" IS NULL 
                  OR "syncStatus" IN ( 'ACTIVE', 'PENDING' ) ) 
    GROUP  BY "syncStatus"`,

File: lib/worker/index.js (hunk 6 above)
Comment (zmoazeni) [reply lib/worker/index.js]: From what I read, `error` is thrown when bull has trouble reading from the queue. So I avoided it in that case. There isn't a job available to read the `startTime` from.

Comment (tebriel) [reply lib/worker/index.js]: Makes perfect sense!

File: lib/worker/index.js (hunk 4 above)
Comment (zmoazeni) [reply lib/worker/index.js]: Yeah, I'm not a big fan of the Hungarian notation for the statsd names. I'll change the datatype in datadog once we start collecting metrics.

File: lib/worker/metrics-job.js (hunk 5 above)
Comment (zmoazeni) [reply lib/worker/metrics-job.js]: Can you give me a hand, what sort of code are you thinking @tebriel? These queries are fast, so I'm really going for the easiest to grok here.

File: lib/worker/index.js (hunk 4 above)
Comment (tebriel) [reply lib/worker/index.js]: I understand that. The default for datadog is `ms` if I recall, maybe if we just make sure that we're reporting this in ms (I think we are) then that's good!

Comment (zmoazeni) [reply lib/worker/index.js]: > I understand that. The default for datadog is ms if I recall, maybe if we just make sure that we're reporting this in ms (I think we are) then that's good!

Ahh yeah! We are! ms is the default:

```javascript
var a = new Date();
undefined
var b = new Date();
undefined
b - a
811
```

lib/worker/metrics-job.js
tebriel: Yeah! I was thinking we could issue these SQL queries concurrently using [`Promise.all()`](https://developer.mozilla.org/en-US/docs/web/javascript/reference/global_objects/promise/all)
(I didn't test this, I just wrote it in the comment box, so there may be a syntax error)

```js
Promise.all([Subscription.syncStatusCounts(), Subscription.agedSyncCounts()]).then((values) => {
  values[0].forEach((row) => {
    statsd.gauge('syncs', row.count, { status: row.syncStatus })
  })
  values[1].forEach((row) => {
    statsd.gauge('aged_syncs', row.count, { status: row.syncStatus })
  })
})
```

tebriel: If those queries are bound to be fast, then I think it's fine to do them sequentially, just thinking we could parallelize sql queries and lean on Postgres to save some time/work.

zmoazeni: Ahh so the `values` is an array of the return values from each of the promises. Neat!

I can make that change in a followup PR. Great suggestion

lib/models/subscription.js
```diff
@@ -119,6 +119,20 @@ module.exports = class Subscription extends Sequelize.Model {
     return results
   }
 
+  static async agedSyncCounts () {
```
tebriel: should we add a comment here to note that if you change the `interval` you need to change the metric name too?

zmoazeni: Hrm. I think paramaterizing the day probably makes more sense. Then the "7" will live only in the job code. I'll hit that in the followup too.

PR #323, opened 2020-03-04
Add documentation about database-migrations

tebriel:
Describe how migrations are done in jira's heroku world and update the Actions yaml to fix the Postgres issue.

## Rollback Strategy

re-deploy the last commit via chatops `.deploy jira-integration to production`

### DIFF
--- a/docs/database-migrations.md
+++ b/docs/database-migrations.md
@@ -1,0 +1,19 @@
+# Migrating the Database
+
+This is a bit complicated because of how we connect these things and how it's deployed. This document should walk you through most of the sharp edges.
+
+## Where
+
+You'll want to use the environment appropriate for your migration:
+
+*  jira-integration-staging
+*  jira-integration-production
+
+
+## Running the Migration
+
+1.  Login to heroku's cli `heroku login`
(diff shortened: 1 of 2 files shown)


docs/database-migrations.md
```diff
@@ -0,0 +1,24 @@
...
+
+We deploy into Heroku right now, and heroku doesn't give us a one-off spot for deploying scripts. To that end, there are three apps in Heroku for Jira:
+
+*  jira-integration-staging
+*  jira-integration-migrations
+*  jira-integration-production
+
+You can deploy -staging and -production with chatops, but for now, -migrations isn't involved with Heaven.
+
+So you will need to go to the jira-integration-migrations page, click on the Deploy tab, and then choose "manual deploy" from your branch.
```
zmoazeni: I'm confused, why do we need to deploy to `-migrations`? Why can't we just deploy to production and run `./node_modules/.bin/sequelize db:migrate`?

tebriel: Unfortunately not, because if you deploy to `-production` you will require production to then restart to be `ssh`-able (this is required every time you re-deploy, it turns off ssh ability).

PR #327, opened 2020-03-23
Add more new relic tracing

zmoazeni:
@tebriel I'm curious what you think about this approach. This is just adding more segments in newrelic so that our traces like https://rpm.newrelic.com/accounts/205/applications/154434972/transactions#id=5b225765625472616e73616374696f6e2f457870726573736a732f504f53542f2f222c22225d&app_trace_id=821171b1-6cf9-11ea-a8d9-0242ac110009_2716_4351 will have a lot more specific timing detail.

I ran this locally and verified the webhook handlers are working. I'm not using newrelic locally though so I'll test this in staging first.

### DIFF
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -53,6 +53,7 @@ module.exports = function middleware (callback) {
     enhanceOctokit(context.github, context.log)
 
     if (newrelic !== undefined) {
+      newrelic.setControllerName(`github/events.${context.name}`)
       newrelic.addCustomAttributes({
         'Webhook ID': context.id,
         'Webhook Event': `${context.name}.${context.payload.action}`,
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -6,6 +6,12 @@ const { Subscription } = require('../models')
 const getJiraClient = require('../jira/client')
 const getJiraUtil = require('../jira/util')
 const enhanceOctokit = require('../config/enhance-octokit')
+
+let newrelic
+if (process.env.NEWRELIC_KEY) {
+  newrelic = require('newrelic')
+}
+
 // Returns an async function that reports errors errors to Sentry.
 // This works similar to Sentry.withScope but works in an async context.
 // A new Sentry hub is assigned to context.sentry and can be used later to add context to the error message.
@@ -37,6 +43,13 @@ const isFromIgnoredRepo = (payload) => {
 module.exports = function middleware (callback) {
   return withSentry(async (context) => {
     enhanceOctokit(context.github, context.log)
+
+    if (newrelic !== undefined) {
+      newrelic.addCustomAttributes({
+        'Webhook ID': context.id,
+        'Webhook Event': `${context.name}.${context.payload.action}`
+      })
+    }
 
     context.sentry.setExtra('GitHub Payload', {
       event: context.name,
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -80,7 +80,7 @@ module.exports = function middleware (callback) {
       return
     }
 
-    const subscriptions = await withNRSegment('middleware:Subscription.getAllForInstallation', async () => {
+    const subscriptions = await withNRSegment('Middleware: Subscription.getAllForInstallation', async () => {
       return Subscription.getAllForInstallation(gitHubInstallationId)
     })
 
@@ -99,7 +99,7 @@ module.exports = function middleware (callback) {
 
       context.log = context.log.child({ jiraHost })
 
-      const jiraClient = await withNRSegment('middleware:getJiraClient', async () => {
+      const jiraClient = await withNRSegment('Middleware: getJiraClient', async () => {
         return getJiraClient(jiraHost, gitHubInstallationId, context.log)
       })
       if (jiraClient == null) {
@@ -110,7 +110,7 @@ module.exports = function middleware (callback) {
       const util = getJiraUtil(jiraClient)
 
       try {
-        await withNRSegment('middleware:webhook handler', async () => callback(context, jiraClient, util))
+        await withNRSegment('Middleware: webhook handler', async () => callback(context, jiraClient, util))
       } catch (err) {
         context.sentry.captureException(err)
       }
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -47,7 +47,8 @@ module.exports = function middleware (callback) {
     if (newrelic !== undefined) {
       newrelic.addCustomAttributes({
         'Webhook ID': context.id,
-        'Webhook Event': `${context.name}.${context.payload.action}`
+        'Webhook Event': `${context.name}.${context.payload.action}`,
+        'Repository': context.payload.repository
       })
     }
 
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -6,11 +6,7 @@ const { Subscription } = require('../models')
 const getJiraClient = require('../jira/client')
 const getJiraUtil = require('../jira/util')
 const enhanceOctokit = require('../config/enhance-octokit')
-
-let newrelic
-if (process.env.NEWRELIC_KEY) {
-  newrelic = require('newrelic')
-}
+const newrelic = require('newrelic')
 
 // Returns an async function that reports errors errors to Sentry.
 // This works similar to Sentry.withScope but works in an async context.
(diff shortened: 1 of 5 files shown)


lib/github/middleware.js
```diff
@@ -38,11 +52,21 @@ module.exports = function middleware (callback) {
...
-      event: context.name,
-      action: context.payload.action,
-      id: context.id,
-      repo: (context.payload.repository) ? context.repo() : undefined
+    if (newrelic !== undefined) {
+      newrelic.addCustomAttributes({
+        'Webhook ID': context.id,
+        'Webhook Event': `${context.name}.${context.payload.action}`,
+        'Repository': context.payload.repository
+      })
```
zmoazeni: More information about custom attributes:
* https://docs.newrelic.com/docs/agents/manage-apm-agents/agent-data/collect-custom-attributes

lib/github/middleware.js
```diff
@@ -24,6 +30,14 @@ const withSentry = function (callback) {
   }
 }
 
+function withNRSegment (segmentName, callback) {
+  if (newrelic !== undefined) {
+    return newrelic.startSegment(segmentName, true, callback)
+  } else {
+    return callback()
+  }
+}
```
zmoazeni: More information about segments:
* https://docs.newrelic.com/docs/agents/nodejs-agent/api-guides/nodejs-agent-api#getTransaction
* https://docs.newrelic.com/docs/agents/nodejs-agent/supported-features/nodejs-custom-instrumentation#expanding-instrumentation

lib/github/middleware.js
```diff
@@ -58,7 +82,10 @@ module.exports = function middleware (callback) {
       return
     }
 
-    const subscriptions = await Subscription.getAllForInstallation(gitHubInstallationId)
+    const subscriptions = await withNRSegment('middleware:Subscription.getAllForInstallation', async () => {
```
tebriel: Not really important: NewRelic names these slightly different: `Middleware: linkShared /slack/events.link_shared` should we follow this pattern of capital M and a space after the `:`? Just wondering if it's worth trying to be consistent here.

lib/github/middleware.js
```diff
@@ -6,6 +6,12 @@ const { Subscription } = require('../models')
 const getJiraClient = require('../jira/client')
 const getJiraUtil = require('../jira/util')
 const enhanceOctokit = require('../config/enhance-octokit')
+
+let newrelic
+if (process.env.NEWRELIC_KEY) {
```
tebriel: I know you're just copying the pattern, but I'm wondering if we should use the [agent-enabled](https://docs.newrelic.com/docs/agents/nodejs-agent/installation-configuration/nodejs-agent-configuration#agent-enabled) configuration in [newrelic.js](https://github.com/integrations/jira/blob/master/newrelic.js#L8) and have it be `agent_enabled: !!process.env.NEWRELIC_KEY,` this would avoid this check and let use call out a noop to the newrelic agent (also we wouldn't have to do the if/else block below). What do you think? The other place I know this pattern is used is in [run.js](https://github.com/integrations/jira/blob/master/lib/run.js#L6).

We did this in [slack's newrelic.js](https://github.com/github/slack/blob/master/newrelic.js#L17-L18) and it's worked really well.

lib/github/middleware.js
```diff
@@ -38,11 +52,21 @@ module.exports = function middleware (callback) {
...
-      repo: (context.payload.repository) ? context.repo() : undefined
+    if (newrelic !== undefined) {
+      newrelic.addCustomAttributes({
+        'Webhook ID': context.id,
+        'Webhook Event': `${context.name}.${context.payload.action}`,
+        'Repository': context.payload.repository
+      })
+    }
+
+    withNRSegment('middleware:sentry.setExtra', () => {
```
tebriel: I'm not sure we need to trace this, it isn't a web call, it just sets keys on the sentry scope object.

lib/github/middleware.js
zmoazeni: Oh I definitely can. Good call!

lib/github/middleware.js
zmoazeni: Yeah I really didn't like this whole conditional stuff. That looks like a great way to simplify this code! I'll take a stab at it.

lib/github/middleware.js
zmoazeni: I thought that, however if you look down we actually run `context.repo()`. So I'm wondering what the execution time would be for that. I think that's actually making an API request against GitHub.

lib/github/middleware.js
```diff
@@ -38,11 +52,21 @@ module.exports = function middleware (callback) {
...
+      context.sentry.setExtra('GitHub Payload', {
```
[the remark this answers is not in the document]
zmoazeni: Oh I didn't add these. I just moved them into the Segment.

lib/github/middleware.js
```diff
@@ -83,7 +112,9 @@ module.exports = function middleware (callback) {
       const util = getJiraUtil(jiraClient)
 
       try {
-        await callback(context, jiraClient, util)
+        withNRSegment('middleware:webhook handler', async () => {
+          await callback(context, jiraClient, util)
+        })
```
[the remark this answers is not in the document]
zmoazeni: Yeah, I believe you're right. I think I still need an explicit `return callback(context, jiraClient, util)` here right?

zmoazeni:
I'm seeing build failures locally too. I'm going to work on resolving them as well.

lib/github/middleware.js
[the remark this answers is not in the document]
tebriel: Ahh yeah, didn't notice that, good point! After you figuring out the `context.id` thing the other day I thought you were adding it here but I mis-read the diff.

lib/github/middleware.js
tebriel: Looking at [the source](https://github.com/probot/probot/blob/6e45ad7/src/context.ts#L106) (note this is the newer probot source) it's just a lookup in the object. That explains why this works without an `await` too!

lib/github/middleware.js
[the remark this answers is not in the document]
tebriel: not if you leave the `{}` out! It returns the value:
```js
a = () => 'tacos'
[Function: a]
> a() === 'tacos'
true
```

lib/github/middleware.js
zmoazeni: Oh thanks! Good call. I'll revert this back then. I agree setting those values shouldn't need to be instrumented then.

tebriel: Found the same set in the [7.x source](https://github.com/probot/probot/blob/7.x/src/context.ts#L63-L74) that's definitely a synchronous object property lookup!

lib/github/middleware.js
zmoazeni: @tebriel I think I'm going to punt this to a followup. Since it will just change how we load/conditional-ize this stuff. But the logic should still run.

lib/jira/client/index.js
```diff
@@ -141,6 +142,8 @@ module.exports = async (jiraHost, gitHubInstallationId, logger) => {
   return client
 }
 
+module.exports = async (...args) => newrelic.startSegment('lib/jira/client: getJiraClient', true, async () => getJiraClient(...args))
```
zmoazeni: @tebriel and I talked on Zoom about how this is more appropriate to have the tracing code be part of the function definition rather than the consuming code. I totally agree with him.

This isn't the prettiest code, but it's a reusable away for us to wrap tracing external to the function definition. Rather than pushing it INSIDE the definition of `getJiraClient` here. I could see this potentially being abstracted to be reused across the codebase.

zmoazeni:
@tebriel lots of things have changed here per our discussions. Let me know what you think.

lib/jira/client/index.js
tebriel: Great work! I think once we get a few more of these under our belt we can probably create a function to wrap this to make it prettier. like `wrapAsync(name, func)` thinking like the following:

```js
module.exports = NRWrapAsync('lib/jira/client: getJiraClient', getJiraClient);
```

But not important now and doesn't really apply yet!

zmoazeni: Yeah 100%. I'm sure we'll start reusing this very soon, and I'm all for a way to push this cruft into one spot instead of copy/paste boilerplate all over.

PR #328, opened 2020-03-25
Stop logging undefined actions

zmoazeni:
Some webhook events are printing like `push.undefined` which we don't need.

Others make sense like `pull_request.closed` so lets keep that

### DIFF
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -42,7 +42,7 @@ module.exports = function middleware (callback) {
 
     newrelic.setControllerName(`github/events.${context.name}`)
     let webhookEvent = context.name
-    if (context.payload.action !== undefined) {
+    if (context.payload.action) {
       webhookEvent = `${webhookEvent}.${context.payload.action}`
     }
     newrelic.addCustomAttributes({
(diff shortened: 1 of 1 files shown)


lib/github/middleware.js
```diff
@@ -41,9 +41,13 @@ module.exports = function middleware (callback) {
     enhanceOctokit(context.github, context.log)
 
     newrelic.setControllerName(`github/events.${context.name}`)
+    let webhookEvent = context.name
+    if (context.payload.action !== undefined) {
+      webhookEvent = `${webhookEvent}.context.payload.action`
```
jules2689:
```suggestion
      webhookEvent = `${webhookEvent}.${context.payload.action}`
```

I think you wanted this?

On another note, would it make sense to make this `Webhook Name: context.name` and `Action: context.payload.action` (if applicable)?

zmoazeni: Lol yeah that should be changed.

> On another note, would it make sense to make this Webhook Name: context.name and Action: context.payload.action (if applicable)?

I thought about that. But I preferred the single line dot notation myself. It aligns with how we subscribe to events with the robot.

lib/github/middleware.js
```diff
@@ -41,9 +41,13 @@ module.exports = function middleware (callback) {
     enhanceOctokit(context.github, context.log)
 
     newrelic.setControllerName(`github/events.${context.name}`)
+    let webhookEvent = context.name
+    if (context.payload.action !== undefined) {
```
tebriel: I think this should be `!= null` in case the key exists in the dictionary but has a null value.

zmoazeni: I just updated to check for truthiness instead.

PR #329, opened 2020-03-25
Adds ability for us to timeout webhook handlers

zmoazeni:
As far as I can tell, we can't kill the request from the outside, but instead we have to cooperatively kill from within the logic. Meaning we need to find and add checks like this wherever we have long running code.

This doesn't currently have any tests. I only tested it locally by adding:

```
    function sleep (ms) {
      return new Promise((resolve) => {
        setTimeout(resolve, ms)
      })
    }
    await sleep(5000)
```

And tweaking the `REQUEST_TIMEOUT_MS` in my .env

### DIFF
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -79,18 +79,16 @@ module.exports = function middleware (callback) {
 
     context.sentry.setTag('transaction', `webhook:${context.name}.${context.payload.action}`)
     for (let subscription of subscriptions) {
-      if (context.timedout) {
-        // also send this to sentry
-        context.log.error({ timeout: true, timeoutElapsed: context.timedout }, `Timing out at after ${context.timedout}ms`)
-        return
-      }
       const jiraHost = subscription.jiraHost
-
       context.sentry.setTag('jiraHost', jiraHost)
       context.sentry.setTag('gitHubInstallationId', gitHubInstallationId)
       context.sentry.setUser({ jiraHost, gitHubInstallationId })
-
       context.log = context.log.child({ jiraHost })
+      if (context.timedout) {
+        Sentry.captureException(new Error('Timed out jira middleware iterating subscriptions'));
+        context.log.error({ timeout: true, timeoutElapsed: context.timedout }, `Timing out at after ${context.timedout}ms`)
+        return
+      }
 
       const jiraClient = await getJiraClient(jiraHost, gitHubInstallationId, context.log)
       if (jiraClient == null) {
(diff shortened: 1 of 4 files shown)


lib/github/middleware.js
```diff
@@ -80,12 +80,15 @@ module.exports = function middleware (callback) {
     context.sentry.setTag('transaction', `webhook:${context.name}.${context.payload.action}`)
     for (let subscription of subscriptions) {
       const jiraHost = subscription.jiraHost
-
       context.sentry.setTag('jiraHost', jiraHost)
       context.sentry.setTag('gitHubInstallationId', gitHubInstallationId)
       context.sentry.setUser({ jiraHost, gitHubInstallationId })
-
       context.log = context.log.child({ jiraHost })
+      if (context.timedout) {
```
zmoazeni, line 87: Timing out needs to be handled cooperatively. So we have to opt-in and check if we're timed out each time.

lib/github/middleware.js
```diff
@@ -80,12 +80,15 @@ module.exports = function middleware (callback) {
...
+        Sentry.captureException(new Error('Timed out jira middleware iterating subscriptions'));
```
zmoazeni, line 88: This isn't the greatest, but I'd like to send this to sentry too. This to me is an actionable error for us to work on.

I think there is value in abstracting this later if/when we start reusing this pattern more.

test/middleware/webhook-timeout.test.js
```diff
@@ -0,0 +1,32 @@
+const webhookTimeout = require('../../lib/middleware/webhook-timeout')
+
+describe('Webhook Timeout', () => {
+  it('sets timedout context with seconds', async () => {
+    const context = {}
+    await webhookTimeout(async () => { await sleep(3) }, 1)(context)
```
zmoazeni, line 6: Just a reminder that `sleep(3)` means sleep 3 milliseconds. I'm not a huge fan of sleeping within tests, but sleeping milliseconds feels alright.

jules2689: I thought it was sleep 3 seconds for a moment 😆 phew! yea, that seems fine to me!

zmoazeni: yeah I used `ms` as the argument to sleep, but I thought about changing it to `sleepMs` instead

zmoazeni: Ah I actually like that better and that's a little more Node-y too. Thanks!

zmoazeni: I talked with @jules2689 in slack. I think this is a good suggestion. But I'm struggling with how to change up the tests. I definitely welcome any help here.

There's effectively two callbacks I care about here:

1. That the timedout is set within the setTimeout
2. That the timer is cleared in both a succesful and error case

I think at least the first one needs to be some sort of async function that we wait on?

I may also still need to use `sleep` at least for the timed out scenario (since that's being called within the closure being wrapped).

I'll mull on it overnight and see if any inspiration comes to me.

❤️ for the suggestion @jules2689

PR #332, opened 2020-04-07
Sets originalException

zmoazeni:
The hope here is that Sentry lists out both stack traces. I believe I've seen that before

The current issue is that sentry stack traces point to where we wrap the new OctokitError rather than where the error originated:

:ref responded with 401 2020-04-07 16-09-30

More information on Sentry hints: https://docs.sentry.io/platforms/node/

> originalException
>
>    The original exception that created the event. This is useful for changing how events are grouped, or to extract additional information.

I believe I've seen this before in Go-land. I think sentry will list both stack traces and this allows us to avoid munging any error's stack trace.

---

I should probably try this out locally and see it happen in Sentry but I feel comfortable enough yolo'ing this and checking the results in production.


zmoazeni:
This unfortunately didn't work. I'll revert and try again. This time testing from local.

PR #334, opened 2020-04-08
Also clean metrics queue

zmoazeni:
We have lots of keys regarding Metrics jobs, this should start purging them.

I'm still planning following up to use `removeOnComplete` separately https://github.com/OptimalBits/bull/issues/840 so this happens automatically


zmoazeni:
This dropped about 27.5k keys.

Issue #280, opened 2019-09-23
Switch all bull jobs to use `removeOnComplete`


zmoazeni:
It looks like you have to opt-in for bull (our queuing library) to delete completed jobs.

* https://github.com/integrations/jira/pull/278#pullrequestreview-292071108
* https://github.com/OptimalBits/bull/issues/840

This is a fun sized issue that can help us keep Redis's memory bounded.

PR #337, opened 2020-04-17
Remove jobs from redis by default

zmoazeni:
We no longer need to manually clean these jobs from redis.

I tested this out locally by verifying the keys were trimmed appropriately.

Before I deploy this I will drop the scheduled job in Heroku.

fixes: #280

### DIFF
--- a/lib/worker/index.js
+++ b/lib/worker/index.js
@@ -22,12 +22,14 @@ function measureElapsedTime(startTime, tags) {
   statsd.histogram('job_duration', timeDiff, tags);
 }
 
+const queueOpts = { defaultJobOptions: { removeOnComplete: true } };
+
 // Setup queues
 const queues = {
-  discovery: new Queue('Content discovery', REDIS_URL),
-  installation: new Queue('Initial sync', REDIS_URL),
-  push: new Queue('Push transformation', REDIS_URL),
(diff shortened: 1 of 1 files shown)


zmoazeni:
@tebriel From reading the history of https://github.com/OptimalBits/bull/issues/840 it seems like it was added to `bull` later in the history. But it is NOT a default. I personally disagree with that stance. 

I suppose that leaving it in does allow some observability within redis if we want to explore past jobs. Though I haven't used it for that.

I just don't feel good about using redis that way long term. I would prefer we move those types of observability to either the logs or datadog, or something else entirely. Or bull adds an EXPIRE to these keys so they are eventually purged.

It also ends up being a forgotten dumping ground of redis keys for things like https://github.com/integrations/jira/pull/334 where we just forget to also add a new queue to be cleaned and end up having 27.5k keys being purged at once.

tebriel:
Awesome! Just didn’t know if we’d be breaking any of our debug steps or something. Good to go from my end.

zmoazeni:
@tebriel Yeah I think this change is innocuous on that front. Thanks for checking!!

zmoazeni:
Just a heads up. I force pushed to keep the script and the `clean` function for now. I'm going to run it once more manually after this goes live. Then I'll PR to remove it too.

PR #340, opened 2020-04-23
Records worker code in NewRelic

zmoazeni:
This allows us to keep an eye on worker jobs and profile slow code.

Here are a couple screenshots to help see what this adds in NewRelic:

Transactions - slack-app-tebriel - New Relic 2020-04-23 15-17-19

Transactions - slack-app-tebriel - New Relic 2020-04-23 15-18-02

Transactions - slack-app-tebriel - New Relic 2020-04-23 15-25-22


lib/worker/index.js
```diff
@@ -83,16 +88,53 @@ const sentryMiddleware = function (jobHandler) {
...
+
+/**
+ * Return an async function that sends timing data to NewRelic
+ *
+ * @param {Promise<object>} jobHandler - The normal job handler code function that we are wrapping
+ */
+function newrelicMiddleware(jobHandler) {
+  return async (job) => {
+    newrelic.startBackgroundTransaction(`job ${job.queue.name}`, 'worker queue', async () => {
+      app.log('ZMOTEST', { here: 'true', data: job.data, opts: job.opts });
```
srt32, line 101:
```suggestion
```

lib/worker/index.js
```diff
@@ -83,16 +88,53 @@ const sentryMiddleware = function (jobHandler) {
...
+}
+
+/**
+ * Return an async function that sends timing data to NewRelic
+ *
+ * @param {Promise<object>} jobHandler - The normal job handler code function that we are wrapping
+ */
+function newrelicMiddleware(jobHandler) {
+  return async (job) => {
+    newrelic.startBackgroundTransaction(`job ${job.queue.name}`, 'worker queue', async () => {
```
srt32, line 100: Is there any concern about this call failing for weird reasons?

lib/worker/index.js
zmoazeni: Gah! Thanks! Sorry!

lib/worker/index.js
zmoazeni: @srt32 "Shouldn't", but those are famous last words. This is wrapping the bull job, and this is using the standard API https://github.com/OptimalBits/bull/blob/develop/REFERENCE.md

That said, this is _also_ wrapped by our Sentry middleware. So if it does fail, it should be captured and raised to Sentry.

Great question.

lib/worker/index.js
```diff
@@ -59,8 +60,12 @@ Object.keys(queues).forEach(name => {
   });
 });
 
-// Return an async function that assigns a Sentry hub to `job.sentry` and sends exceptions.
-const sentryMiddleware = function (jobHandler) {
+/**
+ * Return an async function that assigns a Sentry hub to `job.sentry` and sends exceptions.
+ *
+ * @param {Promise<object>} jobHandler - The normal job handler code function that we are wrapping
```
tebriel, line 66: I love the use of JSDoc here but I'm not quite sure this is a `Promise<object>` I think it's a callback.

A freestanding comment of the following could define the callback. I used `unknown` which is a valid type and is a stand-in for "we'll figure out what it is later". Feel free to specify the type if you know what it is, or if you want to say "it can be anything", an `any` type would be more appropriate I think!
```js
/**
 * This callback is displayed as a global member.
 * @callback JobHandler
 * @param {unknown} job
 * @returns {Promise<unknown>}
 */
```

Then your `@param` would look like:
```js
/**
 * Return an async function that assigns a Sentry hub to `job.sentry` and sends exceptions.
 *
 * @param {JobHandler} jobHandler - The normal job handler code function we are wrapping
 */
```

zmoazeni: @tebriel Hrm, I'll have to check on some types here. But I thought `Promise<object>` was appropriate based on https://devhints.io/jsdoc

I was treating this as "a Promise fulfilled by an object" in this case the "object" is the `job` from bull.

Here's an example of what we're annotating: https://github.com/integrations/jira/blob/94cad140cea8266976265d9b1e5ce96625eefa6c/lib/sync/discovery.js#L11

tebriel: Right! So `{Promise<object>}` is the `@returns` result of `jobHandler`. `jobHandler` (if I'm reading this right) is a `Function`. Which could be documented as `{Function}` as well (but doesn't convey as much information). It's an `object` definitely, because, well, everything is an object in javascript. To indicate that a Function is async, you set the `returns` value as a `Promise` which says "I'm an async function, or at least I return something `await`-able".

zmoazeni: I paired with @tebriel last week and he leveled me up here. Thanks!

zmoazeni: I've dropped the types for now. There's a few ways @tebriel and I talked about documenting them and I'd like for us to pair once more.

I'll reintroduce them as a followup. I'd like to get this NewRelic behavior in production so we can begin recording performance.

PR #341, opened 2020-04-23
Move webhook handlers directly to bg job for pullRequest

zmoazeni:
This moves the processing of just `pullRequest` events to a background job. There's still lots that we can do in this realm, but I'm hoping this can be a one-by-one transition of our webhook handlers to address our load issues.

### DIFF
--- a/lib/github/index.js
+++ b/lib/github/index.js
@@ -17,8 +17,11 @@ module.exports = (robot) => {
 
   robot.on('push', middleware(push));
 
-  // robot.on(['pull_request.opened', 'pull_request.closed', 'pull_request.reopened', 'pull_request.edited'], middleware(pullRequest));
-  robot.on(['pull_request.opened', 'pull_request.closed', 'pull_request.reopened', 'pull_request.edited'], workerMiddleware('pullRequest'));
+  if (process.env.NO_BACKGROUND_WEBHOOKS === 'true') {
+    robot.on(['pull_request.opened', 'pull_request.closed', 'pull_request.reopened', 'pull_request.edited'], middleware(pullRequest));
+  } else {
+    robot.on(['pull_request.opened', 'pull_request.closed', 'pull_request.reopened', 'pull_request.edited'], workerMiddleware('pullRequest'));
+  }
 
   robot.on('create', middleware(createBranch));
   robot.on('delete', middleware(deleteBranch));
--- a/lib/github/index.js
+++ b/lib/github/index.js
@@ -1,6 +1,6 @@
 const issueComment = require('./issue-comment');
 const issue = require('./issue');
-const middleware = require('./middleware');
+const { middleware, workerMiddleware } = require('./middleware');
 const pullRequest = require('./pull-request');
 const push = require('./push');
 const { createBranch, deleteBranch } = require('./branch');
@@ -17,7 +17,8 @@ module.exports = (robot) => {
 
   robot.on('push', middleware(push));
 
-  robot.on(['pull_request.opened', 'pull_request.closed', 'pull_request.reopened', 'pull_request.edited'], middleware(pullRequest));
+  // robot.on(['pull_request.opened', 'pull_request.closed', 'pull_request.reopened', 'pull_request.edited'], middleware(pullRequest));
+  robot.on(['pull_request.opened', 'pull_request.closed', 'pull_request.reopened', 'pull_request.edited'], workerMiddleware('pullRequest'));
 
   robot.on('create', middleware(createBranch));
   robot.on('delete', middleware(deleteBranch));
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -135,18 +135,18 @@ function workerMiddleware(callbackName) {
     context.log = context.log.child({ gitHubInstallationId });
 
     if (context.payload.sender.type === 'Bot') {
-      context.log({ noop: 'bot', botId: context.payload.sender.id, botLogin: context.payload.sender.login }, 'Halting futher execution since the sender is a bot');
+      context.log.info({ noop: 'bot', botId: context.payload.sender.id, botLogin: context.payload.sender.login }, 'Halting futher execution since the sender is a bot');
       return;
     }
 
     if (isFromIgnoredRepo(context.payload)) {
-      context.log({ noop: 'ignored_repo', installation_id: context.payload.installation.id, repository_id: context.payload.repository.id }, 'Halting futher execution since the repository is explicilty ignored');
+      context.log.info({ noop: 'ignored_repo', installation_id: context.payload.installation.id, repository_id: context.payload.repository.id }, 'Halting futher execution since the repository is explicilty ignored');
       return;
     }
 
     const count = await Subscription.count({ where: { gitHubInstallationId } });
     if (count < 1) {
-      context.log({ noop: 'no_subscriptions' }, 'Halting futher execution since no subscriptions were found');
+      context.log.info({ noop: 'no_subscriptions' }, 'Halting futher execution since no subscriptions were found');
       return;
     }
 
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -7,6 +7,7 @@ const getJiraClient = require('../jira/client');
 const getJiraUtil = require('../jira/util');
 const enhanceOctokit = require('../config/enhance-octokit');
 const newrelic = require('newrelic');
+const { queues } = require('../worker');
 
 // Returns an async function that reports errors errors to Sentry.
 // This works similar to Sentry.withScope but works in an async context.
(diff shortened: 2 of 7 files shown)


lib/github/index.js
```diff
@@ -17,7 +17,8 @@ module.exports = (robot) => {
 
   robot.on('push', middleware(push));
 
-  robot.on(['pull_request.opened', 'pull_request.closed', 'pull_request.reopened', 'pull_request.edited'], middleware(pullRequest));
+  // robot.on(['pull_request.opened', 'pull_request.closed', 'pull_request.reopened', 'pull_request.edited'], middleware(pullRequest));
+  robot.on(['pull_request.opened', 'pull_request.closed', 'pull_request.reopened', 'pull_request.edited'], workerMiddleware('pullRequest'));
```
zmoazeni, line 21: In my mind, the ideal here would be to start with one handler (on of our bigger ones) and move over bit by bit to the background queue.

lib/sync/webhook.js
```diff
@@ -0,0 +1,75 @@
...
+
+    const jiraClient = await getJiraClient(jiraHost, gitHubInstallationId, context.log);
+    if (jiraClient == null) {
+      // Don't call callback if we have no jiraClient
+      context.log.error({ noop: 'no_jira_client' }, `No enabled installation found for ${jiraHost}.`);
+      return;
+    }
+    const util = getJiraUtil(jiraClient);
+
+    const callback = pullRequest;
```
zmoazeni, line 65: Yuuup this is ugly, but the idea is that with `callbackName` we can switch on the appropriate handler.

I haven't tried serializing a function into Bull, but that seems somewhat dangerous since the function body _could_ change between queue and run. (if a deploy happens)

lib/worker/index.js
```diff
@@ -98,14 +100,20 @@ function sentryMiddleware(jobHandler) {
 function newrelicMiddleware(jobHandler) {
   return async (job) => {
     newrelic.startBackgroundTransaction(`job ${job.queue.name}`, 'worker queue', async () => {
-      app.log('ZMOTEST', { here: 'true', data: job.data, opts: job.opts });
       const transaction = newrelic.getTransaction();
+
+      const jobData = Object.assign({}, job.data);
+      // Don't log webhook payload to newrelic
+      if (jobData.context) {
+        jobData.context = '[FILTERED]';
```
zmoazeni, line 108: I made sure to filter out the payload info out of newreilc

lib/github/middleware.js
```diff
@@ -105,4 +106,56 @@ module.exports = function middleware(callback) {
       }
     }
   });
+}
+
+
+function workerMiddleware(callbackName) {
```
zmoazeni, line 112: This is basically a copy/paste of the `middleware` function above. Minus a different ending.

I would think that duplication would live only a short while during our transition.

lib/github/middleware.js
```diff
@@ -105,4 +106,56 @@ module.exports = function middleware(callback) {
...
+      context.log({ noop: 'ignored_repo', installation_id: context.payload.installation.id, repository_id: context.payload.repository.id }, 'Halting futher execution since the repository is explicilty ignored');
+      return;
+    }
+
+    // TODO just check the count. Don't fetch all of them.
+    const subscriptions = await Subscription.getAllForInstallation(gitHubInstallationId);
+    if (!subscriptions.length) {
+      context.log({ noop: 'no_subscriptions' }, 'Halting futher execution since no subscriptions were found');
+      return;
+    }
```
zmoazeni, line 152: I am keeping at least one query here in this handler, but it's to check if we have any subscriptions for the installation id. I imagine this will cut down the queuing _considerably_.

lib/github/index.js
```diff
@@ -17,7 +17,11 @@ module.exports = (robot) => {
 
   robot.on('push', middleware(push));
 
-  robot.on(['pull_request.opened', 'pull_request.closed', 'pull_request.reopened', 'pull_request.edited'], middleware(pullRequest));
+  if (process.env.NO_BACKGROUND_WEBHOOKS === 'true') {
```
zmoazeni, line 20: I'm going to keep both paths in production code, but we can quickly back out my worker changes if we see something bad happen.

lib/github/middleware.js
```diff
@@ -105,4 +106,55 @@ module.exports = function middleware(callback) {
...
+
+    const count = await Subscription.count({ where: { gitHubInstallationId } });
+    if (count < 1) {
+      context.log.info({ noop: 'no_subscriptions' }, 'Halting futher execution since no subscriptions were found');
+      return;
+    }
+
+    await queues.webhook.add({ callbackName, context });
```
tebriel, line 153: instead of having two copies of this function, would it be possible to to the if env var here instead? Basically:
```
If use_bull_queue:
  await queues.webhook.add
else:
  original logic
```

Since all the logic is the same, it's just encapsulated in another file, you could call that other file for either. It makes me feel safer considering possible drift where one function gets updated but not the other, also means that all the original tests will be touching the code that will be running, no matter the flag.

zmoazeni: @tebriel can you expand a bit more on what you're thinking here and the motivation? The way we call these functions are pretty different. I did add an ENV switch but it's to use the inline vs async versions.

My approach here is to temporarily keep this not-DRY in a short term as we transition, then deleting the original `middleware` function.

lib/sync/webhook.js
```diff
@@ -0,0 +1,75 @@
...
+    const util = getJiraUtil(jiraClient);
+
+    const callbacks = {
+      pullRequest,
+    };
+
+    const callback = callbacks[callbackName];
+    if (!callback) {
+      throw new Error(`Unknown callback: ${callbackName}. Known callbacks: ${Object.keys(callbacks)}`);
+    }
```
zmoazeni, line 66: This should raise some visibility if we end up enqueuing jobs with improper names for the callbacks.

test/setup/env.js
```diff
@@ -12,6 +12,10 @@ const defaults = Object.assign({
   // Generate a secure, random one with `openssl rand -hex 32` in your .env file
   STORAGE_SECRET: '8cad66340bc92edbae2ae3a792d351f48c61d1d8efe7b2d9408b0025c1f7f845',
   SETUP: 'yes', // indicates that the setup did run
+
+  // use a different database than development
+  // that way tests don't enqueue jobs that development tries to run
+  REDIS_URL: 'redis://127.0.0.1:6379/2',
```
zmoazeni, line 18: I found this out while testing. Our tests currently enqueue all redis jobs in the same space that the development server looks to work them off. I'm using a different database here (`2`) to isolate it. I'm also clearing the database between individual test runs.

zmoazeni:
Hrm. Looking into these push test failures.

lib/github/middleware.js
tebriel: Yeah, so my thought was: we moved all of the logic from `middleware.js` after the subscription select, to another file. It's doing the same thing, but one has some extra decoration. If we could encapsulate that logic into a single function, then the original path could call it and the new path could also call it (but maybe with some pre-work). This way we make sure there weren't any typos/changes/missed scope variables/etc from the logic that gets moved into the new file. This is just the core business logic of the endpoint, so it seems like it should be the same for sync or bull-queue processing.

zmoazeni: @tebriel and I talked over zoom on this. Recapping what we talked about here.

I totally agree with him that copy/pasting code is not a good long term solution. I'm trying to figure out the simplest change (and trying to avoid any refactors if I can) to have production. My gut says that backgrounding these requests will lead to the Jira integration being more stable, but I'm not certain. And I'm open to being wrong (and adjusting).

I'm hoping that next week I'll get that confidence. And I'll also expand this to:

* branch creation
* commit pushes

Which lead our "time spent" in NewRelic traces.

I also want to put together a datadog monitor before this goes out so that I can keep tabs on the queue in case we are not working it off fast enough.

Ideally this original method will be deleted by the end of next week. I do have a calendar reminder for two weeks though for me to hold me accountable to that. I do not want this code to be duplicated long-term.

@tebriel feel free to mention anything I may have missed! Thanks for the chat! and the suggestion!

lib/sync/webhook.js
tebriel: Should we do this at the beginning of the function? No reason to do all this work to get down here to just find out we can't process it, right?

zmoazeni: Holy cow. 100%. Great call @tebriel

zmoazeni:
Just a heads up for anyone following this. We ware going to temporarily deprioritize this effort in favor of some other Jira Observability. I'm going to leave this open, but I hope to return to it next week.

zmoazeni:
Closing since we decided on a different direction.

PR #342, opened 2020-04-24
Profile all Incoming Probot Events in NewRelic

tebriel:
Currently we have a number of events that are just bucketed under `/ post` which take up a lot of time. Let's bucket them all so we know which events are coming in that we're ignoring (and how much time we're spending with them).

Additionally, `package.json` specifies the latest nodejs within the 12.x LTS series, `nodemon` wouldn't install for me on 12.16.2 because of an `fsevents` dependency. I bumped `nodemon` to the 2x series (latest) to get a newer version that worked (plus has lower memory requirements).


lib/github/index.js
```diff
@@ -6,8 +7,15 @@ const push = require('./push');
...
 
+/**
+ * @param {import('probot').Application} robot - The probot instance
+ */
 module.exports = (robot) => {
+  /**
+   * @param {import('probot').Context} context - The incoming request context.
+   */
   robot.on('*', (context) => {
+    newrelic.setControllerName(`github/events.${context.name}`);
```
srt32, line 18: Would we want `context.event` in here too?

tebriel: Ahh looks like `name` is deprecated in favor of `event` https://github.com/probot/probot/blob/6e45ad7/src/context.ts#L90-L92 I'll update the pr

PR #343, opened 2020-04-28
Upgrade Probot to v9

tebriel:
## Overview

Probot 7.5.3 was last updated about a year ago, and probot 9 (8 was skipped) has a whole lot of under-the-hood octokit improvements. Not the least of which is [@octokit/plugin-throttling](https://www.npmjs.com/package/@octokit/plugin-throttling) which uses bottleneck. We can configure this bottleneck plugin to use redis to collaborate on shared limits (which will be nice to prevent one worker from running away with the farm).

### Scope

I tried to scope this PR to a few things:

1.  Follow any removals listed on the [Probot Version 9.0.0 Changelog](https://github.com/probot/probot/releases/tag/v9.0.0)
2.  Replace all `@octokit/rest.js` deprecations from [v15.18.0](https://github.com/octokit/rest.js/releases/tag/v15.18.0#deprecations) as they're fully removed from probot.

### Unrelated (sorta) changes

There are a couple of non-related changes that were super helpful for me while debugging the upgrade, but I can be convinced to revert:
1.  Upgrading Jest (no code changes, but gave better errors/stack traces)
1.  Upgrading SuperTest (same reason)
1.  Upgrading eslint-plugin-import (the console complained about it a lot)

### Key Changes

There are a few major changes in this release:
1.  Method names of rest.js APIs that changed
1.  The way we instantiate `@octokit/app.js` instances. We no longer do a deep import from probot internals to create an `App`.
1.  The way we instantiate `@octokit/rest`. Almost all the time we specify `auth` as it handles most of our cases.

### Side Effects

Upgrading probot and a few other deps gives us a *lot* of security fixes.

#### Probot v7 (current package.json)

```
found 42249 vulnerabilities (42227 low, 18 moderate, 4 high)
  run `npm audit fix` to fix them, or `npm audit` for details
```

#### Probot v9 (And misc package.json changes)
```
found 11 vulnerabilities (8 low, 3 moderate)
  run `npm audit fix` to fix them, or `npm audit` for details
```

### Up Next

In trying to keep the scope of these changes as small as possible, I did not connect the bottleneck throttling to redis, nor did I modify any of the throttling settings. That will come in a follow-up PR to try to keep things separate (and make debugging if something goes wrong easier).


lib/frontend/github-client-middleware.js
```diff
@@ -1,21 +1,18 @@
...
-
   if (req.session.githubToken) {
-    res.locals.github.authenticate({
-      type: 'token',
-      token: req.session.githubToken,
+    res.locals.github = new Octokit({
+      auth: req.session.githubToken,
     });
+  } else {
+    res.locals.github = new Octokit();
```
tebriel, line 9: this is an unauthenticated client, which was the case before if `req.session.githubToken` didn't exist.

lib/run.js
```diff
@@ -29,7 +29,11 @@ function start() {
   probot.start();
 }
 
-throng({
-  workers,
-  lifetime: Infinity,
-}, start);
+if (workers === 1) {
```
tebriel, line 32: This is what we do in the Slack integration to make debugging locally easier (only 1 process, throng with 1 worker still starts 2 processes).

test/middleware/webhook-timeout.test.js
```diff
@@ -5,7 +5,7 @@ describe('Webhook Timeout', () => {
     const context = {};
     await webhookTimeout(async () => { await sleep(3); }, 1)(context);
     expect(context.timedout).toBeGreaterThan(0);
-    expect(context.timedout).toBeLessThan(15);
+    expect(context.timedout).toBeLessThan(100);
```
tebriel, line 8: Timing was all over the place on my box, this just gives the test more headroom. I'm not sure we should be testing that it's less than anything, since I'm not sure what that buys us (and jest times out after 5s) but I didn't want to remove it.

test/setup/app.js
```diff
@@ -46,14 +46,17 @@ beforeEach(() => {
...
       id: 12257,
-      cert: findPrivateKey(),
+      privateKey: findPrivateKey(),
     }),
     cache: cacheManager.caching({
       store: 'memory',
       ttl: 60 * 60, // 1 hour
     }),
+    throttleOptions: {
+      enabled: false,
```
tebriel, line 58: Don't throttle the tests! This is a huge speedup in test runs.

lib/run.js
```diff
@@ -10,7 +10,7 @@ const { findPrivateKey } = require('probot/lib/private-key');
 const { createProbot } = require('probot');
 const app = require('./configure-robot');
 
-const workers = process.env.WEB_CONCURRENCY || 1;
+const workers = Number(process.env.WEB_CONCURRENCY || 1);
```
srt32, line 13: This worked before?

.gitignore
```diff
@@ -9,3 +9,4 @@ log/*
 .envrc
 .tool-versions
 newrelic_agent.log
+.vscode
```
zmoazeni, line 12: This is minor (and ignorable). I've worked with teams who prefer not to have editor files as part of .gitignore, and instead for each person using the editor to add these entries to `.git/info/exclude` (a local-only .gitignore file).

I don't have a strong preference one way or another, but wanted to call it out in case that's a similar path we want to take.

test/middleware/webhook-timeout.test.js
zmoazeni: I'm working on revamping these assertions so we don't sleep at all. But I would be fine with us dropping this particular assertion. You're right it doesn't really give us much confidence and it's brittle.

lib/run.js
tebriel: Yes! `throng` will cast it to a `Number` if we pass in a string, but since I wanted to do the comparison at the end of this file, I needed it to be a `Number` ahead of passing it to `throng`.

.gitignore
tebriel: Hmm, yeah. Maybe we need a wider team discussion on that. `slack` has `.vscode` in the `.gitignore`. Generally I feel like editor configurations are very user-specific. That's partially because I end up running the app differently than the rest of the team on both slack and jira haha. I can leave it in and locally ignore it, but I significantly changed the `launch.json` file for my own needs and I had to hard-code the path to my local `node` install (still haven't figured out why yet) so that was my motivation for removing. LMK if you want me to add it back.

zmoazeni: I think this is all good for now. Yeah let's bubble this up to a team discussion to see if there's consensus one direction or another.

Honestly it doesn't seem like there's a ton of negative consequences one way or another (at least that I'm seeing, but I could be missing something).

jasonkarns: There's also the system level excludes file. IDE (and OS) -specific ignores will likely be identical for a given user on a given machine, so adding it to the system level ignore file means you don't need to add it manually to `.git/info/exclude` in every repo on your system.

I'm also a pretty staunch supporter of minimal project gitignores in order to elevate the signal-to-noise ratio.

tebriel: I think that's a reasonable stance!

PR #348, opened 2020-04-30
Unify GitHub API Invocations

tebriel:
We instantiate the `@octokit/rest` `Oktokit` object a **lot**. In addition to that, we aren't using `probot`'s one, which has some nice rate-limiting, retry, and other features built in. This starts the foundation for letting us have a unified feature/config set for those octokit APIs by fetching them from one place. It special cases some configuration for the `test` node environment, but I'm expecting that we'll soon add some profiling to the `throttle` options which will go in here too.

Also fun that was had: the probot octokit auto-retries if a non-200 response was returned. Which made nock very sad because it was like "you told me there'd only be 1 request, but there were 3" and then stuff was failing. This should make the test more consistent.

### DIFF
--- a/test/unit/api/api.test.js
+++ b/test/unit/api/api.test.js
@@ -43,15 +43,19 @@ const createApp = (locals) => {
 
 describe('API', () => {
   describe('Authentication', () => {
-    const app = createApp();
+    let app;
 
-    it('should return 404 if no token is provided', () => supertest(app)
-      .get('/')
-      .set('Authorization', 'Bearer xxx')
-      .expect(404)
-      .then(response => {
-        expect(response.body).toMatchSnapshot();
-      }));
+    beforeAll(() => {
+      app = createApp();
+    });
+
+    it('should return 404 if no token is provided', () =>
+      supertest(app)
+        .get('/api')
+        .expect(404)
+        .then(response => {
+          expect(response.body).toMatchSnapshot();
+        }));
 
     it('should return 200 if a valid token is provided', () => {
       nock('https://api.github.com').post('/graphql').reply(200, successfulAuthResponseWrite);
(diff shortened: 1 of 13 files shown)


test/unit/api/api.test.js
```diff
@@ -43,15 +43,19 @@ const createApp = (locals) => {
...
-      .expect(404)
-      .then(response => {
-        expect(response.body).toMatchSnapshot();
-      }));
+    beforeAll(() => {
+      app = createApp();
+    });
+
+    it('should return 404 if no token is provided', () =>
+      supertest(app)
```
tebriel, line 53: This test passed due to the fact that the old url it was testing was `/` and that route wasn't ever set up. Now setting to `/` and removed the `Authorization` header to test what the test meant to test.


....Test.

test/unit/api/api.test.js
```diff
@@ -153,14 +157,6 @@ describe('API', () => {
           expect(response.body.HttpError).toMatchSnapshot();
         });
     });
-
-    it('should return 500 if an error happens', () => supertest(app)
```
tebriel, line 157: Probot won't throw an error on a 500 like it did in 7x. I couldn't find a way to replicate this situation anymore.

test/unit/jira/client/axios.test.js
```diff
@@ -64,7 +64,7 @@ describe('Jira axios instance', () => {
             status: '500',
             env: 'test',
           },
-          value: (value) => value > 0 && value < 100,
+          value: (value) => value > 0 && value < 1000,
```
jules2689, line 67: what is this?

tebriel: this is a test of how long it took the test framework to actually run the call and return the response. I *really* dislike that we're testing the duration of the request here, but I'm not ready to rip these out, so I'm just widening the value because as long as it's > 0 it's technically OK, anything over 5k will trigger jest to fail the test.

PR #350, opened 2020-05-01
Use Bull's suggestions on limiting number of connections

tebriel:
We are limited in # of allowed connections in Heroku's Redis, let's use Bull's suggestion on how to limit connections:
https://github.com/OptimalBits/bull/blob/master/PATTERNS.md#reusing-redis-connections

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
     "express-sslify": "^1.2.0",
     "helmet": "^3.21.2",
     "hot-shots": "^7.4.1",
+    "ioredis": "^4.16.3",
     "markdown": "^0.5.0",
     "moment": "^2.24.0",
     "moo": "^0.5.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -30,6 +30,7 @@
     "express-sslify": "^1.2.0",
     "helmet": "^3.21.2",
     "hot-shots": "^7.4.1",
+    "ioredis": "^4.16.3",
```
tebriel, line 33: bull already requires ioredis, but since we `require` it now in our code, we need it to be in package.json

tebriel:
|Environment|Connections Before|Connections After|Web Throng Workers|Worker Throng Workers|Bull Concurrent Workers|
|-|-|-|-|-|-|
|Dev|21|11|1|1|1|
|Staging|29|19|1|1|1|
|Prod|441|313|36|6|10|
Saw Max of 837 during Prod Deploy (previously had been > 1000)

PR #353, opened 2020-05-04
Consolidate Redis Settings to set Client Name

tebriel:
In an effort to determine why we're using so many redis connections, this PR sets the `connectionName` parameter everywhere we're connecting to Redis. This should allow us to run `CLIENT LIST` in redis and aggregate based on what we're using.

Also:

*  Lint `bin/worker` with `npm run lint --fix` (but also include `bin/worker` in the `npm run lint` list of files.
*  Explicitly list `Bottleneck` in the project requirements because we now require it directly.

Side notes:
This will not support a database specification (for now) we can add it when we start using one.


tebriel:
The worker for some reason isn't connecting to REDIS, will need to investigate

PR #355, opened 2020-05-06
Documentation and Dep Improvements

tebriel:
*  Bump Axios to 0.19.2 for bugfixes
*  JSDoc with types in axios.js to better understand usage
*  Use `allSettled` instead of `all` to ensure we complete all promises.
*  Logs out errors from Jira API so that we can better understand why there was an error.

### DIFF
--- a/lib/jira/client/axios.js
+++ b/lib/jira/client/axios.js
@@ -76,7 +76,7 @@ function getErrorMiddleware(logger) {
         }, `Jira request: ${error.request.method} ${error.request.path} - ${status} ${statusText}`);
 
         if (status in JiraErrorCodes) {
-          logger.error(error.response.data, JiraErrorCodes[error.response.status]);
+          logger.error(error.response.data, JiraErrorCodes[status]);
         }
 
         return Promise.reject(new JiraClientError(error));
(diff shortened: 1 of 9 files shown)


lib/jira/client/axios.js
```diff
@@ -8,80 +8,139 @@ const JiraClientError = require('./jira-client-error');
...
+     * @returns {Promise<Error>} The rejected promise
+     */
+    (error) => {
+      if (error.response) {
+        const { status, statusText } = error.response || {};
+        logger.debug({
+          params: error.config.fields,
+        }, `Jira request: ${error.request.method} ${error.request.path} - ${status} ${statusText}`);
+
+        if (status in JiraErrorCodes) {
```
jules2689, line 78: whats the diff between status and error.response.status?

tebriel: `status` actually comes from `error.response` on line 73 above! Same thing! I just forgot to use it further down.

lib/jira/client/axios.js
```diff
@@ -8,80 +8,139 @@ const JiraClientError = require('./jira-client-error');
...
+          logger.error(error.response.data, JiraErrorCodes[error.response.status]);
```
tebriel, line 79:
```suggestion
          logger.error(error.response.data, JiraErrorCodes[status]);
```

PR #356, opened 2020-05-06
Enable CodeQL on Jira Integration

tebriel:


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+
+    # Initializes the CodeQL tools for scanning.
+    - name: Initialize CodeQL
+      uses: github/codeql-action/init@v1
+      # Override language selection by uncommenting this and choosing your languages
+      # with:
+      #   languages: go, javascript, csharp, python, cpp, java
```
srt32, line 26: We don't need this line at all?

tebriel: Nope, looks like it autodetects!

.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,45 @@
...
+    # ℹ️ Command-line programs to run using the OS shell.
+    # 📚 https://git.io/JvXDl
+
+    # ✏️ If the Autobuild fails above, remove it and uncomment the following three lines
+    #    and modify them (or add more) to build your code if your project
+    #    uses a compiled language
+
+    #- run: |
+    #   make bootstrap
+    #   make release
```
srt32, line 42: Shall we remove the comments?

tebriel: We can, I installed this through the marketplace + actions without modification. What do you think is best?

srt32: I'd propose we delete all the comments but that's just an opinion ;)

tebriel: done!

PR #357, opened 2020-05-06
CodeQL fixes

tebriel:
This fixes all of our CodeQL discovered vulnerabilities. These were lower-priority because they're in the admin api which is gated by a GraphQL check to verify that the user is a GitHub staff member, but it's still good practice to avoid leaving these around.

Included in this PR:

1.  Reflected XSS Possibility: An auth'd Staff Member could pass in an invalid InstallationID, which we would then return, un-escaped. Not really a problem, but if someone exposed a fault in the verification API it could be bad. Mitigated by checking the types of the parameters before continuing with the request.
2.  Missing Rate Limiting: The Staff APIs had no rate limiting on them, meaning that someone could hammer away and degrade the service. Mitigated by adding the express-rate-limit to the Staff APIs and coordinating rate limiting with redis.
3.  Exposing internal information. One of our endpoints would just return the entire exception as the response body on a 401 from the GraphQL API. This now returns only the error body from the response instead of the stack trace and other data.


lib/api/index.js
```diff
@@ -117,7 +147,10 @@ app.get('/', (req, res) => {
   res.send({});
 });
 
-app.get('/:installationId', async (req, res) => {
+app.get('/:installationId', [
+  check('installationId').isInt(),
```
zmoazeni, line 151: oh TIL! this is neat!

tebriel: Yeah this seems to be a very standard express module to add in, and I love the safety it brings. Plus if this were a regular public API, users would get back a response about the section of the validation that failed, but our stafftools client and Jira are the only consumers, so that's less helpful except in dev!

PR #360, opened 2020-05-07
Query for comment count out of band

tebriel:
the `/graphql` endpoint seems to time out on large requests with `totalCount`, this is an attempt to mitigate that by querying for it outside of the graphql query.

Tried the old query in the graphql explorer and it timed out, same query without the `totalCount` succeeded, though it's unclear if my second query had a cached result.


tebriel:
@zmoazeni good point. We currently rely on the OctokitError wrapper for these metrics/error reporting but I'll add some specific ones as I know you're currently working on enhancing those.

lib/sync/pull-request.js
```diff
@@ -9,12 +12,33 @@ const { getPullRequests: getPullRequestQuery } = require('./queries');
...
+
+  let asyncTags = [];
+  await statsd.asyncTimer(
+    async () => {
+      try {
+        ({ edges } = (await github.graphql(getPullRequestQuery, vars)).repository.pullRequests);
+        asyncTags.push('status:200');
+      } catch (error) {
+        if (error instanceof RequestError) {
+          asyncTags.push(`status:${error.status}`);
```
zmoazeni, line 31: It's pretty safe to say that we don't have many variance in the `status` right? I'm just thinking of the cardinality here.

lib/sync/pull-request.js
```diff
@@ -9,12 +12,33 @@ const { getPullRequests: getPullRequestQuery } = require('./queries');
...
+  };
+
+  let asyncTags = [];
+  await statsd.asyncTimer(
+    async () => {
+      try {
+        ({ edges } = (await github.graphql(getPullRequestQuery, vars)).repository.pullRequests);
+        asyncTags.push('status:200');
+      } catch (error) {
+        if (error instanceof RequestError) {
```
zmoazeni, line 30: We may be able to use duck-typing here. But this is 100% ignorable:

```suggestion
        if (error.status) {
```

This _may_ be able to drop the `RequestError` dependency and import? I dunno if that's coming in for a different reason though!

tebriel:
This PR attempted to fix an issue some of the users on https://github.com/integrations/jira/issues/318 have experienced but we're still seeing the graphql timeout.

PR #362, opened 2020-05-11
Switch pull request sync flow to rest api

tebriel:
The graphql endpoint is timing out and it's breaking some customers with very large repos (> 100k pull requests). The rest api is a bit more performant (in local testing) and doesn't error out on large repos like graphql does. This PR rewrites the logic to use the Rest API for listing Pull Requests as a workaround.

### DIFF
--- a/lib/sync/pull-request.js
+++ b/lib/sync/pull-request.js
@@ -1,57 +1,89 @@
+const url = require('url');
 const transformPullRequest = require('./transforms/pull-request');
-const { getPullRequests: getPullRequestQuery } = require('./queries');
-const { logger } = require('probot/lib/logger');
-
 const statsd = require('../config/statsd');
 
 /**
+ * @typedef {object} RepositoryObject
+ * @property {number} id
+ * @property {string} name
+ * @property {object} owner
+ * @property {string} owner.login
+ * @property {string} html_url
+ * @property {string} full_name
+ */
+
+/**
+ * Find the next page number from the response headers.
+ *
+ * @param {{link: string}} headers - The headers from the API response.
+ * @returns {number|undefined} The next page number.
+ */
+function getNextPage(headers = {}) {
+  const nextUrl = ((headers.link || '').match(/<([^>]+)>;\s*rel="next"/) || [])[1];
+  if (!nextUrl) {
+    return;
+  }
+  const parsed = url.parse(nextUrl).query.split('&');
+  let nextPage;
+  parsed.forEach((query) => {
+    const [key, value] = query.split('=');
+    if (key === 'page') {
+      nextPage = Number(value);
+    }
+  });
+  return nextPage;
+}
+
+/**
  * @param {import('probot').GitHubAPI} github - The GitHub API Object
- * @param {any} repository - TBD
- * @param {any} cursor - TBD
+ * @param {RepositoryObject} repository - TBD
+ * @param {string|number} cursor - TBD
  * @param {any} perPage - TBD
  * @returns {any} TBD
  */
-exports.getPullRequests = async (github, repository, cursor, perPage) => {
+async function getPullRequests(github, repository, cursor, perPage) {
+  /** @type {number} */
+  let status;
+  /** @type {{link: string}} */
+  let headers;
+  /** @type {import('@octokit/rest').Octokit.PullsListResponse} */
   let edges;
+  if (!cursor) {
+    cursor = 1;
+  } else {
+    cursor = Number(cursor);
+  }
   const vars = {
     owner: repository.owner.login,
     repo: repository.name,
     per_page: perPage,
-    cursor,
+    page: cursor,
   };
 
   let asyncTags = [];
   await statsd.asyncTimer(
     async () => {
-      try {
-        ({ edges } = (await github.graphql(getPullRequestQuery, vars)).repository.pullRequests);
-        asyncTags.push('status:200');
-      } catch (error) {
-        logger.error('GraphQL Query Error for Sync Pull Request', error);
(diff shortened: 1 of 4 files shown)


lib/sync/pull-request.js
```diff
@@ -1,57 +1,89 @@
...
+ */
+
+/**
+ * Find the next page number from the response headers.
+ *
+ * @param {{link: string}} headers - The headers from the API response.
+ * @returns {number|undefined} The next page number.
+ */
+function getNextPage(headers = {}) {
+  const nextUrl = ((headers.link || '').match(/<([^>]+)>;\s*rel="next"/) || [])[1];
```
tebriel, line 22: I stole this directly from the paginate plugin https://github.com/octokit/plugin-paginate-rest.js/blob/master/src/iterator.ts#L40-L42

lib/sync/transforms/pull-request.js
```diff
@@ -25,40 +33,52 @@ function getAuthor(author) {
...
+ */
+module.exports = async (payload, author, github) => {
+  const { pullRequest, repository } = payload;
   const { issueKeys } = parseSmartCommit(pullRequest.title);
 
   if (!issueKeys) {
     return {};
   }
 
+  // Comment Count is required by the Jira API, but only do this second fetch if we have issue keys.
```
jules2689, line 55: Why? I dont get this sentence 🙈

lib/sync/pull-request.js
```diff
@@ -1,57 +1,91 @@
...
+ * @property {string} full_name
+ */
+
+/**
+ * Find the next page number from the response headers.
+ *
+ * @param {{link: string}} headers - The headers from the API response.
+ * @returns {number|undefined} The next page number.
+ */
+function getNextPage(headers = {}) {
```
jules2689, line 21: Does probot's octokit not autopaginate?

lib/sync/pull-request.js
```diff
@@ -1,57 +1,91 @@
...
-      repo: repository.name,
-      pull_number: pull.number,
-    });
-    pull.comments = { totalCount: pr.data.comments };
-    const { data } = transformPullRequest({ pull_request: pull, repository }, pull.author);
+  // Force us to go to a non-existant page if we're past the max number of pages
+  const nextPage = getNextPage(headers) || cursor + 1;
+  console.log(`CURSOR: ${cursor}, NEXTPAGE: ${nextPage}`);
+
+  edges.forEach((edge) => {
```
jules2689, line 83: I don't get why we need to do this

lib/sync/pull-request.js
```diff
@@ -1,57 +1,91 @@
...
+    edge.cursor = nextPage;
+  });
+
+  let pullRequests = await Promise.all(edges.map(async (pull) => {
```
jules2689, line 87: Why do we paginate everything and then only take the first PR?

lib/sync/transforms/pull-request.js
tebriel: We must tell Jira the count of comments on the pull request. But there's no reason to make this second API call if there isn't a commit in the title. That's why it occurs after the `if (!issueKeys)`

lib/sync/pull-request.js
tebriel: It does! But we're chunking the work ourselves, every 20 items we save our state and queue up another job. We do this to avoid having a job that runs for 20 hours and the fails.

lib/sync/transforms/pull-request.js
jules2689: > if there isn't a commit

That makes sense

> Comment Count is required by the Jira API, but only do this second fetch if we have issue keys.

I'm just not sure how commits apply to this sentence

lib/sync/pull-request.js
jules2689: Do we pick up where it left off?

lib/sync/pull-request.js
tebriel: this is one of those "mind blowing" things in there. the `transformPullRequest` builds this object that has an array with a single pull request in it, we return the data and that single PR and then it goes as an item in the `pullRequests` array. So we're using all the PRs but building them each individually in a really weird way.

lib/sync/pull-request.js
tebriel: incoming comment on this line

lib/sync/pull-request.js
tebriel: We do! That's what we're using the page number (in the cursor) for.

lib/sync/transforms/pull-request.js
tebriel: if there isn't an **Issue Key** in the title. Wow that was confusing.

lib/sync/pull-request.js
```diff
@@ -1,57 +1,97 @@
...
+
+  // Attach the "cursor" (next page number) to each edge, because the function that uses this data
+  // fetches the cursor from one of the edges instead of letting us return it explicitly.
+  edges.forEach((edge) => {
+    edge.cursor = nextPage;
+  });
+
+  let pullRequests = await Promise.all(edges.map(async (pull) => {
+    const { data } = await transformPullRequest({ pullRequest: pull, repository }, pull.user, github);
     return data && data.pullRequests[0];
```
jules2689, line 95: This returns a PR? It reads like a boolean to me 😆

tebriel: oh, no, you're right. lolol. It returns a boolean. WHY IS THIS CODE SO CONFUSING

PR #367, opened 2020-06-04
Run build jobs on pull requests

tebriel:
If someone issues a PR from a fork, we don't run the build job, which then prevents deployment/merging.


tebriel:
i don't like that it builds twice, closing for now.

PR #387, opened 2020-09-14
Set SameSite 'none' for cookie-session cookies

tebriel:
Allows us to use XHR requests from jira to change the sync status of requests

When users click the "submit" to refresh the status of a sync, it's failing silently (you can see it in the console). This is due to a csrf token validation failure.
The cookies are not allowed to be sent as part of the XHR request because they had a default policy of `SameSite: 'lax'`. This means that `*.atlassian.com` was the only remote URL allowed for sending those credentialed cookies (that contain the CORS secret within them!). Since we didn't have a signed cookie with a secret, the CORS validation always failed.

As part of my debugging I bumped `csurf` to the latest version and also decreased the version of `cookie-session` to a stable, non-beta release.


tebriel:
Amazingly chrome reports this to us, but I didn't know enough to know what this meant or that I should be paying attention to it until I figured out what the problem was.

Chrome Error Report

jules2689 (approved):

PR #388, opened 2020-09-16
Make issue key parsing case insensitive

justinoue:
This modifies the issue key regex to match lowercase characters.

In addition it adds a transform to the moo.js issue key tokenizer to maintain the same all caps output in the event that anything Jira side is case sensitive (though I don't believe it is).

I also added a lowercase test.

### DIFF
--- a/lib/smart-commit-tokenizer.js
+++ b/lib/smart-commit-tokenizer.js
@@ -3,7 +3,7 @@ const newrelic = require('newrelic');
 
 // RDVCS Connector regexp provided by Atlassian
 const punct = '!"\\#$%&\'()*+,\\-./:;<=>?@\\[\\\\\\]^_`{|}~';
-const issueKeysRegex = new RegExp(`(?:(?<=[\\s${punct}])|^)(?:[A-Z][A-Z\\d]+-\\d+)(?:(?=[\\s${punct}])|$)`);
+const issueKeysRegex = new RegExp(`(?:(?<=[\\s${punct}])|^)(?:[A-Za-z][A-Za-z\\d]+-\\d+)(?:(?=[\\s${punct}])|$)`);
 
 module.exports = function () {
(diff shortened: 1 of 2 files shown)


justinoue:
Related issue: https://github.com/integrations/jira/issues/220

tebriel:
@justinoue thanks for this contribution! The jira project is currently transitioning ownership within GitHub so I can't merge this right away, but I'm going to pass this on to the new team to take care of and make sure there aren't any unintended consequences with this! I really appreciate this work!

justinoue:
> @justinoue thanks for this contribution! The jira project is currently transitioning ownership within GitHub so I can't merge this right away, but I'm going to pass this on to the new team to take care of and make sure there aren't any unintended consequences with this! I really appreciate this work!

Cool, thanks for the update. Here's to hoping nothing breaks. 🤞

pravinbashyal:
is there any update to this?

justinoue:
Hey @rachellerathbone , while I don't use jira anymore, this is still a little disappointing.
This does not require the Jira application to support mixed case or lowercase issue keys.

All this does is parse things in brackets case-insensitively. The URL it generates will still be all uppercase. (Although for what it's worth, Jira does redirect lower/mixed case issue keys to uppercase.)

[Test-1]
[test-1]
[TEST-1]

Would all then be parsed and link to https://myjira.atlassian.net/browse/TEST-1

Without this change, only [TEST-1] works, and require you to yell all the time about your issues.

rachellerathbone:
Hey @justinoue. I completely understand your frustrations. Unfortunately, while it seems like a simple change, it's a little more complicated on our end. 

This is actually a significant change that we would have to consider across all our providers. We could end up in situations where a customer's code is case insensitive, but their CI/CD integration is case sensitive. Then there is the issue with false positives. This is potentially a change that we could introduce in the future but it would require thorough investigation (i.e evaluating existing data to see what percentage of new hits are true positives vs false positives).

I've passed this onto the team and we're going to add a feature request for making this improvement into our backlog.

PR #390, opened 2020-10-07
Use express-rate-limiter for rate limiting

tebriel:
Recently we've been having issues with malicious users scanning looking for random files on the jira application, this causes us to exceed our SLOs and to fail to process a lot of valid requests. This PR adds [express-rate-limiter](https://github.com/nfriedly/express-rate-limit) to try to alleviate those problems.

To avoid rate limiting incoming requests from webhooks, we query the [GitHub Meta Endpoint](https://docs.github.com/en/free-pro-team@latest/rest/reference/meta#get-github-meta-information) to get a list of the current webhook CIDRs. These should not change frequently, so fetching them on startup seems reasonable. Based on our incoming query volume from things other than github webhooks, we do not see anything greater than 5 requests per minute as an average, I've set the max to 100 just to give busy users plenty of room.

These rate limits are coordinated through our Redis instance, so even when incoming requests are balanced across the workers, we should still benefit from the load balancing.

There is a new feature flag (as an env var): `USE_RATE_LIMITING` must be `'true'` to enable this rate limiting, otherwise the app should perform as it did before.

There are new metrics:
`jira_integration.express.rate_limited` will show up in datadog counting the number of times we've rate limited requests, this will need to be monitored when the flag is enabled.

Additionally there is a newrelic tracing segment named `isIp4InCidr` which should let us know how efficient (in practice) the CIDR validation function is, it appears to be very quick, but it's good to see it under load as well.

### DIFF
--- a/lib/configure-robot.js
+++ b/lib/configure-robot.js
@@ -51,7 +51,7 @@ async function getGitHubCIDRs(logger) {
   });
   const metaResp = await api.meta.get();
   const GitHubCIDRs = metaResp.data.hooks;
-  logger.info({ GitHubCIDRs }, 'Whitelisted CIDRs for rate limiting');
+  logger.info({ GitHubCIDRs }, 'CIDRs that can skip rate limiting');
   return GitHubCIDRs;
 }
 
@@ -73,7 +73,7 @@ module.exports = async (app) => {
        * NOTE: This is only expected to be done for IP CIDRs we trust (github)
        *
        * @param {import('express').Request} req
-       * @returns {boolean} if the IP is within our whitelisted cidrs
+       * @returns {boolean} if the IP is within our non-rate-limited cidrs
        */
       skip(req) {
         return isIp4InCidrs(req.ip, GitHubCIDRs);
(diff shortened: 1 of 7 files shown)


lib/configure-robot.js
```diff
@@ -1,16 +1,108 @@
...
+    type: 'installation',
+    installationId: inst.data[0].id,
+  });
+  api = GitHubAPI({
+    auth: token,
+    logger,
+  });
+  const metaResp = await api.meta.get();
+  const GitHubCIDRs = metaResp.data.hooks;
+  logger.info({ GitHubCIDRs }, 'Whitelisted CIDRs for rate limiting');
```
jules2689, line 54: Allowlisted?

lib/run.js
```diff
@@ -27,6 +27,8 @@ const probot = createProbot({
  * Start the probot worker.
  */
 function start() {
+  // We are always behind a proxy, but we want the source IP
+  probot.server.set('trust proxy', true);
```
jules2689, line 31: What will this do?

jules2689 (approved):

lib/configure-robot.js
tebriel: Oh gosh thank you!

lib/run.js
tebriel: This will cause express to use the originating IP address from the client instead of the most recent IP address which is that of the internal load balancer.

tebriel: Details: https://expressjs.com/en/guide/behind-proxies.html

PR #393, opened 2020-10-09
Use a unified JiraClientError message

tebriel:
Because Sentry aggregates based on the message field of an error, we shouldn't
have this have a very high cardinality. The details are available in the
"request" and "response" fields in the Sentry Metadata

Example sentry event with request/response data in it https://sentry.io/organizations/github-integrations/issues/1946869803/?project=1240711&query=is%3Aunresolved&statsPeriod=1h

### DIFF
--- a/lib/jira/client/jira-client-error.js
+++ b/lib/jira/client/jira-client-error.js
@@ -3,9 +3,7 @@
  */
 class JiraClientError extends Error {
   constructor(error) {
-    const { status, statusText } = error.response;
-    const { method, path } = error.response.request;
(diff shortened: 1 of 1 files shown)


jules2689 (approved):

tebriel:
JiraClientErrors are merged now with a type in the server side rule described here: https://docs.sentry.io/product/sentry-basics/guides/grouping-and-fingerprints/
it's `error.type:JiraClientError -> jira-client-error` set in the [project settings](https://sentry.io/settings/github-integrations/projects/jira/)

PR #399, opened 2020-10-20
Build out framework for sending hydro events

tebriel:
This framework exists to send events to the Hydro HTTP Gateway, using the hmac signed payloads. This does not send the events yet, just builds out the initial framework, protobuf, and tests. A future PR will send the events.

The `TRACKING_DISABLED` env var is set to `true` in staging and prod vault configurations.

### DIFF
--- a/lib/tracking/index.js
+++ b/lib/tracking/index.js
@@ -72,7 +72,11 @@ async function submitProto(proto) {
     return true;
   }
   const data = {
-    events: [{ schema: proto.schema, value: JSON.stringify(proto) }],
+    events: [{
+      schema: proto.schema,
+      value: JSON.stringify(proto),
+      cluster: 'potomac',
+    }],
   };
 
   const dataStr = JSON.stringify(data);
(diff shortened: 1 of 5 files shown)


lib/tracking/index.js
```diff
@@ -0,0 +1,142 @@
...
+    status = resp.status;
+    logger.debug('Hydro Protobuf Accepted', data);
+  } catch (err) {
+    if (err.response == null) {
+      // This is not an AxiosError
+      logger.error(err);
+      status = 'exception';
+    } else {
+      /** @type {import('axios').AxiosError} */
+      const axError = err;
```
jules2689, line 112: this will always be an axios error?

jules2689 (approved):

lib/tracking/index.js
tebriel: Great question!! Axios' error isn't a subclass of the `AxiosError` class, so I can't say `if (err isinstance AxiosError)` unfortunately. Checking for the `response` field within the error is how the axios library does it in its own code, so it feels fairly safe, I'm not aware of any root `Error` instances that would attach a `response` object (outside of axios). I think it's safe to assume this is an axios error in this case.

PR #401, opened 2020-10-22
Instrument actions taken via the Jira Console

tebriel:
These are things users do like delete/add/edit subscriptions and installations.

Previous PR #400 with Webhook events, PR #399 set up the original framework.

This is the 3rd of 3 PRs to track create/disable/delete actions taken on subscriptions and installations.

### DIFF
--- a/lib/frontend/delete-github-subscription.js
+++ b/lib/frontend/delete-github-subscription.js
@@ -88,7 +88,6 @@ module.exports = async (req, res) => {
       });
     }
   } catch (err) {
-    console.log(err);
     req.log.error(err);
     res.sendStatus(400);
   }
--- a/lib/frontend/delete-github-subscription.js
+++ b/lib/frontend/delete-github-subscription.js
@@ -1,22 +1,37 @@
-const { Subscription } = require('../models');
+const {
+  ActionType,
+  ActionSource,
+  ActionFromSubscription,
+} = require('../proto/v0/action');
+const { submitProto } = require('../tracking');
+const { Installation, Subscription } = require('../models');
 
+/**
+ * Handle the when a user adds a repo to this installation
+ *
+ * @param {import('express').Request} req
+ * @param {import('express').Response} res
+ * @returns {Promise<void>}
+ */
 module.exports = async (req, res) => {
   if (!req.session.githubToken) {
-    return res.sendStatus(401);
+    res.sendStatus(401);
+    return;
   }
 
   if (!req.body.installationId || !req.body.jiraHost) {
     res.status(400);
-    return res.json({
+    res.json({
       err: 'installationId and jiraHost must be provided to delete a subscription.',
     });
+    return;
   }
 
   /**
    * Returns the role of the user for an Org or 'admin' if the
    * installation belongs to the current user
    *
-   * @returns {string}
+   * @returns {Promise<string>}
    */
   async function getRole({ login, installation }) {
     if (installation.target_type === 'Organization') {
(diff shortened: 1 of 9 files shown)


lib/frontend/delete-github-subscription.js
```diff
@@ -48,24 +64,32 @@ module.exports = async (req, res) => {
...
+      res.sendStatus(202);
     } catch (err) {
       res.status(403);
-      return res.json({
+      res.json({
         err: `Failed to delete subscription to ${req.body.installationId}. ${err}`,
       });
     }
   } catch (err) {
+    console.log(err);
```
jules2689, line 91: do we want to console.log?

lib/frontend/delete-github-subscription.js
```diff
@@ -48,24 +64,32 @@ module.exports = async (req, res) => {
...
+        return;
       }
 
-      await Subscription.uninstall({
-        installationId: req.body.installationId,
-        host: req.body.jiraHost,
-      });
-      return res.sendStatus(202);
+      /** @type {import('../models/subscription')} */
+      const subscription = await Subscription.getSingleInstallation(req.body.jiraHost, req.body.installationId);
```
jules2689, line 74: I dont get why we deleted uninstall

lib/frontend/post-github-configuration.js
```diff
@@ -33,21 +49,28 @@ module.exports = async (req, res) => {
...
-        return res.json({
+        res.json({
           err: `Failed to add subscription to ${req.body.installationId}. User is not an admin of that installation`,
         });
+        return;
       }
     }
 
-    await Subscription.install({
+    const subscription = await Subscription.install({
```
jules2689, line 59: we also deleted install?

lib/frontend/delete-github-subscription.js
tebriel: err no, and the linter should've caught that, dang.

lib/frontend/delete-github-subscription.js
tebriel: We move it down to line 81 where we call `destroy`. `uninstall()` does a `get` and then a `destroy` but we need to know the subscriptions details to record it before we destroy it.

lib/frontend/post-github-configuration.js
tebriel: oh dang I missed the `.destroy()` call here, good catch!

tebriel: I didn't delete it, I just capture the output and use it to create an event!

lib/frontend/delete-github-subscription.js
jules2689: Would it make sense to make uninstall and install do the logging so we don't have to duplicate uninstall logic?

lib/frontend/post-github-configuration.js
jules2689: Misread!

lib/frontend/delete-github-subscription.js
tebriel: yeah that rule appears to be off, I saw them when I did a project-wide search. I'll file an issue but I figure we can probably just leave them :p

lib/frontend/delete-github-subscription.js
tebriel: Hmm, maybe, the reason I didn't was because install/uninstall don't know the source of the request, lemme look at the possible collateral damage of adding a parameter to install/uninstall.

tebriel: `Subscription` models don't have any explicit linkage to `Installation` models (which feels like a design flaw, but it's what we have right now). I can link the two up together, but it means importing the `Installation` model in or having an optional `installationId` parameter for uninstalling to be able to report the install ID. I think the method I've implemented here keeps the Subscription from needing knowledge about Installations. I looked into implementing `hasMany` and `belongsTo` but because there are no uniqueness guarantees in the database for anything but primary key, this process also breaks down.

Summary:
 I think that this process outlined here is the safest method without a larger refactor to improve the data model/data integrity.

jules2689 (approved):

PR #403, opened 2020-11-13
Let Github bots events trigger changes in Jira

### LINKED ISSUE #394: PRs not displaying in Jira
Came across an issue where a PR wouldn't display in Jira.

The PR was created with the github-actions bot as the author and wouldn't be detected. If a real user commented on the PR, it would show up. If a real user changed the status, it would show up. etc etc.

Changing the author of the PR to a real user fixed the issue. We did this by not using ${{ secrets.GITHUB_TOKEN }} in the action and changing that to a custom token for a real user.

It should really just work... however and whoever creates the PR.

mat007:
With Github pushing for using Github Apps and Actions instead of bot users this will only become more common to have user type «Bot» triggering webhooks.
I don’t see why there should be a difference when a bot or a real user makes some changes on Github that should be reflected in Jira.

Fixes #394

### DIFF
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -60,11 +60,6 @@ module.exports = function middleware(callback) {
     const gitHubInstallationId = context.payload.installation.id;
     context.log = context.log.child({ gitHubInstallationId });
 
-    if (context.payload.sender.type === 'Bot') {
-      context.log({ noop: 'bot', botId: context.payload.sender.id, botLogin: context.payload.sender.login }, 'Halting futher execution since the sender is a bot');
-      return;
-    }
-
     if (isFromIgnoredRepo(context.payload)) {
       context.log({ noop: 'ignored_repo', installation_id: context.payload.installation.id, repository_id: context.payload.repository.id }, 'Halting futher execution since the repository is explicilty ignored');
       return;


tebriel:
Hi @mat007 Thanks for this submission. The reason this protection exists is to avoid bot feedback loops where two bots are triggering off each others events, causing a cascade of webhooks. Such as the following (admittedly contrived) example:
1. User submits a PR with a link to a Jira Issue
2. Jira Integration receives the PR webhook, updates the PR Text to replace the JIRA issue ID as a link to the Jira Issue
3. Bot 2 receives a PR Update Webhook, Updates the Issue removing the link and reformatting the Jira Issue ID for some other purpose
4. Goto 2

We've had multiple internal discussions about ways to work around this, but currently we have not settled on a solution.

mat007:
Thanks for providing some context @tebriel, I understand your concern.
It’s not easy to prevent loops without storing state, is the integration entirely stateless?

However it seriously limits automation as it is now.
Maybe this could be an option that could be opted out?

tebriel:
The integration is not entirely stateless, no. I think that allowing a user to opt out of this would potentially allow them to use bot automations, but would still open up the potential for a feedback loop. The Jira integration is going through an ownership transfer within GitHub at the moment, I'm hoping the new team will be ready soon to look at issues such as this one.

atrigueiro:
Closing this PR in favour of https://github.com/integrations/jira/pull/437

PR #405, opened 2020-12-10
Update CONTRIBUTING.md

tebriel:


CONTRIBUTING.md
```diff
@@ -18,13 +18,16 @@ Contributions to this product are [released][releases] to the public under the [
...
+Please ensure that you have [homebrew](https://brew.sh/) installed. Instructions for setting this application up on an OS other than OSX are currently not outlined.
+
 This app is written in [ES6 JavaScript](https://nodejs.org/en/docs/es6/) and runs on [Node.js](https://nodejs.org/). After cloning the repository, install the dependencies by running:
 
 ```
 $ script/bootstrap
 ```
 
-This will install `node` and `postgres`, which are required to run the app. You will need to start up the postgres instance, and then run:
+This will install `node` and `mysql`, which are required to run the app. You will need to start up the postgres instance, and then run:
```
jules2689, line 30: Second postgres in there

jules2689 (approved):
Good to go except for that second postgres

CONTRIBUTING.md
tebriel: we don't use mysql at all, I've lost my dang mind.

rachellerathbone (approved):

PR #425, opened 2021-04-14
ARC-69: Fix branch deletion/creation when PR is merged

atrigueiro:
This PR stops sending the branch on the payload when the Pull Request Merged event is called because if the "Automatically delete head branches" option is enabled, the branch deleted and PR merged events might be sent out “at the same time” and received out of order, which causes the branch being created again.

**Current scenario**:
- **Automatically delete head branches** is enabled under Settings
- Create Jira issue
- Create branch using the Jira issue key
- Commit and Merge PR
Expected result: Branch is deleted
Bug: Branch was deleted but then recreated

Scenarios after implementing this fix:

**Scenario 1:**
- **Automatically delete head branches** is _enabled_ under Settings
- Create Jira issue
- Create branch using the Jira issue key
- Commit and Merge PR
Result: Branch is deleted

**Scenario 2:**
- **Automatically delete head branches** is _disabled_ under Settings
- Create Jira issue
- Create branch using the Jira issue key
- Commit and Merge PR
Result: Branch is not deleted


lib/transforms/pull-request.js
```diff
@@ -79,4 +81,10 @@ module.exports = (payload, author) => {
       updateSequenceId: Date.now(),
     },
   };
+
+  if (pullRequestStatus === 'MERGED') {
```
tebriel, line 85: Could we add a comment here to indicate why this is the case to avoid someone changing it where they might not understand the consequences?

tebriel (commented):
Thanks so much for this PR, this looks like a great quality of life improvement for users. Could you add a test to validate this functionality and prevent future regressions?

lib/jira/client/index.js
```diff
@@ -170,7 +170,7 @@ const batchedBulkUpdate = async (data, options, instance, installationId) => {
       },
     });
   });
-  await Promise.allSettled(batchedUpdates);
+  await Promise.all(batchedUpdates);
```
tebriel, line 173: Ooo this is an interesting change. This will immediately reject if any of the `batchedUpdates` fails, instead of waiting for all of them to complete (failing or not). Since we don't really care about the result either way I wonder if there is even a reason to `await` this call.

atrigueiro: Yeah probably there isn't. Happy to remove it.

tebriel (approved):
I'm sure users will appreciate this improvement! Thank you for adding/improving tests and adding comments for future developers.

saquibkhan (approved):

PR #427, opened 2021-04-20
Feature/arc 76 update contributing

nsurendran1991:
updated CONTRIBUTING.md ngrok installation command.

Any version of brew of 2.7 or higher requires users to run the command in the format of brew install --cask ngrok vs the format of brew cask install ngrok for versions 2.6 and below.  

To save users the time of having to google the error Error: Unknown command: cask when running the command we’ve provided, we should list both commands, and the corresponding versions, in our doc.


tebriel (approved):
Thanks for this informative change!

PR #428, opened 2021-04-26
ARC-7 progress sync status

rachellerathbone:
When a customer installs a GitHub organization in Jira, there is no indicator of how the sync is progressing. The Sync Status simply gives a status e.g. ACTIVE. If a customer is installing a handful of repos the sync times is minutes, if they’re installing 100s (possibly 1000s) this installation sometimes takes days.

To improve the customer experience, we have worked with our design team to start implementing small changes including:

- reordering columns to give a better flow of data
- providing a **Repos Synced** value which shows the number of repos installed / total number of repos
- merged the repositories column values (selected/all) into the Repos Synced column to combine data and prevent making the table too 'noisy'
- adding colour coding to the Sync Status labels to make statuses easier to see.

Screen Shot 2021-05-02 at 4 16 27 pm


mboudreau (approved):
Great job on this, it looks much nicer and has good feedback for the user :+1:

lib/frontend/app.js
```diff
@@ -63,6 +63,12 @@ module.exports = (appTokenGenerator) => {
   app.set('view engine', 'hbs');
   app.set('views', path.join(rootPath, 'views'));
 
+  // Handlebars helpers
+  const hbs = require('hbs');
```
tebriel, line 67: typically we `require` modules at the top of the file, is there a pre-loading reason to avoid requiring this until within this exported function?

lib/frontend/get-jira-configuration.js
```diff
@@ -25,25 +28,34 @@ const formatDate = function (date) {
...
+      host: jiraHost,
+      connections,
+      failedConnections,
+      hasConnections: connections.length > 0 || failedConnections.length > 0,
+      APP_URL: process.env.APP_URL,
+      csrfToken: req.csrfToken(),
+    });
+    return 'Jira configuration rendered successfully.';
+  } catch (error) {
+    console.error(error);
```
tebriel, line 58: the `req` object has a probot logger instance attached to it in https://github.com/integrations/jira/blob/master/lib/middleware/log-middleware.js. We should probably use that for the extra context/json formatting.

lib/frontend/get-jira-configuration.js
```diff
@@ -25,25 +28,34 @@ const formatDate = function (date) {
...
+  try {
+    res.render('jira-configuration.hbs', {
+      host: jiraHost,
+      connections,
+      failedConnections,
+      hasConnections: connections.length > 0 || failedConnections.length > 0,
+      APP_URL: process.env.APP_URL,
+      csrfToken: req.csrfToken(),
+    });
+    return 'Jira configuration rendered successfully.';
```
tebriel, line 56: Do we render this string somewhere? Would it be better to return something other than a string? Normally I'd expect this function wouldn't return anything and would `throw` in the case of an error, I see the rendering errors are being `catch`'d below, does this improve the user's experience? Without the original `error` object (and instead a string being returned in the case of an error) there's no easy way to branch on what type of error it is, if later we decide we want to do that.

lib/sync/installation.js
```diff
@@ -11,8 +11,21 @@ const tasks = {
 };
 const taskTypes = Object.keys(tasks);
 
+const updateNumberOfReposSynced = async (repos, subscription) => {
+  const repoIds = Object.keys(repos);
+
+  const syncedRepos = repoIds.filter((id) => {
+    // all 3 statuses need to be complete for a repo to be fully synced
+    const { pullStatus, branchStatus, commitStatus } = repos[id];
+    return pullStatus && branchStatus && commitStatus === 'complete';
```
tebriel, line 20: I think this will test if `pullStatus` is truthy, if `branchStatus` is truthy, and if `commitStatus` is exactly equal to `complete`. Maybe you'd prefer something like this?
```suggestion
    return pullStatus === 'complete' && branchStatus === 'complete' && commitStatus === 'complete';
```

static/css/jira-configuration.css
```diff
@@ -0,0 +1,43 @@
+.jiraConfigurationTable {
```
tebriel, line 1: Have these css changes been tested cross-browser? In just one browser? What's our check for compatibility?

tebriel (commented):
Great start! Love that there is a test for this and the clarity this will bring to the user. I have some questions and possible improvements for debug-ability and a possible logic issue.

static/css/jira-configuration.css
rachellerathbone: I've checked in several browsers. Original screen shot is from Chrome. 

Safari:

safari

Firefox:

firefox

lib/sync/installation.js
rachellerathbone: Good point will update.

lib/frontend/get-jira-configuration.js
rachellerathbone: Cleaned this up a little and passed the error to `next` so the error page will render in the event where an error occurs.

tebriel (approved):

PR #429, opened 2021-04-27
ARC-36: fix failing test

atlassian-test-account:
Currently, when you follow `CONTRIBUTING.md` in Github for Jira and get to the `npm test` step, one of the tests fail.

Failing test: updates all Subscriptions for a given jira clientKey when a site is renamed in test/unit/models/index.test.js
Fails with: 

```
Expected: "https://renamed-user.atlassian.net"
Received: "https://existing-instance.atlassian.net"
```

This was happening because we were calling Jest's `beforeEach` function which executes before every test in the scope. This meant that the renamed url of `https://renamed-user.atlassian.net` was being reset:

Screen Shot 2021-04-27 at 4 48 55 pm

By replacing `beforeEach` with `beforeAll`, we only execute this logic once, before any test is run. Now all tests pass:

Screen Shot 2021-04-27 at 4 50 39 pm

### DIFF
--- a/test/unit/models/index.test.js
+++ b/test/unit/models/index.test.js
@@ -45,7 +45,7 @@ describe('test installation model', () => {
     eventType: 'installed',
   };
 
-  beforeEach(async () => {
+  beforeAll(async () => {
     // Clean up the database
     await Installation.truncate({ cascade: true, restartIdentity: true });
     await Subscription.truncate({ cascade: true, restartIdentity: true });
--- a/test/unit/models/index.test.js
+++ b/test/unit/models/index.test.js
@@ -45,11 +45,13 @@ describe('test installation model', () => {
     eventType: 'installed',
   };
 
-  beforeAll(async () => {
+  beforeEach(async () => {
     // Clean up the database
     await Installation.truncate({ cascade: true, restartIdentity: true });
     await Subscription.truncate({ cascade: true, restartIdentity: true });
+  });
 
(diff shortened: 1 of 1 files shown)


rachellerathbone:
Above comment was by me. Didn't realize I was signed into my secondary account.

tebriel (approved):
This test has plagued me for some time! I'm so happy to see a fix for it!

My thoughts are that the `truncate` calls are in a `beforeEach` to ensure that the database is "clean" before every test. Is there a way we could separate out things into both a `beforeEach` and a `beforeAll` and do the reset only once in the `beforeAll` but clean up the DB each time?

Does that make sense to do it that way? Maybe the tests are formatted in such a way that they're creating and utilizing unique objects and so that won't be an issue. I haven't dived deep into the tests here, so I'm open to your knowledge of this section of the code!

I'm going ahead and approving this PR in case that the `beforeAll` is appropriate and the `beforeEach` strategy I line out above isn't useful, the tests pass so that's at least a good indication!

rachellerathbone:
Just updated: pulled the truncate calls out into `beforeEach` and left the rest in `beforeAll`. All tests are passing still 🎉

rachellerathbone:
Accidentally opened in wrong account. Closing and while open a new PR.

rachellerathbone:
New PR https://github.com/integrations/jira/pull/431

PR #434, opened 2021-04-30
ARC-36 add await to installation update

rachellerathbone:
This is a secondary (and hopefully final) PR to fix the flaky test in the app. Turns out, we weren't awaiting update in the install method so our super speedy tests were receiving the returned valued for the existing subscription instead of the renamed one.

### DIFF
--- a/lib/models/installation.js
+++ b/lib/models/installation.js
@@ -95,11 +95,7 @@ class Installation extends Sequelize.Model {
         jiraHost: payload.host,
       }).then(async (record) => {
         const subscriptions = await Subscription.getAllForClientKey(record.clientKey);
-        for (const subscription of subscriptions) {
-          await subscription.update({
-            jiraHost: record.jiraHost,
-          });
-        }
+        await Promise.all(subscriptions.map(subscription => subscription.update({ jiraHost: record.jiraHost })));
 
         return installation;
       });
--- a/lib/models/installation.js
+++ b/lib/models/installation.js
@@ -95,11 +95,12 @@ class Installation extends Sequelize.Model {
         jiraHost: payload.host,
       }).then(async (record) => {
         const subscriptions = await Subscription.getAllForClientKey(record.clientKey);
-        subscriptions.forEach(subscription => {
-          subscription.update({
+        for (const subscription of subscriptions) {
+          await subscription.update({
             jiraHost: record.jiraHost,
           });
-        });
+        }
+
         return installation;
       });
     }
--- a/test/unit/models/index.test.js
+++ b/test/unit/models/index.test.js
@@ -46,12 +46,6 @@ describe('test installation model', () => {
   };
 
   beforeEach(async () => {
-    // Clean up the database
-    await Installation.truncate({ cascade: true, restartIdentity: true });
-    await Subscription.truncate({ cascade: true, restartIdentity: true });
-  });
-
-  beforeAll(async () => {
     const installation = await Installation.install({
       host: existingInstallPayload.baseUrl,
       sharedSecret: existingInstallPayload.sharedSecret,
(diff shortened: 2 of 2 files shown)


mboudreau:
@tebriel 

@rachellerathbone and I found the problem this morning while pairing.  Essentially it comes down to the tests being too fast and the `Installation.install` function not waiting for the subscriptions to be done updating before continuing.  Definitely a tricky one to solve :)

lib/models/installation.js
```diff
@@ -95,11 +95,12 @@ class Installation extends Sequelize.Model {
...
       }).then(async (record) => {
         const subscriptions = await Subscription.getAllForClientKey(record.clientKey);
-        subscriptions.forEach(subscription => {
-          subscription.update({
+        for (const subscription of subscriptions) {
+          await subscription.update({
             jiraHost: record.jiraHost,
           });
-        });
+        }
```
mboudreau, line 102: I actually was thinking, if there's a lot of subscriptions, it would be slower than having everything in parallel since the loop would do everything sequentially.  An alternative would be to use `Promise.all` :)

```
await Promise.all(
            subscriptions.map(
                subscription => subscription.update({jiraHost: record.jiraHost})
            )
        );
```

mboudreau (changes requested):
Small little change to do while we're there.  I also thought of looking at the `Subscription.install` function to see if it has the same issue, and it kinda does for the line `Subscription.findOrStartSync(subscription);` where it's missing an await before this.  Maybe a quick add? :)

lib/models/installation.js
rachellerathbone: Good point. Let's not make that slower.

mboudreau (approved):
Awesome :D

test/unit/models/index.test.js
```diff
@@ -72,6 +66,12 @@ describe('test installation model', () => {
     });
   });
 
+  afterEach(async () => {
+    // Clean up the database
+    await Installation.truncate({ cascade: true, restartIdentity: true });
+    await Subscription.truncate({ cascade: true, restartIdentity: true });
```
tebriel, line 72: The only time I can see this being a problem is if the tests failed due to an exception and did not exit cleanly. This means the next run of the tests might have state from the previous run. Is there anything we can/should do about that?

lib/models/installation.js
```diff
@@ -95,11 +95,8 @@ class Installation extends Sequelize.Model {
         jiraHost: payload.host,
       }).then(async (record) => {
         const subscriptions = await Subscription.getAllForClientKey(record.clientKey);
-        subscriptions.forEach(subscription => {
-          subscription.update({
-            jiraHost: record.jiraHost,
-          });
-        });
+        await Promise.all(subscriptions.map(subscription => subscription.update({ jiraHost: record.jiraHost })));
```
tebriel, line 98: Nice find, great improvement!

test/unit/models/index.test.js
mboudreau: Even if the test fails, the `afterEach` is still called.  It's better to do it this way as it's the recommended testing approach for all testing frameworks - start with clean slate, end with clean slate.  

The problem with having it in the `beforeEach` is that when another test suite starts it will have some state already there, where the assumption is that there would be none.  If another suite installed a models that this one doesn't know about, you'll have to truncate that one also, leading to adding extra code just for cleaning up another suite.

tebriel: Yes, agreed on all counts, I meant in the case of a catastrophic exception (for example a ctrl-c). We probably don’t need to optimize for that case, but I wanted to raise the idea just in case.

tebriel (approved):

test/unit/models/index.test.js
mboudreau: Right.  We were going to address that later when rewriting the tests so that it doesn't have a dependency on an external postgres service for them to run :)

PR #436, opened 2021-05-03
ARC-105: add block to handle rejected promise

rachellerathbone:
When a user removes an org during an active sync an unhandled error is thrown in the console.

unhandled-promise-rejection

I've added a conditional block to simply log a message to the console when a subscription is `null` and return. All other active syncs continue.

https://user-images.githubusercontent.com/37155488/116833288-cd091e80-abfb-11eb-9daf-326cef6067eb.mov

### DIFF
--- a/lib/sync/installation.js
+++ b/lib/sync/installation.js
@@ -37,6 +37,12 @@ module.exports.processInstallation = (app, queues) => {
     // Get a fresh subscription instance
     const subscription = await Subscription.getSingleInstallation(jiraHost, installationId);
 
+    // handle promise rejection when an org is removed during a sync
+    if (subscription == null) {
+      console.error('Organization has been deleted. Other active syncs will continue.');
+      return;
+    }
+
     const status = edges.length > 0 ? 'pending' : 'complete';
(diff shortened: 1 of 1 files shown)


mboudreau (approved):
This is a hard one to test without refactoring the sync as we need to change the state in the middle of the sync and be guaranteed it happened before continuing.  Only way I can think to do that is to break apart this function so that each repo sync is an individual call.

At this point, for such a simple change, I think it's easier just to move forward for now.

lib/sync/installation.js
```diff
@@ -37,6 +37,12 @@ module.exports.processInstallation = (app, queues) => {
     // Get a fresh subscription instance
     const subscription = await Subscription.getSingleInstallation(jiraHost, installationId);
 
+    // handle promise rejection when an org is removed during a sync
+    if (subscription == null) {
+      console.log('Organization has been deleted. Other active syncs will continue.');
```
tebriel, line 42: This might be enhanced by using `app.log` instead of `console.log` as the App logger has extra context in it and will follow the json output patterns, generally we try to avoid `console.log` within the application.

tebriel (approved):

PR #437, opened 2021-05-03
ARC-114: Allow bots to trigger changes

atrigueiro:
This PR adds tests to https://github.com/integrations/jira/pull/413.

Points to help us understand why having an exception to allow bots events only on PR state changes sound reasonable:

- We don't see how we can end up in a bot feedback loop situation if we allow bot events processing only for those, given the fact that this integration does not update PR states by itself.

- The act of merging - which is inherently a one-time action - won’t cause a loop so the bot-filter can be dropped for merge hooks, but left in place for other types of hooks so that the status could be properly posted in JIRA.

- The infinite loop issue may only happen on comment/description edition because it is the only case where this Jira integration actually writes data in GitHub and can trigger events. For that reason, we won’t add edited to the list of actions. I.e. If two bots are both set to auto-respond to all comments, you'd get an infinite loop.

### DIFF
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -65,13 +65,15 @@ module.exports = function middleware(callback) {
     const gitHubInstallationId = context.payload.installation.id;
     context.log = context.log.child({ gitHubInstallationId });
 
+    // Edit actions are not allowed because they trigger this Jira integration to write data in GitHub and can trigger events, causing an infinite loop.
+    // State change actions are allowed because they're one-time actions, therefore they won’t cause a loop.
     if (context.payload.sender.type === 'Bot' && !isStateChangeAction(context.payload.action)) {
       context.log({ noop: 'bot', botId: context.payload.sender.id, botLogin: context.payload.sender.login }, 'Halting further execution since the sender is a bot and action is not a state change');
       return;
     }
 
     if (isFromIgnoredRepo(context.payload)) {
-      context.log({ noop: 'ignored_repo', installation_id: context.payload.installation.id, repository_id: context.payload.repository.id }, 'Halting further execution since the repository is explicilty ignored');
+      context.log({ noop: 'ignored_repo', installation_id: context.payload.installation.id, repository_id: context.payload.repository.id }, 'Halting further execution since the repository is explicitly ignored');
       return;
     }
 
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -66,12 +66,12 @@ module.exports = function middleware(callback) {
     context.log = context.log.child({ gitHubInstallationId });
 
     if (context.payload.sender.type === 'Bot' && !isStateChangeAction(context.payload.action)) {
-      context.log({ noop: 'bot', botId: context.payload.sender.id, botLogin: context.payload.sender.login }, 'Halting futher execution since the sender is a bot and action is not a state change');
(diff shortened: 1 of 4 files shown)


mboudreau (approved):
Looks good! Maybe you should close https://github.com/integrations/jira/pull/413 and add a comment pointing to this one to notify the original PR author :)

tebriel (approved):
Looks like a good compromise, should we document in the readme this new behavior?

lib/github/middleware.js
```diff
@@ -60,13 +65,13 @@ module.exports = function middleware(callback) {
     const gitHubInstallationId = context.payload.installation.id;
     context.log = context.log.child({ gitHubInstallationId });
 
-    if (context.payload.sender.type === 'Bot') {
-      context.log({ noop: 'bot', botId: context.payload.sender.id, botLogin: context.payload.sender.login }, 'Halting futher execution since the sender is a bot');
+    if (context.payload.sender.type === 'Bot' && !isStateChangeAction(context.payload.action)) {
```
thombergs, line 68: @atrigueiro can we add a comment somewhere explaining why we're doing this? I.e. explain that the app can trigger pull request `edit` events which can cause bot loops and why it's safe to process bot events that are not in the list above.

lib/github/middleware.js
```diff
@@ -35,6 +35,11 @@ const isFromIgnoredRepo = (payload) =>
   // Repository: https://admin.github.com/stafftools/repositories/seequent/lf_github_testing
   payload.installation.id === 491520 && payload.repository.id === 205972230;
 
+const isStateChangeAction = (action) => [
+  'opened',
```
thombergs, line 39: @atrigueiro in the Jira issue you mention that we should add `edited` to this list. Is that no longer true? Why can these three events trigger bot loops, but not any others?

atrigueiro: Yeah I've added an explanation for it on the description of this PR.

- The act of merging - which is inherently a one-time action - won’t cause a loop so the bot-filter can be dropped for merge hooks, but left in place for other types of hooks so that the status could be properly posted in JIRA.

- The infinite loop issue may only happen on comment/description edition because it is the only case where this Jira integration actually writes data in GitHub and can trigger events. For that reason, we won’t add edited to the list of actions. I.e. If two bots are both set to auto-respond to all comments, you'd get an infinite loop.

lib/github/middleware.js
atrigueiro: @thombergs I think you might be confused with the logic here. The opposite is actually happening. We won't allow edit events because they can cause bot loops and we're also not processing any other event - except the ones from that list -which are a one-time action therefore they won’t cause a loop.

atrigueiro:
> Looks like a good compromise, should we document in the readme this new behavior?

@tebriel To be honest, I think the users expected this behaviour already, this is a bug fix. I've added a comment explaining the logic implemented on the `lib/github/middleware.js` file though.

thombergs (approved):

Raul6469 (approved):
🎉

InteNs:
is this fix live right now? or do we need to wait for a release somewhere? ('i'm looking at the environments in Github but they seem abandoned).

mboudreau:
@InteNs Hey mate, we're deploying it to staging tomorrow and doing a round of testing, then if all works well, it should go to production later in the week.  Hang tight and we'll notify you when deployed.  We're in the process of changing ownership of this integration from Github to Atlassian, where we'll be able to deploy more often with better visibility as to what's being release.

Cheers :)

PR #441, opened 2021-05-05
ARC-103 change ACTIVE to IN PROGRES

rachellerathbone:
We would like to change the term 'ACTIVE' to 'IN PROGRESS' because:

- the term 'active' can be misleading, particularly for a first time user. 'Active' may indicate to a user that their installation is now active and ready to use when, in fact, they may have a long sync ahead of them.
- to be consistent with naming conventions at Atlassian.

I have tested and confirmed adding an org shows 'IN PROGRESS' and still completes. Additionally, I noted the updated status was present in the db for an org that had a sync in progress.

Screen Shot 2021-05-05 at 1 36 53 pm

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,14 @@
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+charset = utf-8
+trim_trailing_whitespace = true
+insert_final_newline = true
+end_of_line = lf
+# editorconfig-tools is unable to ignore longs strings or urls
+max_line_length = off
+
+[CHANGELOG.md]
+indent_size = unset
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -57,7 +57,7 @@
     "jest": true
   },
   "globals": {
-    "td": "writeable"
+    "td": "writable"
   },
   "plugins": [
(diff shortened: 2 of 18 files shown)


mboudreau (changes requested):
You changed the DB value?  I'm not sure this is going to work because it means that all the values currently in the DB won't update and will stay with 'ACTIVE'.  This will probably cause some issues on deployment as you now have 2 different value types.

Best way to do it right now is just doing the change in the frontend without changing the DB, change `ACTIVE` for `IN PROGRESS` in the in the javascript code itself after you get the data.  Then we should create a ticket to change the DB value once we have ownership transfer is complete.

EDIT: Actually, scratch that, I forgot that Sequalize has a migration strategy for deployments.  You just need to add a new one there that updates the values in the DB :)

rachellerathbone:
Updated so now we only make the change on the frontend. 

Screen Shot 2021-05-07 at 10 51 54 am

DB remains the same.

Screen Shot 2021-05-07 at 10 51 47 am

mboudreau (approved):
Looks good, it'll do for now and we'll change the data later :)

PR #454, opened 2021-05-11
ARC-74 fix for error sending the complete event to Jira

rachellerathbone:
[Jira ticket.](https://softwareteams.atlassian.net/jira/software/projects/ARC/boards/385?assignee=604894ffe19f910069497c6a&selectedIssue=ARC-74)

When a sync completes when see the following error in the terminal: `ERROR probot: Error sending the `complete` event to JIRA` which comes from making a request to `github/migrationComplete` which returns a 404.

Both the [migrationComplete](https://stash.atlassian.com/projects/JIRACLOUD/repos/jira/browse/jira-components/jira-plugins/jira-development-status/jira-development-integration-plugin/src/main/java/com/atlassian/jira/plugin/devstatus/devinfo/rest/v010/DevInfoGithubMigrationResource010.java#62) and the [undoMigration](https://stash.atlassian.com/projects/JIRACLOUD/repos/jira/browse/jira-components/jira-plugins/jira-development-status/jira-development-integration-plugin/src/main/java/com/atlassian/jira/plugin/devstatus/devinfo/rest/v010/DevInfoGithubMigrationResource010.java#82) check to see if `GITHUB_CONNECT_APP_IDENTIFIER` equals `com.github.integration.production` which it obviously wouldn't in dev.

I've checked the logs in Splunk and confirmed this only happens when working with a local instance and not when installing an org in prod. I've updated the code to only make this request in prod for both endpoints. Additionally, I changed `app.log.error` to `app.log.warn` to correspond with the logging in Jira: `cleanLog.warn("Received a request to undo the migration from an app that wasn't Github");`

Screen Shot 2021-05-11 at 5 12 56 pm


jest/setEnvVars.js
```diff
@@ -0,0 +1 @@
+process.env.NODE_ENV = 'test';
```
mboudreau, line 1: is jest not getting the `NODE_ENV` from the `.env` file?  if not, we should probably use that instead to make it consistent.

lib/api/index.js
```diff
@@ -359,12 +359,26 @@ app.post('/:installationId/migrate/:undo?', [
 
   try {
     if (req.params.undo) {
-      await jiraClient.devinfo.migration.undo();
+      /**
+       * Only call github/migrationUndo in prod. Undo will only be called in Jira if
+       * GITHUB_CONNECT_APP_IDENTIFIER is equal to com.github.integration.production
+       */
+      if (process.env.NODE_ENV === 'production') {
```
mboudreau, line 366: Having logic to see if we're in production in several places sprinkled through the code is not great.  How about we create our own `jiraClient` wrapper so we can centralize that logic into a single file?

lib/sync/installation.js
```diff
@@ -86,9 +86,17 @@ module.exports.processInstallation = (app, queues) => {
...
         try {
-          await jiraClient.devinfo.migration.complete();
+          /**
+           * Only call github/migrationComplete in prod. Complete will only be called in Jira if
+           * GITHUB_CONNECT_APP_IDENTIFIER is equal to com.github.integration.production
+           */
+          if (process.env.NODE_ENV === 'production') {
+            await jiraClient.devinfo.migration.complete();
+          } else {
+            app.log('Successfully completed installing GitHub org.');
```
mboudreau, line 96: this will only get logged if we're in non-prod?

test/unit/api/api.test.js
```diff
@@ -372,6 +372,14 @@ describe('API', () => {
     });
 
     describe('undo', () => {
+      beforeEach(() => {
+        process.env.NODE_ENV = 'production';
```
mboudreau, line 376: we should probably test the logic doesn't get called when in development mode.  It would be much easier to do this if we had the wrapper as we'd only need to test the wrapper itself.

mboudreau (changes requested):

lib/sync/installation.js
rachellerathbone: It should be any environment other than prod right?

jest/setEnvVars.js
rachellerathbone: Yep. Good point.

lib/sync/installation.js
mboudreau: well, I guess it depends?  Right now, in prod, you'll have the migration actually complete, but no log, while in anything but prod, you get the log but it's a 'lie' because there was no migration.  I guess I'm questioning if it's needed at all for non-prod environments or should it always display in any environment?

lib/api/index.js
```diff
@@ -359,12 +360,27 @@ app.post('/:installationId/migrate/:undo?', [
...
     if (req.params.undo) {
-      await jiraClient.devinfo.migration.undo();
+      /**
+       * Only call github/migrationUndo in prod. Undo will only be called in Jira if
+       * GITHUB_CONNECT_APP_IDENTIFIER is equal to com.github.integration.production
+       */
+
+      if (isProd()) {
+        await jiraClient.devinfo.migration.undo();
+      }
```
mboudreau, line 370: Sorry, I might not have been clear.  I meant doing a file called `jiraClient` or something along those lines and you could add the logic of 'isProd' in there so that out here, it just looks like you're doing `jiraClient.undo()` and it handles the logic by itself

mboudreau: One way to do that is have our own file that wraps out the jira client so that you have a function of 'undo' which does the logic and then calls the real jiraClient, or you can override the actual `undo` function and add the logic in there and still calling the old function after the logic passes.

lib/api/index.js
```diff
@@ -359,14 +360,33 @@ app.post('/:installationId/migrate/:undo?', [
...
-      await jiraClient.devinfo.migration.undo();
+      /**
+       * Only call github/migrationUndo in prod. Undo will only be called in Jira if
+       * GITHUB_CONNECT_APP_IDENTIFIER is equal to com.github.integration.production
+       */
+      if (isProd()) {
+        await jiraClient.devinfo.migration.undo();
+        res.send(`Successfully called migration undo for ${installationId}`);
+      } else {
+        res.send('migrationUndo will only be called in prod');
```
rachellerathbone, line 371: Looking for thoughts on what to send here/alternatives on how to handle. We don't want to send a response with `Successfully called migration undo for ${installationId}` unless in prod. Same goes for the complete response below.

lib/api/index.js
```diff
@@ -360,13 +361,19 @@ app.post('/:installationId/migrate/:undo?', [
   try {
     if (req.params.undo) {
       await jiraClient.devinfo.migration.undo();
+
+      const responseMsg = isProd() ? 'Successfully called migrationUndo' : 'Sync failed';
```
mboudreau, line 365: 'Sync Failed' isn't right though just because you're in non-prod. I think it would be good to just leave it as `Successfully called migrationUndo` for all calls, no matter if in prod or not.

Any errors will be caught by the try/catch, which now thinking about it, we probably shouldn't wrap it around both undo or complete call, but just have a try/catch for each one so we can handle the failure event differently.

mboudreau (changes requested):
sorry for being a pain :grimacing:

lib/api/index.js
rachellerathbone: How would you recommend testing? We now call `jiraClient.devinfo.migration.undo()` and `jiraClient.devinfo.migration.complete()` for all environments and only call `instance.post` if in prod. From Splunk, I was able to confirm this request was only made from prod, but wasn't able to write a test for it (always failed saying it wasn't called).

mboudreau: you can set `process.env.NODE_ENV` in the `beforeEach` and set it back to the original in `afterEach`.

rachellerathbone: Already doing that.

mboudreau: and no bueno?

mboudreau (approved):

PR #456, opened 2021-05-12
ARC-118 support merge commits

rachellerathbone:
[Jira ticket](https://softwareteams.atlassian.net/jira/software/projects/ARC/boards/385?assignee=604894ffe19f910069497c6a&selectedIssue=ARC-118).

This extends on the work from PR [346](https://github.com/integrations/jira/pull/346) and address issue [345](https://github.com/integrations/jira/issues/345).

Currently, if a user creates a merge commit in Github this is not recognised in Jira. However, if someone does the same thing in Bitbucket, this is detected.

GitHub vs Bitbucket:

Screen Shot 2021-05-12 at 8 39 12 pm

With this addition, merge commits from GitHub are now recognised accordingly by Jira.

Screen Shot 2021-05-12 at 5 36 02 pm

**Additions I made:**

- wrapped logic in a try/catch block
- added a comment to the `isMergeCommit` var
- cleaned up spacing and destructuring
- added tests (this was something the contributor asked for help with).


lib/transforms/push.js
```diff
@@ -31,12 +31,16 @@ function mapFile(githubFile) {
...
-    id: payload.repository.id,
-    name: payload.repository.name,
-    full_name: payload.repository.full_name,
-    html_url: payload.repository.html_url,
-    owner: payload.repository.owner,
+    id,
+    name,
+    full_name,
+    html_url,
+    owner,
```
mboudreau, line 43: cool little trick with this, we can do something like:
```
const repository = {
    id,
    name,
    full_name,
    html_url,
    owner,
} = payload.repository;
```
:)

lib/transforms/push.js
```diff
@@ -75,81 +79,97 @@ async function enqueuePush(payload, jiraHost) {
...
-            url: username ? `https://github.com/${username}` : undefined,
-          },
-          authorTimestamp: githubCommit.author.date,
-          displayId: data.sha.substring(0, 6),
-          fileCount: files.length, // Send the total count for all files
-          files: filesToSend.map(mapFile),
-          id: data.sha,
-          issueKeys: sha.issueKeys,
-          url: data.html_url,
+    try {
```
mboudreau, line 82: What's the reason for the try/catch? did we have an issue with one of the calls?

mboudreau (changes requested):
few small changes, but for the most part, this is great work :)

lib/transforms/push.js
rachellerathbone: Just a precautionary measure given the new change. Think I should remove it?

lib/transforms/push.js
rachellerathbone: When this change I get `Unexpected use of 'name'  no-restricted-globals` as name is a reserved word. Should I refactor the naming of that variable or leave with the current changes?

lib/transforms/push.js
mboudreau: No,  I think it's good considering what happened with the last publish, I was just curious if there was an explicit reason for it that we needed to bring up with github.

lib/transforms/push.js
mboudreau: does it work with `name:name` instead?

mboudreau: actually, nvm, apparently you can't do it in one like as it doesn't omit the other properties.  Weird.  Leave it as is :)

mboudreau (approved):

PR #460, opened 2021-05-18
ARC-102 postlogin redirect fix

mboudreau:
Doing a quick fix as there's an issue that was found in the redirect.


rachellerathbone (changes requested):
I noticed we aren't using `events` anymore in `github-oauth`. Can you delete it?

rachellerathbone (approved):

PR #461, opened 2021-05-19
ARC-102 postlogin redirect last fix

mboudreau:
This is the last fix needed for the redirect after multiple tests with staging.


rachellerathbone (changes requested):
Similar to my nitpick the other day: unused var - `events`

mboudreau:
> Similar to my nitpick the other day: unused var - `events`

Where do you see this?

rachellerathbone (approved):

PR #467, opened 2021-05-25
ARC-108 micros deployment

rachellerathbone:
…ling to ddeploy)


app.json
```diff
@@ -0,0 +1,48 @@
+{
```
thombergs, line 1: This file is the Heroku descriptor, right? I think we can ditch it in the micros branch.

project-descriptor.yml
```diff
@@ -0,0 +1,13 @@
+service:
```
thombergs, line 1: Is this used by Nebulae to spin up the app?

thombergs (approved):

project-descriptor.yml
rachellerathbone: Yep.

PR #478, opened 2021-06-03
ARC-267 Updated the app descriptor so the app will add an item in the…

KAbakumovAtl:
In this PR I've changed the app descriptor to add another item in the navigation bar, so GitHub App configuration will be easily accessible from the Apps configuration pages:
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,4 +10,5 @@ log/*
 .tool-versions
 newrelic_agent.log
 .vscode
+.idea
 db/config.json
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -10,4 +10,5 @@ log/*
 .tool-versions
 newrelic_agent.log
 .vscode
+.idea
```
KAbakumovAtl, line 13: Seems I am the first one who uses Idea :)

rachellerathbone: Hmmm weird. That was in there before 😕

rachellerathbone (approved):

KAbakumovAtl:
Declined, created a new PR for the `micros` branch  https://github.com/integrations/jira/pull/499

PR #479, opened 2021-06-03
Adding env vars to service descriptor

rachellerathbone:


github-for-jira.sd.yml
```diff
@@ -1,4 +1,4 @@
-buildNumber: 0.0.0
+buildNumber: 0.0.3
```
thombergs, line 1: This should always stay `0.0.0` ... the deployment pipeline always deploys this tag
```suggestion
buildNumber: 0.0.0
```

github-for-jira.sd.yml
```diff
@@ -63,4 +57,28 @@ environmentOverrides:
       remoteDebug: true
     config:
       environmentVariables:
+        APP_URL: https://github-for-jira.dev.services.atlassian.com
+        WEBHOOK_SECRET: development
```
thombergs, line 61: Not sure what this secret is used for, but if it's a secret, it shouldn't be in here, but instead stashed securely

rachellerathbone: The APP_URL or the WEBHOOK_SECRET? The WEBHOOK_SECRET is publicly documented in CONTRIBUTING.md... so not very secret.

Screen Shot 2021-06-07 at 11 21 27 am

thombergs: But the value of the WEBHOOK_SECRET variable should be secret, I'm sure. Anyways, it's good to list them in the service descriptor, but it should be commented out or something.

github-for-jira.sd.yml
```diff
@@ -42,13 +42,11 @@ config:
...
     NODE_ENV: production
     NODE_OPTIONS: "--no-deprecation"
-    PORT: "8080"
-
-scaling:
-  min: 1
-  max: 1
-  instanceCount: 1
-  instance: t3.small
+    LOG_LEVEL: debug
```
KAbakumovAtl, line 45: Should we use DEBUG log level by default? 

Let's put INFO as default and DEBUG as ddev override

KAbakumovAtl (changes requested):

github-for-jira.sd.yml
```diff
@@ -63,4 +61,27 @@ environmentOverrides:
...
         NODE_OPTIONS: "--no-deprecation --inspect=0.0.0.0:5005"
+        PORT: "8080"
+        # * Commented out env vars are to be saved in secrets
+        # APP_URL: app url for dev instance
+        # WEBHOOK_SECRET: development
+        # ATLASSIAN_URL: Jira instance url
+  staging:
+    config:
+      environmentVariables:
+        APP_URL: INVALID
```
KAbakumovAtl, line 75: btw, should we put proper links here? Or we put "INVALID" to prevent the app from being used in these envs?

rachellerathbone: This was just a placeholder while working on the dev deployment. The staging & prod vars will be updated when we do the deployment work for those stages.

KAbakumovAtl (approved):

PR #480, opened 2021-06-04
ARC-226 Renamed _ping endpoint to deepcheck. Added deepcheck config t…

KAbakumovAtl:
* Renamed /_ping endpoint to /deepcheck
* Put deepcheck and healthcheck together in one file
* Added deepcheck to the app descriptor


package-lock.json
```diff
@@ -1,18068 +1,7 @@
 {
```
thombergs, line 1: Why are there so many changes in this file? Looks suspicious.

thombergs (changes requested):

KAbakumovAtl:
Not sure why, there should be none. Maybe it has something to do with my node version?

What Node version should we use for GH App?

rachellerathbone:
> Not sure why, there should be none. Maybe it has something to do with my node version?
> 
> What Node version should we use for GH App?

In this JS version the app needs to run on a node version of v12.x. The upcoming TS version will need to be on 14.x

KAbakumovAtl:
I am currently on v12.15.0 

I'll try to check it out from micros and rebuild

KAbakumovAtl:
OK, we need some help from an expert now. 

I've just checked out package-lock.json from `micros` branch. Then I run npm install. It updated the package-lock.json to the one I have in this PR.

@mboudreau  How can I help it?

thombergs (approved):

KAbakumovAtl:
I found what caused changes for package lock. Mismatch of npm versions. 

This package-lock.json was build with npm 7.x. which uses lockfile version 2. While I had npm 6.x installed locally, which uses lockfile version 1.

I updated npm to 7.x and it stopped changing package-lcok.json.

package-lock.json
mboudreau: It's fine.  I think I might have mistakenly committed `lockfileVersion: 2` by mistake while playing around with yarn.  Leave it in the PR as we should use version 1 as that's what npm supports right now.

KAbakumovAtl:
So I discussed with @mboudreau . We'll keep lockfile on version 1.x for now. We'll keep it to 1.x until we decide if we want to switch to 2.x or even to yarn.

If you people use npm 7.x, then downgrade it using:

```
sudo npm install -g npm@6.14.13  
```

KAbakumovAtl (commented):
Oh turned out I didn't press "submit review" button to make my comments visible

PR #491, opened 2021-06-17
ARC-179 Change authorisation for the app admin endpoints

KAbakumovAtl:
For admin endpoints I've changed the permissions fetch query. So it will check permissions for our repo inside Fusion-Arc Org.


src/api/index.ts
```diff
@@ -109,11 +109,11 @@ router.use(async (req: Request, res: Response, next: NextFunction): Promise<void
     }
 
     if (!data.viewer.organization) {
-      req.log.info(`Non-GitHub scoped token attempted to access staff routes: login=${data.viewer.login}, isEmployee=${data.viewer.isEmployee}`);
+      req.log.info(`Non Atlassian scoped token attempted to access staff routes: login=${data.viewer.login}, isEmployee=${data.viewer.isEmployee}`);
       res.status(401)
         .json({
           error: 'Unauthorized',
-          message: 'Token provided does not have access to the `github` organization',
+          message: 'Token provided does not have required access',
```
KAbakumovAtl, line 116: Unified the message to make it more secure

thombergs (approved):

mboudreau:
@KAbakumovAtl wish you would have waited for this merge as it should be reviewed by at least 2 people.

- You don't need the `isEmployee` fields in the graphql query as it's specific for Github Employees and that we don't actually do any checks on that field either way.  It should also be removed from the logs.
- We don't need to check the permissions of the repo itself, we just need organization level permissions for `fusion-arc` to be admin.  The resulting query would look like: 
```
{
  viewer {
    login
    organization(login: "fusion-arc") {
      viewerCanAdminister
    }
  }
}
```
And then change the `validViewerPermission` function:
```
function validViewerPermission(viewer) {
  return viewer.organization.viewerCanAdminister === "ADMIN";
}
```

PR #499, opened 2021-06-23
Arc 267 add a new navigation item to configure GitHub app

KAbakumovAtl:
In this PR I've changed the app descriptor to add another item in the navigation bar, so GitHub App configuration will be easily accessible from the Apps configuration pages:
image


mboudreau (approved):
easy :+1:

PR #501, opened 2021-06-23
ARC-253 add more metrics

rachellerathbone:
Updated existing metrics with tags recommended in [documentation](https://hello.atlassian.net/wiki/spaces/OBSERVABILITY/pages/402392597/How+to+-+Send+custom+metrics+to+SignalFx#Tagging-your-custom-metrics) and added new metrics that will hopefully prove useful once we have dashboards.

Didn't add  error messages to any of the metrics. We are planning to configure Sentry which will provide us with a more robust debugging interface. Dashboards can be used to identify errors are occurring, Sentry can be used to help identify the cause. Work for that will be done on the next [issue](https://softwareteams.atlassian.net/jira/software/projects/ARC/boards/385?assignee=604894ffe19f910069497c6a&selectedIssue=ARC-249) I'm about to start on.


atrigueiro (approved):

src/api/index.ts
```diff
@@ -167,25 +173,46 @@ router.get('/:installationId',
     } catch (err) {
       req.log.error(err);
       res.status(500).json(err);
+    } finally {
+      const elapsed = Date.now() - requestStart;
+      const tags = {
+        path:'/:installationId',
+        method: 'GET',
+        status: res.status.toString(),
```
KAbakumovAtl, line 181: In case of timeout we could get it like 500 in metrics. I'd rather make a separate `client_error` status for exceptions

src/api/index.ts
```diff
@@ -167,25 +173,46 @@ router.get('/:installationId',
...
+
+      statsd.histogram('get.installationId', elapsed, tags);
     }
   });
 
 router.get('/:installationId/repoSyncState.json',
   check('installationId').isInt(),
   returnOnValidationError,
   async (req: Request, res: Response): Promise<void> => {
+    const requestStart = Date.now();
```
KAbakumovAtl, line 192: We need to add it as a middleware function to avoid code duplication.
As it is done, for example, in: src/config/enhance-octokit.ts:33

KAbakumovAtl: Would be also great to unify it across endpoints, so we'll have a common code which emits app response time metrics

src/frontend/delete-github-subscription.ts
```diff
@@ -71,14 +75,24 @@ export default async (req: Request, res: Response): Promise<void> => {
...
+      statsd.increment('delete-github-subscription.failure', tags);
+
       res.status(403)
         .json({
           err: `Failed to delete subscription to ${req.body.installationId}. ${err}`,
         });
     }
   } catch (err) {
+    tags.push(`status:400`)
+    statsd.increment('delete-github-subscription.failure', tags);
```
mboudreau, line 94: what's the standard for these magic strings?

src/frontend/github-oauth.ts
```diff
@@ -91,12 +89,23 @@ export default (opts: OAuthOptions): GithubOAuth => {
       );
 
       req.session.githubToken = response.data.access_token;
+
       if (!req.session.githubToken) {
         return next(new Error('Missing Access Token from Github OAuth Flow.'));
       }
-      return res.redirect(redirectUrl);
+
+      return res.redirect(redirectUrl.toString());
```
mboudreau, line 97: It's already a string.

src/frontend/github-oauth.ts
```diff
@@ -58,13 +55,14 @@ export default (opts: OAuthOptions): GithubOAuth => {
...
-  async function callback(req, res, next): Promise<void> {
+  async function callback(req: Request, res: Response, next: NextFunction): Promise<void> {
+    const requestStart = Date.now();
     const { query } = url.parse(req.url, true);
     const code = query.code as string;
     const state = query.state as string;
 
     // Take save redirect url and delete it from session
-    const redirectUrl = req.session[state];
+    const redirectUrl = req.session[state] || '';
```
mboudreau, line 65: Did this give you issues?  It should always be there when the callback gets called...

src/frontend/delete-github-subscription.ts
```diff
@@ -71,14 +75,24 @@ export default async (req: Request, res: Response): Promise<void> => {
 
       await subscription.destroy();
       await submitProto(action);
+
+      tags.push(`status:202`)
+      statsd.increment('delete-github-subscription.success', tags);
```
KAbakumovAtl, line 80: We can potentially extract it as a middleware function and apply it to all our endpoints. 

Also, many of these metrics are coming out of the box from load balancers and service proxy. I believe we can just use existing platform metrics to count success/failures. 

For example: some client metrics emitted by Service Proxy: https://atlassian.signalfx.com/#/chart/v2/EwuFjPTA0AE?groupId=EUe6Xq7A4AA&configId=EUe6Kd_A0AA&startTime=-12h&endTime=Now

Load Balancer metrics: https://atlassian.signalfx.com/#/chart/v2/EUe5LRSAwAA?groupId=EUe6Xq7A4AA&configId=EUe6GqPA0AA&startTime=-12h&endTime=Now

src/github/index.ts
```diff
@@ -5,14 +5,24 @@ import pullRequest from './pull-request';
 import push from './push';
 import { createBranch, deleteBranch } from './branch';
 import webhookTimeout from '../middleware/webhook-timeout';
+import bunyan from 'bunyan';
```
mboudreau, line 8: We really should centralize this as our own `Logger` so we can abstract the implementation and change options globally.

mboudreau (changes requested):

src/frontend/github-oauth.ts
rachellerathbone: I was getting an error that it is  `UNKNOWN`

src/frontend/delete-github-subscription.ts
rachellerathbone: What do you mean?

src/github/index.ts
rachellerathbone: Ha. I actually had a centralized logger until I merged everything into your TS branch. Between my logger changes and your logger changes it got too messy so I ripped it out. I think that should be done as a separate PR, however.

src/frontend/github-oauth.ts
rachellerathbone: See above comment.

src/api/index.ts
```diff
@@ -133,14 +134,18 @@ router.get('/:installationId',
...
```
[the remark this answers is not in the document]
rachellerathbone: image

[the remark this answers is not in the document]
rachellerathbone: https://media.giphy.com/media/pCO5tKdP22RC8/giphy.gif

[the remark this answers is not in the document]
rachellerathbone: <iframe src="https://giphy.com/embed/pCO5tKdP22RC8" width="480" height="294" frameBorder="0" class="giphy-embed" allowFullScreen></iframe><p><a href="https://giphy.com/gifs/smash-getting-pro-pCO5tKdP22RC8">via GIPHY</a></p>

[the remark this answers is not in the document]
rachellerathbone: giphy

src/frontend/delete-github-subscription.ts
mboudreau: `delete-github-subscription.failure`, is there a convention?  What if we wanted to change them? etc

src/frontend/github-oauth.ts
mboudreau: really?  Huh.  Still, you have the `|| ""` up above which will make it a string

env.ts
```diff
@@ -6,8 +6,8 @@ const filepath = ".env";
 // Check to see if ngrok is up and running
 (async function main() {
 
-  // Does .env exists?
-  if (!fs.existsSync(filepath)) {
+  // Does .env exist and is the app running locally?
+  if (!fs.existsSync(filepath) && !process.env.MICROS_ENV) {
```
mboudreau, line 10: I'm not sure I understand the point of this?

rachellerathbone: There's no env file in micros so the deployment fails.

src/frontend/delete-github-subscription.ts
rachellerathbone: Any convention would need to be decided upon by us as a team.

src/api/index.ts
rachellerathbone: Can you clarify this a little more? Do you mean you want to change `status: res.status.toString()` to something like `status: client_error`?

env.ts
mboudreau: yeah, but micros doesn't call `npm start` :)

src/frontend/github-oauth.ts
rachellerathbone: That's what I thought but I see this when I remove `toString()`

Screen Shot 2021-06-24 at 7 00 46 pm

rachellerathbone: I can have just `redirectUrl.toString()` and omit the `|| ''` Error goes away with that.

src/config/statsd.ts
```diff
@@ -35,6 +38,22 @@ function hrtimer() {
   };
 }
 
+export const elapsedTimeMetrics = (path: string) => {
```
KAbakumovAtl, line 41: Nice!

src/github/index.ts
```diff
@@ -5,14 +5,24 @@ import pullRequest from './pull-request';
...
+    const { name, payload } = context;
+    logger.info({ event: name, action: payload.action }, 'Event received');
+
+    const tags = [
+      `name: webhooks`,
+      `event: ${name}`,
+      `action: ${payload.action}`,
+    ];
+
+    statsd.increment('jira_configuration_error', tags);
```
KAbakumovAtl, line 25: I wonder would `jira_configuration_error` be a metric name or there will be some kind of prefix?

Our metrics should follow some kind of naming convention

For exmaple in DSS we call them `counter.dss.commit.processor.webhook.truncated`

I think here we can also  come up with somethning like:

`github.for.jira.app.server.error.jira.configuration`

`github.for.jira.app.server.http.request.handling.duration`

src/sync/installation.ts
```diff
@@ -370,6 +371,9 @@ export const processInstallation =
       }
 
       await subscription.update({ syncStatus: 'FAILED' });
+
+      statsd.increment('installation-sync.failed');
```
KAbakumovAtl, line 375: Do we count successful syncronisations as well?

Would be great to count total sync durations (if we have the start time in our queue messages). It is nice to have. We should add it in the future

src/frontend/delete-github-subscription.ts
mboudreau: Sure, I was just thinking it would be nice to have them all in a single spot.  Not important for now.

src/frontend/github-oauth.ts
mboudreau: wait.  Is this an error only showing up in your IDE?  Does it pass with when you do `npm run build`?

env.ts
KAbakumovAtl: Oh I think it is because of

`ENV NODE_ENV development` in Dockerfile.prod

It was failing for me as well. I am fixing it in my PR: https://github.com/integrations/jira/pull/503/files

src/frontend/github-oauth.ts
mboudreau: Hm, might be something wrong with your cookie session typings?  Either way, my current version returns `any` which isn't good anyways.  please change the line to `const redirectUrl = req.session[state] as string;` and it should work.

src/frontend/github-client-middleware.ts
```diff
@@ -37,9 +40,7 @@ export const isAdmin = (githubClient: GitHubAPI) =>
       } = await githubClient.orgs.getMembership({ org, username });
       return role === "admin";
     } catch (err) {
-      console.log(err);
-      console.log(`${org} has not accepted new permission for getOrgMembership`);
-      console.log(`error=${err} org=${org}`);
+      logger.error(`${org} has not accepted new permission for getOrgMembership - error: ${err}`);
```
KAbakumovAtl, line 43: Nice!

src/frontend/github-oauth.ts
```diff
@@ -91,12 +92,16 @@ export default (opts: OAuthOptions): GithubOAuth => {
...
       if (!req.session.githubToken) {
         return next(new Error('Missing Access Token from Github OAuth Flow.'));
       }
-      return res.redirect(redirectUrl);
+
+      return res.redirect(redirectUrl.toString());
     } catch (e) {
       return next(new Error('Cannot retrieve access token from Github'));
+    } finally {
+      elapsedTimeMetrics(`https://${host}/login/oauth/access_token`);
```
KAbakumovAtl, line 104: Not sure what does it do here.

elapsedTimeMetrics will return a function which won't be executed anywhere. 

If we want to measure the endpoint time we should add it as a middleware function to the endpoint itself, not here.

Also we shouldn't use `host` as part of tags parameter. If we'll have too many hosts we'll explode tags cardinality (We should keep variants of each tag value low, otherwise it will cause problems in SignalFX)

src/tracking/index.ts
```diff
@@ -125,10 +126,13 @@ export const submitProto = async (
...
     const [name, count] = stats;
-    statsd.increment(submissionMetricName, count as number, [
+
+    const tags = [
       `schema:${name}`,
       `status:${status}`,
-    ]);
+    ];
+
+    statsd.increment(submissionMetricName, count as number, tags);
```
KAbakumovAtl, line 135: where does `submissionMetricName` coming from? I can't see it in this file, seems it will be `undefined` in runtime 🤔

KAbakumovAtl (changes requested):
Sorry i missed it during the first review cycle
* We need to check metrics naming and standartise it somehow (having some kind of naming conventions, include our service name in metirc names)
*There are a few comments in the code itself, I afraid it could fail in a few places (Need some checking)

Please let me know if you want to catch up and talk through the comments. It could safe some time on discussion. Happy to do it in any available time

env.ts
mboudreau: ah right, sorry.  Just change it to `production` and it'll work

src/frontend/app.ts
```diff
@@ -137,42 +138,74 @@ export default (octokitApp: App): Express => {
...
     oauth.checkGithubAuth,
     getGitHubSetup,
+    elapsedTimeMetrics('/github/setup'),
+  );
+
+  app.post(
+    '/github/setup',
+    csrfProtection,
+    postGitHubSetup,
+    elapsedTimeMetrics('/github/setup'),
```
mboudreau, line 148: I'm fairly certain that this won't work. Your timer start when the app initializes, and then only checks the different when the result is sent.  It should be before the function you want to check the time of and only start the timer when it gets called through express before you call `next()`

mboudreau: Also, did the library I linked to didn't work?

src/config/statsd.ts
mboudreau: Do you even need the path? you can get that from the request

src/config/statsd.ts
```diff
@@ -35,6 +38,22 @@ function hrtimer() {
...
+  const expressStatsdLogger = bunyan.createLogger({ name: 'elapsedTimeInMs' });
+  const elapsedTimeInMs = hrtimer();
```
mboudreau, line 43: this will only get the difference in time from the start of the app to when each function is called, not giving an accurate elapsed time.

src/github/index.ts
mboudreau: Agreed.

src/tracking/index.ts
mboudreau: If that was the case, PR wouldn't build.  It's at the top of the file.

mboudreau (changes requested):
Just a few things to fix but looking a lot better :)

src/frontend/github-oauth.ts
rachellerathbone: Screen Shot 2021-06-28 at 12 46 13 pm

Git blame says it's your typing 😛  but I can change it.

mboudreau: well, my IDE is saying that it's `any` and is fine without errors, so I think this is a vscode issue :)

env.ts
mboudreau: and of course, remove the added conditional in env.ts

src/github/index.ts
rachellerathbone: Screen Shot 2021-06-28 at 12 52 49 pm

All metrics are prefixed with `github-for-jira.` as defined in the global tags. I'll update the individual metrics themselves, however, to follow a better convention.

src/frontend/app.ts
rachellerathbone: I couldn't find the metrics in SFX. Additionally, the package itself says `This middleware can be used either globally or on a per-route basis (preferred)`  and then gives the example of 

```function statsd (path) {
  return function (req, res, next) {
    var method = req.method || 'unknown_method';
    req.statsdKey = ['http', method.toLowerCase(), path].join('.');
    next();
  };
}
 
app.get('/', statsd('home'), function (req, res) {
  res.send('Hello World!');
});
 ```
 
 which doesn't use the actual package. I figured it didn't make sense to install a package when we could basically use the example middleware function.

File: src/config/statsd.ts (hunk 17 above)
Comment (rachellerathbone) [reply src/config/statsd.ts:43]: Do you have a suggestion on how to fix it?

File: src/frontend/github-oauth.ts (hunk 4 above)
Comment (rachellerathbone) [reply src/frontend/github-oauth.ts:100]: You IDE-ist.

File: src/frontend/app.ts (hunk 16 above)
Comment (mboudreau) [reply src/frontend/app.ts:148]: oh, thought it was integrated with the library... how disappointing..

File: src/config/statsd.ts (hunk 17 above)
Comment (mboudreau) [reply src/config/statsd.ts:43]:
```
const expressStatsdLogger = bunyan.createLogger({ name: 'elapsedTimeInMs' });
export const elapsedTimeMetrics = (req: Request, res: Response, next: NextFunction) => {
    const path = req.path;
    const tags = { path, method: req.method };

    res.on('finish', () => {
      expressStatsdLogger.info(`${path} : ${elapsedTimeInMs()}`);
      statsd.histogram('elapsedTimeInMs', elapsedTimeInMs(), tags);
    });

    next();
};
```
And you can use it by doing `app.get(authFunction, elapsedTimeMetrics, yourGetFunction);` It has the be right before the main function itself.

src/frontend/github-oauth.ts
mboudreau: :P

src/config/statsd.ts
rachellerathbone: Oh nice. Thanks for this and the `once` tip.

mboudreau (approved):
Looks good :+1:

PR #503, opened 2021-06-24
ARC-294 ARC-301 update deployment configuration to make up running in multiple nodes

KAbakumovAtl:
In this PR 2 things are done:
* Now app runs in micros in multiple nodes
* Worker and WebServer nodes are separated (We currently have 1 of each)


.env.example
```diff
@@ -19,6 +19,7 @@ STORAGE_SECRET=
 
 # DEVELOPMENT VARIABLES
 TUNNEL_PORT=8080
+WORKER_PORT=4003
```
KAbakumovAtl, line 22: When running app locally we need to have one port open for the worker health-check endpoint (But we wouldn't use it during for normal local devloop)

Dockerfile.prod
```diff
@@ -16,6 +16,6 @@ FROM node:14.17-alpine
 USER node
 COPY --chown=node:node --from=build /app /app
 WORKDIR /app
-ENV NODE_ENV development
+ENV NODE_ENV production
```
KAbakumovAtl, line 19: Not sure why it was set to "development", would be good to know

github-for-jira.sd.yml
```diff
@@ -87,9 +118,3 @@ environmentOverrides:
       environmentVariables:
         APP_URL: https://github-for-jira.prod.services.atlassian.com
         INSTANCE_NAME: github-for-jira-production
-
-scaling:
-  min: 1
-  max: 1
-  instanceCount: 1
```
KAbakumovAtl, line 94: It is not recommended to use 'instanceCount' property so I removed it

package.json
```diff
@@ -13,8 +13,8 @@
     "start": "run-p start:main start:worker",
     "start:main": "ts-node-dev --inspect=0.0.0.0:9229 --respawn --transpile-only --rs src/main",
     "start:worker": "ts-node-dev --inspect=0.0.0.0:9230 --respawn --transpile-only --rs src/worker",
-    "start:production": "run-p start:main:production start:worker:production",
-    "start:main:production": "node .",
+    "start:production": "run-p start:WebServer:production start:worker:production",
+    "start:WebServer:production": "node .",
```
KAbakumovAtl, line 17: WebServer is the name of the default Micros group and there is no way to change it. So I had to rename the default command so the docker command would be able to pick it up.  I know it looks ugly now, so I am open for suggestions

src/worker/app.ts
```diff
@@ -2,21 +2,31 @@
...
   webhookPath: undefined,
   webhookProxy: undefined,
 });
 
+const App = async (app: Application): Promise<Application> => {
+  // Deepcheck and Healthchecks endpoints for checking app health
+  setupDeepcheck(app);
+  return app;
+};
+
```
KAbakumovAtl, line 25: I had to add health-check endpoints to worker node, because it is required by Micros

Dockerfile.prod
rachellerathbone: I think this was missed when the Dockerfiles were broken out. It was originally `production` in the micros Dockerfile branch but `development` in the TS Dockerfile branch.

.env.example
mboudreau: can we maybe keep it consistent with the other port?  Maybe make it 8081?

Dockerfile.prod
mboudreau: it was probably my bad when I moved docker files around, creating this .prod version.

package.json
mboudreau: wait, so there's no way of changing the default in micros and it's using Pascal-case? :|

src/worker/app.ts
```diff
@@ -2,21 +2,31 @@
...
 
-const probot = createProbot({
+export const probot = createProbot({
   id: Number(process.env.APP_ID),
   cert: findPrivateKey(),
 
   // These aren't needed by worker process
   secret: undefined,
-  port: undefined,
+  port: Number(process.env.WORKER_PORT) || 8080,
```
mboudreau, line 15: port 8080 would interfere with the main API port.

mboudreau (changes requested):

package.json
thombergs: I was thinking we could introduce some kind of "environment variable mapping" in the Dockerfile. That would map all Micros-specific env vars like this one to generic env vars that are used by the app. Maybe you can look into it if something like this is possible?

Dockerfile.prod
```diff
@@ -16,6 +16,6 @@ FROM node:14.17-alpine
...
 
-CMD ["npm", "run", "start:production"]
+CMD ["sh", "-c", "npm run start:${MICROS_GROUP}:production"]
```
thombergs, line 21: How many worker and web threads are we starting on each node? I guess the default is 1? Are we using the resources effectively with only 1 worker on each node?

.env.example
KAbakumovAtl: makes sense

package.json
KAbakumovAtl: I was initially thinking of something like that, but I afraid it can overcomplicate the situation.

However what can we do - map MICROS_GROUP to a nodejs command:

WebServer: "start:main:production"
Worker:    "start:worker:production"

Anything else - fail the start

src/worker/app.ts
KAbakumovAtl: No it wouldn't, when running in Micros as a "worker", there will be no main "API". When we run the main version, this Probot wouldn't start.

When we run locally the "WORKER_PORT" env variable will be used, which will be different from the main API port

Dockerfile.prod
KAbakumovAtl: Good question. We might optimise it in the future by starting more threads. For now we can control the scale with the amount of nodes.

@mboudreau How many threads are used in Heroku?

mboudreau: I think github said 8 for the API and 3-4 for the worker

src/worker/app.ts
mboudreau: I'm talking running it locally with `npm start`.  It should work out of the box without the need for the env var.

package.json
mboudreau: nah, I'd want to separate the app from the deployment as much as possible.  I agree that having it in the Dockerfile would make more sense.  Something like `CMD ["sh", "-c", "CMD=$(echo \"${MICROS_GROUP}\" | tr '[:upper:]' '[:lower:]' | sed 's/webserver/main/'); npm run start:$CMD:production"]` as the command for the dockerfile.  I just tested this in alpine docker and it seems to work.

src/worker/app.ts
KAbakumovAtl: Our process of running things locally involves copying .env.example file to .env and stting up env variables. So we should be good

package.json
KAbakumovAtl: Let's keep the script, it is easier to read

mboudreau: What script?

src/worker/app.ts
mboudreau: I know we can be, but I'm saying that environment variables, by nature, are optional.  You are setting up a potential code path that will make the port conflict with each other.  All you need to do is set the default port for the worker to 8081 and it won't be a problem.

bin/start-server-micros.sh
```diff
@@ -0,0 +1,18 @@
...
+fi
+
+if [ "$MICROS_GROUP" = "Worker" ]; then
+  COMMAND="start:worker:production"
+fi
+
+if [ -z "$COMMAND" ]; then
+  echo "Wrong MICROS_GROUP environment parameter: ${MICROS_GROUP}"
+  exit 1
+fi
```
mboudreau, line 16: Wouldn't it make more sense to use case statements?

Dockerfile.prod
```diff
@@ -16,6 +16,7 @@ FROM node:14.17-alpine
 USER node
 COPY --chown=node:node --from=build /app /app
 WORKDIR /app
-ENV NODE_ENV development
+ENV NODE_ENV production
 
-CMD ["npm", "run", "start:production"]
+RUN chmod +x ./bin/start-server-micros.sh
```
mboudreau, line 21: the file should already be executable, you don't need to add this in the dockerfile.  If it's not, make it so and commit/push the file.

.env.example
KAbakumovAtl: `bin/start-server-micros.sh`

package.json
KAbakumovAtl: `bin/start-server-micros.sh`

bin/start-server-micros.sh
KAbakumovAtl: Good point. Didn't know bashscript has it

Dockerfile.prod
KAbakumovAtl: It seems our current docker images for heroku doesn't use threads (we use the same command). I think we need to make an issue for the further threads/nodes optimisation and move on

mboudreau: Sorry, I misread that, thought you said nodes.  Threads depended on the size of their instance.  I just pushed a change to micros to use throng's default (it's a node clustering app) which will use the maximum number of threads - 1 as one needs the be the 'coordinator'.

mboudreau (approved):

PR #504, opened 2021-06-28
ARC-329 Proxy

mboudreau:
Created Global Edge proxy for the following domains:

github.stg.atlassian.com => jira-staging.githubapps.com
github.atlassian.com => jira.github.com

There is no 'dev' environment on the github side, so I removed the proxy for it.


github-for-jira.sd.yml
```diff
@@ -10,7 +10,7 @@ compose:
     image: ${DOCKER_IMAGE_NAME}
     tag: ${DOCKER_IMAGE_TAG}
     ports:
-    - 8080:8080
+      - 8080:8080
```
thombergs, line 13: https://twitter.com/brunoborges/status/1194494132380741632

github-for-jira.sd.yml
```diff
@@ -113,8 +107,28 @@ environmentOverrides:
...
         # * Commented out env vars are to be saved in secrets
         # WEBHOOK_SECRET: for signed signature when syncing an org
+    resources:
+      - type: globaledge
+        name: proxy
+        attributes:
+          default_vanity_dns: false
+          domain:
+            - github.stg.atlassian.com
+          upstream_address:
```
thombergs, line 117: So I guess we just remove the `upstream_address` when we want to make the switch and then the traffic goes to our service?

thombergs (approved):

github-for-jira.sd.yml
mboudreau: I think only one of the globaledge can create the `default_vanity_dns`.  There's a couple of options from here on how to switch over which we can discuss once the migration is done :)

PR #505, opened 2021-06-28
ARC-293 Adjusted Content Security Policy in accordance with Atlassian Security Guidelines

KAbakumovAtl:
Adjusted Content Security Policy in accordance with Atlassian Security Guidelines

### DIFF
--- a/src/frontend/index.ts
+++ b/src/frontend/index.ts
@@ -2,19 +2,38 @@ import sslify from 'express-sslify';
 import helmet from 'helmet';
 import getFrontendApp from './app';
 import {Application} from 'probot';
-import {Express, Router} from 'express';
+import {Express, NextFunction, Request, Response, Router} from 'express';
+import crypto from 'crypto';
 
 function secureHeaders(router: Router, frontendApp: Express) {
+
+  router.use((_: Request, res: Response, next: NextFunction): void => {
+    res.locals.nonce = crypto.randomBytes(16).toString('hex');
+    next();
+  });
+
+
   // Content Security Policy
   router.use(helmet.contentSecurityPolicy({
     directives: {
       defaultSrc: ["'none'"],
+      //Doesn't allow embeddable objects
+      objectSrc: ["'none'"],
       // Allow <script> tags hosted by ourselves and from atlassian when inserted into an iframe
-      scriptSrc: ["'self'", process.env.APP_URL, 'https://*.atlassian.net', 'https://*.jira.com', 'https://connect-cdn.atl-paas.net/'],
+      scriptSrc: ["'self'", process.env.APP_URL, 'https://*.atlassian.net', 'https://*.jira.com', 'https://connect-cdn.atl-paas.net/',
+        "'unsafe-inline'", "'strict-dynamic'", (_: Request, res: Response): string => {
+          return `'nonce-${res.locals.nonce}'`
+        }],
       // Allow XMLHttpRequest/fetch requests
       connectSrc: ["'self'", process.env.APP_URL],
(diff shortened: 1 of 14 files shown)


src/frontend/index.ts
```diff
@@ -2,19 +2,38 @@ import sslify from 'express-sslify';
...
+      scriptSrc: ["'self'", process.env.APP_URL, 'https://*.atlassian.net', 'https://*.jira.com', 'https://connect-cdn.atl-paas.net/',
+        "'unsafe-inline'", "'strict-dynamic'", (_: Request, res: Response): string => {
+          return `'nonce-${res.locals.nonce}'`
+        }],
       // Allow XMLHttpRequest/fetch requests
       connectSrc: ["'self'", process.env.APP_URL],
       // Allow <style> tags hosted by ourselves as well as style="" attributes
       styleSrc: ["'self'", "'unsafe-inline'"],
+      // Allow using github-for-jira pages as iframes only in jira
+      frameAncestors: ['https://*.atlassian.net', 'https://*.jira-dev.com'],
```
mboudreau, line 32: would it ever be needed to also use atlassian.com?

src/frontend/index.ts
```diff
@@ -2,19 +2,38 @@ import sslify from 'express-sslify';
...
   router.use(helmet.contentSecurityPolicy({
     directives: {
       defaultSrc: ["'none'"],
+      //Doesn't allow embeddable objects
+      objectSrc: ["'none'"],
       // Allow <script> tags hosted by ourselves and from atlassian when inserted into an iframe
-      scriptSrc: ["'self'", process.env.APP_URL, 'https://*.atlassian.net', 'https://*.jira.com', 'https://connect-cdn.atl-paas.net/'],
+      scriptSrc: ["'self'", process.env.APP_URL, 'https://*.atlassian.net', 'https://*.jira.com', 'https://connect-cdn.atl-paas.net/',
+        "'unsafe-inline'", "'strict-dynamic'", (_: Request, res: Response): string => {
+          return `'nonce-${res.locals.nonce}'`
```
mboudreau, line 25: can make the function return inline to simplify

src/frontend/index.ts
```diff
@@ -2,19 +2,38 @@ import sslify from 'express-sslify';
...
+
+  router.use((_: Request, res: Response, next: NextFunction): void => {
+    res.locals.nonce = crypto.randomBytes(16).toString('hex');
+    next();
+  });
+
+
   // Content Security Policy
   router.use(helmet.contentSecurityPolicy({
     directives: {
```
mboudreau, line 18: Personally, I wouldn't fill out every directive option for Helmet since they have very sensible defaults:
```
default-src 'self';
base-uri 'self';
block-all-mixed-content;
font-src 'self' https: data:;
frame-ancestors 'self';
img-src 'self' data:;
object-src 'none';
script-src 'self';
script-src-attr 'none';
style-src 'self' https: 'unsafe-inline';
upgrade-insecure-requests
```
set `useDefaults: true` before the directives, then only change the directives that are specific to us. Docs here: https://github.com/helmetjs/helmet

mboudreau (changes requested):
small changes, but good PR for security :+1:

src/frontend/index.ts
KAbakumovAtl: Good point. I've checked with the connect team. Currently we only have *.atlassian.net and *.jira.com in prod

mboudreau: :+1:

mboudreau (approved):
looking good :)

PR #507, opened 2021-06-29
ARC-249 configure sentry

rachellerathbone:
Configured Sentry so that we have better visibility into errors in our app via our Sentry project: [github-to-jira-integration](https://sentry.prod.atl-paas.net/atlassian/github-to-jira-integration/)

Changes:
* upgraded version of @sentry/node to 5.5.0 (version recommended it Atlassian Sentry docs)
* removed bunyan-sentry-stream and @sentry/integrations as neither were being used (create a follow-up [issue](https://softwareteams.atlassian.net/jira/software/projects/ARC/boards/385/backlog?selectedIssue=ARC-338) to determine if either could be of value)
* started pulling relevant env vars out into a centralized file (follow-up [issue](https://softwareteams.atlassian.net/jira/software/projects/ARC/boards/385/backlog?selectedIssue=ARC-340) to move remaining env vars to this file)
* added types to Sentry.Handlers.requestHandler()
* added one additional error metric (created a follow-up [issue](https://softwareteams.atlassian.net/jira/software/projects/ARC/boards/385/backlog?selectedIssue=ARC-341) to add more metrics and clean up the current implementation)

**Example errors in Sentry:**

Screen Shot 2021-06-29 at 12 56 00 pm


atrigueiro (approved):

src/config/environment-variables.ts
```diff
@@ -0,0 +1,6 @@
+export const envVars = {
+  microsEnv: process.env.MICROS_ENV,
+  microsServiceVersion: process.env.MICROS_SERVICE_VERSION,
+  nodeEnv: process.env.NODE_ENV,
+  sentryServerDSN: process.env.SENTRY_DSN,
+};
```
mboudreau, line 6: We already have something for that in `env.ts`, we just need to add the typings to it.

src/config/sentry.ts
```diff
@@ -1,8 +1,11 @@
...
+const { sentryServerDSN, microsEnv, microsServiceVersion } = envVars;
 
-export default (): void => {
-  Sentry.init({
-    dsn: process.env.SENTRY_DSN,
-    release: process.env.HEROKU_SLUG_COMMIT,
+export const initializeSentry = (): void => {
+  return Sentry.init({
+    dsn: sentryServerDSN,
+    environment: microsEnv || 'development',
```
mboudreau, line 8: Wouldn't you want to have that default set in a single place like the envVars?

src/frontend/app.ts
```diff
@@ -65,8 +66,9 @@ export default (octokitApp: App): Express => {
   const app = express();
   const rootPath = path.resolve(__dirname, '..', '..');
 
+  initializeSentry();
   // The request handler must be the first middleware on the app
-  app.use(Sentry.Handlers.requestHandler());
+  app.use(Sentry.Handlers.requestHandler() as express.RequestHandler);
```
mboudreau, line 71: You shouldn't need this if you already have `@types/sentry` included in your project because the type is the same as `express.RequestHandler`.  Can you check to make sure your vscode isn't ignoring the `@types` directory because by [default they should be included](https://www.typescriptlang.org/tsconfig#types).

mboudreau: This would also explain the issue you had with `req.session[state]`.

src/jira/index.ts
```diff
@@ -41,8 +42,9 @@ const authenticate = async (
 export default (robot: Application): void => {
   const router = robot.route('/jira');
 
+  initializeSentry();
```
mboudreau, line 45: Do we need to initialize sentry for every route?  Couldn't we just do it the highest level like the start function of main and worker?

mboudreau (changes requested):

src/config/environment-variables.ts
rachellerathbone: Do you mean that I should move this to env.ts?

mboudreau: Yeah :)

It tries to read the .env and return the parsed env vars from that, but we could combine it to try to simplify and centralize all environment variables.

I'd also recommend we use the same name as the env vars we're passing in to remove confusion or typos: `SENTRY_DSN: process.env.SENTRY_DSN`

src/worker.ts
```diff
@@ -1,16 +1,18 @@
-import throng from "throng";
-import { start } from "./worker/main";
-import InitializeSentry from "./config/sentry";
+import throng from 'throng';
+import { initializeSentry } from './config/sentry';
+import { start } from './worker/main';
 
 const isProd = process.env.NODE_ENV === 'production';
-InitializeSentry();
+initializeSentry();
```
mboudreau, line 6: This needs to be in the `worker/main.ts` start function or else it won't work with node clustering.

src/config/sentry.ts
```diff
@@ -1,8 +1,12 @@
...
+  envVars;
 
-export default (): void => {
-  Sentry.init({
-    dsn: process.env.SENTRY_DSN,
-    release: process.env.HEROKU_SLUG_COMMIT,
+export const initializeSentry = (): void => {
+  return Sentry.init({
+    dsn: sentryServerDSN,
+    environment: microsEnv || microsEnvDefault,
```
mboudreau, line 9: Sorry, I might have not been clear.  Instead of having `microsEnv || microsEnvDefault` in multiple places (which we already do), we can just set `microsEnv: process.env.MICROS_ENV || "development"` in the env vars so we never have to deal with that logic again.

src/config/environment-variables.ts
rachellerathbone: What about in micros deployments? Won't it fail if I move envVars there? env.ts is at the root. Wouldn't we need to keep envVars inside src?

mboudreau: To be clear, I mean `src/config/env.ts` :)

rachellerathbone: oh haha. Didn't even realise that was there 🤦

src/config/env.ts
```diff
@@ -1,13 +1,22 @@
...
-const filename = nodeEnv === "test" ? ".env.test" : ".env";
+import dotenv from 'dotenv';
+import path from 'path';
+
+export const envVars = {
+  MICROS_ENV: process.env.MICROS_ENV || 'development',
+  MICROS_SERVICE_VERSION: process.env.MICROS_SERVICE_VERSION,
+  NODE_ENV: process.env.NODE_ENV,
+  SENTRY_DSN: process.env.SENTRY_DSN,
+};
```
mboudreau, line 9: This export const should probably be exported as part of default.  Something like:
```
export default {
  ...env.parsed
}
```

This way, it captures all environment variables that are set in .env as well :)

src/worker.ts
```diff
@@ -1,16 +1,19 @@
...
-  }, start);
+  initializeSentry();
+  throng(
+    {
+      lifetime: Infinity,
+    },
+    start,
+  );
 } else {
+  initializeSentry();
```
mboudreau, line 17: wouldn't need to initialize it twice if you put it in the `worker/main` start function :)

package.json
```diff
@@ -39,8 +39,7 @@
     "@atlaskit/css-reset": "^2.0.5",
     "@atlaskit/reduced-ui-pack": "^8.19.0",
     "@octokit/auth-app": "^2.6.0",
-    "@sentry/integrations": "^5.15.5",
-    "@sentry/node": "^5.15.5",
+    "@sentry/node": "^5.5.0",
```
KAbakumovAtl, line 42: It looks like we are actually downgrading from 5.15.x to 5.5.x 

Should we use the latest version instead?

src/worker.ts
KAbakumovAtl: You sure it is there? I can't see it in the code.

However I agree, we should move sentry initialisation to `/worker/main.ts` hence some sentry configuration code is already there

KAbakumovAtl:
Nice! Happy to see Sentry is already useful for us :)

mboudreau (approved):
image

PR #508, opened 2021-06-29
Introduce a "maintenance mode"

thombergs:
This PR adds a "maintenance" warning to the main configuration page of the app, if the environment variable `MAINTENANCE_MODE` is set to `true`. Also, the buttons are deactivated so that a user cannot add an organization or trigger a sync. 

This maintenance mode is to be used during the expected downtime during the migration to the Atlassian infrastructure. 

It can be enabled via the Heroku CLI like this:

```
heroku config:set MAINTENANCE_MODE=true
```

And disabled via:

```
heroku config:unset MAINTENANCE_MODE
```

According to the [Heroku docs](https://devcenter.heroku.com/articles/config-vars#using-the-heroku-cli), a change in env vars will automatically restart the application so it should automatically take effect afterwards.

To test locally, run `MAINTENANCE_MODE=true npm run start`.


lib/frontend/get-jira-configuration.js
```diff
@@ -52,6 +53,7 @@ module.exports = async (req, res, next) => {
       connections,
       failedConnections,
       hasConnections: connections.length > 0 || failedConnections.length > 0,
+      maintenanceMode: process.env.MAINTENANCE_MODE === 'true' || false,
```
mboudreau, line 56: We can probably just do a check if _anything_ is set for maintenance mode instead with `maintenanceMode: !!process.env.MAINTENANCE_MODE` since the data being set on the flag isn't important.

lib/frontend/get-jira-configuration.js
```diff
@@ -42,6 +42,7 @@ module.exports = async (req, res, next) => {
         isGlobalInstall: data.repository_selection === 'all',
         installedAt: formatDate(data.updated_at),
         syncState: data.syncState,
+        maintenanceMode: process.env.MAINTENANCE_MODE === 'true' || false,
```
mboudreau, line 45: Do you need this?  Maintenance mode is global and not per installation, right?

mboudreau (changes requested):

lib/frontend/get-jira-configuration.js
thombergs: nope, just removed it and used `../maintenanceMode` in the template ... that seems to access to top-level field.

lib/frontend/get-jira-configuration.js
thombergs: I did this on purpose to avoid misunderstandings (like when someone sets the env var to `MAINTENANCE_MODE=false`, this would still evaluate to `true` if I don't explicitly compare to the string "true".

lib/frontend/get-jira-configuration.js
mboudreau: yep, correct :)

mboudreau (approved):
Looking good :+1:

mboudreau:
@thombergs I just had a thought.  What happens if the user is on another page that isn't the dashboard? they can still continue doing their thing...

Maybe we should instead have something higher level like in `frontend/app.js` which would do an `app.use(...)` to check for maintenance for all routes

thombergs:
> @thombergs I just had a thought. What happens if the user is on another page that isn't the dashboard? they can still continue doing their thing...

The user would have to be on that page right in the moment in which we enable the maintenance mode, though. But yeah, maybe it would be best if we redirect every UI to the dashboard page showing the maintenance mode. I'll have a think about that.

mboudreau:
Don't forget about APIs as well :)  It probably should be top-top level.

lib/frontend/get-jira-configuration.js
KAbakumovAtl: `|| false` - is it also to improve readability?

KAbakumovAtl (approved):

lib/frontend/get-jira-configuration.js
mboudreau: ha, that's a good point, I didn't even think of that

thombergs: yeah, I'm a big fan of code that's a bit more verbose but leaves little to your imagination :)

thombergs:
I'm going to merge this and we'll later decide if we need to "protect" other endpoints than the configuration page with the maintenance mode, as well.

mboudreau:
@thombergs I'm going to push for it as I can envision multiple instances where not locking down all endpoints are going to cause issues and "data loss".

PR #509, opened 2021-06-30
ARC-360 upgrade issue key parser

rachellerathbone:
Upgrade issue key parsing library which now ignores issue keys with a single char e.g. F-67


mboudreau (approved):
Easy peasy :)

PR #517, opened 2021-07-01
ARC-10 Use outbound proxy for calls to GitHub API

thombergs:
This PR introduces two environment variables `EXTERNAL_ONLY_PROXY_HOST` and `EXTERNAL_ONLY_PROXY_PORT`. These environment variables are provided by Micros and contain the host and port of the "outbound proxy".

With this change, all calls to the GitHub API will now go through that proxy, which guarantees that all calls are made from a fixed range of IP addresses (see section "Outgoing Connections" on [this page](https://support.atlassian.com/organization-administration/docs/ip-addresses-and-domains-for-atlassian-cloud-products/)). 

If the env vars are left empty, no proxy is used for calls to the GitHub API.


mboudreau (changes requested):
There's no need for another dependency.  All API calls are done through Axios which can give us better flexibility in the long term, but for now, you can set the proxy globally for axios by doing
```
axios.defaults.proxy = {
  host: string;
  port: number;
  auth?: {
    username: string;
    password:string;
  };
  protocol?: string;
}
```

thombergs:
@mboudreau but that would route ALL outbound requests through the proxy, which we don't want. Outbound proxy restricts outbound calls to make requests back to protected Atlassian resources, so in a staging environment, calls to Jira would fail. 

Or is there a way to tell axios to only proxy certain outbound URLs?

mboudreau:
@thombergs you make a good point and then I started to look more into it and it's more complex than that as we have to include the packaged `GithubAPI` that's in `probot`.

I think we need to look into a solution that intercepts all HTTP requests and uses a proxy if it matches a domain that we specify, like `*.github.com` or something.

mboudreau:
I can't find any proxy agent solutions where we have the control over which domain/url get proxied.  Seems to be all or nothing.  Current solution won't work as there's other GithubAPI objects inside of Probot.  We'll need to do it globally by doing:
```
import * as http from "http";
http.globalAgent = new HttpsProxyAgent(PROXY)
```
We should add this at the very top level so it's one of the first things that runs.  It should probably be set in the `start` function of both the main and worker process or call a function that sets it.

src/config/env.ts
```diff
@@ -7,11 +7,25 @@ const env = dotenv.config({
...
 });
 
+function getProxyFromEnvironment(): string{
+  let proxyHost: string = process.env.EXTERNAL_ONLY_PROXY_HOST;
+  let proxyPort: string = process.env.EXTERNAL_ONLY_PROXY_PORT;
+
+  if(!proxyHost || !proxyPort){
+    return '';
+  }
+  const proxyAddress = `http://${proxyHost}:${proxyPort}`;
```
mboudreau, line 17: Shouldn't this be https?

src/config/env.ts
```diff
@@ -7,11 +7,25 @@ const env = dotenv.config({
...
+  return proxyAddress;
+}
+
 export default {
   MICROS_ENV: process.env.MICROS_ENV || 'development',
   MICROS_SERVICE_VERSION: process.env.MICROS_SERVICE_VERSION,
   NODE_ENV: process.env.NODE_ENV,
   SENTRY_DSN: process.env.SENTRY_DSN,
+  // The proxy to use for outbound calls to GitHub. Leave empty to not use a proxy.
+  GITHUB_PROXY: getProxyFromEnvironment(),
```
mboudreau, line 28: Would probably be good to change this to `GLOBAL_PROXY`

src/config/env.ts
thombergs: no, our proxy is explicitly available via `http` (but the environment variable doesn't include the prefix, so I had to work around that like this)

src/config/env.ts
thombergs: Why? It's explicitly a proxy only for calls to GitHub. But I'll rename it to `GITHUB_API_PROXY` instead, I guess, to make it more meaningful.

thombergs:
Tested successfully with the app deployed to our staging app and a GitHub Enterprise organization with enabled IP allowlisting.

mboudreau:
You've only modified one `GithubAPI` object.  There are multiple, and there's still no guarantee that there isn't some that happen behind the scenes in our libraries like probot or octokit.  Here's an image of all creations of the `GithubAPI` object throughout the codebase:
image

Hence, to be certain that all calls to github go through the proxy we have to change the `http.globalAgent` for all requests.

mboudreau (approved):

mboudreau:
@thombergs when are we thinking of merging this? before or after migration?

thombergs:
closing as it has been resolved otherwise

PR #519, opened 2021-07-04
ARC-254 Editing Metrics

rachellerathbone:
Add `elapsedTimeMetrics` to /healthcheck and /deepcheck endpoints and renamed metric `sync` (copied from original metrics) to `pullRequestSync`. Also updated `errorStatusCode` so prevent us from getting undefined values in SFX.

As a follow-up: https://softwareteams.atlassian.net/jira/software/projects/ARC/boards/385/backlog?selectedIssue=ARC-365&text=wishlist

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -429,7 +429,11 @@ router.post(
       res.send('Successfully called migrationComplete');
       await subscription.update({ syncStatus: 'COMPLETE' });
 
-      statsd.increment(metricSyncStatus.complete);
+      const tags = {
+        installationId: githubInstallationId.toString()
+      };
+
+      statsd.increment(metricSyncStatus.complete, tags);
     } catch (err) {
       res.send(`Error trying to complete migration: ${err}`).status(500);
     }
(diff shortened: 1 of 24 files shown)


src/api/index.ts
```diff
@@ -429,7 +429,11 @@ router.post(
       res.send('Successfully called migrationComplete');
       await subscription.update({ syncStatus: 'COMPLETE' });
 
-      statsd.increment(metricSyncStatus.complete);
+      const tags = {
+        installationId: githubInstallationId.toString()
```
KAbakumovAtl, line 433: Unfortunately, we can't use "InstallationID" as a tag, because we'll explode tags cardinality.

Tags should have limited amount of variants. For example "response code" is OK, beause we are generally getting not more of dozen of them. InstallationID or Jira Host name are bad, because we have tens of thousands of Jira Instances connected to GitHub

src/jira/install.ts
```diff
@@ -28,7 +28,9 @@ export default async (req: Request, res: Response): Promise<void> => {
   action.type = ActionType.CREATED;
   action.actionSource = ActionSource.WEBHOOK;
 
-  statsd.increment(metricHttpRequest().install);
+  const tags = { installationId: installation.id.toString() }
```
KAbakumovAtl, line 31: Same here :(

KAbakumovAtl (approved):

PR #521, opened 2021-07-06
ARC-289 remove projects table

KAbakumovAtl:
I've removed the projects table and related code.

I am thinking of testing it in Ddev, however if I deploy it there there will be no turning back, we won't be able to use previous versions because the table won't be there anymore and we'll probably get SQL errors. So I am going to do after the review when I'll be confident we can merge it after testing


mboudreau:
Looks good, my only thought is that, do we have a `sequelize db:migrate` running anywhere when we deploy a new version?  Heroku has a 'prestart' script which would run before the deployment, but I'm curious what's the steps here with micros?  I'm not a big fan of a migration running before a deployment because the 'old' version is going to break until the new app is fully deployed and that doesn't really work when you're scaling or doing a slower switch over.

Might be smarter to have the DB migration stuff be done manually or maybe have different scripts depending if we want to run pre-deploy or post-deploy.  When you want to delete something, it should be done post-deploy, while adding can be done pre-deploy.

KAbakumovAtl:
Good point Michel, currently I saw the code in the main.js which will trigger the migration via system call. However it doesn't seem to be the right way to do it. I'll try to research if we can do it in a different way using Micros. However if we won't be able to, we can still trigger the migrations on startup.

mboudreau:
We can only do it in micros because there's nobody using the service.  This will not be the case moving forward. 
I'd also recommend holding off actually removing the table until after the migration since the table will come with it.  We can remove the code in the meantime though.

KAbakumovAtl:
Makes sense

mboudreau (approved):

KAbakumovAtl:
Tested in ddev, it works

PR #526, opened 2021-07-08
ARC-8: Fix GitHub Actions for Jira mapping.

atrigueiro:


lib/transforms/deployment.js
```diff
@@ -44,7 +40,6 @@ function mapEnvironment({ environment }, isProdEnvironment) {
   const jiraEnv = Object.keys(environmentMapping).find(key => isEnvironment(environmentMapping[key]));
 
   if (!jiraEnv) {
-    app.log.info(`Unmapped environment detected for deployment. Unmapped value is ${environment}. Sending it as unmapped to Jira.`);
```
atrigueiro, line 47: Couldn't access `app` inside this function, so moved logging to use `context` below.

lib/transforms/workflow.js
```diff
@@ -52,19 +66,14 @@ module.exports = (payload) => {
         state: mapStatus(workflow_run),
         lastUpdated: workflow_run.updated_at,
         issueKeys,
-        references: workflow_run.pull_requests.map(pr => (
```
atrigueiro, line 55: `map` throws an error is `pull_requests` is empty. Also, `references` can't be an empty array, so moved this check to the bottom.

lib/transforms/workflow.js
```diff
@@ -30,15 +30,29 @@ function mapStatus({ status, conclusion }) {
...
+        name: pr.head.ref,
+        uri: `${pr.head.repo.url}/tree/${pr.head.ref}`,
+      },
+    }
+  ));
+}
+module.exports = async (payload) => {
   const { workflow_run } = payload;
-  const { issueKeys } = parseSmartCommit(workflow_run.head_branch);
+  const { issueKeys } = parseSmartCommit(`${workflow_run.head_branch}\n${workflow_run.head_commit.message}`);
```
atrigueiro, line 49: Get issue key from commit message, as well as the branch name.

lib/transforms/deployment.js
```diff
@@ -80,7 +80,7 @@ module.exports = async (context) => {
         environment: {
           id: deployment_status.environment,
           displayName: deployment_status.environment,
-          type: mapEnvironment(deployment_status, deployment.production_environment),
```
atrigueiro, line 83: `production_environment` is not a valid property of the `deployment` object.

mboudreau (approved):

PR #527, opened 2021-07-09
ARC-10: configuring proxy for outbound calls

thombergs:
Finally, a working solution. Tested in `ddev`, and the sync is going through with enabled IP allowlist.

This proxies ALL outbound calls now, not only the ones towards GitHub. This is not a problem when we're working with production Jira instance. When we're working with Jira dev instances, we have to disable the proxy, because the proxy doesn't allow calls to protected Atlassian resources.

### DIFF
--- a/src/config/env.ts
+++ b/src/config/env.ts
@@ -7,11 +7,24 @@ const env = dotenv.config({
   path: path.resolve(process.cwd(), filename),
 });
 
+function getProxyFromEnvironment(): string{
+  let proxyHost: string = process.env.EXTERNAL_ONLY_PROXY_HOST;
+  let proxyPort: string = process.env.EXTERNAL_ONLY_PROXY_PORT;
+
+  if(!proxyHost || !proxyPort){
+    return '';
+  }
+  const proxyAddress = `http://${proxyHost}:${proxyPort}`;
+
+  return proxyAddress;
+}
+
 export default {
   MICROS_ENV: process.env.MICROS_ENV || 'development',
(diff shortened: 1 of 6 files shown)


src/config/env.ts
```diff
@@ -7,11 +7,24 @@ const env = dotenv.config({
...
+
+  return proxyAddress;
+}
+
 export default {
   MICROS_ENV: process.env.MICROS_ENV || 'development',
   MICROS_SERVICE_VERSION: process.env.MICROS_SERVICE_VERSION,
   NODE_ENV: process.env.NODE_ENV,
   SENTRY_DSN: process.env.SENTRY_DSN,
+  PROXY: getProxyFromEnvironment(),
```
thombergs, line 27: Translating the Micros-specific environment variable to a "neutral" variable named `PROXY`.

src/config/proxy.ts
```diff
@@ -0,0 +1,15 @@
+import envVars from './env';
+import {bootstrap} from 'global-agent';
+import Logger from 'bunyan';
+
+const logger = new Logger({name:'proxy'});
+
+if(envVars.PROXY){
+  logger.info(`configuring proxy: ${envVars.PROXY} for outbound calls`);
+  process.env.GLOBAL_AGENT_HTTP_PROXY = envVars.PROXY;
```
thombergs, line 9: Setting the environment variable required by `global-agent` from the "neutral" env var `PROXY`.

mboudreau (approved):

PR #528, opened 2021-07-12
ARC-271 DB Migration

mboudreau:
Added DB Migration bash script to make the process repeatable.  Tested it with ddev to local migration and it works well.

### DIFF
--- a/db-migration.sh
+++ b/db-migration.sh
@@ -1,0 +1,91 @@
+#!/usr/bin/env bash
+
+BOLD=$(tput bold)
+RED=$(tput setaf 1)
+GREEN=$(tput setaf 2)
+YELLOW=$(tput setaf 3)
+RESET=$(tput sgr0)
+
+## VARS
+OUTPUT=/tmp/github-for-jira.dir
+JOBS=$(($(nproc --all)-2)) # Number of cpu cores minus 2 to do parallel restore jobs
+SOURCE_URL="$1"
+TARGET_URL="$2"
+
+if ! [ -x "$(command -v pg_dump)" ]; then
+  echo "${RED}postgresql-client is not installed, please install it first.${RESET}"
+  exit 1
+fi
+
+function check_db {
+  if [ -z "$1" ]
+  then
+    echo "${RED}$2 database URL must be specified.${RESET}"
+    exit 1
+  fi
+  echo "${YELLOW}Checking connection to $2 database...${RESET}"
+  pg_isready --timeout 60 --dbname "$1"
+  if [ "$?" == "0" ]
+  then
+    echo "${GREEN}Successful connection to $2 database!${RESET}"
+  else
+    echo "${RED}Cannot connect to $2 database with URL '$1'. Exiting.${RESET}"
+    exit 1
+  fi
+}
+
+read -p "Check that you are connected to the Atlassian VPN (Press any key to continue)" PROMPT
+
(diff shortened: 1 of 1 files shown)


db-migration.sh
```diff
@@ -0,0 +1,91 @@
...
+RED=$(tput setaf 1)
+GREEN=$(tput setaf 2)
+YELLOW=$(tput setaf 3)
+RESET=$(tput sgr0)
+
+## VARS
+OUTPUT=/tmp/github-for-jira.dir
+JOBS=$(($(nproc --all)-2)) # Number of cpu cores minus 2 to do parallel restore jobs
+SOURCE_URL="$1"
+TARGET_URL="$2"
```
thombergs, line 13: The URL is just `hostname:port`?

db-migration.sh
```diff
@@ -0,0 +1,91 @@
...
+then
+  echo "${RED}Database dump failure, stopping migration.${RESET}"
+  exit 1
+fi
+
+echo "${GREEN}Source database dump to '$OUTPUT' successful.${RESET}"
+echo "${YELLOW}Restoring target database with data from '$OUTPUT' to '$TARGET_URL'...${RESET}"
+
+# creates database before restore, no ownership, use directory format, jobs for parallel tasks
+pg_restore -v --clean --if-exists --no-owner --no-privileges --jobs "$JOBS" --dbname "$TARGET_URL" "$OUTPUT"
```
thombergs, line 82: Is there anything we need to do after this migration to properly set owners and privileges? Anything we need to do so that the Micros app can access the database?

thombergs (approved):

db-migration.sh
mboudreau: It's a DB connection string, like you can find for the `PG_DATABASE_URL` env var.

db-migration.sh
mboudreau: Nope, it'll just default to the currently logged in user (of the db connection string).  This is to prevent having the user of the source DB be the 'owner' of the data on the target DB.

Since this is a migration and not a DB backup, we can remove all that extra metadata as they're not needed and we essentially need a new 'owner'

db-migration.sh
mboudreau: example `postgres://postgres:postgres@127.0.0.1:5432/jira-dev`

db-migration.sh
```diff
@@ -0,0 +1,136 @@
...
+then
+  read -p "Enter the Target Database URL: " TARGET_URL
+fi
+check_db $TARGET_URL "Target"
+
+echo "Databases ready for migration from '$SOURCE_URL' to '$TARGET_URL'."
+
+# Confirm that the user wants to continue
+read -p "${YELLOW}${BOLD}This will overwrite Target Database, are you sure you want to continue? [y/N] ${RESET}" PROMPT
+if ! [ "${PROMPT^^}" == "Y" ]
```
atrigueiro, line 98: Is bash case sensitive? The line above suggests user to type "y" but we're checking for "Y".

mboudreau: it is, but the `^^` in `"${PROMPT^^}"` turns it to uppercase instead of having a check for both `y` and `Y` :)

atrigueiro: Oh sweet!

atrigueiro (approved):

KAbakumovAtl (commented):
Looks good to me!

PR #531, opened 2021-07-13
Arc 290 arc 334 centralise logging and cleanup logs

KAbakumovAtl:
In this pull request I did the following:

1. Created a logger.ts file with shared logger for the application (Same changes as @rachellerathbone  made before)
2. Used this logger everywhere where we could. (Some Probot logs are still use its custom logger, hence we can't override it)
3. Added a filtering stream to the logger, to remove HTTP access logs which hardcoded in Probot
4. Added overrides for console.log (and its other logging functions) to use our common logging and ES Lint rule to prohibit console.log usage

I am currently testing this in ddev. In the next PR I'll get through all our logs and make sure that they comply our logging standard.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -24,8 +24,8 @@ TUNNEL_SUBDOMAIN=<subdomain>
 PRIVATE_KEY_PATH=<your-private-key>.pem
 ATLASSIAN_URL=https://<your-instance>.atlassian.net
 TRACKING_DISABLED=true
-HYDRO_BASE_URL=https://hydro-base-url.com/api/v1/events
-HYDRO_APP_SECRET=2dd220c366ec5b86104efd7324c2d405
+HYDRO_BASE_URL=
+HYDRO_APP_SECRET=
 
 # Use `trace` to get verbose logging or `info` to show less
 LOG_LEVEL=debug
--- a/.env.test
+++ b/.env.test
@@ -28,5 +28,5 @@ HYDRO_BASE_URL=https://hydro-base-url.com/api/v1/events
 HYDRO_APP_SECRET=2dd2testsecret2d405
 
 # Use `trace` to get verbose logging or `info` to show less
-LOG_LEVEL=info
+LOG_LEVEL=fatal
 WEBHOOK_PROXY_URL=http://localhost:8080/github/events
--- a/.env.test
+++ b/.env.test
@@ -25,7 +25,7 @@ PRIVATE_KEY_PATH=./test/setup/test-key.pem
 ATLASSIAN_URL=https://test-atlassian-instance.net
 TRACKING_DISABLED=true
 HYDRO_BASE_URL=https://hydro-base-url.com/api/v1/events
-HYDRO_APP_SECRET=2dd2testsecretc2d405
+HYDRO_APP_SECRET=2dd2testsecret2d405
 
 # Use `trace` to get verbose logging or `info` to show less
 LOG_LEVEL=info
--- a/.env.test
+++ b/.env.test
@@ -25,7 +25,7 @@ PRIVATE_KEY_PATH=./test/setup/test-key.pem
 ATLASSIAN_URL=https://test-atlassian-instance.net
 TRACKING_DISABLED=true
 HYDRO_BASE_URL=https://hydro-base-url.com/api/v1/events
-HYDRO_APP_SECRET=2dd220c366ec5b86104efd7324c2d405
+HYDRO_APP_SECRET=2dd2testsecretc2d405
 
 # Use `trace` to get verbose logging or `info` to show less
 LOG_LEVEL=info
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -83,7 +83,7 @@ The first time you run the app, simply run:
 ```
 npm install # installs node modules
 docker-compose up # Spin up docker containers
-npm run db:init # Creates DBs and initializes tables
+npm run db # Creates DBs and initializes tables
 ```
 
 From then on, to just run the app locally in development mode, just do `npm start`.
--- a/package.json
+++ b/package.json
@@ -61,6 +61,7 @@
     "bottleneck": "^2.19.5",
     "bull": "^3.12.1",
     "bunyan": "^1.8.15",
+    "bunyan-format": "^0.2.1",
     "cookie-session": "^2.0.0-rc.1",
     "csurf": "^1.11.0",
     "date-fns": "^1.29.0",
@@ -84,7 +85,8 @@
     "sequelize-encrypted": "^1.0.0",
     "throng": "^4.0.0",
     "tslib": "~2.2.0",
-    "typescript": "~4.2.4"
+    "typescript": "~4.2.4",
+    "uuid": "^8.3.2"
   },
   "devDependencies": {
     "@typescript-eslint/eslint-plugin": "~4.23.0",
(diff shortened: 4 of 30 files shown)


.env.example
```diff
@@ -24,8 +24,8 @@ TUNNEL_SUBDOMAIN=<subdomain>
 PRIVATE_KEY_PATH=<your-private-key>.pem
 ATLASSIAN_URL=https://<your-instance>.atlassian.net
 TRACKING_DISABLED=true
-HYDRO_BASE_URL=https://hydro-base-url.com/api/v1/events
-HYDRO_APP_SECRET=2dd220c366ec5b86104efd7324c2d405
+HYDRO_BASE_URL=
+HYDRO_APP_SECRET=
```
KAbakumovAtl, line 28: I've removed this from our code, hence we don't need it anyway

.env.test
```diff
@@ -25,7 +25,7 @@ PRIVATE_KEY_PATH=./test/setup/test-key.pem
 ATLASSIAN_URL=https://test-atlassian-instance.net
 TRACKING_DISABLED=true
 HYDRO_BASE_URL=https://hydro-base-url.com/api/v1/events
-HYDRO_APP_SECRET=2dd220c366ec5b86104efd7324c2d405
+HYDRO_APP_SECRET=2dd2testsecret2d405
```
KAbakumovAtl, line 28: Highlighting that it is not a real secret

CONTRIBUTING.md
```diff
@@ -83,7 +83,7 @@ The first time you run the app, simply run:
 ```
 npm install # installs node modules
 docker-compose up # Spin up docker containers
-npm run db:init # Creates DBs and initializes tables
+npm run db # Creates DBs and initializes tables
```
KAbakumovAtl, line 86: Changed hence `db:init` was renamed to `db`

src/util/filteringStream.ts
```diff
@@ -0,0 +1,33 @@
...
+        out.write(chunk, encoding)
+      }
+      next();
+    }
+  });
+  return writable
+}
+
+
+const shouldBeFiltered = (chunk: any) : boolean => {
```
KAbakumovAtl, line 27: @mboudreau How can I fix this warning? What should I use if I don't know the "chunk" type?

.env.example
thombergs: We can remove the lines completely!

.env.test
thombergs: We can remove the HYDRO stuff completely!

KAbakumovAtl: I tried but tests fail without it

src/jira/client/axios.ts
```diff
@@ -15,22 +16,20 @@ declare module 'axios' {
...
 function getAuthMiddleware(secret: string) {
-  const logger = bunyan.createLogger({ name: 'AXIOS' });
   return (
     /**
      * @param {import('axios').AxiosRequestConfig} config - The config for the outgoing request.
      * @returns {import('axios').AxiosRequestConfig} Updated axios config with authentication token.
      */
     (config) => {
-      logger.info(`AXIOS:`, config.url);
-      logger.info(`AXIOS URL:`, typeof config.url);
```
KAbakumovAtl, line 32: It seemed to be debug logging , so I cleaned it up

.env.test
rachellerathbone: I think we probably want to look at ripping HYDRO out as a separate issue. Maybe create a Jira issue and follow up later?

src/util/filteringStream.ts
rachellerathbone: Isn't the chunk a string here?

src/main.ts
```diff
@@ -2,23 +2,25 @@ import './config/env'; // Important to be before other dependencies
...
 import { exec } from 'child_process';
 import { initializeSentry } from './config/sentry';
+import { getLogger } from './config/logger'
 
 const isProd = process.env.NODE_ENV === 'production';
 const { redisOptions } = getRedisInfo('probot');
 
-const probot = createProbot({
+
+const probot = new Probot({
```
mboudreau, line 16: Probably keep this as `createProbot` unless you're using something with the probot constructor options directly.

src/util/filteringStream.ts
```diff
@@ -0,0 +1,33 @@
+import stream from 'stream';
```
mboudreau, line 1: just `import {Writable} from 'stream'`

src/util/filteringStream.ts
mboudreau: Pretty sure the `chunk` in this case is based on the `out` type, which I believe, since you're inputting the output of bunyan format, is going to be a string.

package.json
```diff
@@ -61,6 +61,7 @@
     "bottleneck": "^2.19.5",
     "bull": "^3.12.1",
     "bunyan": "^1.8.15",
+    "bunyan-format": "^0.2.1",
```
mboudreau, line 64: should add `@types/bunyan-format` too for type definition :)

src/util/filteringStream.ts
```diff
@@ -0,0 +1,33 @@
...
+      next();
+    }
+  });
+  return writable
+}
+
+
+const shouldBeFiltered = (chunk: any) : boolean => {
+  return chunk.includes && (chunk.includes("GET /")
+    || chunk.includes("POST /") || chunk.includes("DELETE /"))
```
mboudreau, line 29: I'm not a fan of this since we won't be able to log anything with express either.  Can't you check for the logger name?  if `probot`, just exclude?

mboudreau (changes requested):

src/util/filteringStream.ts
KAbakumovAtl: That's what I thought in the beginning. But turned out that chunk has `Buffer` type here. I am using `any` because chunk has `any` type in `write` method parameter in `WritableOptions`

src/util/filteringStream.ts
```diff
@@ -0,0 +1,36 @@
...
+      next();
+    }
+  });
+  return writable
+}
+
+//TODO Remove this code when there will be convenient way to do it in Probot.
+//See https://github.com/probot/probot/issues/1577
+const shouldBeFiltered = (chunk: any) : boolean => {
+  return chunk.includes && (chunk.includes("GET /")
```
mboudreau, line 30: This could be simplified as 
```
return chunk.toString().match(`${middlewareLoggerName}.*(GET|POST|DELETE|PUT|PATCH) \/`)
```

mboudreau (approved):
Everything looks great except one place where an improvement could be made.

src/util/filteringStream.ts
KAbakumovAtl: Hence every single log in our app will come through this check, I was trying to avoid creating extra objects during this check. So that's why I used only methods Buffer provided me. 

Maybe it will be OK to spawn an extra string here, but I'd prefer to stay safe. Logging is in general is pretty resource-intensive part of a service

mboudreau: creating a string from a buffer will be quicker than doing a loop of the buffer multiple times for the same logic.  The string creation from the buffer is done at the node vm level, as should be very quick.

KAbakumovAtl: Ok, in thus case I agree

PR #533, opened 2021-07-15
ARC-367 resync all incomplete subscriptions

mboudreau:
As part of the migration, we want a way to restart the sync for those that were in the middle of one when we enable maintenance mode.  Simplest way is to just restart sync for any subscription that isn't already complete. 

I've tested this with up to 1000 subscriptions without any database load issue.  This might trigger some old sync that never finished to start again, but we could always add a limit as to how far back we go (last week? last 24h?).


src/api/index.ts
```diff
@@ -374,33 +376,33 @@ router.post(
...
-      installation.enabled ? 'Verification successful' : 'Verification failed',
+      installation.enabled ? "Verification successful" : "Verification failed"
     );
-  },
+  }
 );
 
 router.post(
-  '/:installationId/migrate/:undo?',
+  "/:installationId/migrate/:undo?",
```
KAbakumovAtl, line 395: This endpoint can be completely removed. It was done for DVCS to GitHub App migration testing and it doesn't exist in Jira anymore afaik. 

It is better to remove it to not confuse is during the current migration

src/api/index.ts
```diff
@@ -228,109 +226,113 @@ router.get(
...
     }
-  },
+  }
 );
 
-// Grab the last n failed syncs and trigger a sync
+// RESYNC ALL INSTANCES
 router.post(
-  '/resyncFailed',
+  "/resync",
```
KAbakumovAtl, line 269: Would we need the original resyncFailed endpoint? (Should we consider `statusTypes` parameter instead?)

src/api/index.ts
mboudreau: :+1:

src/api/index.ts
mboudreau: yeah, don't need that one anymore as I've rolled it into the current one with sensible defaults.  If you wanted the same functionality, you can just set `statusTypes:['FAILED']` in the body.

rachellerathbone (approved):

KAbakumovAtl (approved):

PR #536, opened 2021-07-19
ARC-290 Make logs more informative and make sure no sensitive data logged

KAbakumovAtl:
In this Pull Request I've cleaned up logging. Added some logs to make sure we log sufficient information, changed logger names.

I've also changed some HTTP responses, so we'll return 500 error codes on unknown errors (instead of 400)

### DIFF
--- a/src/config/enhance-octokit.ts
+++ b/src/config/enhance-octokit.ts
@@ -2,11 +2,9 @@ import OctokitError from '../models/octokit-error';
 import statsd from './statsd';
 import {extractPath} from '../jira/client/axios';
 import {GitHubAPI} from 'probot';
-import {LoggerWithTarget, wrapLogger} from 'probot/lib/wrap-logger';
-import defaultLogger from './logger';
 import {metricHttpRequest} from './metric-names';
 
-const instrumentRequests = (octokit: GitHubAPI, log: LoggerWithTarget) => {
+const instrumentRequests = (octokit: GitHubAPI) => {
   octokit.hook.wrap('request', async (request, options) => {
     const requestStart = Date.now();
     let responseStatus = null;
@@ -31,7 +29,6 @@ const instrumentRequests = (octokit: GitHubAPI, log: LoggerWithTarget) => {
       };
 
       statsd.histogram(metricHttpRequest().github, elapsed, tags);
-      log.debug(tags, `GitHub request time: ${elapsed}ms`);
     }
   });
 };
@@ -42,9 +39,9 @@ const instrumentRequests = (octokit: GitHubAPI, log: LoggerWithTarget) => {
  * This acts like an Octokit plugin but works on Octokit instances.
  * (Because Probot instantiates the Octokit client for us, we can't use plugins.)
  */
-export default (octokit: GitHubAPI, originalLogger?: LoggerWithTarget): GitHubAPI => {
-  const logger = originalLogger || wrapLogger(defaultLogger);
+export default (octokit: GitHubAPI): GitHubAPI => {
+
   OctokitError.wrapRequestErrors(octokit);
-  instrumentRequests(octokit, logger);
+  instrumentRequests(octokit);
   return octokit;
 };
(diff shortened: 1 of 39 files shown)


src/config/enhance-octokit.ts
```diff
@@ -31,7 +29,6 @@ const instrumentRequests = (octokit: GitHubAPI, log: LoggerWithTarget) => {
       };
 
       statsd.histogram(metricHttpRequest().github, elapsed, tags);
-      log.debug(tags, `GitHub request time: ${elapsed}ms`);
```
KAbakumovAtl, line 34: Removed it to not spam console output when testing. We have a metric anyway.

src/config/statsd.ts
```diff
@@ -11,14 +11,19 @@ export const globalTags = {
...
 
 const statsd = new StatsD({
   prefix: 'github-for-jira.',
   host: 'platform-statsd',
   port: 8125,
   globalTags,
-  errorHandler: (err) => logger.warn({ err }, 'error writing metrics'),
+  errorHandler: (err) => {
+    if (process.env.NODE_ENV !== 'development') {
+      logger.warn(err, 'error writing metrics')
```
KAbakumovAtl, line 23: I've added a check to not log these errors when we test locally, hence there will be a lot

src/frontend/app.ts
```diff
@@ -241,10 +241,9 @@ export default (octokitApp: App): Express => {
       'Not Found': 404,
     };
 
-    const errorStatusCode = errorCodes[err.message] || 400
+    const errorStatusCode = errorCodes[err.message] || 500
```
KAbakumovAtl, line 244: If we were unable to resolve the error code by exception, we'll return 500 status code, hence the error is unknown. ( We can't blame the client if we not sure 🙂 )

src/frontend/delete-github-subscription.ts
```diff
@@ -79,7 +82,7 @@ export default async (req: Request, res: Response): Promise<void> => {
         });
     }
   } catch (err) {
-    req.log.error(err);
-    res.sendStatus(400);
+    req.log.error(err, 'Error while processing delete subscription request');
+    res.sendStatus(500);
```
KAbakumovAtl, line 86: Same here, return 500 if we caught an unknown error

src/frontend/github-client-middleware.ts
```diff
@@ -40,7 +39,7 @@ export const isAdmin = (githubClient: GitHubAPI) =>
       } = await githubClient.orgs.getMembership({ org, username });
       return role === "admin";
     } catch (err) {
-      logger.error(`${org} has not accepted new permission for getOrgMembership - error: ${err}`);
+      logger.warn(err, `${org} has not accepted new permission for getOrgMembership`);
```
KAbakumovAtl, line 42: We'll pass req.log here, so we log it with the request Id

src/github/middleware.ts
```diff
@@ -81,7 +81,6 @@ export default (
         {
           noop: 'bot',
           botId: context.payload.sender.id,
-          botLogin: context.payload.sender.login,
```
KAbakumovAtl, line 84: I was unable to reproduce this log, however `botLogin` doesn't sound like something we can log

src/github/index.ts
```diff
@@ -10,12 +10,15 @@ import { getLogger } from '../config/logger'
...
 export default (robot) => {
-  const logger = getLogger('github');
+  const logger = getLogger('github.webhooks');
 
   // TODO: Need ability to remove these listeners, especially for testing...
   robot.on('*', (context) => {
     const { name, payload } = context;
-    logger.info({ event: name, action: payload.action }, 'Event received');
+
+    context.log = logger.child({id: context.id})
```
KAbakumovAtl, line 19: Later we should add a proper tracing in our app. We need to put this ID in our queue payloads, and use it to trace our logs. Same for Installation jobs, we need to have a trace ID there, so we can query all logs related to the same sync. But this is quite a big task and out of scope of this PR

rachellerathbone (approved):
Looks good to me :)

PR #541, opened 2021-07-20
ARC-387 redirect to github/configuration

rachellerathbone:
This is first of many steps to improve the painful installation flow we have in place. These changes focus on improving the experience for users who are installing the GitHub for Jira app from the Jira marketplace and do not have their Jira instance installed in their GitHub org. As seen in the screenshot below, after a user clicks the green install button from the 'Install Jira' page, they are then redirected to the Jira marketplace. As the app would've already been installed, the steps in the purple box are redundant (another redundant path is to go marketplace > apps in navbar > manage apps > get started).

Screen Shot 2021-07-20 at 9 54 24 am

With the updated url redirect, users will skip these steps and land back on the github/configuration page with their org now displayed in the list with the option to 'Install' (see last screen to the right in above screenshot). The install flow is still far too long 🥴 but it's a start.

**Additional changes on this PR:** I cleaned up the marketplace urls and moved both the marketplace and github/configuration url out into a central location so we could remove some of our cleanup comments.

**What's not included in this PR**: updating the redirect url when a user installs the app from GitHub. Ideally, if a user installs from GitHub:
- if they don't have the GitHub installed in Jira, they get directed from github/setup to the jira marketplace
- if they do have it installed (maybe they installed it but never completed the setup or the Jira side) instead of being directed to the marketplace, they get redirected to github/configuration.
The current stream of work focuses on improving the installation flow from Jira. Improving the flow from GitHub is a separate stream we will work on later.

### DIFF
--- a/src/util/getUrl.ts
+++ b/src/util/getUrl.ts
@@ -1,0 +1,14 @@
+interface IUrlParams {
+  githubHost: string;
+  jwt: string;
+  jiraHost: string;
+}
+
+export const getGitHubConfigurationUrl = (urlParams: IUrlParams): string => {
+  const { githubHost, jwt, jiraHost } = urlParams;
+
+  return `https://${githubHost}/github/configuration?jwt=${jwt}&xdm_e=${jiraHost}`;
+};
+
+export const getJiraMarketplaceUrl = (jiraHost: string): string =>
+  `https://${jiraHost}/plugins/servlet/upm/marketplace/plugins/com.github.integration.production`;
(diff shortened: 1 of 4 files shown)


src/util/getUrl.ts
```diff
@@ -0,0 +1,14 @@
+interface IUrlParams {
+  githubHost: string;
+  jwt: string;
+  jiraHost: string;
+}
+
```
KAbakumovAtl, line 6: Nice!

KAbakumovAtl (approved):

PR #544, opened 2021-07-20
ARC-385 Adding global maintenance mode for all endpoints

mboudreau:
Any URL that's not a healthcheck or admin API will return a status of `503` with this message when `MAINTENANCE_MODE` environment variable is set to `true`:

Screenshot from 2021-07-20 18-07-22


rachellerathbone (approved):
Looks good to me! Lock it in Eddie.

PR #551, opened 2021-07-22
ARC-389 update add orgs page

rachellerathbone:
This is the first pass at updating the connect an org page. Changes included in this PR:

- display all/selected repos
- display # of repos per org
- update text for header
- move jira instance URL below header and remove from bottom
- add titles to table columns
- move 'Manage' into middle column and convert to image
- change text of button from 'Install' to 'Connect' change styling of button

More work will need to be done on this page after we convert to React to complete the design spec and use Atlaskit components.

Below:
Current implementation > Connect an org page with updates from this PR > Final design to be completed later in the stream.

Screen Shot 2021-07-23 at 11 54 18 am

Side note: can't wait to rewrite in React so I don't have to battle with existing styling.


src/config/logger.ts
```diff
@@ -3,6 +3,7 @@ import bformat from 'bunyan-format';
 import filteringStream from '../util/filteringStream'
 
 // add levelInString to include DEBUG | ERROR | INFO | WARN
+// outputMode options: short|long|simple|json|bunyan
 const formatOut = filteringStream(bformat({outputMode: 'bunyan', levelInString: true}));
```
mboudreau, line 7: I was thinking, maybe we should check to see if `NODE_ENV` is dev and make it human readable?  Either that, or use a different stream for stdout that makes it human readable in console, but uses the bunyan format (or json) to output to file.

Not really relevant to your PR, just figure it would be a nice quality of life thing.

src/frontend/get-github-configuration.ts
```diff
@@ -28,6 +30,16 @@ export default async (req: Request, res: Response, next: NextFunction): Promise<
         username: login,
         type: installation.target_type,
       });
+
+      const authedApp = await app.auth(installation.id);
+      enhanceOctokit(authedApp);
+
+      const repositories = await authedApp.paginate(
```
mboudreau, line 37: are you getting all repos just to get the number?  You should maybe consider using the graphql endpoint so you don't have to do multiple calls.

views/github-configuration.hbs
```diff
@@ -1,70 +1,123 @@
-<!doctype html>
 <html>
```
mboudreau, line 1: Should probably keep the doctype

src/frontend/get-github-configuration.ts
mboudreau: We should also consider using `Promise.all` so that all calls are parallelized instead of sequential. It'll reduce the amount of time to show a UI to the user.

mboudreau (changes requested):
Few things to change.  Some of the larger stuff can probably be done as a separate PR.

views/github-configuration.hbs
rachellerathbone: That was an accident. Didn't mean to delete that so good catch!

src/config/logger.ts
rachellerathbone: Totally agree. I might open up a separate PR from this one to update that. I'll remove the comment from this as well.

mboudreau: :+1:

src/frontend/get-github-configuration.ts
rachellerathbone: Not sure we can utilise graphql here. To get the number of repos we need to separate queries.

For orgs - the main account, so 'rachellerathbone' in my case, isn't 'considered an org so I get an error if I try to pass `installation.account.login` to this query. It does, however, give the right data back for other orgs:
```organization(login:$login) {
    repositories {
      totalCount
    }
  }```

For the primary account a second query is needed, however, it returns the incorrect data. It should return **1** (selected 1 repo) but it returns **35**. Not even sure where this number is coming from as I have 94 repos. Had a look at the list of repos coming back and they are completely random. Some I use frequently, others I haven't touched in years.
```user(login:$login) {
    repositories {
      totalCount
    }
  }```

Comment (mboudreau) [reply src/frontend/get-github-configuration.ts]: I think the proper way to do the query to the user is by using the `viewer` type which returns the authenticated user, in this case, the app at that specific installation ID:
```
query { 
  viewer { 
    repositories {
      totalCount
    }
  }
}
```

Try that to see if it returns the correct count. You can play around graphql with the explorer: https://docs.github.com/en/graphql/overview/explorer

mboudreau: I just asked saquib about the discrepancies.  I've noticed a few myself between the REST api and graphql, which is extremely concerning :/

rachellerathbone: Unfortunately the viewer query is just as whack and returns the wrong number for my primary account. Orgs are correct but 'rachellerathbone' returns the mysterious 35...

mboudreau: that's really concerning... we really need to look more into this.

rachellerathbone: Do we want to hold off on merging this in until we can use a graphql endpoint? Or make a note to update when graphql returns the correct data?

mboudreau: yeah, probably not good to have this blocked because github can't get their APIs in order :P
Let's more forward without this and I'll look this up further as it will cause issues in the future.

mboudreau (approved):

PR #552, opened 2021-07-23
ARC-385 Micros Maintenance

mboudreau:
The micros version of this PR: https://github.com/atlassian/github-for-jira/pull/544


src/config/env.ts
```diff
@@ -1,34 +1,75 @@
...
+const envVars: EnvVars = {
+  MICROS_ENV: EnvironmentEnum[process.env.MICROS_ENV || EnvironmentEnum.development],
   MICROS_SERVICE_VERSION: process.env.MICROS_SERVICE_VERSION,
-  NODE_ENV: process.env.NODE_ENV,
+  NODE_ENV: nodeEnv,
   SENTRY_DSN: process.env.SENTRY_DSN,
   PROXY: getProxyFromEnvironment(),
-  ...env.parsed,
-};
+  MAINTENANCE_MODE: BooleanEnum[process.env.MAINTENANCE_MODE],
```
KAbakumovAtl, line 41: I believe we should control this with a feature flag rather than env variable. Not sure if we integrated our service with Launch Darkly already, but we should definitely look into it for this case.

(However if we are not integrated, we can do it in the next PR, hence this one is quite big already)

src/config/env.ts
```diff
@@ -1,34 +1,75 @@
...
+export enum EnvironmentEnum {
+  test = "test",
+  development = "development",
+  production = "production",
+}
 
-function getProxyFromEnvironment(): string{
-  let proxyHost: string = process.env.EXTERNAL_ONLY_PROXY_HOST;
-  let proxyPort: string = process.env.EXTERNAL_ONLY_PROXY_PORT;
+export enum BooleanEnum {
```
KAbakumovAtl, line 11: I wonder why do we need this enum? Can't we just cast/convert our env variable to boolean?

src/config/statsd.ts
```diff
@@ -19,7 +20,7 @@ const statsd = new StatsD({
   port: 8125,
   globalTags,
   errorHandler: (err) => {
-    if (process.env.NODE_ENV !== 'development') {
+    if (process.env.NODE_ENV !== EnvironmentEnum.development) {
```
KAbakumovAtl, line 23: Nice 👍

src/frontend/app.ts
```diff
@@ -129,81 +138,95 @@ export default (octokitApp: App): Express => {
...
-  app.post('/jira/sync', verifyJiraMiddleware, elapsedTimeMetrics, retrySync);
+  app.post("/jira/sync", verifyJiraMiddleware, elapsedTimeMetrics, retrySync);
+
+
+  // Set up event handlers
+  app.get("/jira/atlassian-connect.json", getJiraConnect);
+  app.post("/jira/events/disabled", jiraAuthenticate, postJiraDisable);
+  app.post("/jira/events/enabled", jiraAuthenticate, postJiraEnable);
+  app.post("/jira/events/installed", postJiraInstall); // we can't authenticate since we don't have the secret
+  app.post("/jira/events/uninstalled", jiraAuthenticate, postJiraUninstall);
```
KAbakumovAtl, line 227: I wonder why did we move it here from `jira.ts` ? 

I find it more confusing than it was before because these are not frontend endpoints and now they are in `frontend/app.ts` file. (We also inlining stuff to the already big method)

KAbakumovAtl (changes requested):
I am a bit concerned with inlining `setupJira` and `setupDeepcheck` into the frontend app. I guess it is better to keep it as it is for now.

src/config/env.ts
mboudreau: We don't have feature flags enabled yet.  Environment variables should be easy enough to change though for the time being.

src/config/env.ts
mboudreau: yes, but no.  There's a larger change that needs to happen here for how to handle env vars.  The main issue is that the type on `process.env` is a `Record<string, string>` so we can't add booleans to it and we have to be able to use the output to recast into `process.env` in the tests for the time being.

This will change in the future, but we need to leave it as in for now.  The `BooleanEnum` just makes it a bit simpler to not screw up with typos.

src/frontend/app.ts
mboudreau: There's  a large refactor that needs to happen.  The reason for the move is that this app is spaghetti and aren't using the best express practices on top of how probot has it's own kinda server that we're using on top, but just uses the `express.Router`.  It's a mindfield.  

I moved it to frontend to simplify and be able to clearly see at what point maintenance mode stops any following routes.  When they were separate, anything that was inside of frontend couldn't stop routes that were external to it and vice versa.  We needed a way to make it more granular without having to duplicate code everywhere.

There's a much larger task here to how to split up the app correctly using express best practices.  This is the middle ground until that change is done.

src/frontend/app.ts
```diff
@@ -46,78 +54,79 @@ const oauth = GithubOAuth({
   githubClient: process.env.GITHUB_CLIENT_ID,
   githubSecret: process.env.GITHUB_CLIENT_SECRET,
   baseURL: process.env.APP_URL,
-  loginURI: '/github/login',
-  callbackURI: '/github/callback',
+  loginURI: "/github/login",
```
rachellerathbone, line 57: Think we need to agree on linting. The prettierrc file you added sets quotes to single with a trailing comma. Linting changes in this PR change quotes to double and no trailing commas 😕 

Screen Shot 2021-07-27 at 10 56 39 am

mboudreau: I'll create another PR to fix styling issues.

KAbakumovAtl (approved):

PR #555, opened 2021-07-26
ARC-413 readable logs

rachellerathbone:
Minor update to the logger so that logs are human-readable when locally but still print in JSON format for all Micros envs.

### DIFF
--- a/src/config/logger.ts
+++ b/src/config/logger.ts
@@ -2,8 +2,12 @@ import Logger, {levelFromName} from 'bunyan';
 import bformat from 'bunyan-format';
 import filteringStream from '../util/filteringStream'
 
+// For any Micros env we want the logs to be in JSON format.
+// Otherwise, if local development, we want human readable logs.
+const outputMode = process.env.MICROS_ENV ? 'json' : 'short';
+
 // add levelInString to include DEBUG | ERROR | INFO | WARN
-const formatOut = filteringStream(bformat({outputMode: 'bunyan', levelInString: true}));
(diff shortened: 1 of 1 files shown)


src/config/logger.ts
```diff
@@ -2,8 +2,12 @@ import Logger, {levelFromName} from 'bunyan';
 import bformat from 'bunyan-format';
 import filteringStream from '../util/filteringStream'
 
+// For any Micros env we want the logs to be in JSON format.
+// Otherwise, if local development, we want human readable logs.
+const outputMode = process.env.MICROS_ENV ? 'json' : 'short';
+
 // add levelInString to include DEBUG | ERROR | INFO | WARN
-const formatOut = filteringStream(bformat({outputMode: 'bunyan', levelInString: true}));
+const formatOut = filteringStream(bformat({ outputMode, levelInString: true }));
```
KAbakumovAtl, line 10: I wonder what's the difference between 'json'  and 'bunyan' here? Did you test it in micros?

KAbakumovAtl:
That's a great idea! Happy with the change as long as logs have the same format in Splunk

src/config/logger.ts
mboudreau: Looking at the code for bunyan-format, they're the same except for `json` which has indentation (multiple line) vs bunyan which is all one line.  Literally `JSON.stringify(log)` vs `JSON.stringify(log, null, 2)`.

mboudreau (approved):
yay!

PR #558, opened 2021-07-28
ARC-416 Display connected state of orgs

rachellerathbone:
Updated the connect an org page to reflect the different states of orgs: connected, connecting, not connected. 

https://user-images.githubusercontent.com/37155488/127426746-414344c1-54ce-4dbd-8ee9-c6bbff81f639.mov

This is an **intermediary step.** The final implementation after we rewite the frontend in React will be to add a link to the 'Manage Jira Subscriptions' page from a dropdown on the GitHub config page (currently accessible by the pencil editing icon on the connect an org page) and to hide connected/connecting orgs from the connect an org page.


src/frontend/app.ts
```diff
@@ -102,20 +102,36 @@ export default (octokitApp: App): Express => {
...
 	hbs.registerHelper(
 		"ifAllReposSynced",
 		(numberOfSyncedRepos, totalNumberOfRepos) =>
 			numberOfSyncedRepos === totalNumberOfRepos
 				? totalNumberOfRepos
 				: `${numberOfSyncedRepos} / ${totalNumberOfRepos}`
 	);
+
+	hbs.registerHelper("repoAccessType", (repository_selection) =>
+		repository_selection === "all" ? "All repos" : "Only select repos"
```
atrigueiro, line 117: Do we mean only `selected` repos here?

rachellerathbone: This is from Pat's design spec:

Screen Shot 2021-07-30 at 1 09 38 pm

which correlates with the terminology used by GH in the app settings:

Screen Shot 2021-07-30 at 1 11 05 pm

atrigueiro: I think this terminology is a bit confusing but it makes sense to be the same as GH.

atrigueiro (dismissed):

mboudreau (approved):

atrigueiro (approved):

PR #560, opened 2021-07-28
Update organization permission

s-anupam:
This PR updates the organization permissions for the app mentioned in contributing guide.

To install organization accounts on Jira, the app needs Read-only permission to **Members** under **Organization permissions**. It is required to check if the user installing the app is an admin of the organization or not.

The app in marketplace already takes this permission while installation.
Installing Jira on org account


mboudreau:
This change is already present in https://github.com/atlassian/github-for-jira/blob/micros/CONTRIBUTING.md which is our migration branch until ownership is completely transferred :)

PR #562, opened 2021-07-29
ARC-418 Collapse STALLED sync state into FAILED

rachellerathbone:
According to Chris from GitHub, if a STALLED sync is the result of a 'temporary technical issue' our app will try and kick the sync off again from where it got stuck. Otherwise, the user needs to resubmit the sync. It's not clear to users whether their issue is temporary or not so, to minimise confusion, we have decided to remove the STALLED state and bucket it in with the FAILED state. 

If our app does pick up a STALLED sync and is able to continue the statuses will now go FAILED > IN PROGRESS > COMPLETE. If not, it will remain in the FAILED state. Right now, a user will need to manually retry the sync but, as we gather metrics, we will work on automating this process.

### DIFF
--- a/SUPPORT.md
+++ b/SUPPORT.md
@@ -1,6 +1,6 @@
 # Questions? Need help? You've come to the right place
 
-This file will help you troubleshoot the common issues that can occur with the GitHub for Jira integration. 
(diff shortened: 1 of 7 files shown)


atrigueiro (approved):

mboudreau (approved):
Nice.

PR #568, opened 2021-08-02
ARC-419 Update logs and metrics FAILED syncStatus

rachellerathbone:
Only made some minor modifications to what we are already sending as I don't think there is a whole lot more we can add. This is mostly going to be a waiting game - we need traffic and people hitting errors that result in a FAILED status so we can start to think about how to automate resolving those issues.

The main change I made was to pass the error message in as a tag to statsd so we can have a chart 2 charts for the FAILED sync status - 1 that displays the total number of failed syncs and 1 that groups the error by error message. However, I don't think this is the end solution. We currently don't know what the error messages are and some/all may be long and not ideal for labels on a chart or could contain information we don't want to display in a UI. 

I thought this could be an intermediary step where we can use the new chart to easily collect data on the different error messages and their frequency. Once we have collected enough data, we could map each error to an error code and message that's defined by us and clearly and succinctly describes the problem.

Interested to hear everyone's thoughts. I'll create a ticket to follow-up with anything we think needs to be done after we have collected enough data (in addition to mapping errors we may want to add alerts via Sentry?). Also, feel free to point out if you think there is anywhere that metrics/logs can be added that I've missed.


src/sync/installation.ts
```diff
@@ -372,12 +387,14 @@ export const processInstallation =
...
 				await subscription.update({ syncStatus: "FAILED" });
 
-				statsd.increment(metricSyncStatus.failed);
 				logger.warn(err, `Sync failed: installationId=${installationId}`);
 
 				job.sentry.setExtra("Installation FAILED", JSON.stringify(err, null, 2));
 				job.sentry.captureException(err);
 
+				const tags = { errorMsg: JSON.stringify(err, null, 2) };
+				statsd.increment(metricSyncStatus.failed, tags);
```
KAbakumovAtl, line 396: So the problem with this approach is that we are risking to explode the metric cardinality. 

We don't know the shape of `err` at this point and it can simply include things like "Installation ID" which will change from customer to customer. We potentially have here a set of unbounded tag variations which we should avoid.

If we want to have a cause in metrics, we should map the error to some kind of Enum, where we strictly control possible variations. And then use this enum as a tag value. (Or, In Java we could use an exception class name hence we don't have many possible classes of errors, but never an exception message)

KAbakumovAtl (changes requested):

src/frontend/get-jira-configuration.ts
```diff
@@ -4,10 +4,25 @@ import { Subscription } from "../models";
...
 
+const sendFailedStatusMetrics = (installationId: string): void => {
+	const syncError = "No updates in the last 15 minutes"
+	logger.warn(syncError, `Sync failed: installationId=${installationId}`);
+
+	Sentry.setExtra("Installation FAILED", syncError);
+	Sentry.captureException(syncError);
+
+	const tags = { errorMsg: syncError };
+	statsd.increment(metricSyncStatus.failed, tags);
```
thombergs, line 23: Just a bit of trivia on metrics: we're not supposed to create metrics with tags that can have a large number of different values (like an error message). The reason is that SignalFX creates a time series for each of the values in that tag. However, given that this is error case doesn't happen too often (hopefully), I think this use case is totally fine, because it won't generate much data.

src/sync/installation.ts
rachellerathbone: So, until we have enough insight into what the errors are and can map them accordingly, would you recommend just sending logs to splunk and capturing the error with sentry? I can make a follow-up issue in Jira to map errors and add an errorMsg chart after we've collected enough data.

thombergs: Ah, I just commented on that in a separate comment :).

I think it's fine to explode cardinality here, because this error (hopefully) doesn't happen that often. 

I agree that ideally, we would map the errors to an enum, but that's only possible if we know the arrors, as @rachellerathbone pointed out in the description of this PR. 

We could maybe just cut the error message off after 100 characters or so, for now? Then see where we land with the metrics, and then at some point classify them properly and change the metric again.

KAbakumovAtl: I don't think cutting the message is an option hence even in 100 characters there is N^100 variations of the string. (We never know what will be there and identifiers can be in this first 100 characters)

I've send a link explaining the cardinality issue to Slack

KAbakumovAtl: So I think the purpose of SFX metrics is a bit different. They can indicate that we have a failures spike. We use it to detect anomalies or watch trends. If we see in SFX that something is not right, we go to Splunk and query the exact errors. Bucketing should be enough. If we see too many errors in "Other" bucket we go to Splunk an then investigating if we should make a new bucket, fix error, etc.

rachellerathbone: Just removed the tags. We can use splunk/sentry for error messages.

KAbakumovAtl (approved):

PR #570, opened 2021-08-03
ARC-385 Maintenance fix

mboudreau:
Moving atlassian-connect.json to not be affected by maintenance mode so people can still install it but can't add organizations.


mboudreau:
Created another branch to fix tests

PR #571, opened 2021-08-03
ARC-392 Add link to logout from GitHub account

rachellerathbone:
Added a link so users can logout from their GH account. Unfortunately, there doesn't appear to be a nicer way to do this so the flow is:

- **click link that opens a new tab to https://github.com/logout**: from here a user is then directed to the GItHub login page. They need to manually return to the app.
- **close the connect an org page**: as the logout/login happens totally independently from our app, the connect an org page has no idea that a user has switched accounts. We need to close this tab and click 'Add a GitHub organization' as this goes through our auth flow and provides us with updated GitHub state. Closing this tab happens 100ms after we send the user to the logout page (cannot happen in parallel)

When users return to the connect an org page they will then see the account and associated orgs for the account they just signed into.

https://user-images.githubusercontent.com/37155488/127974476-7048371f-3d9a-4e30-90fd-c4d1fa754f23.mov


static/js/github-configuration.js
```diff
@@ -28,3 +28,13 @@ $('.delete-link').click(function (event) {
...
 })
+
+$('.logout-link').click(function (event) {
+  event.preventDefault();
+
+	window.open('https://github.com/logout','_blank');
+
+	window.setTimeout(function() {
+		this.close()
+	}, 100)
```
mboudreau, line 39: If I read this right, you click on the logout link, ignore default behavior, open up a new tab for the github logout (which we don't save the reference to), then set a timeout to close our current window (/github/configuration) and leave the github logout window open?

KAbakumovAtl: Is it to make sure the window won't stay opened with the old token?

rachellerathbone: The logout window, which renders the logout button, directs to the login page. This stays open as there is no way for us to redirect from this point.

We're closing github/configuration because this page doesn't know that there has been an account logout/login. We need users to go back to the github config page so we pick up the account change.

Would there be a way to do this differently? If we wanted to keep github/configuration open we'd need to make a request to get a new token right?

mboudreau (approved):

PR #573, opened 2021-08-03
ARC-404: Adding CICD to micros branch.

atrigueiro:


src/github/index.ts
```diff
@@ -49,6 +51,10 @@ export default (robot) => {
 		middleware(pullRequest)
 	);
 
+	robot.on("workflow_run", middleware(workflow));
+
+	robot.on("deployment_status", middleware(deployment));
```
mboudreau, line 56: As per your changes above, shouldn't there be a `deployment` event as well?

src/github/middleware.ts
```diff
@@ -82,16 +82,12 @@ export default (
...
 		) {
-			context.log(
-				{
-					noop: "bot",
-					botId: context.payload.sender.id
-				},
-				"Halting further execution since the sender is a bot and action is not a state change"
-			);
-			return;
+			if (!isStateChangeOrDeploymentAction(context.name)) {
```
mboudreau, line 87: shouldn't these 2 conditions be merged into a single if statement? or is there something missing after the second if?

src/jira/client/axios.ts
```diff
@@ -114,7 +114,7 @@ function getSuccessMiddleware(logger) {
 				},
 				`Jira request: ${response.config.method.toUpperCase()} ${
 					response.config.originalUrl
-				} - ${response.status} ${response.statusText}`
+				} - ${response.status} ${response.statusText}\\n${JSON.stringify(response.data)}\`);`
```
mboudreau, line 117: the JSON.stringify bit doesn't look right.  First, you have a `\\n` which will escape and show literally `\n` in the string.  Then you have an errant escaped tilde and a `);` in the string.

src/transforms/deployment.ts
```diff
@@ -0,0 +1,89 @@
...
+			return "successful";
+		case "error":
+		case "failure":
+			return "failed";
+		default:
+			return "unknown";
+	}
+}
+
+function mapEnvironment(environment: string) {
```
mboudreau, line 26: same here.

src/transforms/deployment.ts
```diff
@@ -0,0 +1,89 @@
+import { Context } from "probot/lib/context";
+import issueKeyParser from "jira-issue-key-parser";
+
+// We need to map the state of a GitHub deployment back to a valid deployment state in Jira.
+// https://docs.github.com/en/rest/reference/repos#list-deployments
+// Deployment state - GitHub: Can be one of error, failure, pending, in_progress, queued, or success
+// https://developer.atlassian.com/cloud/jira/software/rest/api-group-builds/#api-deployments-0-1-bulk-post
+// Deployment state - Jira: Can be one of unknown, pending, in_progress, cancelled, failed, rolled_back, successful
+function mapState(state: string) {
```
mboudreau, line 9: I'd probably set the return type here.

src/transforms/deployment.ts
```diff
@@ -0,0 +1,89 @@
...
+	const jiraEnv = Object.keys(environmentMapping).find(key => isEnvironment(environmentMapping[key]));
+
+	if (!jiraEnv) {
+		return "unmapped";
+	}
+
+	return jiraEnv;
+}
+
+export default async (context: Context) => {
```
mboudreau, line 50: Would be great to get the return type here.

src/github/deployment.ts
```diff
@@ -0,0 +1,13 @@
+import transformDeployment from "../transforms/deployment";
+import { Context } from "probot/lib/context";
+
+export default async (context: Context, jiraClient) => {
+	const { data: jiraPayload } = await transformDeployment(context);
```
mboudreau, line 5: Is there a reason that we call a function that returns an object with a key of `data` that we then completely ignore that data key?  Why doesn't the function just return the jiraPayload object?  No need to faff about.

src/github/workflow.ts
```diff
@@ -0,0 +1,13 @@
+import transformWorkflow from "../transforms/workflow";
+import { Context } from "probot/lib/context";
+
+export default async (context: Context, jiraClient) => {
+	const { data: jiraPayload } = await transformWorkflow(context);
```
mboudreau, line 5: Same problem here, we completely discard the `data` key, hence we don't need to have it in the function either.

src/transforms/workflow.ts
```diff
@@ -0,0 +1,75 @@
+import issueKeyParser from "jira-issue-key-parser";
+import { Context } from "probot/lib/context";
+
+// We need to map the status and conclusion of a GitHub workflow back to a valid build state in Jira.
+// https://docs.github.com/en/rest/reference/actions#list-workflow-runs-for-a-repository
+// Workflow status - GitHub: Can be one of queued, in_progress, or completed.
+// Workflow conclusion - GitHub: Can be one of action_required, cancelled, failure, neutral, success, skipped, stale, or timed_out
+// https://developer.atlassian.com/cloud/jira/software/rest/api-group-builds/#api-builds-0-1-bulk-post
+// Build state - Jira: Can be one of pending, in_progress, successful, failed, cancelled, unknown
+function mapStatus(status: string, conclusion: string) {
```
mboudreau, line 10: return type and conclusion is optional

src/transforms/workflow.ts
```diff
@@ -0,0 +1,75 @@
...
+		case "completed.stale":
+			return "cancelled";
+		case "completed.action_required":
+			return "pending";
+		default:
+			return "unknown";
+	}
+}
+
+function mapPullRequests(pull_requests) {
```
mboudreau, line 34: parameter and return type would be nice.

mboudreau (changes requested):
Looking good, just need a few typings here and there :)

src/github/branch.ts
```diff
@@ -9,7 +9,7 @@ export const createBranch = async (context: Context, jiraClient) => {
 	if (!jiraPayload) {
 		context.log(
 			{ noop: "no_jira_payload_create_branch" },
-			"Halting futher execution for createBranch since jiraPayload is empty"
+			"Halting further execution for createBranch since jiraPayload is empty"
```
KAbakumovAtl, line 12: nice!

src/jira/client/index.ts
```diff
@@ -213,7 +213,46 @@ async function getJiraClient(
...
+					const subscription = await Subscription.getSingleInstallation(jiraHost, gitHubInstallationId);
+					await subscription.update({ syncWarning: "Exceeded issue key reference limit. Some issues may not be linked." });
+				}
+				const payload = {
+					deployments: data.deployments,
+					properties: {
+						gitHubInstallationId,
+					},
+				};
+				logger.info(`Sending deployments payload to jira. Payload: ${payload}`);
```
KAbakumovAtl, line 252: I don't think we should log payloads hence they contain UGC. If we need it for testing we should change to `debug`

- [ ] remove payloads logging in prod

KAbakumovAtl: Still good to have a logline which indicates that the request was made. So we just need to remove the ${payload} part

src/github/index.ts
atrigueiro: Nope, we only listen to `deployment_status`. I'll comment the `deployment` event from https://github.com/atlassian/github-for-jira/pull/573/files#diff-4f81d94913f59db9100bd467a208d1a30a2c9d548aa71d6a54e4fd710c3a3018R21

src/jira/client/axios.ts
atrigueiro: Copying/pasting fail 🤦‍♀️

src/github/deployment.ts
atrigueiro: No need, returning only jiraPayload now.

src/transforms/commit.ts
```diff
@@ -29,33 +30,6 @@ function mapCommit(githubCommit, author): Commit {
 	};
 }
 
-interface Commit {
```
atrigueiro, line 32: Moved to new file `src/interfaces/jira.ts`

src/transforms/deployment.ts
```diff
@@ -0,0 +1,113 @@
...
+			return "in_progress";
+		case "success":
+			return "successful";
+		case "error":
+		case "failure":
+			return "failed";
+		default:
+			return "unknown";
+	}
+}
```
atrigueiro, line 24: Don't know why the tabbing looks messed here but it looks good on my IDE
Screen Shot 2021-08-09 at 1 20 21 pm

src/github/deployment.ts
```diff
@@ -0,0 +1,13 @@
+import transformDeployment from "../transforms/deployment";
+import { Context } from "probot/lib/context";
+
+export default async (context: Context, jiraClient) => {
```
mboudreau, line 4: Should have a return type of `:Promise<void>`

src/github/workflow.ts
```diff
@@ -0,0 +1,13 @@
+import transformWorkflow from "../transforms/workflow";
+import { Context } from "probot/lib/context";
+
+export default async (context: Context, jiraClient) => {
```
mboudreau, line 4: return `Promise<void>`

src/interfaces/jira.ts
```diff
@@ -0,0 +1,81 @@
...
+		id: string;
+		repositoryUri: string;
+	};
+	ref: {
+		name: string;
+		uri: string;
+	}
+}
+
+export interface Build {
```
mboudreau, line 12: Probably be good to prefix all interfaces with `Jira` to not cause any confusion.

src/transforms/deployment.ts
mboudreau: it's fine :)
It's because it's using tabs and for some reason github shows tabs as 8 spaces, which is massive...

mboudreau (approved):
:crossed_fingers:

PR #581, opened 2021-08-04
ARC-370 Fixed JWT Authentication for iframe endpoints

KAbakumovAtl:
I've added validation to JWT tokens for iframe and calls we are making from iFrame

Also changed iframe script to generate context tokens for the calls (instead of reusing the iframe token itself)


src/jira/util/jwt.ts
```diff
@@ -10,6 +10,23 @@ const TOKEN_KEY_HEADER = `X-${TOKEN_KEY_PARAM}`;
...
+ * Context tokens are tokens which are generated by App iframes for the authentication with app backend.
+ * They don't require sqh verification and their qsh is set to a fixed `context-qsh` value.
+ *
+ * When building endpoints we should specify which type of tokens they require
+ *
+ * See details at: https://community.developer.atlassian.com/t/action-required-atlassian-connect-vulnerability-allows-bypass-of-app-qsh-verification-via-context-jwts/47072
+ */
+export enum TokenType {
+	normal = "normal",
+	context = "context"
```
KAbakumovAtl, line 27: I don't know why it is triggered hence `normal` value is used in the app.js

src/jira/util/jwt.ts
```diff
@@ -94,25 +131,16 @@ export const hasValidJwt = (secret: string, baseUrl: string, req: Request, res:
...
-			expectedHash = jwt.createQueryStringHash(req, true, baseUrl);
-			signatureHashVerified = verifiedClaims.qsh === expectedHash;
-			if (!signatureHashVerified) {
-				jwt.createCanonicalRequest(req, true, baseUrl); // eslint-disable-line
-
-				// Send the error message for the first verification - it's 90% more likely to be the one we want.
-				sendError(res, 401, "Authentication failed: query hash does not match.");
-				return false;
-			}
+		if (tokenType === TokenType.context) {
```
KAbakumovAtl, line 135: So these changes we can feature flag:
1. We always require qsh to be set
2. If token type is "context" we expect it to be equal to "context-qsh"

src/jira/util/jwt.ts
```diff
@@ -94,25 +131,16 @@ export const hasValidJwt = (secret: string, baseUrl: string, req: Request, res:
...
+			//validate query string hash
+			return verifyQsh(verifiedClaims.qsh, baseUrl, req, res);
 		}
+	} else {
+		sendError(res, 401, "JWT tokens without qsh are not allowed");
+		return false
 	}
-
-	return true;
 };
```
KAbakumovAtl, line 146: Surprisingly this file never had unit tests, so

- [ ] I'll add some tests for it

src/jira/util/jwt.ts
KAbakumovAtl: [] add a feature flag when Tom's changes are merged

src/jira/authenticate.ts
```diff
@@ -22,7 +22,7 @@ export default async (
 	res.locals.installation = installation;
 
 	// TODO: Should the express response logic be inside 'hasValidJwt'?
-	if (hasValidJwt(sharedSecret, jiraHost, req, res)) {
+	if (hasValidJwt(sharedSecret, jiraHost, req, res, TokenType.normal)) {
```
KAbakumovAtl, line 25: Using `jiraHost` url here is wrong and it was passing by accident. The reason is that Jira URL and the app URL have the same path and `atlassian-jwt` ignores the hostname (See https://bitbucket.org/atlassian/atlassian-jwt-js/src/master/lib/jwt.ts#lines-410). However to be correct we need to use the app URL here

[ ] fix base URL to be the app URL

src/jira/util/jwt.ts
```diff
@@ -57,7 +74,27 @@ function sendError(res, code, msg) {
...
+
+			// Send the error message for the first verification - it's 90% more likely to be the one we want.
+			sendError(res, 401, "Authentication failed: query hash does not match.");
+			return false;
+		}
+	}
+	return true
+}
+
+export const hasValidJwt = (secret: string, baseUrl: string, req: Request, res: Response, tokenType: TokenType) => {
```
mboudreau, line 97: I'm really not a fan of passing res/req to internal function as it creates spaghetti code and you lose who's in control of what.

All things related to the web server should be handled by the route controller to prevent stepping on toes and reducing complexity.

src/frontend/app.ts
```diff
@@ -230,19 +231,19 @@ export default (octokitApp: App): Express => {
 	app.get(
 		"/jira/configuration",
 		csrfProtection,
-		verifyJiraMiddleware,
+		verifyJiraMiddleware(TokenType.normal),
```
mboudreau, line 234: How do we know which ones are normal and which ones are not?

src/jira/util/jwt.ts
mboudreau: Do we really need the token type?  Couldn't you just do `return verifiedClaims.qsh === "context-qsh" || verifyQsh(verifiedClaims.qsh, baseUrl, req, res);`?  Why do we need to specify it?

mboudreau (commented):
Have we considered using [ACE](https://bitbucket.org/atlassian/atlassian-connect-express/src/master/)?

src/jira/util/jwt.ts
KAbakumovAtl: Yes, we do. For the security reason we need to accept either one or another way of authentication. 

It is done so we'll have only dedicated endpoints accessible with the context token "context-qsh". If somebody steal token with "context-qsh" qsh, then this person can get access to any endpoint if we don't differentiate. Otherwise other endpoints are protected by requiring strict qsh.

There are more details about it here: https://community.developer.atlassian.com/t/action-required-atlassian-connect-vulnerability-allows-bypass-of-app-qsh-verification-via-context-jwts/47072

src/jira/util/jwt.ts
KAbakumovAtl: I am a bit reluctant to refactor this part here, hence I would like to minimise the changes. However I can use it as a deleage of middleware function, but adding "next" parameter and calling it from there. Then we can rename it to "validateJwtMiddleware"

src/frontend/app.ts
KAbakumovAtl: So endpoints which are called from Jira they contain a proper "qsh" parameter. So for those endpoints we use Normal tokens. 

Tokens which we generate inside iFrame they are so-called "Context" tokens and they don't have "qsh" set to a constant value "context-qsh". It is not empty to prevent this token being used as a universal authentication token and accessing any endpoints.

So for endpoints which are accessible from the iFrame we use context token type.

static/js/jira-configuration.js
```diff
@@ -34,39 +34,47 @@ $('.configure-connection-link').click(function (event) {
...
-    url: `/jira/configuration?jwt=${encodeURIComponent(params.get('jwt'))}&xdm_e=${encodeURIComponent(params.get('xdm_e'))}`,
-    data: {
-      installationId: $(event.target).data('installation-id')
-    },
-    success: function (data) {
-      AP.navigator.reload()
-    }
-  })
+	window.AP.context.getToken(function(token){
+		$.ajax({
```
KAbakumovAtl, line 38: Now we'll generate context tokens, instead of reusing JWT passed to the iframe.

src/jira/util/jwt.ts
```diff
@@ -2,50 +2,56 @@
...
 		if (token) {
 			const foundIn = tokenInQuery ? "query" : "request body";
-			req.log(`JWT token found in ${foundIn} and in header: using ${foundIn} value.`);
+			req.log.info(`JWT token found in ${foundIn} and in header: using ${foundIn} value.`);
 		} else {
 			token = authHeader.substring(4);
 		}
 	}
 
-	// TODO: Remove when we discontinue the old token middleware
```
KAbakumovAtl, line 46: (View as side-by-side diff)
I removed this hence this part of the code already was unreachable and we don't use these parameter names for our tokens

src/jira/util/jwt.ts
```diff
@@ -2,50 +2,56 @@
...
+	const tokenInBody = req.body ? req.body[JWT_PARAM] : undefined;
 	if (tokenInQuery && tokenInBody) {
 		req.log("JWT token can only appear in either query parameter or request body.");
 		return;
 	}
 	let token = tokenInQuery || tokenInBody;
 
 	// if there was no token in the query-string then fall back to checking the Authorization header
-	const authHeader = req.headers[AUTH_HEADER];
+	const authHeader = req.headers ? req.headers[AUTH_HEADER] : undefined;
```
KAbakumovAtl, line 41: Surprisingly, but this code was unreachable because if this check above `if (!tokenInQuery && !req.body)`

src/frontend/app.ts
mboudreau: In that case, I'd recommend you use 2 different functions for the middleware and not having to input the token type to a function.

src/jira/util/jwt.ts
mboudreau: should be `req.headers?.[AUTH_HEADER] `.

I also don't see what's you're referencing with `(!tokenInQuery && !req.body)` as I don't see this change?

KAbakumovAtl: I tried `req.headers?.[AUTH_HEADER] : unknown`  but typescript gives me errors. 

I removed this part `(!tokenInQuery && !req.body)` and now check `if(!token)`  in the end

mboudreau:
Isn't there some kind of library we can use so we don't have to implement our own JWT verification?  Seems like something that would be standardized and used a lot.

src/jira/util/jwt.ts
KAbakumovAtl: todo 
 - [] add typing everywhere

KAbakumovAtl:
Closing this PR hence its changes were incorporated in this one https://github.com/atlassian/github-for-jira/pull/599

PR #584, opened 2021-08-04
ARC-423: Add support for Feature Flags

thombergs:
This adds feature flag support with LaunchDarkly. In LaunchDarkly, we can target feature flags per jiraHost URL, so we can do percentage rollouts and the like.

This PR also removes the `MAINTENANCE_MODE` env var, because the maintenance mode is now controlled via the `maintenance-mode` feature flag.


src/config/feature-flags.ts
```diff
@@ -0,0 +1,138 @@
...
+	createLaunchDarklyUser(jiraHost: string): LDUser {
+		const hash = crypto.createHash("sha1");
+		hash.update(jiraHost)
+		const hashAsHex = hash.digest("hex");
+		return {
+			key: hashAsHex
+		}
+	}
+
+	booleanValue(flag: string, jiraHost: string, defaultValue: boolean): Promise<boolean> {
```
KAbakumovAtl, line 60: Hence we not always having Jira context, should we have a method which doesn't accepts jiraHost and returns value for "default user"? 

For example, during webhook processing we don't have jira instance.

thombergs: I thought about adding such a function, but decided against it because I wanted to force the developer into the default of having a feature flag that is scoped via jiraHostUrl. 

During webhook processing, the maintenance window feature flag activates only after it knows which jiraHost the webhook belongs to.

src/config/env.ts
```diff
@@ -40,11 +40,9 @@ const envVars: EnvVars = {
 	NODE_ENV: nodeEnv,
 	SENTRY_DSN: process.env.SENTRY_DSN,
 	PROXY: getProxyFromEnvironment(),
-	MAINTENANCE_MODE: BooleanEnum[process.env.MAINTENANCE_MODE] || BooleanEnum.false,
+	LAUNCHDARKLY_KEY: process.env.LAUNCHDARKLY_KEY,
```
thombergs, line 43: I stashed the LaunchDarkly keys as secrets for the `ddev`, `stg-west`, and `prod-west` environments so that it connects to the proper LaunchDarkly project for the given environment.

src/config/feature-flags.ts
KAbakumovAtl: Ok, I guess we can add if we'll really need it in the future

KAbakumovAtl (approved):

src/config/feature-flags.ts
```diff
@@ -0,0 +1,138 @@
...
+		const user = this.createLaunchDarklyUser(jiraHost);
+		logger.info(`resolving feature flag "${flag}" for jiraHost "${jiraHost}" (hash: "${user.key}")`);
```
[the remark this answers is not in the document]
thombergs: good catch. that was just for testing. I'll remove it completely.

mboudreau (changes requested):
Seems a bit complicated for a feature flagging system that only has 2 functions.  I'd go for a simpler approach:

```
import LaunchDarkly from "launchdarkly-node-server-sdk";
import { getLogger } from "./logger";
import envVars from "./env";

const logger = getLogger("feature-flags");

const launchdarklyClient = LaunchDarkly.init(envVars.LAUNCHDARKLY_KEY, {
	logger,
	offline: !envVars.LAUNCHDARKLY_KEY
});

export enum BooleanFlags {
	MAINTENANCE_MODE = "MAINTENANCE_MODE",
	OTHER_BOOLEAN_FLAG = "OTHER_FLAG",
}

export enum StringFlags {
	OTHER_STRING_FLAG = "OTHER_STRING_FLAG",
}

interface FeatureFlagUser {
	id: string;
	secondary?: string;
	name?: string;
}


const getLaunchDarklyValue = async (flag: BooleanFlags | StringFlags, defaultValue: boolean | string, user?: FeatureFlagUser): Promise<boolean | string> => {
	try {
		await launchdarklyClient.waitForInitialization();
		const userhash = user ? launchdarklyClient.secureModeHash({
			key: user.id,
			secondary: user.secondary,
			name: user.name
		}) : "global";
		logger.info(`Resolving boolean value "${flag}" (hash: "${userhash}")`);
		return launchdarklyClient.variation(flag, { key: userhash }, defaultValue);
	} catch (err: unknown) {
		logger.error(`Error resolving boolean value "${flag}"`, err);
		return defaultValue;
	}
}

export const booleanValue = async (flag: BooleanFlags, defaultValue: boolean, user?: FeatureFlagUser): Promise<boolean> => 
	Boolean(await getLaunchDarklyValue(flag, defaultValue, user));

export const stringValue = async (flag: StringFlags, defaultValue: string, user?: FeatureFlagUser): Promise<string> => 
	String(await getLaunchDarklyValue(flag, defaultValue, user));
```
For the unit tests/mocking, we would just need to do `jest.mock("../../sc/config/feature-flags")` then mock the implementation for the specific parameters coming in.

src/config/feature-flags.ts
```diff
@@ -0,0 +1,50 @@
...
+	logger,
+	offline: !envVars.LAUNCHDARKLY_KEY
+});
+
+export enum BooleanFlags {
+	MAINTENANCE_MODE = "maintenance-mode",
+}
+
+export enum StringFlags {
+	OTHER_STRING_FLAG = "other-string-flag",
```
mboudreau, line 18: Should probably remove this for now until we have a flag for it.

src/frontend/app.ts
```diff
@@ -172,7 +173,15 @@ export default (octokitApp: App): Express => {
 	app.get("/jira/atlassian-connect.json", getJiraConnect);
 
 	// Maintenance mode view
-	app.use((req, res, next) => (isMaintenanceMode() ? getMaintenance(req, res) : next()));
+	app.use(async (req, res, next) => {
+		if (!req.session.jiraHost) {
+			// we only activate the maintenance mode for requests that are bound to a jira host
+			next();
+		} else {
```
mboudreau, line 180: This doesn't mate sense.  There's other calls that might not use or have access to the session.  And you'd think if the maintenance mode is global that it would display the maintenance page.

src/frontend/app.ts
```diff
@@ -172,7 +173,15 @@ export default (octokitApp: App): Express => {
...
+			await booleanFlag(BooleanFlags.MAINTENANCE_MODE, false, req.session.jiraHost) ? getMaintenance(req, res) : next();
```
mboudreau, line 181: are we just the await is waiting for `booleanFlag` here? or is just waiting for the ternary operator output?

src/github/middleware.ts
```diff
@@ -123,6 +117,12 @@ export default (
 		);
 		for (const subscription of subscriptions) {
 			const { jiraHost } = subscription;
+
+			if (await booleanFlag(BooleanFlags.MAINTENANCE_MODE, false, jiraHost)) {
+				context.log(`Maintenance mode ENABLED for jira host ${jiraHost} - Ignoring event of type ${webhookEvent}`);
+				continue;
+			}
```
mboudreau, line 124: Why are we putting this in subscriptions if the feature flag is based on the installation?

src/worker/main.ts
```diff
@@ -135,7 +135,7 @@ const sentryMiddleware = (jobHandler) => async (job) => {
 
 const commonMiddleware = (jobHandler) => sentryMiddleware(jobHandler);
 
-export const start = (): void => {
+export const start = async () => {
```
mboudreau, line 138: This isn't async... or at least, if it is, it's missing awaits

test/unit/maintenance/__snapshots__/maintenance.test.ts.snap
```diff
@@ -1,105 +1,5 @@
...
-  ],
-  "vendor": Object {
-    "name": "Atlassian",
-    "url": "https://atlassian.com",
-  },
-}
-`;
-
-exports[`Maintenance Frontend Maintenance should return 503 for any frontend routes 1`] = `Object {}`;
-
```
mboudreau, line 102: Where'd this all go?  Your tests are probably not working as it should as the snapshot should be the same.

mboudreau (commented):
There's some major problems in the diff of `test/unit/maintenance/maintenance.test.ts`

src/config/feature-flags.ts
thombergs: TypeScript complains when I remove it. I'll just leave it in for future reference.

src/frontend/app.ts
thombergs: Since the feature flags now support an empty jiraHost, I removed the special case.

src/frontend/app.ts
thombergs: I split it up now to make sure that we're only awaiting the `booleanFlag`.

src/github/middleware.ts
thombergs: Because the `Installations` table doesn't have the `githubInstallationId`, funnily enough :). So I can't resolve the `jiraHost` via the `Installations` table.

test/unit/maintenance/__snapshots__/maintenance.test.ts.snap
thombergs: No clue why GitHub shows this as deleted. It's still there: https://github.com/atlassian/github-for-jira/blob/feature-flags/test/unit/maintenance/__snapshots__/maintenance.test.ts.snap

thombergs:
@mboudreau addressed most comments. Not sure what you mean with "major issues" in `test/unit/maintenance/maintenance.test.ts`. The warning that we're using hardcoded credentials? I think that can be safely ignored.

mboudreau:
If you look at the diff, maintenance mode is missing several tests and the snapshot is essentially completely gone.  Your change shouldn't have affected this as drastically, especially not the snapshot.

thombergs:
@mboudreau I don't see an issue with that. I removed the test that checks the content of the atlassian-connect.js, because there is another test that checks for the status code, which is a more stable test. Since this test is removed, it doesn't need the snapshot for atlassian-connect.js anymore, and that's why the snapshot was modified.

mboudreau:
Status code can be deceiving and I specifically looked at the contents of connect.json because it needs to return the correct information for the installation while in maintenance mode, which is a specific use case.   There could be a case where a 200 status is returned with the maintenance mode HTML instead - which I need to do as part of the marketplace interceptor.

thombergs:
@mboudreau alright, re-added the test case

mboudreau (approved):

PR #587, opened 2021-08-05
ARC-433: resync per installation

thombergs:
Extends the `/resync` endpoint to allow to sync only specific installation IDs. We can use this endpoint after the migration to trigger resyncs of installations that have lost webhooks during the downtime.


mboudreau (changes requested):
Remove feature-flag branch changes.

CONTRIBUTING.md
```diff
@@ -101,6 +101,15 @@ In your Jira instance, in the `Manage Apps` section, click on your App's button,
 
 After this is done, you should see your repos starting to sync in the App's dashboard.
 
+### Accessing Admin Endpoints
+
+There are some admin endpoints (see [/src/api/index.ts](/src/api/index.ts)) that require special permissions.
+
+To call these endpoints, you must:
+
+* be an admin in a special GitHub org
```
mboudreau, line 110: You can mention the org here since it's in the code.

src/api/index.ts
```diff
@@ -1,33 +1,32 @@
-import express, { NextFunction, Request, Response } from "express";
-import { check, oneOf, validationResult } from "express-validator";
+import express, {NextFunction, Request, Response} from "express";
+import {check, oneOf, validationResult} from "express-validator";
```
mboudreau, line 2: Can you make sure your IDE has the editorconfig and eslint plugin installed? All these whitespace changes shouldn't happen.

src/api/index.ts
```diff
@@ -271,21 +270,22 @@ router.post(
...
-		});
+		let subscriptions: SubscriptionModel[];
+
+		if (installationIds.length > 0) {
+			// we want to trigger a sync for the given installation IDs only
+			subscriptions = await Subscription.getAllForInstallations(installationIds, limit, offset);
+		} else {
+			// we want to trigger a sync for all installations with one of the given status types
+			subscriptions = await Subscription.getAllForStatusTypes(statusTypes, limit, offset);
+		}
```
mboudreau, line 288: Shouldn't it also include the status types? Seems weird that one path will do all installations specified and another is only by status types.   It would make more sense to do both but by default all instances are included.

src/models/subscription.ts
```diff
@@ -72,6 +72,39 @@ export default class Subscription extends Sequelize.Model {
...
+			where: {
+				// Does a OR check on status types
+				[Op.or]: statusTypes.map(status => ({syncStatus: status}))
+			},
+			limit,
+			offset,
+			order: [["updatedAt", "DESC"]]
+		});
+	}
+
```
mboudreau, line 107: Combine these 2 functions.

mboudreau (changes requested):
While you're changing the api/index.ts file, could you please move the `/:installationId*` routes to the bottom of the file right before the export?  In express, you want to go by most specific route at the top to the least specific ones (like vars, which is essentially a wildcard).

CONTRIBUTING.md
thombergs: yeah, I know. It's just one more place to change if it ever changes. I'll just link to the code, I guess.

src/api/index.ts
```diff
@@ -388,4 +323,60 @@ router.post(
 	}
 );
 
+router.get(
```
thombergs, line 326: This was just moved down from above.

KAbakumovAtl (approved):

src/models/subscription.ts
```diff
@@ -72,6 +72,41 @@ export default class Subscription extends Sequelize.Model {
...
+	static getAllFiltered(
+		installationIds: number[],
+		statusTypes: string[],
+		limit: number,
+		offset: number
+	): Promise<Subscription[]> {
+
+		const andFilter = [];
+
+		if (statusTypes && statusTypes.length > 0) {
```
mboudreau, line 84: can use optional chaining `statusTypes?.length`.  Also, if statusTypes might be undefined, you need to set the parameter to optional by adding `?` at the end of the name.  I think the end result should be something like:

```
statusTypes: string[] = ["FAILED", "PENDING", "ACTIVE"],
installationIds: number[] = [],
offset: number = 0,
limit?: number,
```
This way, you don't even need the 2 conditional to check for length.

src/api/index.ts
```diff
@@ -271,21 +213,14 @@ router.post(
 		const syncType = req.body.syncType || "partial";
 		// Defaults to anything not completed
 		const statusTypes = req.body.statusTypes as string[] || ["FAILED", "PENDING", "ACTIVE"];
+		// Defaults to any installation
+		const installationIds = req.body.installationIds as number[] || [];
 		// Can be limited to a certain amount if needed to not overload system
 		const limit = Number(req.body.limit) || undefined;
 		// Needed for 'pagination'
 		const offset = Number(req.body.offset) || 0;
```
mboudreau, line 221: Since the function itself has been moved to the model, we might want to add the defaults there too?

mboudreau (changes requested):
so close :)

src/models/subscription.ts
thombergs: I moved the defaults to the model and simplified the if-conditions, but we can't remove the conditional checks, because they're building the SQL query with an IN operator and if the list we pass into the IN operator is empty, we get an empty result

mboudreau (approved):

PR #593, opened 2021-08-06
ARC-429 fixing db:migrate in prod

mboudreau:
Moving db:migrate from typescript code and making sure it works in prod.


package.json
```diff
@@ -14,8 +14,8 @@
     "start:main": "ts-node-dev --inspect=0.0.0.0:9229 --respawn --transpile-only --rs src/main",
     "start:worker": "ts-node-dev --inspect=0.0.0.0:9230 --respawn --transpile-only --rs src/worker",
     "start:production": "run-p start:main:production start:worker:production",
-    "start:main:production": "node .",
-    "start:worker:production": "node src/worker",
+    "start:main:production": "run-s db:migrate:prod && node .",
+    "start:worker:production": "run-s db:migrate:prod && node src/worker",
```
KAbakumovAtl, line 18: Should we run it on both worker and main? Should we just leave it only for main?

mboudreau: I think it should be done on both as we don't know which one will spin up first and should assume it could be either.  I did confirm that the migration only happens once even though called in multiple processes because of the `SequelizeMeta` table.

KAbakumovAtl: Doesn't really matter if one of the nodes will spin up before the db migration executed. We should never have a code which will rely that the migration will happen on the deployment before service startup. That mean our code should work perfectly with old and with the new state of the database. Only on the next step after the migration is completed we deploy the code change (or turn on a feature flag)

mboudreau: Well, yes, it does.  Your new code needs to know the state of the database - if your code depends on a new table or column, you need to know that it's going to be there before the code runs.  This is how we can achieve that consistently.  The old code shouldn't break on an addition.  Since we can't directly change the DB schema from our local computers, we need something that can run side by side with the code.  This is a industry standard of doing DB maintenance.

package.json
```diff
@@ -24,12 +24,14 @@
...
-    "db:create": "run-p db:create:*",
+    "db:create": "run-p db:create:dev db:create:test",
     "db:create:dev": "dotenv -- sequelize db:create || exit 0",
     "db:create:test": "dotenv -e .env.test -- sequelize db:create --env test || exit 0",
-    "db:migrate": "run-p db:migrate:*",
+    "db:create:prod": "sequelize db:create --env production || exit 0",
+    "db:migrate": "run-p db:migrate:dev db:migrate:test",
     "db:migrate:dev": "dotenv -- sequelize db:migrate",
     "db:migrate:test": "dotenv -e .env.test -- sequelize db:migrate --env test",
+    "db:migrate:prod": "sequelize db:migrate --env production",
```
thombergs, line 34: How will we run this script in prod when we have changed the DB schema? I would like to have it outside the code as well, but with the restrictions we have in place we'd always have to get a special permission from the Micros team to get production credentials with write access. 

We decided to put the DB migrate code into the application because the application has all the credentials already.

The long-term solution I had in mind was to move the code into an admin endpoint so we can decouple the database migrations from a deployment.

thombergs (changes requested):

package.json
mboudreau: couple of lines ups, it runs on production start for both worker and web threads.

thombergs: Ah, missed that.

Ok. I guess we're not going to have many or risky database changes any time soon. If we do, I think we should pull it into the code again and trigger it via HTTP, so it's decoupled from startup.

thombergs (approved):

package.json
mboudreau: Can't say I'm a fan of that method as it'll harder to test, harder to code around, harder to maintain and would potentially be a security issue. This method of doing migrations before startup is fairly common practice and has worked well.

The only other way I would recommend it would be to completely separate DB maintenance from the app and deploy it at will. However, there's still the problem that the code is essentially having to assume that everything works fine. It also makes it harder to roll back.

PR #595, opened 2021-08-10
ARC-429 Fixing connect.json to use APP_URL

mboudreau:
`atlassian-connect.json` needs to return the `APP_URL` for the `baseUrl` key or else it will make the marketplace app migration useless because it'll still point to the old heroku app url.

Also needed to fix maintenance mode to return a 200 status for `/jira/configure` when being called from Marketplace because it tries to intercept the status and doesn't show the page if not a 200.

### DIFF
--- a/lib/jira/connect.js
+++ b/lib/jira/connect.js
@@ -14,7 +14,7 @@ module.exports = async (req, res) => {
       name: `GitHub${isProd ? '' : (instance ? (` (${instance})`) : '')}`,
       description: 'Application for integrating with GitHub',
       key: `com.github.integration${instance ? `.${instance}` : ''}`,
-      baseUrl: `${isHttps ? 'https' : 'http'}://${req.get('host')}`,
+      baseUrl: process.env.APP_URL,
       lifecycle: {
         installed: '/jira/events/installed',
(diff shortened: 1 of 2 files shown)


lib/jira/connect.js
```diff
@@ -14,7 +14,7 @@ module.exports = async (req, res) => {
       name: `GitHub${isProd ? '' : (instance ? (` (${instance})`) : '')}`,
       description: 'Application for integrating with GitHub',
       key: `com.github.integration${instance ? `.${instance}` : ''}`,
-      baseUrl: `${isHttps ? 'https' : 'http'}://${req.get('host')}`,
+      baseUrl: process.env.APP_URL,
```
thombergs, line 17: Maybe we should check with Saquib that the `APP_URL` env var has the correct value in the Heroku environments before merging?

thombergs (approved):

lib/jira/connect.js
mboudreau: `APP_URL` is needed for the app to run.  Currently staging is set to `github.stg.atlassian.com` while prod is currently set to their old domain `jira.github.com`, which we will soon ask to change to use our proxy.

PR #596, opened 2021-08-10
ARC-429 Fixing maintenance mode to show in marketplace

mboudreau:
Need to be able to show it in marketplace.  I also replaced a graphic that was using a PNG where we could use an SVG for better clarity.


atrigueiro (approved):

static/css/jira-configuration.css
```diff
@@ -40,7 +40,7 @@
 .jiraConfiguration__empty {
     margin: auto;
     max-width: 570px;
-    padding: 8.5em 0;
+    padding: 90px 0;
```
rachellerathbone, line 43: What's the reasoning for switching to px? Using a relative unit of measure means the size will adjust as the screen size changes. This, however, is a fixed size.

rachellerathbone (changes requested):

static/css/jira-configuration.css
mboudreau: I can change it back, but I'm not a big fan of font-size specific padding/margin/whatever because different places might have different font sizes - like if you were to put padding on in an H3 or something else that might use slightly larger text, your consistency for padding will be all off.  Like currently, on my screen, the padding was 138px based on em and made the whole message fairly close to the bottom and if I resize my screen, the messaging does come out of view.

Not sure what the standard is here, but in the design it's at 90px.  I find it much more explicit of using pixels combined with media queries so it can change based on screen size.  Would be even better if all of this was done at the global level for consistency.  However, we're about to redo everything to react anyways, so it's a moot point as I'm guessing there's a standard at that point we can use.

I'm happy to change it back, but will have to change the values to reflect the design padding.

rachellerathbone: Generally, best practice for flexible design is to use relative units so that elements adjust as their containers shrink/grow. Using fixed measurements means more code (additional media queries to manually change elements sizing). Media queries fall into the same realm as `!important`: they should be used sparingly and when necessary. They'll never be too much of an issue in our app given the small size of our frontend but in larger applications, more media queries means longer styling download times in the browser. So, even though our app is small, I think we should stick with best practices and allow for scalable stylesheets that are easier to maintain when we need to modify our frontend.

Using ems for padding shouldn't cause an issue with something like an H3 element as the padding should be relative to that container.  If we use a fixed measurement, as you move from a larger to smaller screen, or vice versa, the container will likely expand/shrink but the padding will remain unchanged and change lead to weird styling side-effects/the need to hackily throw in more and more media queries at more and more break points to get everything in the right place.

As far as the design goes - I don't think I've ever seen designers use anything other than px in designs. As a frontend engineer it took me a while to wrap my head around relative measurements (rem and em) and am not sure it's something that makes it's way into the designer realm. I've found it usually falls on the frontend devs to make the conversion.

mboudreau: Understood, reverted it back to `em` values, but reduced it to follow design.

rachellerathbone (approved):

KAbakumovAtl (approved):

PR #597, opened 2021-08-11
ARC-436 Adding better webhook logs for maintenance mode

mboudreau:
We need to do this to be able to get `installationId` for any webhooks that got triggered while we're doing the migration.  Afterwards, we can start a resync for those installations so no data is lost in the process.

### DIFF
--- a/lib/github/middleware.js
+++ b/lib/github/middleware.js
@@ -46,12 +46,6 @@ const isStateChangeOrDeploymentAction = (action) => [
 
 module.exports = function middleware(callback) {
   return withSentry(async (context) => {
-    // Ignore all incoming webhooks if maintenance mode is enabled
-    if (isMaintenanceMode()) {
-      context.log('Maintenance mode ENABLED - Ignoring event');
-      return;
-    }
-
     enhanceOctokit(context.github, context.log);
 
     let webhookEvent = context.name;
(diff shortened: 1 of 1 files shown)


lib/github/middleware.js
```diff
@@ -75,6 +69,14 @@ module.exports = function middleware(callback) {
     const gitHubInstallationId = context.payload.installation.id;
     context.log = context.log.child({ gitHubInstallationId });
 
+    // Ignore all incoming webhooks if maintenance mode is enabled
+    if (isMaintenanceMode()) {
+      context.sentry.setTag('maintenanceMode', true);
+      Sentry.captureMessage('Maintenance mode ENABLED - Ignoring event');
+      context.log({ maintenanceMode: true }, 'Maintenance mode ENABLED - Ignoring event');
```
thombergs, line 76: How do we get the installation ID out of this log event? What's `context.log.child({ gitHubInstallationId })` in line 70 doing? Is it adding the installation to the JSON structure of the log event?

It would probably be safer to just include the installation ID in the log message itself to avoid any issues in extracting the log data from Splunk later.

mboudreau: that's correct, line 70 is making `context.log` always include this specific installationId with all requests as a piece of filterable data.  It's already in splunk and we can test it after deploying it to staging to by getting all the `installationId` of all messages with `maintenanceMode:true`.

thombergs (approved):

lib/github/middleware.js
thombergs: Ok, I just tested the export in Splunk and it returns all fields in the CSV export, so we should be good.

mboudreau: nice :)

PR #599, opened 2021-08-11
Arc 369 add jwt validation for install endpoint

KAbakumovAtl:
This pull request incorporates changes from https://github.com/atlassian/github-for-jira/pull/581 and addresses some feedbacks from that pull request. My apologies for this pull request being too large

Apart from that changes, In this pull request I am adding jwt tokens validation for Install and Uninstall callback. To make it work I had to enable an API change in the connect app descriptor which changes JWT tokens format for install and uninstall callback. (More details about why it was needed here: https://community.developer.atlassian.com/t/action-required-atlassian-connect-installation-lifecycle-security-improvements/49046).

To verify these tokens in new format the substantial refactoring of jwt code and `atlassian-jwt` bump were required.

Now short summary of what was made:
----------------
JWT Refactoring

Now we have the following types of JWT tokens:
* Tokens signed by symmetrical sharedSecret (same secret is used to decrypt and encrypt them)
    * Symmetric tokens can be normal tokens, which require `qsh` being set and follow all normal JWT rules to verify this `qsh` by comparing it with hashed canonical request string. These tokens are used when Jira makes a request to our service (except install and uninstall cases), or when we want to call Jira API.
    * Context tokens, which have `qsh` set to `context-qsh`. Such tokens are used to authenticate calls made from iFrame within Jira to our backend
* Tokens signed asymmetrically. Jira signes this token using its own private key and we verify them by public key, hosted at  `https://connect-install-keys.atlassian.com`. Such tokens we use to verify install callback (which Jira uses to transfer us the shared secret) and uninstall callback. We need to use this way because we don't have shared secret at the point of installation and we can't use symmetric signatures to verify the callback authenticity.

I've made a substantial refactoring of the existing jwt code to support all cases and added tests. You'd probably want to ask, why don't we use existing JWT libraries? So, Atlassian had `atlassian-connect-express` (ACE) framework which handles all authentication for us with all atlassian-specific features. Unfortunately github-for-jira app doesn't use ACE, so we have to either extend the existing jwt handling code, or migrate the app to ACE. To minimise changes we'll just extend the existing code.

----------------
Part 1. 
Fixing JWT Authentication for iFrame endpoints.

Feature flag: `fix-jwt-authentication-for-iframe-endpoints`

For endpoints:
* GET /jira/configuration
* DELETE /jira/configuration
* POST /jira/sync

App only verified that the JWT token we passed was signed with correct sharedSecret. The app didn't verify expiry date or qsh. Also app used to pass jwt token from the iframe query instead of requesting a new one from the frontend. 

To fix the following changes has been made:
* app now requests tokens from Jira in iframe script (These are "Context Tokens") - not feature flagged, hence these tokens will work perfectly with way of authentication 
* app validates these context tokens properly (feature flagged)

-------------------
Part 2.
Adding authentication to installation endpoint

* added `"signed-install": true` to ApiMigrations section of the connect app descriptor. This will make Jira sending asymmetrically signed JWT tokens for install and uninstall endpoints. (Feature flagged)
* for install and uninstall endpoints changed authentication to asymmetric jwt verification. (Feature flagged)

-----------------
What's missing in this PR?

Metrics to make sure the feature flag rollout will go well and we won't break anything in prod. So these metrics will be added in the following PR (this one is too big already)


src/frontend/verify-jira-jwt-middleware.ts
```diff
@@ -0,0 +1,57 @@
...
+		verifySymmetricJwtTokenMiddleware(installation.sharedSecret, tokenType, req, res, next);
+	} else {
+		verifyTokenIsDecodableMiddleware(installation.sharedSecret, req, next);
+	}
+
+};
+
+export const verifyJiraJwtTokenMiddleware = verifyJiraJwtMiddleware(TokenType.normal);
+
+export const verifyJiraContextJwtTokenMiddleware = verifyJiraJwtMiddleware(TokenType.context);
```
KAbakumovAtl, line 57: @mboudreau As you requested, exported different versions of middleware functions without parameters

src/jira/authenticate.ts
```diff
@@ -1,28 +1,41 @@
...
+	if (await booleanFlag(BooleanFlags.USE_JWT_SIGNED_INSTALL_CALLBACKS, false) ) {
+		await authenticateJiraEventMiddleware(req, res, next, true)
+	} else {
+		await authenticateJiraEventMiddleware(req, res, next, false)
 	}
+}
 
-	const { jiraHost, sharedSecret, clientKey } = installation;
 
-	req.addLogFields({
```
KAbakumovAtl, line 18: I've extracted fetching installation and adding it to the local context to a separate middleware function. See `src/jira/extract-installation.ts`

src/jira/util/queryAtlassianConnectPublicKey.ts
```diff
@@ -0,0 +1,23 @@
+import axios from "axios";
+
```
KAbakumovAtl, line 2: Util function which fetches public key by key ID from Atlassian public keys storage

static/js/jira-configuration.js
```diff
@@ -34,39 +34,47 @@ $('.configure-connection-link').click(function (event) {
...
-    type: 'DELETE',
-    url: `/jira/configuration?jwt=${encodeURIComponent(params.get('jwt'))}&xdm_e=${encodeURIComponent(params.get('xdm_e'))}`,
-    data: {
-      installationId: $(event.target).data('installation-id')
-    },
-    success: function (data) {
-      AP.navigator.reload()
-    }
-  })
+	window.AP.context.getToken(function(token){
```
KAbakumovAtl, line 37: So for our frontend calls we'll now use Context tokens instead of reusing the one passed with iframe query

test/unit/frontend/verify-jira-middleware.test.ts
```diff
@@ -29,129 +47,235 @@ describe("#verifyJiraMiddleware", () => {
...
+			log: logger
 		};
+	};
 
-		it("should call next with a valid token and secret", async () => {
-			mocked(Installation.getForHost).mockResolvedValue(installation);
-			const req = buildRequest("test-host", "secret");
-			jwt.decode(req.query.jwt, "secret");
+	//Tests which should pass regardless if USE_JWT_SIGNED_INSTALL_CALLBACKS feature flag is on or off
+	const commonTests = () => {
```
KAbakumovAtl, line 76: Not sure if this follows best javascript practices, but I extracted some tests in a function to repeat them with and without feature flag.

I'd like to know what is the best way to do it, but I'll leave it as is hence we'll remove it anyway after the feature flag cleanup.

test/unit/jira/__snapshots__/connect.test.ts.snap
```diff
@@ -0,0 +1,230 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
KAbakumovAtl, line 1: One of these snapshots is duplicated from the maintenance mode tests. Maybe it would make sense to combine the tests. But I'dnt worry too much hence we'll remove maintenance mode tests on the feature flag celanup.

src/jira/util/queryAtlassianConnectPublicKey.ts
```diff
@@ -0,0 +1,23 @@
...
+const CONNECT_INSTALL_KEYS_CDN_URL =
+	"https://connect-install-keys.atlassian.com";
```
KAbakumovAtl, line 4: We need to be aware of one thing. This is a link for pord. After this change we won't be able to use the app with `jira-dev` 

And why would you need hence we can easily can use prod jira for the app local testing? But I am letting you know in case if anyone needs it

src/jira/connect.ts
```diff
@@ -21,13 +22,17 @@ export default (_: Request, res: Response): void => {
 		}
 	];
 
+	let apiMigrations: any = {gdpr: false}
+
+	if (await booleanFlag(BooleanFlags.USE_JWT_SIGNED_INSTALL_CALLBACKS, false)) {
+		apiMigrations = {gdpr: false, "signed-install": true}
+	}
```
mboudreau, line 29: you could do it as just `const apiMigrations = {gdpr: false, "signed-install": await booleanFlag(BooleanFlags.USE_JWT_SIGNED_INSTALL_CALLBACKS, false)}`

src/jira/connect.ts
```diff
@@ -21,13 +22,17 @@ export default (_: Request, res: Response): void => {
...
+
 	res.status(200)
 		.json({
 			// Will need to be set to `true` once we verify the app will work with
 			// GDPR compliant APIs. Ref: https://github.com/github/ce-extensibility/issues/220
-			apiMigrations: {
-				gdpr: false
-			},
+			apiMigrations: apiMigrations,
```
mboudreau, line 35: don't need to set both the key and value, just `apiMigrations,` will do.

src/jira/util/jwt.ts
```diff
@@ -1,72 +1,146 @@
...
 
-export const hasValidJwt = (secret: string, baseUrl: string, req: Request, res: Response) => {
+function decodeAsymmetricToken(token, publicKey, noVerify) {
+	return decodeAsymmetric(
+		token,
+		publicKey,
+		getAlgorithm(token),
+		noVerify
+	);
+}
```
mboudreau, line 79: Should have typings.

mboudreau (changes requested):
`heroku.json` shouldn't be there.  Delete it.

mboudreau:
So, this PR is massive and I'm still reviewing it locally in my IDE.  I would suggest that we only merge this once the migration is done, or is the migration blocked by this because it doesn't meet our security standard until it's done?  I wouldn't want to introduce something as major as this if we don't need to because if this doesn't work, it'll break everything...

src/jira/connect.ts
KAbakumovAtl: I don't want to change descriptor in any way if the feature flag is off. If I do it it might trigger a version update when we'll be performing migration.

KAbakumovAtl:
Yes, that's the plan. We can merge it shortly after the migration is done

src/frontend/get-github-configuration.ts
```diff
@@ -101,7 +102,7 @@ export default async (req: Request, res: Response, next: NextFunction): Promise<
 			// we can get the jira client Key from the JWT's `iss` property
 			// so we'll decode the JWT here and verify it's the right key before continuing
 			const installation = await Installation.getForHost(req.session.jiraHost);
-			const { iss: clientKey } = JWT.decode(req.session.jwt, installation.sharedSecret);
+			const { iss: clientKey } = decodeSymmetric(req.session.jwt, installation.sharedSecret, getAlgorithm(req.session.jwt));
```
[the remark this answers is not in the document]
KAbakumovAtl: Unfortunately no, algorithm is required parameter for this method

PR #602, opened 2021-08-12
ARC-358 Changed scaling rules

KAbakumovAtl:
Updated scaling rules.

Tested CPU and Network Delay in staging env, it works 👍

### DIFF
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -78,10 +78,10 @@ scaling:
   max: 5
   metrics: &CpuMemScalingRules
     complexScalingRule:
-      EvaluationPeriods: 3
+      EvaluationPeriods: 1
       Threshold:
         Lower: 30
-        Upper: 85
+        Upper: 80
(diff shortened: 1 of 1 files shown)


github-for-jira.sd.yml
```diff
@@ -225,16 +225,16 @@ environmentOverrides:
                       Value: { "Fn::GetAtt": [ "ALBTargetGroup", "TargetGroupFullName" ] }
                   MetricName: TargetResponseTime
                   Namespace: AWS/ApplicationELB
-                Period: 60
-                Stat: Average
+                Period: 300
+                Stat: Maximum
```
KAbakumovAtl, line 229: The only thing which concerns me here is that Maximum might be a bit noisy here. However Micros doesn't allows percentiles and Average wouldn't work well (due to healtcheck, deepcheck). We can go with it anyway and then adjust if it create noise.

mboudreau (approved):

PR #605, opened 2021-08-13
ARC-438 app restructure

rachellerathbone:
First of all, apologies about the size. Lots of things moved around equals lots of pathname changes. Reviewing the restructure will probably be easiest if you pull this branch locally.

Restructured the existing tree in preparation for the rewrite of our frontend. To align with Atlassian guidelines, and adhere to architecture 'guidelines' of Tangerine, I've made the following changes:

- create backend and client directories in /src for a clearer separation
- renamed utils > common: I currently have one common dir in src but, as I begin making additions to the frontend, it may make more sense to have separate common dirs for the backend and client
- renamed static > public
- deleted the frontend dir
- moved all the middleware files into backend/middleware
- moved all testing setup, util, and shared files into common/test-utils
- grouped tests and the files they are testing into folders (structure specified in guidelines)

I haven't deleted the views folder yet and the goal here is to keep the app in a working state and get everything in its new place. That will be done in one of my upcoming PRs.


mboudreau:
Do you think we should hold this PR until the migration is done to reduce the amount of risk involved?

rachellerathbone:
> Do you think we should hold this PR until the migration is done to reduce the amount of risk involved?

I'm merging all of this work into `ARC-107-react-rewrite` to prevent that and to break the work involved into smaller PRs. The idea is that all my feature branches get approved, get merged back into `ARC-107-react-rewrite` while I keep that branch up-to-date with `micros` (which will probably become master while I'm doing this). When all the work is done and all PRs against `ARC-107-react-rewrite` are approved, then we can look at merging that branch in.

src/backend/github/deployment/deployment.test.ts
```diff
@@ -1,13 +1,13 @@
 /* eslint-disable @typescript-eslint/no-var-requires */
```
thombergs, line 1: I'm not sure I like it that the test files now live in the same folders as the production code. This can lead to unwanted dependencies and mixups. I know this is kind of a default in Node apps, but my gut tells me it's better to have the tests in a separate folder structure. 

@mboudreau what's your experience with tests in their own folder structure in Node?

src/backend/github/branch/index.ts
```diff
@@ -1,7 +1,7 @@
-import transformBranch from "../transforms/branch";
+import transformBranch from "../../transforms/branch";
 import { Context } from "probot/lib/context";
 import issueKeyParser from "jira-issue-key-parser";
-import { isEmpty } from "../jira/util/isEmpty";
+import { isEmpty } from "../../../common/isEmpty";
```
thombergs, line 4: I like that the new structure makes each "module" have its own folder with a dedicated `index.ts` file that serves as an API into that module. 

We probably should enforce that more in the future. Currently, we are still referencing functions like `isEmpty()` from anywhere without going through the "API" of the `common` module. Although, that's probably alright for the `common` module, but not for modules with business logic in them.

src/backend/github/middleware/middleware.test.ts
```diff
@@ -1,11 +1,11 @@
 import { logger } from "probot/lib/logger";
 import { mocked } from "ts-jest/utils";
-import { Installation, Subscription } from "../../../src/models";
-import GitHubAPI from "../../../src/config/github-api";
-import middleware from "../../../src/github/middleware";
-import { mockModels } from "../../utils/models";
+import { Installation, Subscription } from "../../models";
+import GitHubAPI from "../../../config/github-api";
+import middleware from ".";
+import { mockModels } from "../../../common/mocks/models";
```
thombergs, line 6: This is why I don't like tests mixed into production code: the `mockModels` are included in the production bundle. I think the build will not include any files ending with `.test.ts` into the bundle, but since this mock file doesn't have the `.test.ts` in the name, it will be included. That increases the bundle size unnecessarily and makes it easier to accidentally use test code in production code.

thombergs (commented):
Tested it and it works. Looks good, except me having an issue with test code next to production code. But I defer to @mboudreau's judgement on this

src/backend/github/deployment/deployment.test.ts
rachellerathbone: I grouped files like this because frontend guidelines stipulate that this is the structure that should be followed. I know these are all backend files but it seems kind of messy to have the backend tests pulling living in /tests while the frontend tests are grouped with the source files. That said, I don't think it's common practice to group test files for node apps with source files 🤷  

When I chatted to some folks in the Fusion team they said all new frontend apps need to be created within the atlassian-frontend monorepo and follow the 'guidelines' (which are really more of a mandate. I told them we can't do this as our backend and frontend are grouped together and, because we have an open source app with such a tiny UI, it's unlikely that they'll split them. That got us out of living in the monorepo but they said we should try and follow the guides as much as possible.

Would love some ideas here.

thombergs (approved):

src/backend/github/deployment/deployment.test.ts
mboudreau: As I mentioned to Tom already, adding the tests next to prod code is a common practice with node apps.  It makes it much easier to update code and tests together without having to go looking in a different directory.  When we're building the typescript, we can just exclude `.test.ts` files so they're not built as part of the final package.  Leave all other commonalities in a `test` folder separate from `src`

src/backend/github/middleware/middleware.test.ts
mboudreau: It's not a front-end app.  The end compilation (you can do it yourself with `npm run release`) is separate files that are included with require.  If a random file gets compiled and not included, it just doesn't matter.  the 'bundle size' will maybe be kilobytes bigger, but the file itself will never be ran anyways.

There are extra checks we can do as well as part of tsc that checks for dangling files (files not required by anyone) which can break the build if not following conventions.  The only issue that could occur is if someone were to accidentally require a test file in a production file (which we can do right now) and there are other checks we can add in place to prevent this from happening.

All in all, I don't think it's an issue and will makes development/testing easier.

PR #618, opened 2021-08-18
Bump storage

thombergs:
Bumping the storage from the default of 5GB to 10GB.


github-for-jira.sd.yml
```diff
@@ -313,6 +313,7 @@ environmentOverrides:
         attributes:
           parameters:
             DBInstanceClass: db.t3.large
+            AllocatedStorage: 10 # 10 GB
```
mboudreau, line 316: Do we need this for all environments or just for production?

thombergs: Just production should be fine.

KAbakumovAtl (approved):

mboudreau (approved):

PR #621, opened 2021-08-18
Prod config

thombergs:
Points the service against the correct prod GitHub app


mboudreau (changes requested):
Bunch of json files not needed

thombergs:
> Bunch of json files not needed

removed them

mboudreau (approved):

PR #623, opened 2021-08-19
ARC-448 webhook logs prod

mboudreau:
Changing APP_URL to point to services domain to test out maintenance mode webhook logs to confirm it's logging installation id.

### DIFF
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -157,7 +157,7 @@ environmentOverrides:
     config:
       environmentVariables:
         APP_URL: https://github-for-jira.dev.services.atlassian.com
-        WEBHOOK_PROXY_URL: https://github-for-jira.dev.services.atlassian.com/github/events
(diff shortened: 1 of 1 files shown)


github-for-jira.sd.yml
```diff
@@ -157,7 +157,7 @@ environmentOverrides:
     config:
       environmentVariables:
         APP_URL: https://github-for-jira.dev.services.atlassian.com
-        WEBHOOK_PROXY_URL: https://github-for-jira.dev.services.atlassian.com/github/events
+        WEBHOOK_PROXY_URL: https://github-for-jira.dev.services.atlassian.com
```
thombergs, line 160: What's this change? Is the `/github/events` now appended in the code itself?

thombergs (approved):

github-for-jira.sd.yml
mboudreau: always was, this was just wrong but the code just got the 'domain' part of that url so it worked.

PR #625, opened 2021-08-20
ARC-451 AUI

rachellerathbone:
Switched a couple of icons out for AUI components, moved the HTML files from /views to /src/client/ui, and cleaned up some styling. Going to work on updating the GitHub config table in the next PR.


src/backend/middleware/index.ts
```diff
@@ -21,8 +21,10 @@ function secureHeaders(router: Router, frontendApp: Express) {
 				"'unsafe-inline'", "'strict-dynamic'", (_: Request, res: Response): string => `'nonce-${res.locals.nonce}'`],
 			// Allow XMLHttpRequest/fetch requests
 			connectSrc: ["'self'", process.env.APP_URL],
-			// Allow <style> tags hosted by ourselves as well as style="" attributes
-			styleSrc: ["'self'", "'unsafe-inline'"],
+			// Allow <style> tags hosted by ourselves and cdnjs, as well as style="" attributes
+			styleSrc: ["'self'", "'unsafe-inline'", "https://cdnjs.cloudflare.com"],
```
KAbakumovAtl, line 25: I wonder what do we need it for? 
I assume it somehow relates to Atlastunnel, I wonder what will be broken if we won't put it here?

src/client/ui/github-configuration.hbs
```diff
@@ -5,6 +5,7 @@
     <meta http-equiv='X-UA-Compatible' content='IE=edge' />
     <meta name='ap-local-base-url' content='{{localBaseUrl}}' />
     <title>{{title}}</title>
+    <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/aui/9.4.0/aui/aui-prototyping.min.css" integrity="DTM1Q+8lU7SzJT+FWr0JFisCSZlwfM0GiAKYy7h1s9vIKa/CIh37s9NuOCqIOgK4tmqrjLK4NuWuIPUQNsikHA==" crossorigin="anonymous" referrerpolicy="no-referrer" />
```
KAbakumovAtl, line 8: I can see now why we had to add above changes to our CSP.

mboudreau: Yeah, it makes it much easier for us.  Actually, I wonder if we can make this non-versioned and just use the latest version at all times?  That way we don't need to ever update AUI.

Actually you can call `GET https://api.cdnjs.com/libraries/aui` and find the latest one.  Maybe we want to call that periodically (every X minutes) to update the value that we use in the handlebars file?

src/client/ui/github-configuration.hbs
```diff
@@ -111,6 +113,7 @@
 
     </section>
     <script src='/public/js/github-configuration.js' nonce='{{nonce}}'></script>
+    <script src="https://cdnjs.cloudflare.com/ajax/libs/aui/9.4.0/aui/aui-prototyping.min.js" integrity="sha512-iUshpSgJ26ryfeS5STpopGIsONtcP9nlIgLU24C8jwGPjsSTO5vqepMFU/KWc3fTIbNhbSTS/arJAGSnMPqQeA==" crossorigin="anonymous" referrerpolicy="no-referrer" nonce='{{nonce}}'></script>
```
mboudreau, line 116: Do we need the JS for anything?  Are we using the AUI components?

rachellerathbone: It's for the loader. The animation for that is handled by JS in AUI, not by the CSS.

mboudreau: :| Well that sucks.  Maybe something to bring up with the AUI team?  Seems like something that's really simple to accomplish in css...

src/client/ui/github-configuration.hbs
KAbakumovAtl: It introduces a risk of getting affected by breaking changes (or bugs)

Btw, is there any better way of versioning it? (Maybe importing it via npm and hosting it in our service?) Can you suggest what's the best practice here?

src/client/ui/github-configuration.hbs
rachellerathbone: https://github.com/atlassian/github-for-jira/pull/625#discussion_r692646954

Do we want to pull in the latest version? Wouldn't we want to have a little more control over specifying the version and manually update it like we would if it were an npm package?

rachellerathbone: I was doing it in CSS. I can revert back to what I had but I'll need the same script on the jira-configuration page when I go to do the dropdown for the table (from what I looked at all animations are JS based).

rachellerathbone: This is what I had before switching to the AUI component:
```
.getConfiguration__loader {
    animation: spin 2s linear infinite;
    border: 2px solid #f3f3f3;
    border-radius: 50%;
    border-top: 2px solid #42526e;
    height: 1.8em;
    width: 1.8em;
}
```

rachellerathbone: It is weird that such simple animations **aren't** handled by css in AUI.

src/backend/middleware/index.ts
```diff
@@ -21,8 +21,10 @@ function secureHeaders(router: Router, frontendApp: Express) {
...
+			// Allow fonts hosted by ourselves and cdnjs
+			fontSrc: ["'self'", "https://cdnjs.cloudflare.com"],
```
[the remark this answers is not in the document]
mboudreau: I'm not sure.  I added that super quickly just to get it working.  I think there might be a better way of doing this though.

src/client/ui/github-configuration.hbs
mboudreau: It's css, I don't think it'll cause issues and even if it did, I'd bet you that it would take less time fixing it than it would take to constantly update the version.  I'd rather not have to think about it if I don't need to.

src/client/ui/github-configuration.hbs
mboudreau: I'm asking the aui channel about this.  I think it's because it's injecting an SVG into it depending on the size, which is kinda silly.  Would make for a great replacement with web components :P

src/client/ui/github-configuration.hbs
KAbakumovAtl: This library has also scripts, there is a <scrpit> tag below

src/backend/middleware/index.ts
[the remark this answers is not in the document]
KAbakumovAtl: Anyway it is not a big deal. We use nonce so I believe we well protected already

KAbakumovAtl (approved):

PR #630, opened 2021-08-22
ARC-464 jira error logging

thombergs:


mboudreau (dismissed):

src/jira/client/axios.ts
```diff
@@ -79,14 +79,15 @@ function getErrorMiddleware(logger) {
...
-				const { status, statusText } = error.response || {};
+				const { status } = error.response || {};
 
-				const errorMessage = status in JiraErrorCodes ? `Error calling Jira API. ${JiraErrorCodes[status]}` :
-					`Error calling Jira API. Response Code: ${status} ${statusText}`;
+				const detailMessage = status in JiraErrorCodes
+					? JiraErrorCodes[status]
+					: JSON.stringify(error?.response?.data);
 
-				logger.warn(error, { message: errorMessage, response_data: error.response.data });
```
thombergs, line 87: removed the WARN, since the `JiraClientError` will be logged on ERROR anyways.

src/jira/client/axios.ts
```diff
@@ -79,14 +79,15 @@ function getErrorMiddleware(logger) {
...
+				const errorMessage = `HTTP Status ${status} - ${detailMessage}`;
```
thombergs, line 88: passing the detailed error message into the `JiraClientError`, so it gets logged. This will result in logs like this:

```
02:57:09.211Z ERROR github-for-jira: Failed to process push (logger=transforms.push)
app_1       |   JiraClientError: Jira client error: HTTP Status 404 - {"message":"null for uri: https://thombergs.atlassian.net/rest/devinfo/0.10/bulk123","status-code":404}
```

mboudreau (approved):

src/jira/client/axios.ts
KAbakumovAtl: We might also want to capture Jira response in case if we are getting 400. So we'll know why the request was rejected

PR #633, opened 2021-08-22
Fixing build/test/deploy workflows

mboudreau:
Removing old heroku deployment and improving build/test/deploy scripts.


.github/workflows/deploy.yml
```diff
@@ -1,39 +1,42 @@
...
-              password $heroku_api_key
-          EOF
-      - name: Deploy to Heroku.
-        env:
-          heroku_app_name: ${{secrets.HEROKU_APP_NAME}}
-        run: |
-          heroku container:login
-          heroku container:push --app=$heroku_app_name web worker --recursive
-          heroku container:release --app=$heroku_app_name web worker
+    - name: Deploy to Micros
```
KAbakumovAtl, line 17: We can remove it hence it doesn't work anyway

thombergs (changes requested):
we should remove the deployment step as we can't use it anyways (because of Bitbucket's allowlisting.

mboudreau:
@thombergs there's no way to change that?  :(

KAbakumovAtl (approved):

thombergs (approved):

PR #634, opened 2021-08-23
Arc 464 jira error logging

thombergs:
proofing the error logging some more: truncating after 200 characters because it seems that Jira returns a whole HTML document in some cases. adding the detail message to HTTP 400s


src/jira/client/axios.ts
```diff
@@ -55,15 +55,13 @@ function getAuthMiddleware(secret: string) {
...
-	403: "The JWT token used does not correspond to an app that defines the jiraDevelopmentTool module, or the app does not define the 'WRITE' scope",
-	413: "Data is too large. Submit fewer devinfo entities in each payload.",
-	429: "API rate limit has been exceeded."
+export const JiraErrorMessages = {
+	400: (detailMessage) => `HTTP 400 - Request had incorrect format. Details: ${detailMessage}`,
+	401: () => "HTTP 401 - Missing a JWT token, or token is invalid.",
+	403: () => "HTTP 403 - The JWT token used does not correspond to an app that defines the jiraDevelopmentTool module, or the app does not define the 'WRITE' scope",
+	413: () => "HTTP 413 - Data is too large. Submit fewer devinfo entities in each payload.",
+	429: () => "HTTP 429 - API rate limit has been exceeded.",
+	"default": (status, detailMessage) => `HTTP ${status} - ${detailMessage}`
```
thombergs, line 64: I collected ALL errors in this single place now to make it more obvious what's happening

src/jira/client/axios.ts
```diff
@@ -55,15 +55,13 @@ function getAuthMiddleware(secret: string) {
...
- * Mapping of Jira Dev Error Codes to nice strings.
- */
-const JiraErrorCodes = {
-	400: "Request had incorrect format.",
-	401: "Missing a JWT token, or token is invalid.",
-	403: "The JWT token used does not correspond to an app that defines the jiraDevelopmentTool module, or the app does not define the 'WRITE' scope",
-	413: "Data is too large. Submit fewer devinfo entities in each payload.",
-	429: "API rate limit has been exceeded."
+export const JiraErrorMessages = {
+	400: (detailMessage) => `HTTP 400 - Request had incorrect format. Details: ${detailMessage}`,
```
KAbakumovAtl, line 59: Might be better to unify all these functions parameters, so all of them accept 2 arguments, so will be more difficult to make a mistake

KAbakumovAtl (approved):

src/jira/client/axios.ts
thombergs: Yeah, I'd like to do that, but our TypeScript linter complains when a function argument isn't used in the function body...

mboudreau (approved):

PR #635, opened 2021-08-23
adding better logging for express responses that aren't successful

mboudreau:


src/frontend/app.ts
```diff
@@ -97,6 +97,17 @@ export default (octokitApp: App): Express => {
 
 	app.use(logMiddleware);
 
+	// Catch non successful responses
+	app.use((req, res, next) => {
+		res.once("finish", () => {
+			// if status is under 200 or higher/equal to 400, but not 503 in maintenance mode, log it to figure out issues
+			if ((res.statusCode < 200 || res.statusCode >= 400) && !(res.statusCode === 503 && booleanFlag(BooleanFlags.MAINTENANCE_MODE, false))) {
+				req.log.warn({ res, req }, `Returning HTTP response of '${res.statusCode}' for path '${req.path}'`);
```
thombergs, line 105: Maybe not in this PR, but later: we should probably expose this as a metric with the path and status as tags, so we can create a chart on top of this.

thombergs (approved):

src/frontend/app.ts
```diff
@@ -97,6 +97,17 @@ export default (octokitApp: App): Express => {
 
 	app.use(logMiddleware);
 
+	// Catch non successful responses
+	app.use((req, res, next) => {
+		res.once("finish", () => {
+			// if status is under 200 or higher/equal to 400, but not 503 in maintenance mode, log it to figure out issues
+			if ((res.statusCode < 200 || res.statusCode >= 400) && !(res.statusCode === 503 && booleanFlag(BooleanFlags.MAINTENANCE_MODE, false))) {
```
KAbakumovAtl, line 104: Do we still need maintenance mode?

src/jira/client/index.ts
```diff
@@ -274,9 +274,9 @@ export default async (
  */
 const batchedBulkUpdate = async (
 	data,
-	options,
 	instance: AxiosInstance,
-	installationId: number
+	installationId: number,
+	options?: { preventTransitions: boolean }
```
KAbakumovAtl, line 279: I wonder is it related to logging or you are also trying to fix passing `preventTransitions` to Jira in this PR?

src/frontend/app.ts
mboudreau: It's to ignore maintenance mode 503s

src/jira/client/index.ts
mboudreau: Not related to logging, was just fixing a type

src/frontend/app.ts
KAbakumovAtl: Yes, but would we need it at all? We've already performed the migration, so eventually we'll cleanup the flag.

KAbakumovAtl (approved):

PR #637, opened 2021-08-23
Add more logging to axios for 404s

mboudreau:
We weren't logging the request side of thing, just that we're getting 404s, but don't know what it's trying to hit.  This will help us identify which hosts are failing and if the hosts are actually offline or not.


src/jira/client/axios.ts
```diff
@@ -55,51 +54,49 @@ function getAuthMiddleware(secret: string) {
...
-	"default": (status, detailMessage) => `HTTP ${status} - ${detailMessage}`
+export const getJiraErrorMessages = (status:number, message?:string) => {
+	switch(status) {
+		case 400: return `HTTP 400 - Request had incorrect format. Details: ${message}`;
+		case 401: return "HTTP 401 - Missing a JWT token, or token is invalid.";
+		case 403: return "HTTP 403 - The JWT token used does not correspond to an app that defines the jiraDevelopmentTool module, or the app does not define the 'WRITE' scope";
+		case 413: return "HTTP 413 - Data is too large. Submit fewer devinfo entities in each payload.";
+		case 429: return "HTTP 429 - API rate limit has been exceeded.";
+		default: return `HTTP ${status} - ${message}`;
+	}
```
KAbakumovAtl, line 65: nice

src/jira/client/axios.ts
```diff
@@ -55,51 +54,49 @@ function getAuthMiddleware(secret: string) {
...
-					: JiraErrorMessages["default"](status, detailMessage)
-
-				return Promise.reject(new JiraClientError(error, errorMessage));
-			} else {
-				return Promise.reject(error);
+				error.message = `Jira Client Error: ${errorMessage}`;
 			}
-		}
-	);
+			logger.error(error);
```
KAbakumovAtl, line 89: 1. Would It cause double-logging if we log this error on the top level as well?
2. Do we log it in case of 401s and 404s and other non-erroneous conditions? 

 [ ] Make sure no double logging occur here, change to warn if it includes 401 and 404s

KAbakumovAtl (changes requested):

thombergs (approved):

src/jira/client/axios.ts
mboudreau: What do you mean at the top level?  Because of the promise rejection from axios?  The only reason it used to capture before is because we used to just throw an error.  This is a bit cleaner. 

But you're right about 400s.  I should check to see what's the status and log according to that.

KAbakumovAtl (approved):

PR #639, opened 2021-08-24
ARC-460 Removed lp 5xx alert, adjusted latency alerts

KAbakumovAtl:
Turned out that we didn't turn off low priority 5xx alerts, which created a lot of noise.

Also forgot to adjust latencies alert threshold for prod. 

Doing it now


github-for-jira.sd.yml
```diff
@@ -176,6 +176,18 @@ environmentOverrides:
...
+        LatencyHigh: null
+        HealthyHostCount: null
+        UnHealthyHostCount: null
+        LowSeverityAlarmWhenTooManyELB5xxErrors: null
+        LowSeverityAlarmWhenTooManyBackend5xxErrors: null
+        WebServerAlarmWhenLowCPUCredits: null
+        WebServerDiskSpaceUtilizationAlarmHigh: null
+        WebServerInstanceVolumeSpaceUtilizationAlarmHigh: null
+        WebServerMemoryAlarmHigh: null
+        WebServerServiceRespawnAlarm: null
```
KAbakumovAtl, line 190: Disable all alarms for ddev

github-for-jira.sd.yml
```diff
@@ -298,12 +314,15 @@ environmentOverrides:
...
+          Threshold: 5
           Priority: Low
           EvaluationPeriods: 3
           Period: 60
           ComparisonOperator: GreaterThanThreshold
-          Statistic: Maximum
+          Statistic: Average
+        LowSeverityAlarmWhenTooManyELB5xxErrors: null
+        LowSeverityAlarmWhenTooManyBackend5xxErrors: null
+        HighSeverityAlarmWhenTooManyELB5xxErrors: null
```
KAbakumovAtl, line 325: We didn't get any ELB alerts surprisingly, but these alerts are also default micros alerts and works for both ALB and ELB (according to the docs). Se wouldn't be wrong to disable them as well.

mboudreau (approved):

PR #640, opened 2021-08-24
Add more logging to octokit calls

mboudreau:


src/config/enhance-octokit.ts
```diff
@@ -1,26 +1,31 @@
...
-				responseStatus = error.responseCode;
-			}
-
+		} catch (err) {
+			error = err;
+			responseStatus = error?.responseCode;
 			throw error;
 		} finally {
+			if(error || responseStatus < 200 || responseStatus >= 400) {
+				logger.warn({request, response, error}, `Octokit error: '${responseStatus}' for request '${options.method} ${options.url}'`);
```
KAbakumovAtl, line 27: I wonder if we have a logger in request.context at this point? So we will have requestId (or job id) for tracing?

src/config/enhance-octokit.ts
```diff
@@ -40,8 +45,7 @@ const instrumentRequests = (octokit: GitHubAPI) => {
  * (Because Probot instantiates the Octokit client for us, we can't use plugins.)
  */
 export default (octokit: GitHubAPI): GitHubAPI => {
-
-	OctokitError.wrapRequestErrors(octokit);
+	// OctokitError.wrapRequestErrors(octokit);
```
KAbakumovAtl, line 48: Should we delete the line if we don't need it?  (Otherwise comment needed explaining why we commented it)

src/util/log-express-error-response.ts
```diff
@@ -0,0 +1,14 @@
+import { NextFunction, Request, Response } from "express";
+import { booleanFlag, BooleanFlags } from "../config/feature-flags";
+import Logger from "bunyan";
+
+export const logExpressErrorResponse = (logger?: Logger) =>
+	(req: Request, res: Response, next: NextFunction) => {
+		res.once("finish", async () => {
+			// if status is under 200 or higher/equal to 400, but not 503 in maintenance mode, log it to figure out issues
+			if ((res.statusCode < 200 || res.statusCode >= 400) && !(res.statusCode === 503 && await booleanFlag(BooleanFlags.MAINTENANCE_MODE, false))) {
+				(logger || req.log).warn({ res, req }, `Returning HTTP response of '${res.statusCode}' for path '${req.path}'`);
```
KAbakumovAtl, line 10: Maybe it should be opposite `(req.log || logger)` so if there is no logger in the request we use the default one

src/github/index.ts
```diff
@@ -11,9 +11,12 @@ import statsd from "../config/statsd";
 import { getLogger } from "../config/logger";
 import { metricWebhooks } from "../config/metric-names";
 import { Application } from "probot";
+import { logExpressErrorResponse } from "../util/log-express-error-response";
 
-export default (robot:Application) => {
+export default (robot: Application) => {
 	const logger = getLogger("github.webhooks");
+	// Catch non successful responses
+	robot.router.use(logExpressErrorResponse(logger));
```
KAbakumovAtl, line 19: I wonder if we can use context.log, which has context id, added to the context on the line 25?

src/config/enhance-octokit.ts
mboudreau: this is for the github api, not express.  There is no `req.context` for it.

src/util/log-express-error-response.ts
mboudreau: logger is being injected in the function.  we should use the one made explicit first, not the default.

src/config/enhance-octokit.ts
KAbakumovAtl: I mean, it would be great if we could pass it to the request where we make a call. And here use (request.log || logger).warn

src/github/index.ts
mboudreau: line 25 is for the webhook handler, this is for the problem express router and we're logging the request which would have the context id I believe.

src/config/enhance-octokit.ts
KAbakumovAtl: However if now we need a quick win to see errors then we can take care of it later.

src/config/enhance-octokit.ts
mboudreau: done.

KAbakumovAtl (approved):

PR #645, opened 2021-08-24
HOT-96452: removing sentry event that we don't need 

thombergs:
Sentry should already get an event through the withSentry() call

### DIFF
--- a/src/github/middleware.ts
+++ b/src/github/middleware.ts
@@ -30,7 +30,7 @@ const withSentry = function (callback) {
 		try {
 			await callback(context);
 		} catch (err) {
-			logger.error(err, "Webhook Error");
+			logger.error({...err}, "Webhook Error");
 			context.sentry.captureException(err);
 			throw err;
 		}
--- a/src/github/middleware.ts
+++ b/src/github/middleware.ts
@@ -9,12 +9,16 @@ import getJiraUtil from "../jira/util";
 import enhanceOctokit from "../config/enhance-octokit";
 import { Context } from "probot/lib/context";
 import { booleanFlag, BooleanFlags } from "../config/feature-flags";
+import { getLogger } from "../config/logger";
 
 // Returns an async function that reports errors errors to Sentry.
 // This works similar to Sentry.withScope but works in an async context.
 // A new Sentry hub is assigned to context.sentry and can be used later to add context to the error message.
 const withSentry = function (callback) {
 	return async (context) => {
+
+		const logger = getLogger("webhook.middleware");
+
 		context.sentry = new Sentry.Hub(Sentry.getCurrentHub().getClient());
 		context.sentry.configureScope((scope) =>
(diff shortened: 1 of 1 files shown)


src/github/middleware.ts
```diff
@@ -151,15 +155,7 @@ export default (
 			}
 			const util = getJiraUtil(jiraClient);
 
-			try {
-				context.sentry.captureMessage(
```
thombergs, line 155: This event is just noisy.

src/github/middleware.ts
```diff
@@ -151,15 +155,7 @@ export default (
...
-					`Middleware: webhook handler - context: ${context}, jiraClient: ${jiraClient}, util: ${util}`
-				);
-				return await callback(context, jiraClient, util);
-			} catch (err) {
-				context.sentry.captureException(err);
```
thombergs, line 160: This event is already covered by calling `withSentry()` at the start of this function.

src/github/middleware.ts
```diff
@@ -151,15 +155,7 @@ export default (
...
-				throw err;
-			}
+			return callback(context, jiraClient, util);
```
mboudreau, line 158: Have you tested this? Do we need the await at all?

thombergs: the linter told me to remove the `await`, because it's not needed in return statements. Which, thinking about it, is not a good linting rule, because it's dangerous for refactorings. I'll re-add the `await`

src/github/middleware.ts
```diff
@@ -9,12 +9,16 @@ import getJiraUtil from "../jira/util";
...
 import { booleanFlag, BooleanFlags } from "../config/feature-flags";
+import { getLogger } from "../config/logger";
 
 // Returns an async function that reports errors errors to Sentry.
 // This works similar to Sentry.withScope but works in an async context.
 // A new Sentry hub is assigned to context.sentry and can be used later to add context to the error message.
 const withSentry = function (callback) {
 	return async (context) => {
+
+		const logger = getLogger("webhook.middleware");
```
mboudreau, line 20: Make this global to the file so it doesn't have to keep creating the logger on every call

src/github/middleware.ts
```diff
@@ -26,6 +30,7 @@ const withSentry = function (callback) {
 		try {
 			await callback(context);
 		} catch (err) {
+			logger.error(err, "Webhook Error");
```
mboudreau, line 33: Should do `{...err}` to explode out all properties so Bunyan logs everything

mboudreau (changes requested):

src/github/middleware.ts
thombergs: done. All other places where Sentry is used we already log a message.

src/github/middleware.ts
thombergs: done

mboudreau (approved):

PR #647, opened 2021-08-24
Remove error event logging

KAbakumovAtl:


src/main.ts
```diff
@@ -35,7 +33,6 @@ async function start() {
 	// We are always behind a proxy, but we want the source IP
 	probot.server.set("trust proxy", true);
 	probot.load(App);
-	probot.webhook.on("error", (err: Error) => logger.error({ ...err }, "Webhook Error"));
```
KAbakumovAtl, line 38: This lead to event payload being logged.

Without this line probot logs an error anyway

thombergs (approved):

PR #648, opened 2021-08-25
ARC-483 fixing type error

mboudreau:
Needed to use more optional chaining to make sure things don't break.

### DIFF
--- a/src/github/branch.ts
+++ b/src/github/branch.ts
@@ -29,7 +29,7 @@ export const deleteBranch = async (context, jiraClient): Promise<void> => {
 	}
 
 	await jiraClient.devinfo.branch.delete(
-		context.payload.repository.id,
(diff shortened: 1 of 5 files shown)


thombergs (approved):

src/github/branch.ts
```diff
@@ -29,7 +29,7 @@ export const deleteBranch = async (context, jiraClient): Promise<void> => {
 	}
 
 	await jiraClient.devinfo.branch.delete(
-		context.payload.repository.id,
+		context.payload.repository?.id,
```
KAbakumovAtl, line 32: I don't think it will work if we don't have repo the branch wouldn't be deleted. (Jira probably return 404). Maybe we can just discard the evnet and log an error in such case?

PR #650, opened 2021-08-25
Added more logs for webhooks processing

KAbakumovAtl:
Adding a bit more logging to the webhooks path so hopefully it will help us to get some insights

### DIFF
--- a/src/github/middleware.ts
+++ b/src/github/middleware.ts
@@ -31,7 +31,7 @@ const withSentry = function (callback) {
 		try {
 			await callback(context);
 		} catch (err) {
-			logger.error({...err}, "Webhook Error");
+			logger.error(err, "Error while processing webhook");
 			context.sentry.captureException(err);
 			throw err;
 		}
@@ -105,7 +105,8 @@ export default (
 		const subscriptions = await Subscription.getAllForInstallation(
 			gitHubInstallationId
 		);
-		if (!subscriptions.length) {
+		const jiraSubscriptionsCount = subscriptions.length
+		if (!jiraSubscriptionsCount) {
 			context.log(
 				{ noop: "no_subscriptions" },
 				"Halting further execution since no subscriptions were found"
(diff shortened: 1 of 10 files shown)


src/github/middleware.ts
```diff
@@ -31,7 +31,7 @@ const withSentry = function (callback) {
 		try {
 			await callback(context);
 		} catch (err) {
-			logger.error({...err}, "Webhook Error");
+			logger.error(err, "Error while processing webhook");
```
KAbakumovAtl, line 34: Here we need to pass the error itself to have a proper stack grace in the log

src/main.ts
```diff
@@ -35,7 +35,9 @@ async function start() {
 	// We are always behind a proxy, but we want the source IP
 	probot.server.set("trust proxy", true);
 	probot.load(App);
-	probot.webhook.on("error", (err: Error) => logger.error({ ...err }, "Webhook Error"));
+	probot.webhook.on("error", (err: Error) => {
+		logger.error({...err, err: JSON.stringify(err)}, "Webhook Error")
```
KAbakumovAtl, line 39: Not sure what magic happens in Probot, but it adds an event to the exception. If we log it like {...err} we'll see an event only and loose the stack trace. If we do JSON.stringify(err) we'll get an error message. 

So added it to have both for now

thombergs (approved):

src/main.ts
mboudreau: it's not probot, it's bunyan.  It only picks a few properties out of Error to display.

src/github/middleware.ts
mboudreau: by just giving it the error, it won't log every property, only stacktrace, message and type I think.  You can do `{...error, err: error}` which should give you best of both worlds.

KAbakumovAtl: No it doesn't work like that, I tried. You need to pass the exact exception as a first argument otherwise no stack trace

PR #651, opened 2021-08-25
trying to fix push webhook by limiting calls to github API

mboudreau:
Reducing APi impact of push and branch webhook where possible.

### DIFF
--- a/src/github/pull-request.ts
+++ b/src/github/pull-request.ts
@@ -5,9 +5,6 @@ import { isEmpty } from "../jira/util/isEmpty";
 import { Context } from "probot/lib/context";
 
 export default async (context: Context, jiraClient, util): Promise<void> => {
-	const author = await context.github.users.getByUsername({
-		username: context.payload.pull_request.user.login
-	});
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any
 	let reviews: any = {};
 	try {
@@ -28,7 +25,7 @@ export default async (context: Context, jiraClient, util): Promise<void> => {
 
 	const jiraPayload = transformPullRequest(
 		context.payload,
-		author.data,
+		context.payload.pull_request?.user,
 		reviews.data
 	);
 	const { pull_request: pullRequest } = context.payload;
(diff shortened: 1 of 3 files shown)


src/github/pull-request.ts
```diff
@@ -28,7 +25,7 @@ export default async (context: Context, jiraClient, util): Promise<void> => {
 
 	const jiraPayload = transformPullRequest(
 		context.payload,
-		author.data,
+		context.payload.pull_request?.user,
```
KAbakumovAtl, line 28: We might loose information like avater_url and url for the author

src/transforms/pull-request.ts
```diff
@@ -19,20 +19,21 @@ function mapReviews(reviews) {
 	// The reduce function goes through all the reviews and creates an array of unique users (so users' avatars won't be duplicated on the dev panel in Jira) and it considers 'APPROVED' as the main approval status for that user.
 	return sortedReviews.reduce((acc, review) => {
 		// Adds user to the usernames object if user is not yet added, then it adds that unique user to the accumulator.
-		if (!usernames[review.user.login]) {
-			usernames[review.user.login] = {
-				name: review.user.login,
+		const user = review?.user;
+		if (!usernames[user?.login]) {
```
KAbakumovAtl, line 23: if user?.login is undefined wouldn't we get an error when trying to acccess a map by undefined key?

KAbakumovAtl: I guess review must have user. According to the previous code. So I wonder why do we use `?.` here?

KAbakumovAtl:
Changes here are in the pullrequest code. So I wonder how it help push and branch webhooks? 

Am I correctly undrestand that we are trying here to reduce GitHub API load by making less calls to users endpoints from Pullrequest webhooks so the branch and push wont be affected by rate limiting?

thombergs (approved):

src/github/pull-request.ts
mboudreau: nope, it's already part of the `user` object on the payload.  We were doing an extra call for no reason.

KAbakumovAtl (approved):

src/transforms/pull-request.ts
mboudreau: you can do `someArray[undefined]`.  It'll returned `undefined` as the key doesn't exist.  I'm adding a lot of optional chaining because we don't have types to protect us and a lot of github docs are wrong as they're just missing parts of the payload depending of the situation.  I was just being cautious

mboudreau:
I won't help the push webhook since haven't done work on that.

PR #652, opened 2021-08-25
Fix logs again

KAbakumovAtl:


src/github/middleware.ts
```diff
@@ -31,7 +27,7 @@ const withSentry = function (callback) {
 		try {
 			await callback(context);
 		} catch (err) {
-			logger.error(err, "Error while processing webhook");
+			context.log.error(err, "Error while processing webhook");
```
KAbakumovAtl, line 30: So this error log turned out to be useless hence it was impossible to connect it with any webhook. 

By using logger from the context we'll also pass Installation ID and webhook ID so we'll have better Idea where it happens

src/main.ts
```diff
@@ -36,7 +36,7 @@ async function start() {
 	probot.server.set("trust proxy", true);
 	probot.load(App);
 	probot.webhook.on("error", (err: Error) => {
-		logger.error({...err, err: JSON.stringify(err)}, "Webhook Error")
+		logger.error(err, `Webhook Error: ${JSON.stringify(err)}`)
```
KAbakumovAtl, line 39: Finally found a magic formula. This will log stack trace and all elements of `err`

thombergs (approved):

PR #653, opened 2021-08-25
remove global edge vanity DNS entry, because it unnecessarily routes …

thombergs:
…traffic back through global edge instead of directly into our service

### DIFF
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -66,12 +66,6 @@ resources:
         # Name of the RDS resource from above
         resource: rds
 
-  - type: globaledge
-    name: ingress
-    attributes:
-      ip_whitelist:
-        - public
-
 scaling:
   instance: t2.small
   min: 1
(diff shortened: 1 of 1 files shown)


github-for-jira.sd.yml
```diff
@@ -276,9 +276,6 @@ environmentOverrides:
           default_vanity_dns: false
           domain:
             - github.stg.atlassian.com
-          upstream_address:
-            - address: github-for-jira.stg.services.atlassian.com
-          rewrite: github-for-jira.stg.services.atlassian.com
```
thombergs, line 281: this is not needed anymore, because by default, global edge will route directly to the service that owns this descriptor

github-for-jira.sd.yml
```diff
@@ -66,12 +66,6 @@ resources:
         # Name of the RDS resource from above
         resource: rds
 
-  - type: globaledge
-    name: ingress
-    attributes:
-      ip_whitelist:
-        - public
-
```
thombergs, line 74: moved to `ddev` overrides

mboudreau (approved):

PR #663, opened 2021-08-27
Adding docker build to github actions

mboudreau:
Cleaning up some dockerfile that aren't needed anymore and also centralizing actions based on trigger.


.github/workflows/on-push.yml
```diff
@@ -0,0 +1,70 @@
...
+        with:
+          path: ./node_modules
+          key: ${{ runner.os }}-${{hashFiles('**/package-lock.json')}}-node_modules
+      - run: npm install
+      - run: npm run db:test
+      - name: Run tests
+        env:
+          NODE_ENV: test
+          REDIS_HOST: localhost
+          REDIS_PORT: ${{ job.services.redis.ports[6379] }}
```
thombergs, line 54: What does this syntax mean? If we use the port 6379 as an index, is there any difference to just hard coding the port into the `REDIS_PORT` env var?

thombergs (approved):

.github/workflows/on-push.yml
mboudreau: It's how [github recommends to do it based on their example](https://github.com/actions/example-services/blob/main/.github/workflows/redis-service.yml).  This syntax is just to get the associated public port based on the internal port.  In case the public port changes (conflicts?), it'll make sure that it won't affect the test.

PR #667, opened 2021-08-29
ARC-470 GitHub config pg

rachellerathbone:
**UPDATE:** there are duplicate css and hbs files, along with some duplicate JS to account for the old and new versions of this page. This will all be removed once we are sure there are no issues with these changes and the FF can be removed.

Also to note: most of the screenshots are from when I originally opened this PR and some of the wording has changed since (e.g. syncing is now backfilling)

**Changes in this PR include**:

- Update to the overall UI of the GH config page (old vs new).

Screen Shot 2021-09-01 at 12 51 07 pm

- Update sync status modal - which previously had a big white gap at the bottom (old vs new)

Screen Shot 2021-09-01 at 12 53 24 pm

- Removal of sync options 'Normal' and 'Full' and removal of retry modal that explained what these 2 types were. 'Resync' can be found in the dropdown and will run a full sync.

- Addition of a retry icon next to and failed sync with 'Retry' label that displays on hover.

Screen Shot 2021-09-01 at 12 57 56 pm

Screen Shot 2021-09-01 at 12 58 04 pm

- Update UI when there is an error (failed connection)

Ugly error we used to render
Screen Shot 2021-09-01 at 12 59 46 pm

New error:

Screen Shot 2021-10-01 at 11 10 52 am


thombergs:
I wonder if this PR is the chance to rename "sync" to "backfilling"? This would make it more clear that this is only for historical data.

Also: in the first screenshot in the column "Repository access" it says "All repos (3/101)". Should that be "Selected repos (3/101)"?

rachellerathbone:
> I wonder if this PR is the chance to rename "sync" to "backfilling"? This would make it more clear that this is only for historical data.

Have we confirmed as a team that this is the terminology we want to adopt? 

> Also: in the first screenshot in the column "Repository access" it says "All repos (3/101)". Should that be "Selected repos (3/101)"?

All repos means I have given access to all repos (which at the time of this screenshot was 101). At the time of the screenshot 3/101 repos were successfully synced.

src/frontend/index.ts
```diff
@@ -15,7 +15,7 @@ function secureHeaders(router: Router, frontendApp: Express) {
 	router.use(helmet.contentSecurityPolicy({
 		useDefaults: true,
 		directives: {
-			defaultSrc: ["'none'"],
+			defaultSrc: ["'self'"],
```
rachellerathbone, line 18: AUI image components don't render when this is set to "none". This is not prohibited by the [CSP guidelines](https://hello.atlassian.net/wiki/spaces/PMP/pages/139161777/Guideline+-+Content+Security+Policy) but would love suggesstions if anyone has a better alterative. Error in the console when this is set to "none" is `Refused to load ... because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled.`

static/css/global-OLD.css
```diff
@@ -0,0 +1,85 @@
+body {
```
rachellerathbone, line 1: If I had FF flags in mind before I made this PR I would've structured the changes in such a way that we don't need 2 copies of this file. Working backwards it's easier & less prone to issues if we just have 2 copies.

static/css/media-queries.css
```diff
@@ -0,0 +1,28 @@
...
+      padding: 2em;
+    }
+
+    .jiraConfiguration__table__cell__settings__ellipsis {
+      right: 0.2em !important;
+    }
+  }
+
+  @media (max-width: 540px) {
+    .getConfiguration__tableHeader {
```
rachellerathbone, line 19: There is a duplicate of this styling in global-OLD.css due to adding the FF after this work was done. Easier to just leave the hbs files pointing at the resources they need than to move styles around.

static/js/jira-configuration.js
```diff
@@ -74,15 +76,44 @@ $('.sync-connection-link').click(function (event) {
...
+			},
+			error: function (error) {
+				console.log(error);
+			},
+		});
+	});
+});
+
+/* ***************************** */
+/* To be removed after FF tested */
```
rachellerathbone, line 109: More code I would've handled differently if I'd planned to (or knew how to) use FF flags before I did this work. Will be cleaned up after these changes are deployed and thoroughly tested.

static/css/media-queries.css
```diff
@@ -0,0 +1,28 @@
+/*  We can't use standard breakpoints as the left panel navigation in Jira hijacks some of the screen */
+@media (max-width: 1530px) {
+    .jiraConfiguration__table__cell__settings__ellipsis {
+      right: 0;
+    }
+  }
```
mboudreau, line 6: Can't say I'm a big fan of putting the media queries in a separate file.  I much prefer placing it in the same file where the original styles are.  Seems like more work when putting them in a different file.

mboudreau (dismissed):
looks good

static/css/media-queries.css
rachellerathbone: Ok, I'll look at moving these styles back out into the other css files in the PR I have open for connect an org pg.

mboudreau (approved):

PR #669, opened 2021-08-30
Setting queue priority to 1 for push webhook events

mboudreau:
If a lot of people are syncing stuff, any push webhooks will be placed at the bottom of the queue and could take hours to sync up.  By setting it to priority 1, it becomes the next item in the queue and will drastically reduce the time it takes to show up in jira and prevent 'false flags' in issues.


thombergs (approved):
Nice, this makes sense. 

We should also work on adding observability into the queues, at some point. I.e. how many jobs of which types are currently waiting.

atrigueiro (approved):

mboudreau:
@thombergs I completely agree, I feel so blind without any data on that.  There are prebundled UIs that support bull that we can add to the app and place it behind a github login just for us :)

PR #677, opened 2021-09-01
ARC-600 fixing webhooks

mboudreau:
Reverting my parallelizing work as it might be causing race conditions.  Adding more logs for errors.

### DIFF
--- a/src/config/feature-flags.ts
+++ b/src/config/feature-flags.ts
@@ -42,7 +42,7 @@ const getLaunchDarklyValue = async (flag: BooleanFlags | StringFlags, defaultVal
 		const user = createLaunchdarklyUser(jiraHost);
 		return launchdarklyClient.variation(flag, user, defaultValue);
 	} catch (err) {
-		logger.error({flag, ...err}, "Error resolving value for feature flag");
+		logger.error({flag, err}, "Error resolving value for feature flag");
 		return defaultValue;
 	}
 }
--- a/src/config/feature-flags.ts
+++ b/src/config/feature-flags.ts
@@ -41,8 +41,8 @@ const getLaunchDarklyValue = async (flag: BooleanFlags | StringFlags, defaultVal
 		await launchdarklyClient.waitForInitialization();
 		const user = createLaunchdarklyUser(jiraHost);
 		return launchdarklyClient.variation(flag, user, defaultValue);
-	} catch (err: unknown) {
-		logger.error(`Error resolving value for feature flag "${flag}"`, err);
+	} catch (err) {
+		logger.error({flag, ...err}, "Error resolving value for feature flag");
 		return defaultValue;
 	}
 }
--- a/src/config/logger.ts
+++ b/src/config/logger.ts
@@ -2,6 +2,7 @@ import Logger, { levelFromName } from "bunyan";
 import bformat from "bunyan-format";
 import filteringStream from "../util/filteringStream";
 import { LoggerWithTarget, wrapLogger } from "probot/lib/wrap-logger";
+import { Request } from "express";
 
 // For any Micros env we want the logs to be in JSON format.
 // Otherwise, if local development, we want human readable logs.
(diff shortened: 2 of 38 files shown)


src/config/feature-flags.ts
```diff
@@ -41,8 +41,8 @@ const getLaunchDarklyValue = async (flag: BooleanFlags | StringFlags, defaultVal
 		await launchdarklyClient.waitForInitialization();
 		const user = createLaunchdarklyUser(jiraHost);
 		return launchdarklyClient.variation(flag, user, defaultValue);
-	} catch (err: unknown) {
-		logger.error(`Error resolving value for feature flag "${flag}"`, err);
+	} catch (err) {
+		logger.error({flag, ...err}, "Error resolving value for feature flag");
```
KAbakumovAtl, line 45: We wouldn't get stack trace if we log like that. Should be 

`logger.error(err, "Error. .... ." )`

src/config/statsd.ts
```diff
@@ -21,7 +21,7 @@ const statsd = new StatsD({
 	globalTags,
 	errorHandler: (err) => {
 		if (process.env.NODE_ENV !== EnvironmentEnum.development) {
-			logger.warn(err, "error writing metrics");
+			logger.warn({...err}, "Error writing metrics");
```
KAbakumovAtl, line 24: I wonder why are we doing it?

src/config/feature-flags.ts
mboudreau: pretty sure we do as the stack trace is a property of the error.

src/config/statsd.ts
mboudreau: because when an `Error` type object is sent to bunyan, it seems to only display 3 properties of the error (stack, type, msg) and ignore the rest.  I'm trying to make sure that it displays everything.

src/config/feature-flags.ts
KAbakumovAtl: No, you can test it. We had this problem a week ago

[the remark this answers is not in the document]
mboudreau: no, it's just an object.

mboudreau: well, it's a good thing that the stack trace isn't helping at all in the logs :/

src/config/statsd.ts
KAbakumovAtl: Yes, but if we do `{...err}` the stack trace will be lost.

We need to do something like this to get all fields:
`	logger.error(err, `Webhook Error: ${JSON.stringify(err)}`)`

(Maybe there is more elegant way to do it, but the err need to be the first parameter to get stack trace)

src/config/feature-flags.ts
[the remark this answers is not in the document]
KAbakumovAtl: Not sure if it is the best idea to add a feature flag as a log param. Hence I don't think it will be a common case to look for the feature flag parameter in Splunk. And having it in params instead of message will make it less visible

KAbakumovAtl (changes requested):
We need to fix stack traces. And I don't think adding all error parameters on the top level of logs is a good idea. We need to do at least something like: {err: error}  (Maybe you can extract a stack trace and add it as a parameter: {err: {stack_trace, error}} ) 

But doing in the way it is now is not OK

src/config/statsd.ts
mboudreau: Fixed this by modifying the global serializers for bunyan.

src/sync/installation.ts
```diff
@@ -157,7 +157,7 @@ const updateJobStatus = async (
 			logger.warn(err, "Error sending the `complete` event to JIRA");
 		}
 	} else {
-		logger.info("Sync status for installationId=%d is pending", installationId);
+		logger.info({ installationId }, "Sync status for installationId=%d is pending");
```
KAbakumovAtl, line 160: This will be logged as "Sync status for installationId=**%d** is pending" , installationId was a string format parameter

src/config/logger.ts
```diff
@@ -13,16 +14,42 @@ const formatOut = filteringStream(bformat({ outputMode, levelInString: true }));
...
 		name: "github-for-jira",
-		logger: "config.logger",
-		stream: formatOut
+		logger: "root-logger",
+		stream: formatOut,
+		serializers: {
+			err: (err) => (!err || !err.stack) ? err : {
+				...err,
+				stack: getFullErrorStack(err)
+			},
```
KAbakumovAtl, line 23: That looks better 👍

KAbakumovAtl (approved):
Approving assuming you've tested new error serializer

src/github/pull-request.ts
```diff
@@ -60,6 +76,6 @@ export default async (context: Context, jiraClient, util): Promise<void> => {
 		return;
 	}
 
-	context.log(`Sending pullrequest update to Jira ${jiraClient.baseURL}`)
+	context.log(`Sending pullrequest update to Jira ${jiraClient.baseURL}`);
```
thombergs, line 79: I'd like to add the pull request ID here, so we can search for a combination of Jira URL and PR id

src/jira/client/axios.ts
```diff
@@ -91,7 +91,7 @@ function getErrorMiddleware(logger: Logger) {
 				const errorMessage = getJiraErrorMessages(status);
 				// Creating an object that isn't of type Error as bunyan handles it differently
 				// Log appropriate level depending on status - WARN: 300-499, ERROR: everything else
-				(status >= 300 && status < 500 ? logger.warn : logger.error)({ ...error, response: error.response, resquest: error.request }, errorMessage);
+				(status >= 300 && status < 500 ? logger.warn : logger.error)(error, errorMessage);
```
thombergs, line 94: This previously didn't log the path name of the failed request:
image

We should add the path so that we know which endpoint causes the errors (and all those 404s)

thombergs (changes requested):

src/jira/client/axios.ts
mboudreau: It will with the latest serializer for the logger.

src/github/middleware.ts
```diff
@@ -118,14 +117,14 @@ export default (
 
 		await Promise.all(subscriptions.map(async (subscription) => {
```
thombergs, line 118: didn't we want to revert this to a for loop

thombergs (approved):

PR #682, opened 2021-09-02
ARC-477 add more logs to diagnose jira 400 failures and fix it

KAbakumovAtl:
A few changes here:

* Fixed request failing with 400 when trying to push commits to Jira
* For push webhook path,  carrying webhookId through the queue and adding to logs as a parameter
* Renamed id to webhookId in logs, where we use webhook id to reduce confusion

### DIFF
--- a/src/github/middleware.ts
+++ b/src/github/middleware.ts
@@ -9,6 +9,10 @@ import getJiraUtil from "../jira/util";
 import enhanceOctokit from "../config/enhance-octokit";
 import { Context } from "probot/lib/context";
 import { booleanFlag, BooleanFlags } from "../config/feature-flags";
+import {getLogger} from "../config/logger";
+
+
+const logger = getLogger("github.webhooks")
 
 // Returns an async function that reports errors errors to Sentry.
 // This works similar to Sentry.withScope but works in an async context.
@@ -72,8 +76,14 @@ export default (
 			payload: context.payload
 		});
 
+		const repoName = context.payload?.repository?.name || "none"
+		const orgName = context.payload?.repository?.owner?.name || "none"
+
 		const gitHubInstallationId = Number(context.payload?.installation?.id);
-		const logger = context.log = context.log.child({ gitHubInstallationId, event: webhookEvent, payload: context.payload });
+
+		//TODO Remove this line and uncomment the next one to get rid of payloads in logs
+		context.log = logger.child({ webhookId: context.id, repoName, orgName, gitHubInstallationId, event: webhookEvent, payload: context.payload });
+		//context.log = logger.child({ webhookId: context.id, repoName, orgName, gitHubInstallationId });
 
 		// Edit actions are not allowed because they trigger this Jira integration to write data in GitHub and can trigger events, causing an infinite loop.
 		// State change actions are allowed because they're one-time actions, therefore they won’t cause a loop.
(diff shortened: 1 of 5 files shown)


src/github/middleware.ts
```diff
@@ -72,8 +76,14 @@ export default (
 			payload: context.payload
 		});
 
+		const repoName = context.payload?.repository?.name || "none"
+		const orgName = context.payload?.repository?.owner?.name || "none"
+
 		const gitHubInstallationId = Number(context.payload?.installation?.id);
-		const logger = context.log = context.log.child({ gitHubInstallationId, event: webhookEvent, payload: context.payload });
+
+		//TODO Remove this line and uncomment the next one to get rid of payloads in logs
```
KAbakumovAtl, line 84: Added TODO so we know what to do

rachellerathbone (dismissed):

src/transforms/push.ts
```diff
@@ -12,20 +12,23 @@ import { JobOptions } from "bull";
...
+
+	const fallbackUrl = `https://github.com/${repoOwner}/${repoName}/blob/${commitHash}/${githubFile.filename}`
+
 	return {
 		path: githubFile.filename,
 		changeType: mapStatus[githubFile.status] || "UNKNOWN",
 		linesAdded: githubFile.additions,
 		linesRemoved: githubFile.deletions,
-		url: githubFile.blob_url
+		url: githubFile.blob_url || fallbackUrl
```
KAbakumovAtl, line 31: The main part. Sometimes we are getting null here, so I decided to construct URL manually in such cases. (Even if manually constructed URL will be wrong, it will be a smaller problem, hence it won't fail the Jira request with 400, and commit files are usually not viewed)

src/transforms/push.ts
```diff
@@ -12,20 +12,23 @@ import { JobOptions } from "bull";
 
 const logger = getLogger("transforms.push");
 
-function mapFile(githubFile) {
+const mapFile = (repoName: string, repoOwner: string, commitHash: string) => (githubFile) => {
```
thombergs, line 15: Can you explain this construct?

mboudreau: Instead of adding complexity of a function returning another function, just create a flat function with all the parameters and use an arrow function below: `filesToSend.map(file => mapFile(file, repo, owner.name, sha.id))`

src/github/middleware.ts
```diff
@@ -123,7 +133,7 @@ export default (
 				gitHubInstallationId.toString()
 			);
 			context.sentry.setUser({ jiraHost, gitHubInstallationId });
-			context.log = logger.child({ jiraHost });
+			context.log = context.log.child({ jiraHost });
```
mboudreau, line 136: The reason I used an external const of `logger` is because all of these function will reference `context.log` then create a child of each other instead of all having the same parent.

mboudreau (changes requested):

test/safe/push.test.ts
```diff
@@ -66,11 +66,11 @@ describe.skip("GitHub Actions", () => {
 
 			githubNock
 				.get("/repos/test-repo-owner/test-repo-name/commits/commit-no-username")
-				.replyWithFile(200, "../fixtures/api/commit-no-username.json");
+				.reply(200, require("../fixtures/api/commit-no-username.json"));
```
KAbakumovAtl, line 69: Fixed some tests

test/safe/push.test.ts
```diff
@@ -245,92 +245,6 @@ describe.skip("GitHub Actions", () => {
 			await expect(processPush(app)(job)).toResolve();
 		});
 
-		// Commenting these out for the moment. DevInfo API runs these
-		// transitions automatially based on the commit message, but we may
```
KAbakumovAtl, line 249: Removing commented code, we don't need these tests, hence we don't have this logic

mboudreau (approved):

PR #685, opened 2021-09-02
ARC-591 error pg

rachellerathbone:
Updated UI for github-error:

Screen Shot 2021-09-03 at 12 11 28 am


src/frontend/app.ts
```diff
@@ -304,10 +304,17 @@ export default (octokitApp: App): Express => {
 
 		statsd.increment(metricError.githubErrorRendered, tags);
 
-		return res.status(errorStatusCode).render("github-error.hbs", {
-			title: "GitHub + Jira integration",
-			nonce: res.locals.nonce
-		});
+		if (await booleanFlag(BooleanFlags.NEW_GITHUB_CONFIG_PAGE, false)) {
+			return res.status(errorStatusCode).render("github-error.hbs", {
```
bgvozdev, line 308: supernit looks like the only difference between if-body and else-body is the name of the template; could probably be simplified by having a ternary operator instead

views/github-error.hbs
```diff
@@ -1,21 +1,37 @@
...
-      <hr />
+      <h3 class="githubError__header">Looks like something went wrong!</h3>
 
-      <p class="text-gray text-center mb-4">We track these errors automatically, but if the problem persists feel free to <a href="https://support.atlassian.com/contact/#/?inquiry_category=technical_issues&is_cloud=true&product_key=third-party-product">contact us</a>. In the meantime, try adding an organization again in Jira.</p>
+      <p class="githubError__message">
+        We track these errors automatically, but if the problem persists feel
+        free to
+        <a
+          class="githubError__link"
+          href="https://support.atlassian.com/contact/#/?inquiry_category=technical_issues&is_cloud=true&product_key=third-party-product"
```
bgvozdev, line 26: (This is not the part of the PR, but I'll ask anyway): shouldn't it lead them to GitHub issues instead?

views/github-error.hbs
```diff
@@ -1,21 +1,37 @@
...
+        >contact us</a>. In the meantime, try adding an organization again in
+        Jira.
+      </p>
+
+      <a
+        class="githubError__link"
+        href="https://github.com/atlassian/github-for-jira/blob/old-master/README.md"
```
bgvozdev, line 33: old-master?

bgvozdev (dismissed):

PR #686, opened 2021-09-02
ARC-590 update connect an org pg

rachellerathbone:
In this PR I predominantly made changes to the connect an org page but also:

- made a few small tweaks to the subscription page so that it now uses AUI components. 
- updated the redirects for github/setup to differentiate between installing from Jira vs installing from GH

We now hide 'Repository access' header in table in empty states (no org with app installed):

Screen Shot 2021-09-03 at 9 35 50 am

When orgs have the app installed:
- we have removed the link from the org name (previously linked to the setting page in GH)
- changed the link for the editing icon (previously linked to the manage subscriptions page but now links to the GH settings pg)
- add a settings icon. On hover it displays the label 'Manage Jira subscriptions' and links to the Manage subscriptions page.

Screen Shot 2021-09-03 at 9 37 37 am

Updated state where user isn't an admin. Previously we rendered nothing which just looks like a bug in the UI. We now render a disabled button which displays the following label on hover.

(Can't see the hover icon in the screenshot so here's a screenshot from the design spec so you know what it is meant to look like)
Screen Shot 2021-09-03 at 9 45 14 am

Screen Shot 2021-09-03 at 9 43 19 am

Screen Shot 2021-10-05 at 4 55 11 pm

Other changes to note/things to confirm:

- You also shouldn't be able to see the editing pencil for orgs where you are not an admin.
- If a sync fails you will now see a warning icon with an option to resync

-----------------------------------------------------------------------------------------------------------------------------

Also tweaked the flow for installing the app to make sure users are redirected to this page when installing from the Jira side (had to revert a previous change last week that had broken the install flow from the GH side).

**When installing from Jira:**

https://user-images.githubusercontent.com/37155488/135967770-8b2cf08b-3cf3-4c5b-adb9-fdb042feefec.mov

**When installing from GH:**

https://user-images.githubusercontent.com/37155488/135967955-4ff1284a-6b38-40af-b8c6-2afd49366a5a.mov


src/frontend/get-github-setup.ts
```diff
@@ -1,22 +1,33 @@
...
+- From Jira: user has already installed the app and is redirected to the connect an org pg
+- From GH:
+		- If we have the users Jira host, redirect to marketplace.
+		- Otherwise, render the setup page.
 */
 export default (req: Request, res: Response): void => {
 	req.log.info("Received get github setup page request");
 
-	if (req.session.jiraHost) {
+	if (req.headers.referer) {
```
rachellerathbone, line 19: This property is only present when we install the app via the Jira flow. Not sure if this is the best way to determine if someone is coming from the Jira or GH side but open to suggestions for distinguishing between the two.

static/js/github-configuration.js
```diff
@@ -38,3 +38,27 @@ $('.logout-link').click(function (event) {
 		this.close()
 	}, 100)
 })
+
+$(".sync-connection-link").click(function (event) {
```
rachellerathbone, line 42: If a sync fails we now display a 'Retry' option on this page. This is the click event that fires it.

views/github-configuration-OLD.hbs
```diff
@@ -0,0 +1,138 @@
+<html>
```
rachellerathbone, line 1: To be deleted when FF is cleaned up.

src/frontend/get-github-setup.ts
bgvozdev: I'm not familiar with these changes, but random idea: could we maybe use different URLs in Jira and in GitHub (e.g. add a query parameter)?

bgvozdev: Question: why are we not feature-flagging this change?

bgvozdev: Answering to myself, probably because this is some code-shuffle refactoring :)

static/js/github-configuration.js
bgvozdev: Lol, I didn't know we use jquery! Ah, good old 2004th :aww:

static/js/github-configuration.js
```diff
@@ -38,3 +38,27 @@ $('.logout-link').click(function (event) {
...
+			installationId,
+			jiraHost,
+			syncType: "full",
+			_csrf: csrfToken,
+		},
+		success: function (data) {
+			window.close();
+		},
+		error: function (error) {
+			console.log(error);
```
bgvozdev, line 61: Not sure if we could: would/should we log a sentry error?

static/js/github-configuration.js
```diff
@@ -38,3 +38,27 @@ $('.logout-link').click(function (event) {
...
+
+$(".sync-connection-link").click(function (event) {
+	event.preventDefault();
+	const installationId = $(event.target).data("installation-id");
+	const jiraHost = $(event.target).data("jira-host");
+	const csrfToken = document.getElementById("_csrf").value;
+
+	$.ajax({
+		type: "POST",
+		url: `/jira/sync`,
```
bgvozdev, line 50: supernit: quotes

views/github-configuration.hbs
```diff
@@ -1,116 +1,200 @@
...
+                      </p>
                     {{/if}}
-                  </div>
+                    {{else}}
+                    <button class="gitHubConfiguration__orgContent__connectBtn--disabled aui-button" aria-disabled="true" disabled>
+                      Connect
+                    </button>
+                  {{/if}}
+
+                  {{#if admin}}
```
bgvozdev, line 136: Hmm... I thought this page is available for admins only... What I'm missing?

bgvozdev (dismissed):

bgvozdev (approved):

static/js/github-configuration.js
rachellerathbone: I mean, it would be helpful if we were doing more than just logging it to the console (lazy copy and paste from jira-configuration.js code). Will take a look into this tomorrow and will probably update in both places.

views/github-configuration.hbs
rachellerathbone: The whole page?? Haha. Nope. See screenshot below for what currently exists is prod. admin just controls whether we can connect or not (like I can't for the Atlassian org).

Screen Shot 2021-10-06 at 6 52 22 pm

PR #687, opened 2021-09-03
revert the Promise.all() to a simple for loop to avoid any race condi…

thombergs:
…tions that we might experience.

### DIFF
--- a/src/github/middleware.ts
+++ b/src/github/middleware.ts
@@ -115,7 +115,7 @@ export default (
 			`webhook:${context.name}.${context.payload.action}`
 		);
 
-		await Promise.all(subscriptions.map(async (subscription) => {
+		for(const subscription of subscriptions) {
 			const { jiraHost } = subscription;
 			context.sentry.setTag("jiraHost", jiraHost);
 			context.sentry.setTag(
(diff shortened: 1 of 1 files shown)


src/github/middleware.ts
```diff
@@ -128,7 +128,7 @@ export default (
 
 			if (await booleanFlag(BooleanFlags.MAINTENANCE_MODE, false, jiraHost)) {
 				context.log(`Maintenance mode ENABLED for jira host ${jiraHost} - Ignoring event of type ${webhookEvent}`);
-				return;
+				continue;
```
thombergs, line 131: We could have kept these 3 `returns`, but it's cleaner and more refactoring-safe if we never return from within the loop.

KAbakumovAtl (approved):

src/github/middleware.ts
KAbakumovAtl: Shame on me, I didnt' check for other returns!

src/github/middleware.ts
```diff
@@ -142,7 +142,7 @@ export default (
 					},
 					`Timing out at after ${context.timedout}ms`
 				);
-				return;
+				continue;
```
thombergs, line 145: Actually, this return was plain wrong. After a timeout we should at least try to send the data to the next Jira. But it was always wrong...

src/github/middleware.ts
KAbakumovAtl: Let's extract the code within for loop to a separate function. I am still getting confused

mboudreau (approved):

PR #691, opened 2021-09-06
ARC-406 Improve categorisation of deployment environments

fizwidget:
At the moment almost 60% of GitHub deployments are being categorised as "unmapped" in Jira. This limits the usefulness of Jira features that rely on these categories, e.g. the Cycle Time & Deployment Frequency reports.

The current logic matches against a list of common names, but does not recognise _variations_ of those names. For example `prod` is categorised correctly but `prod-east` is not.

This PR attempts to improve the situation by recognising variations of common environment names.


src/transforms/deployment.ts
```diff
@@ -26,13 +26,15 @@ function mapState(state: string): string {
...
 	// We need to map the environment of a GitHub deployment back to a valid deployment environment in Jira.
 	// https://docs.github.com/en/actions/reference/environments
 	// GitHub: does not have pre-defined values and users can name their environments whatever they like. We try to map as much as we can here and log the unmapped ones.
 	// Jira: Can be one of unmapped, development, testing, staging, production
-	const isEnvironment = (envNames) =>
-		envNames.some(envName => environment.localeCompare(envName, undefined, { sensitivity: "base", ignorePunctuation: true }) === 0);
+	const isEnvironment = (envNames) => {
+		const envNamesPattern = RegExp(`^(.*[^a-z0-9])?(${envNames.join("|")})([^a-z0-9].*)?$`, "i");
+		return environment.match(envNamesPattern);
+	}
```
fizwidget, line 37: I feel like this is a bit hard to follow...please let me know if you have suggestions for improving readability here :)

src/transforms/deployment.ts
```diff
@@ -26,13 +26,15 @@ function mapState(state: string): string {
...
 }
 
-function mapEnvironment(environment: string): string {
+export function mapEnvironment(environment: string): string {
 	// We need to map the environment of a GitHub deployment back to a valid deployment environment in Jira.
 	// https://docs.github.com/en/actions/reference/environments
 	// GitHub: does not have pre-defined values and users can name their environments whatever they like. We try to map as much as we can here and log the unmapped ones.
 	// Jira: Can be one of unmapped, development, testing, staging, production
-	const isEnvironment = (envNames) =>
-		envNames.some(envName => environment.localeCompare(envName, undefined, { sensitivity: "base", ignorePunctuation: true }) === 0);
```
fizwidget, line 35: Happy to keep the existing `localeCompare` logic around too if we think it's worthwhile. Removing it means things like `intégration` will end up as unmapped, when they previously would have been categorised correctly.

mattcolman:
Nice!

src/transforms/deployment.ts
```diff
@@ -26,13 +26,15 @@ function mapState(state: string): string {
...
+	const isEnvironment = (envNames) => {
+		const envNamesPattern = RegExp(`^(.*[^a-z0-9])?(${envNames.join("|")})([^a-z0-9].*)?$`, "i");
```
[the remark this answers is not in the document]
mboudreau: Also, could you please add typings to the parameter? `string[]`

test/unit/transforms/deployments.test.ts
```diff
@@ -0,0 +1,59 @@
...
+		`${environmentName}:east`,
+		`${environmentName}_east`,
+		`${environmentName}/east`,
+		`${environmentName} east`,
+	]);
+
+	it("should transform commonly used environment names to the expected environment types", () => {
+		const environmentTypesToEnvironmentNames = _.groupBy(exampleEnvironmentNameVariations, mapEnvironment);
+
+		expect(environmentTypesToEnvironmentNames).toMatchSnapshot();
```
mboudreau, line 57: Can't say I'm a fan of having a test just based on the snapshot as it can easily be updated by mistake and there's nothing else to catch it.  It will also fail if any new environments are added or regex might change the capture, but not specifically going through what are the 'rules' of the algorithm.

I also don't see any failure tests, only successful ones.

mboudreau (changes requested):
Thanks for the PR, this is great :D
Just need a few things changed and we should be good to release this :)

test/unit/transforms/deployments.test.ts
fizwidget: Yeah that's fair, snapshot tests do have their drawbacks. I'll rework the tests 👍 

> I also don't see any failure tests, only successful ones.

I was thinking the [unmapped environments](https://github.com/atlassian/github-for-jira/pull/691/files#diff-bd38d697968a2846e77fea1a253983dd9944eb96b2633953f6af5b1e03a319f7R293) could be considered failures in this context. Perhaps that's not what you mean though?

mboudreau: I guess, but those failures are very simplistic.  they don't have 'prod' or any other environments in their naming convention at all.  I'd probably try things like `postage` or have some names with some non-latin characters, etc.  They will return as unmapped, but the more environment names we can throw at it, the more confidence we can have in the regex :)

src/transforms/deployment.ts
```diff
@@ -26,19 +27,30 @@ function mapState(state: string): string {
...
+		);
+		return envNamesPattern.test(_.deburr(environment));
+	};
 
 	const environmentMapping = {
 		development: ["development", "dev", "trunk"],
 		testing: ["testing", "test", "tests", "tst", "integration", "integ", "intg", "int", "acceptance", "accept", "acpt", "qa", "qc", "control", "quality"],
 		staging: ["staging", "stage", "stg", "preprod", "model", "internal"],
-		production: ["production", "prod", "live"],
+		production: ["production", "prod", "prd", "live"],
```
fizwidget, line 53: Snuck in `prd` here as I've seen this used before. Let me know if this is a bad idea for some reason.

test/unit/transforms/deployments.test.ts
fizwidget: Yep good idea, I've added a non-latin example. I think I'm missing the significance of `postage` though?

src/transforms/deployment.ts
```diff
@@ -26,19 +27,30 @@ function mapState(state: string): string {
...
+		// Matches separators within environment names, e.g. "-" in "prod-east" or ":" in "test:mary"
+		const separator = "[^a-z0-9]";
+		// Matches an optional prefix, followed by one of the input names, followed by an optional suffix.
+		// This lets us match variants, e.g. "prod-east" and "prod-west" are considered variants of "prod".
+		const envNamesPattern = RegExp(
+			`^(.*${separator})?(${exactMatch})(${separator}.*)?$`,
+			"i"
+		);
+		return envNamesPattern.test(_.deburr(environment));
+	};
```
fizwidget, line 47: I've split out some helper variables and added comments - hopefully more readable now :)

src/transforms/deployment.ts
mboudreau: nah, it's fine.  At some point, we're probably going to make this whole mapping thing configurable by our customers.

mboudreau (approved):
Looking much better :)

test/unit/transforms/deployments.test.ts
mboudreau: `postage` has `stage` in it :P

mboudreau:
@fizwidget It's deploying to staging.  Feel free to test it on that environment and once it's all good, we'll release to prod :)

PR #693, opened 2021-09-06
ARC-616: Extend app descriptor to include remotelinks

damoster:
Was there anything else I was meant to do?

### DIFF
--- a/src/jira/connect.ts
+++ b/src/jira/connect.ts
@@ -81,7 +81,7 @@ export default async (_: Request, res: Response): Promise<void> => {
 				jiraRemoteLinkInfoProvider: {
 					key: "github-remotelinks-integration",
 					name: {
-						value: "GitHub Security Reports"
+						value: "GitHub"
 					},
 					logoUrl: "https://github.githubassets.com/images/modules/logos_page/GitHub-Mark.png",
 					homeUrl: "https://github.com/features/security"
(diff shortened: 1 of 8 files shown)


src/jira/connect.ts
```diff
@@ -78,6 +78,14 @@ export default async (_: Request, res: Response): Promise<void> => {
 					},
 					homeUrl: "https://github.com/features/actions",
 				},
+				jiraRemoteLinkInfoProvider: {
+					key: "github-remotelinks-integration",
+					name: {
+						value: "GitHub Security Reports"
```
TerryAg, line 84: Needs PM approval

atrigueiro:
You'll also need to add a reference to remote links on the [Jira client](https://github.com/atlassian/github-for-jira/blob/main/src/jira/client/index.ts#L220), so we can do something like `remoteLinks.update` [here](https://github.com/atlassian/github-for-jira/blob/ARC-617-security-reports-remote-links/src/github/code-scanning-alert.ts#L14).

src/jira/connect.ts
damoster: After slack discussion and PM chat, changed to "GitHub" now

atrigueiro (approved):

TerryAg (approved):

PR #695, opened 2021-09-07
ARC-612 Standardizing all author fields

mboudreau:
Using the same function throughout and use common sense defaults so that no required fields are missing.

### DIFF
--- a/src/config/logger.ts
+++ b/src/config/logger.ts
@@ -11,17 +11,17 @@ const outputMode = process.env.MICROS_ENV ? "json" : "short";
 // add levelInString to include DEBUG | ERROR | INFO | WARN
 const formatOut = filteringStream(bformat({ outputMode, levelInString: true }));
 
-const requestSerializer = (req: Request) => (!req || !req.connection) ? req : {
+const requestSerializer = (req: Request) => req && {
 	method: req.method,
 	url: req.originalUrl || req.url,
 	path: req.path,
 	headers: req.headers,
-	remoteAddress: req.connection.remoteAddress,
-	remotePort: req.connection.remotePort,
+	remoteAddress: req.connection?.remoteAddress,
+	remotePort: req.connection?.remotePort,
 	body: req.body
 };
 
-const errorSerializer = (err) => (!err || !err.stack) ? err : {
+const errorSerializer = (err) => err && {
 	...err,
 	response: Logger.stdSerializers.res(err.response),
 	request: requestSerializer(err.request),
@@ -47,7 +47,9 @@ const logger = Logger.createLogger(
 		serializers: {
 			err: errorSerializer,
 			res: Logger.stdSerializers.res,
-			req: requestSerializer
+			response: Logger.stdSerializers.res,
+			req: requestSerializer,
+			request: requestSerializer
 		}
 	}
 );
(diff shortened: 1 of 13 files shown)


src/config/logger.ts
```diff
@@ -11,17 +11,17 @@ const outputMode = process.env.MICROS_ENV ? "json" : "short";
...
 	headers: req.headers,
-	remoteAddress: req.connection.remoteAddress,
-	remotePort: req.connection.remotePort,
+	remoteAddress: req.connection?.remoteAddress,
+	remotePort: req.connection?.remotePort,
 	body: req.body
 };
 
-const errorSerializer = (err) => (!err || !err.stack) ? err : {
+const errorSerializer = (err) => err && {
```
KAbakumovAtl, line 24: What does this && operator doing? 

If err is not set then it will be undefined, otherwise it will be the following object?

src/jira/client/index.ts
```diff
@@ -302,7 +302,7 @@ const batchedBulkUpdate = async (
 			}
 		};
 		return instance.post("/rest/devinfo/0.10/bulk", body).catch((err) => {
-			logger.error({err, body, data}, "Jira Client Error: Cannot update Pull Request")
+			logger.error({ err, res: err.response, req: err.request, body, data }, "Jira Client Error: Cannot update Pull Request");
```
KAbakumovAtl, line 305: Let's change the message, because it is not only related to pull requests

src/util/jira.ts
```diff
@@ -0,0 +1,26 @@
+import { JiraAuthor } from "../interfaces/jira";
+
+export const getJiraAuthor = (author?: Author): JiraAuthor => author ? {
+	avatar: author.avatar_url || author.avatarUrl || `https://github.com/${author.login}.png`,
```
KAbakumovAtl, line 4: So author can either have avatar_url or avatarUrl parameters? What does it depend on?

src/util/jira.ts
```diff
@@ -0,0 +1,26 @@
...
+} : {
+	avatar: "https://github.com/ghost.png",
+	name: "Deleted User",
+	email: "deleted@noreply.user.github.com",
+	url: "https://github.com/ghost"
+};
+
+interface Author {
+	avatar_url?: string;
+	avatarUrl?: string;
```
KAbakumovAtl, line 17: Let's add a comment with clarification why there are both `avatar_url` and `avatarUrl`

src/config/logger.ts
mboudreau: if error is truthy it continues to return the object or else it returns itself which will be serialized out.

src/jira/client/index.ts
mboudreau: yeah, good point.

KAbakumovAtl (dismissed):
Looks good to me, just would be great to fix the message and add comment regarding Author fields

src/util/jira.ts
mboudreau: There were some places in the code that were referencing `avatarUrl` but I'm fairly sure that it's wrong as I'm fairly sure it's all `avatar_url` when it comes from the github data.  Didn't want to change it in case it might work.

mboudreau: actually, the difference is between the REST API and the GraphQL API.  One uses `avatar_url` while the other does `avatarUrl` respectively.

KAbakumovAtl (approved):

PR #697, opened 2021-09-07
ARC-589 Added support for Jira staging

KAbakumovAtl:
Added ability to install the app on staging Jira Instances


test/unit/jira/util/jwt.test.ts
```diff
@@ -388,10 +396,6 @@ describe("jwt", () => {
 				iss: "jira",
 			});
 
-			when(queryAtlassianConnectPublicKey).calledWith(
-				testKid
-			).mockRejectedValueOnce(new Error("404"));
```
KAbakumovAtl, line 393: These tests were wrong. Copy-pasting problem 🤦

src/jira/util/jwt.ts
```diff
@@ -184,6 +184,19 @@ export const verifySymmetricJwtTokenMiddleware = (secret: string, tokenType: Tok
 
 const ALLOWED_BASE_URLS = [BASE_URL]
 
+const isStagingTenant = (req: Request): boolean => {
+	let isStagingTenant = false;
+	const hostBaseUrl = _.get(req.body, "baseUrl");
```
mboudreau, line 189: why are you using `_.get` just to get a property of body?

src/jira/util/jwt.ts
```diff
@@ -184,6 +184,19 @@ export const verifySymmetricJwtTokenMiddleware = (secret: string, tokenType: Tok
...
+	if (hostBaseUrl) {
+		const host = new URL(hostBaseUrl).hostname;
```
mboudreau, line 191: This needs to be wrapped in a try/catch in case the baseUrl isn't a URL.

src/jira/util/jwt.ts
```diff
@@ -184,6 +184,19 @@ export const verifySymmetricJwtTokenMiddleware = (secret: string, tokenType: Tok
...
+		const hostEnvironment = host.substring(host.indexOf(".") + 1);
```
mboudreau, line 192: Couldn't you use regex for this? `return /\.jira-dev\.com$/.test(new URL(hostBaseUrl).host)`

src/jira/util/jwt.ts
```diff
@@ -184,6 +184,19 @@ export const verifySymmetricJwtTokenMiddleware = (secret: string, tokenType: Tok
 
 const ALLOWED_BASE_URLS = [BASE_URL]
 
+const isStagingTenant = (req: Request): boolean => {
+	let isStagingTenant = false;
```
mboudreau, line 188: don't need this var.

src/jira/util/jwt.ts
```diff
@@ -184,6 +184,19 @@ export const verifySymmetricJwtTokenMiddleware = (secret: string, tokenType: Tok
...
+	const hostBaseUrl = _.get(req.body, "baseUrl");
+	if (hostBaseUrl) {
+		const host = new URL(hostBaseUrl).hostname;
+		const hostEnvironment = host.substring(host.indexOf(".") + 1);
+		if (hostEnvironment === "jira-dev.com") {
+			isStagingTenant = true;
+		}
+	}
+	return isStagingTenant
```
mboudreau, line 197: just `return false` if it gets to the end.

src/jira/util/jwt.ts
```diff
@@ -193,7 +206,7 @@ export const verifyAsymmetricJwtTokenMiddleware = async (req: Request, res: Resp
 			return;
 		}
 
-		const publicKey = await queryAtlassianConnectPublicKey(getKeyId(token));
+		const publicKey = await queryAtlassianConnectPublicKey(getKeyId(token), isStagingTenant(req));
```
mboudreau, line 209: I'd prefer if the `isStagingTenant` function didn't have to assume that that `req.body.baseUrl` exists as it's not really it's concern.  Just doing `isStagingTenant(req.body?.baseUrl)` should do.

mboudreau (changes requested):

src/jira/util/jwt.ts
KAbakumovAtl: It already works without baseUrl in the body, in most unit tests it is not there

src/jira/util/jwt.ts
KAbakumovAtl: I copied this code from ACE and left it almost unchanged. I don't think we should change much there to make sure that it will work in the same way as ACE.

src/jira/util/jwt.ts
KAbakumovAtl: Ok, this one makes sense

src/jira/util/jwt.ts
```diff
@@ -184,6 +184,19 @@ export const verifySymmetricJwtTokenMiddleware = (secret: string, tokenType: Tok
 
 const ALLOWED_BASE_URLS = [BASE_URL]
 
+const isStagingTenant = (req: Request): boolean => {
+	try {
+		const hostBaseUrl = req.body?.baseUrl;
+		if (hostBaseUrl) {
```
mboudreau, line 190: You don't need the if at all.  It either parses or it throws an error.

KAbakumovAtl: No, I am not going to use exceptions where condition can be used

mboudreau (approved):

PR #704, opened 2021-09-08
Arc 615 remotelinks transformer

TerryAg:


src/interfaces/jira.ts
```diff
@@ -86,17 +86,29 @@ export interface JiraAssociation {
 }
 
 export interface JiraRemoteLinkData {
-	product: string;
 	remoteLinks: JiraRemoteLink[];
 }
 
+interface IssueKeysAssociation {
```
damoster: /nit change name to `Association` or `JiraAssociation` to match api schema?

src/interfaces/jira.ts
```diff
@@ -86,17 +86,29 @@ export interface JiraAssociation {
...
 	schemaVersion: string;
 	updateSequenceNumber: number;
 	associations: JiraAssociation[];
 	displayName: string;
+	description: string;
 	url: string;
 	type: string;
+	status: JiraRemoteLinkStatus;
 	lastUpdated: number;
+	associations: IssueKeysAssociation[];
```
damoster: As the build says, you've got associations defined twice in this schema

damoster (changes requested):
Fix duplicate identifier. Other than that looks good, left some nit comments

CONTRIBUTING.md
```diff
@@ -44,11 +44,14 @@ Your new GitHub app will need the following repository permissions & events:
 + Issues: Read & write
 + Metadata: Read-only
 + Pull requests: Read & write
++ Security events: Read-only
 
 **Organization Permissions**:
 + Members: Read-Only
 
 **Subscribe to Events**:
++ Security advisory
```
atrigueiro, line 53: Do we need this? Aren't we only listening to the `Code scanning alert` event?

CONTRIBUTING.md
```diff
@@ -101,7 +104,7 @@ Go to your Jira instance that you created earlier and do the following steps:
 
 ### Setting up the App
 
-In your Jira instance, in the `Manage Apps` section, click on your App's button, then click on `Get Started`.  This will bring you to the App's dashboard.  Click the `Add an Organization` button and follow the steps to install the App on Github and allow it permission to view your repos.
+In your Jira instance, in the `Manage Apps` section, click on your App's button, then click on `Get Started`.  This will bring you to the App's dashboard.  Click the `Connect GitHub organisation` button and follow the steps to install the App on GitHub and allow it permission to view your repos.
```
atrigueiro, line 107: nit: z instead of s for organization

src/transforms/code-scanning-alert.ts
```diff
@@ -1,35 +1,80 @@
...
+			return components.slice(2).join("/");
+		case "pull": // pull request
+			return getPullRequestTitle(parseInt(components[2]));
+		default:
+			// log something here
+			return "";
+	}
+}
+
+const transformStatusToAppearance = (status: string): string => {
```
atrigueiro, line 25: Can we add a comment explaining possible values of `status` and valid values for `appearance` please? Similar to what was done [here](https://github.com/atlassian/github-for-jira/blob/main/src/transforms/deployment.ts#L11).

src/transforms/code-scanning-alert.ts
```diff
@@ -1,35 +1,80 @@
...
+		// The webhook contains ALL instances of this alert, so we need to grab the ref from each instance.
+		entityTitles.push(...alert.instances.map((instance) => getEntityTitle(instance.ref)));
+	} else {
+		// The action is associated with a single branch/PR
+		entityTitles.push(getEntityTitle(ref));
+	}
+
+	const parser = issueKeyParser();
+	// Flatten and get all issue keys
+	const issueKeys = [].concat(...entityTitles.map((entityTitle) => parser.parse(entityTitle)));
```
atrigueiro, line 54: You can just do `issueKeyParser().parse(entityTitle)`

src/transforms/code-scanning-alert.ts
```diff
@@ -1,35 +1,80 @@
...
+			id: `${repository.id.toString()}-${alert.number}`,
 			updateSequenceNumber: Date.now(),
-			associations: [],
-			displayName: code_scanning_alert.name,
-			url: code_scanning_alert.html_url,
-			type: 'security',
-			lastUpdated: getLastUpdated(code_scanning_alert)
-		}],
+			displayName: alert.rule.description,
+			description: alert.rule.full_description,
```
atrigueiro, line 66: I was just checking on the UI and it doesn't seem Jira is able to display the whole description. It also has a limit of 255 characters. We should consider displaying just the `alert.rule.description` here and something else as the link. instead.

src/interfaces/jira.ts
TerryAg: Resolved with merge conflict

src/interfaces/jira.ts
TerryAg: Resolved with merge conflict

CONTRIBUTING.md
TerryAg: image

The permission makes me think we need it. It explicitly mentions "code scanning alerts" - I can test what happens without the permission

CONTRIBUTING.md
TerryAg: ewwwww :)

src/transforms/code-scanning-alert.ts
TerryAg: Good idea. I might take this a step further and define an enum type for the `appearance` (like Jira has)

CONTRIBUTING.md
atrigueiro: So that's Security events, which is the permission you added on line 47 above, which is correct. But we don't need to listen to security advisory, just need to listen to the code scanning event on the line below.

src/transforms/code-scanning-alert.ts
TerryAg: Won't this create a whole new object every iteration through `entityTitles` ?

CONTRIBUTING.md
atrigueiro: haha I know, but just to keep the pattern from https://github.com/atlassian/github-for-jira/blob/main/views/jira-configuration.hbs#L24

src/transforms/code-scanning-alert.ts
TerryAg: Both displayname and description can be hovered over to get more information:
image

But you're right, this could go over 255 chars and the request will fail, so I need to probably slice everything after 255.

I'm not against changing the values around but maybe need to ask PM.

atrigueiro: Oh cool I didn't know we could hover over! But yeah, we'll still need to slice it

src/transforms/code-scanning-alert.ts
```diff
@@ -1,35 +1,84 @@
...
+			id: `${repository.id.toString()}-${alert.number}`,
 			updateSequenceNumber: Date.now(),
-			associations: [],
-			displayName: code_scanning_alert.name,
-			url: code_scanning_alert.html_url,
-			type: 'security',
-			lastUpdated: getLastUpdated(code_scanning_alert)
-		}],
+			displayName: `Alert #${alert.number}`,
+			description: alert.rule.description.substring(0, MAX_STRING_LENGTH),
```
atrigueiro, line 70: Noice!

src/transforms/code-scanning-alert.ts
TerryAg: I've added a slice

CONTRIBUTING.md
TerryAg: Yea I think Jira also uses 'z' so makes sense

atrigueiro (approved):

damoster (approved):

PR #705, opened 2021-09-08
ARC-612 fixing author generation throughout the codebase

mboudreau:
More fixes to author names for dev info panel


atrigueiro:
Looks good to me. I'm just confused why we're doing `getGithubUser` on pull request actions in addition to `getJiraAuthor` and all the other actions (commits, branches) we just do `getJiraAuthor`?

mboudreau:
@atrigueiro Yeah, fair, I need to add some comments.

I noticed while doing testing that in some instances, it was using the username as the name, and in other places using the full name.  Apparently, there's some API requests that if it has a `user` or `author` field, it'll return the avatar, id, username....but not the full name of the user.  These are the places where I found that didn't have the name field, so I do an extra call to get the full user information which includes the full name.

It sucks, I know.  Github API isn't very consistent.  This will be fixed once we move over to using GraphQL only instead of REST, but that change is bigger and riskier.  This is an in-between fix for that.

docker-compose.yml
```diff
@@ -36,4 +36,6 @@ services:
     environment:
       DATABASE_URL: postgres://postgres:postgres@postgres:5432/jira-dev
       REDIS_HOST: redis
+      REDIS_PORT: 6379
```
KAbakumovAtl: Does it relate to switching to redisx?

test/fixtures/api/user.json
```diff
@@ -0,0 +1,19 @@
+{
+	"login": "bkeepers",
+	"name": "Bee Keepers",
```
KAbakumovAtl, line 3: It it a real user data?

mboudreau: no, all fake.  Probably a user named that in github though...

src/sync/queries.ts
```diff
@@ -52,6 +52,7 @@ export const getCommits = `query ($owner: String!, $repo: String!, $per_page: In
                   message
                   oid
                   url
+                  changedFiles
```
thombergs, line 55: This is a number and not a list of changed files, right? Just want to prevent us from loading a bunch of data that we end up not needing.

thombergs (approved):
Nice. I had some doubts regarding calling the GitHub API to get the user, but it's all protected with try/catch!

src/sync/queries.ts
mboudreau: Correct, just an `Int`.  We weren't actually sending the commited files to the jira api, just 0.

KAbakumovAtl (approved):

thombergs (approved):

test/fixtures/api/user.json
KAbakumovAtl: Yep. But at least it is all public

https://github.com/bkeepers

KAbakumovAtl (approved):

PR #711, opened 2021-09-10
ARC-637 adding options to remove failed jobs, manually cleaning old failed jobs

mboudreau:
This should remove many jobs that aren't needed anymore in redis and clear up the unstability.

Also removed old datadog folder as we don't use it anymore.


src/worker/main.ts
```diff
@@ -23,15 +23,15 @@ const client = new Redis(getRedisInfo("client").redisOptions);
...
+function measureElapsedTime(job: Queue.Job, tags) {
+	statsd.histogram(metricHttpRequest().jobDuration, job.finishedOn - job.processedOn, tags);
 }
 
 const queueOpts: QueueOptions = {
 	defaultJobOptions: {
-		removeOnComplete: true
+		attempts: 1,
+		removeOnComplete: true,
+		removeOnFail: true
```
thombergs, line 34: whoa. That means redis has been keeping all the failed jobs in memory forever?

src/worker/main.ts
```diff
@@ -23,15 +23,15 @@ const client = new Redis(getRedisInfo("client").redisOptions);
...
-	const timeDiff = endTime - startTime;
-	statsd.histogram(metricHttpRequest().jobDuration, timeDiff, tags);
+function measureElapsedTime(job: Queue.Job, tags) {
+	statsd.histogram(metricHttpRequest().jobDuration, job.finishedOn - job.processedOn, tags);
 }
 
 const queueOpts: QueueOptions = {
 	defaultJobOptions: {
-		removeOnComplete: true
+		attempts: 1,
```
thombergs, line 32: shouldn't this be a value > 1? To catch some intermittent 5xxs from Jira or similar?

thombergs (dismissed):

src/worker/main.ts
mboudreau: I'm not certain because there was some places referencing `removeOnFail`.  The config stuff is everywhere and super confusing.  I've tried to make it a bit more succinct by having the config in only one place.

src/worker/main.ts
mboudreau: yep, your right, I'll fix that.

src/sync/discovery.ts
```diff
@@ -5,12 +5,6 @@ import { Application } from "probot";
 import { SyncStatus } from "../models/subscription";
 import { getLogger } from "../config/logger";
 
-const jobOpts = {
-	removeOnComplete: true,
-	removeOnFail: true,
-	attempts: 3
-};
```
KAbakumovAtl, line 12: So we removing it from here because we put these settings globally on the queue, right?

KAbakumovAtl (commented):
Let's also increase our worker nodes count

thombergs (approved):

KAbakumovAtl (approved):

src/sync/discovery.ts
mboudreau: yes

PR #712, opened 2021-09-10
ARC-241: Send GitHub Security Reviews as Remote Links to Jira

atrigueiro:
We're adding support to sending GitHub Security Reviews as Remote Links to Jira:
image (1)

Some useful links when reviewing this PR: 

GitHub Security Reviews: https://github.com/features/security

Code Scanning Alert webhook: https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#code_scanning_alert

Remote Links POST API: https://developer.atlassian.com/cloud/jira/software/rest/api-group-remote-links/#api-remotelinks-1-0-bulk-post


README.md
```diff
@@ -114,6 +114,7 @@ Read, Write, and Admin for Development Information (branches, commits, and pull
...
 |**Read** and **write** access to actions and deployments|If you want to see build and deployment information in Jira, the app will need read and write permissions for deployments. In order for the GitHub for Jira app to send the correct status (pending, in progress, successful etc) of a deployment to Jira we require the `state` property. This property is only available on the `deployment_status` event which occurs when a deployment is created. The [GitHub API](https://docs.github.com/en/rest/reference/repos#create-a-deployment-status) requires that GitHub apps have read and write access to listen to deployment creation events. <br><br> **Deployments:** Read and write deployment permissions are used for the following webhooks: <br> - deployment status|
+|**Read** access to Security events| To sync GitHub code scanning alerts as Remote Links in Jira|
```
KAbakumovAtl, line 117: /nit We can just simply tell " sync GitHub code scanning alerts with Jira tickets" 

I believe those people who will be reviewing permissions are not aware too much of what "remote links" are

src/github/code-scanning-alert.ts
```diff
@@ -0,0 +1,15 @@
+import transformCodeScanningAlert from "../transforms/code-scanning-alert";
+import { Context } from "probot/lib/context";
+import logger from "../config/logger";
+
+export default async (context: Context, jiraClient): Promise<void> => {
```
KAbakumovAtl, line 5: Should we put it behind a feature flag?

src/jira/connect.ts
```diff
@@ -78,6 +78,14 @@ export default async (_: Request, res: Response): Promise<void> => {
 					},
 					homeUrl: "https://github.com/features/actions",
 				},
+				jiraRemoteLinkInfoProvider: {
+					key: "github-remotelinks-integration",
+					name: {
+						value: "GitHub"
+					},
+					logoUrl: "https://github.githubassets.com/images/modules/logos_page/GitHub-Mark.png",
+					homeUrl: "https://github.com/features/security"
```
KAbakumovAtl, line 87: I wonder if we would have more remote links from GitHub? (like GitHub Issues?) 
Should we put more generic URL like or APP repo URL or GitHub homepage?

src/transforms/code-scanning-alert.ts
```diff
@@ -0,0 +1,100 @@
...
+	})
+}
+
+const getEntityTitle = (ref: string, repoName: string, repoOwner: string, context: Context): Promise<string> => {
+	// ref can either be a branch reference or a PR reference
+	const components = ref.split("/")
+	switch (components[1]) {
+		case "heads": // branch
+			// The branch name may contain forward slashes! Rejoin them
+			return Promise.resolve(components.slice(2).join("/"));
```
KAbakumovAtl, line 31: If branch name is: "issue/abc-124/my-awesome-branch" we'll get "my-awesome-branch" 

I wonder is it what we want?

src/transforms/code-scanning-alert.ts
```diff
@@ -0,0 +1,100 @@
...
+import logger from "../config/logger";
+
+const MAX_STRING_LENGTH = 255;
+
+const getPullRequestTitle = (repoName: string, prId: number, repoOwner: string, context: Context): Promise<string> => {
+	return context.github.pulls.get({
+		owner: repoOwner,
+		repo: repoName,
+		pull_number: prId
+	}).then((response) => {
```
KAbakumovAtl, line 13: According to our codestyle conventions we favour "await" rather than "then". Should change it to await and try/catch?

KAbakumovAtl (commented):
We should add a feature flag I believe. 

Other comments are nits and  questions.

Overall, very good PR. Well done!

src/transforms/code-scanning-alert.ts
TerryAg: We are not splitting slashes on the branch name, but instead on the `ref` string. This `ref` string will look something like:

`refs/heads/<branchname>`

Let's say the branch name is `issue/abc-124/my-awesome-branch`. Then components will be `[refs, heads, issue, abc-124, my-awesome-branch]`. We split up by slash so that we can confirm that the ref is referencing a branch head. But you are right that branches can have slashes, so we need to join back the branch name by joining `components.slice(2)` (grab everything after `heads`)

src/github/code-scanning-alert.ts
TerryAg: I wasn't aware GitHub App supported feature flagging, we should add this yep.

We should be careful to rollout by tenant and not by request (and be careful in our email to accept new permissions)

src/jira/connect.ts
damoster: Ah yep, that's a good point. I'll switch it to the GitHub homepage so that it's more generic in case of other future uses of remote links feature

src/github/code-scanning-alert.ts
```diff
@@ -0,0 +1,20 @@
+import transformCodeScanningAlert from "../transforms/code-scanning-alert";
+import { Context } from "probot/lib/context";
+import logger from "../config/logger";
```
rachellerathbone, line 3: We don't need to import this here. We use `context.log` for all our webhooks.

src/github/code-scanning-alert.ts
```diff
@@ -0,0 +1,20 @@
...
+	const jiraPayload = await transformCodeScanningAlert(context);
+
+	if (!jiraPayload) {
+		logger.info({noop: "no_jira_payload_code_scanning_alert"}, "Halting further execution for code scanning alert since jiraPayload is empty");
+		return;
+	}
+
+	logger.info(`Sending code scanning alert event as Remote Link to Jira: ${jiraClient.baseURL}`);
+	await jiraClient.remoteLink.submit(jiraPayload);
+};
```
rachellerathbone, line 20: We need to emit metrics here. e.g.

```
const { webhookReceived, name, log } = context;

	webhookReceived && emitWebhookProcessedMetrics(
		webhookReceived,
		name,
		log,
		jiraResponse?.status
	);
```

src/transforms/code-scanning-alert.ts
rachellerathbone: Did you miss this change? I'm still seeing .then as opposed to await.

src/transforms/code-scanning-alert.ts
```diff
@@ -0,0 +1,100 @@
...
+	if (action === "closed_by_user" || action === "reopened_by_user") {
+		// These are manual operations done by users and are not associated to a specific Issue.
+		// The webhook contains ALL instances of this alert, so we need to grab the ref from each instance.
+		entityTitlesPromises.push(...alert.instances.map((instance) => getEntityTitle(instance.ref, repository.name, repository.owner.login, context)));
+	} else {
+		// The action is associated with a single branch/PR
+		entityTitlesPromises.push(getEntityTitle(ref, repository.name, repository.owner.login, context));
+	}
+
+	return Promise.all(entityTitlesPromises).then(entityTitles => {
```
rachellerathbone, line 69: This also needs to be updated to use `await`

src/github/code-scanning-alert.ts
```diff
@@ -0,0 +1,28 @@
+import transformCodeScanningAlert from "../transforms/code-scanning-alert";
+import { Context } from "probot/lib/context";
+import { booleanFlag, BooleanFlags } from "../config/feature-flags";
+import {emitWebhookProcessedMetrics} from "./../util/webhook-utils";
+
+export default async (context: Context, jiraClient): Promise<void> => {
```
mboudreau, line 6: No more default usage please :)

src/transforms/code-scanning-alert.ts
```diff
@@ -0,0 +1,93 @@
...
+			return "success"; // green
+		case "dismissed":
+			return "moved"; // yellow
+		default:
+			context.log.info(`Received unknown status from code_scanning_alert webhook: ${status}`);
+			return "default";
+	}
+}
+
+export default async (context: Context): Promise<JiraRemoteLinkData | undefined> => {
```
mboudreau, line 52: Another default export

src/transforms/code-scanning-alert.ts
```diff
@@ -0,0 +1,93 @@
...
+		));
+	} else {
+		// The action is associated with a single branch/PR
+		entityTitles.push(await getEntityTitle(ref, repository.name, repository.owner.login, context));
+	}
+
+	const issueKeys = entityTitles.flatMap((entityTitle) => jiraIssueKeyParser(entityTitle) ?? []);
+	if (issueKeys.length === 0) {
+		return Promise.resolve(undefined);
+	}
```
mboudreau, line 71: `if(!issueKeys.length) { return undefined; }`

src/transforms/code-scanning-alert.ts
```diff
@@ -0,0 +1,93 @@
...
+
+	return Promise.resolve({
```
mboudreau, line 73: You don't have to do `promise.resolve` everywhere, just make the function async and it does it for you

src/transforms/code-scanning-alert.ts
```diff
@@ -0,0 +1,93 @@
...
+		remoteLinks: [{
+			schemaVersion: "1.0",
+			id: `${repository.id.toString()}-${alert.number}`,
```
mboudreau, line 76: don't have to do `toString()` as the interpolation already does it for you

src/transforms/code-scanning-alert.ts
```diff
@@ -0,0 +1,93 @@
...
+			updateSequenceNumber: Date.now(),
+			displayName: `Alert #${alert.number}`,
+			description: alert.rule.description.substring(0, MAX_STRING_LENGTH),
```
mboudreau, line 79: Are we sure that description is always going to be non-null?

src/transforms/code-scanning-alert.ts
```diff
@@ -0,0 +1,93 @@
...
+			url: alert.html_url,
+			type: "security",
+			status: {
+				appearance: transformStatusToAppearance(alert.most_recent_instance.state, context),
+				label: alert.most_recent_instance.state
+			},
+			lastUpdated: alert.created_at,
```
mboudreau, line 86: This should be `alert.updated_at` since it is possible for alerts to update and not only get created

src/jira/client/jira-client.ts
```diff
@@ -421,6 +452,23 @@ const updateIssueKeysFor = (resources, func) => {
...
 };
 
+/**
+ * Runs the mutatingFunc on the association values field for each entity resource
+ * Assumption is that the transformed resource only has one association which is for
+ * "issueIdOrKeys" association.
+ */
+const updateIssueKeyAssociationValuesFor = (resources, mutatingFunc) => {
+	resources.forEach((resource) => {
+		resource.associations[0].values = mutatingFunc(resource.associations[0].values)
```
mboudreau, line 462: We should probably check that associations and non-null and has a length of not 0 or else the whole function will blow up

src/jira/client/jira-client.ts
```diff
@@ -348,6 +367,18 @@ const withinIssueKeyLimit = (resources: { issueKeys: string[] }[]): boolean => {
...
 
+/**
+ * Returns if the max length of the issue key field is within the limit
+ * Assumption is that the transformed resource only has one association which is for
+ * "issueIdOrKeys" association.
+ */
+const withinIssueKeyAssociationsLimit = (resources) => {
+	if (resources == null) return [];
+
+	const issueKeyCounts = resources.map((resource) => resource.associations[0].values.length);
```
mboudreau, line 378: We should check to see if associations are non-null and have a length of not 0

src/transforms/code-scanning-alert.ts
atrigueiro: According to GH docs yes but I've added a `|| undefined` just in case.

src/transforms/code-scanning-alert.ts
atrigueiro: That makes sense however when we were testing locally we didn't get any `updated_at` when the alert was created, so I'll just make the code accept one of the two (payload examples [here](https://hello.atlassian.net/wiki/spaces/PF/pages/1314969066/Innovation+Project+-+Connecting+GitHub+code+scanning+alerts+to+Jira+Issues#Webhook-payload-examples).)

src/jira/client/jira-client.ts
```diff
@@ -348,6 +367,22 @@ const withinIssueKeyLimit = (resources: { issueKeys: string[] }[]): boolean => {
 	return Math.max(...issueKeyCounts) <= ISSUE_KEY_API_LIMIT;
 };
 
+/**
+ * Returns if the max length of the issue key field is within the limit
+ * Assumption is that the transformed resource only has one association which is for
+ * "issueIdOrKeys" association.
+ */
+const withinIssueKeyAssociationsLimit = (resources) => {
```
mboudreau, line 375: there's no typings for the input and output of the function

src/jira/client/jira-client.ts
```diff
@@ -421,6 +456,25 @@ const updateIssueKeysFor = (resources, func) => {
 	return resources;
 };
 
+/**
+ * Runs the mutatingFunc on the association values field for each entity resource
+ * Assumption is that the transformed resource only has one association which is for
+ * "issueIdOrKeys" association.
+ */
+const updateIssueKeyAssociationValuesFor = (resources, mutatingFunc) => {
```
mboudreau, line 464: there's no typings for the input and output of the function

src/jira/client/jira-client.ts
```diff
@@ -348,6 +367,22 @@ const withinIssueKeyLimit = (resources: { issueKeys: string[] }[]): boolean => {
...
+ */
+const withinIssueKeyAssociationsLimit = (resources) => {
+	let issueKeyCounts = [];
+
+	if (resources?.associations?.length > 0) {
+		issueKeyCounts = resources.map((resource) => resource.associations[0].values.length);
+		return Math.max(...issueKeyCounts) <= ISSUE_KEY_API_LIMIT;
+	}
+
+	return issueKeyCounts;
```
mboudreau, line 383: Shouldn't this return `true` and not an array?

src/transforms/transform-code-scanning-alert.ts
```diff
@@ -0,0 +1,93 @@
...
+	const components = ref.split("/")
+	switch (components[1]) {
+		case "heads": // branch
+			// The branch name may contain forward slashes! Rejoin them
+			return Promise.resolve(components.slice(2).join("/"));
+		case "pull": // pull request
+			return await getPullRequestTitle(repoName, parseInt(components[2]), repoOwner, context);
+		default:
+			context.log.error(`Could not interpret reference from code_scanning_alert: ${ref}`);
+			return Promise.resolve("");
```
mboudreau, line 33: just `return "";` will do, however, are you sure you want to continue if you can't find the entity name?

src/transforms/transform-code-scanning-alert.ts
```diff
@@ -0,0 +1,93 @@
...
+
+const getPullRequestTitle = async (repoName: string, prId: number, repoOwner: string, context: Context): Promise<string> => {
+	const response = await context.github.pulls.get({
+		owner: repoOwner,
+		repo: repoName,
+		pull_number: prId
+	});
+
+	if (response.status !== 200) {
+		context.log.warn(`Received error when querying for Pull Request information: ${JSON.stringify(response.data)}`);
```
mboudreau, line 15: don't stringify the data in the message, just add it as data to the log `context.log.warn({response}, "Received error when...")`

src/transforms/transform-code-scanning-alert.ts
atrigueiro: Yup that's fine because if we don't find an entity type we just won't find any issue keys and will return the function on line 70.

src/jira/client/jira-client.ts
atrigueiro: Oops yeah nice catch. Typings would've helped hehe

src/jira/client/jira-client.ts
```diff
@@ -421,6 +454,27 @@ const updateIssueKeysFor = (resources, func) => {
 	return resources;
 };
 
+/**
+ * Runs the mutatingFunc on the association values field for each entity resource
+ * Assumption is that the transformed resource only has one association which is for
+ * "issueIdOrKeys" association.
+ */
+const updateIssueKeyAssociationValuesFor = (resources: JiraRemoteLink[], mutatingFunc: any): JiraRemoteLink[] => {
```
mboudreau, line 462: shouldn't the func be of type `mutatingFunc:(value:string[]) =>  string[]`?

src/jira/client/jira-client.ts
```diff
@@ -421,6 +454,27 @@ const updateIssueKeysFor = (resources, func) => {
...
+	resources?.forEach(resource => {
```
mboudreau, line 463: is it possible that resources is undefined?  if that's the case, the type needs to be `resources?:JiraRemoteLink[]`

src/jira/client/jira-client.ts
```diff
@@ -348,6 +367,20 @@ const withinIssueKeyLimit = (resources: { issueKeys: string[] }[]): boolean => {
 	return Math.max(...issueKeyCounts) <= ISSUE_KEY_API_LIMIT;
 };
 
+/**
+ * Returns if the max length of the issue key field is within the limit
+ * Assumption is that the transformed resource only has one association which is for
+ * "issueIdOrKeys" association.
+ */
+const withinIssueKeyAssociationsLimit = (resources: JiraRemoteLink[]): boolean => {
+	if (!resources) {
```
mboudreau, line 376: if it's possible for resources to be undefined, the type needs to be `resources?:JiraRemoteLink[]`

mboudreau (approved):

src/jira/client/jira-client.ts
atrigueiro: Not really, we're checking for undefined before even calling this function (in [here](https://github.com/atlassian/github-for-jira/pull/712/files#diff-157b18999c0ae6ee36ed20dcfb16e01c5f9574ff54a27df8e5970db851b9bd11R13)) so this is just an extra check.

src/jira/client/jira-client.ts
atrigueiro: Same as above.

src/jira/client/jira-client.ts
atrigueiro: Nope because mutatingFunc can be `dedup` or `truncate` and they both accept any type of array, nothing guarantees those will always be `string[]`

mboudreau: both dedup needs to be a `string[]` input, truncate can be `unknown[]`, but in your case, for this function, it needs to be able to accept a `string[]`

PR #717, opened 2021-09-13
ARC-637 Upgrading to redisx

mboudreau:
Upgrading to redisx to use the latest version and to increase cache size.

### DIFF
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -32,12 +32,13 @@ links:
 cleanup: true
 
 resources:
-  - type: redis
-    name: bottleneck
+  - type: redisx
+    name: cache
     attributes:
-      snapshotRetention: 0
-      config:
-        maxmemory-policy: allkeys-lru
+      version: 6.x
+      size: 500
+      clusterModeEnabled: false
+      transitEncryptionEnabled: true
       dataType:
         - UGC/Label                # name of GitHub org / Jira site
         - PII/IndirectConfidential # name of GitHub org
@@ -329,6 +330,11 @@ environmentOverrides:
           ComparisonOperator: GreaterThanOrEqualToThreshold
           Statistic: Sum
     resources:
+      - type: redisx
+        name: cache
+        attributes:
+          size: 5000
+
       - name: database
         type: postgres-db
         attributes:
--- a/src/config/redis-info.ts
+++ b/src/config/redis-info.ts
@@ -1,8 +1,10 @@
 import Redis from "ioredis";
+import isProd from "../jira/util/isProd";
 
 export default (connectionName: string): Redis.RedisOptions => ({
 	port: Number(process.env.REDISX_CACHE_PORT) || 6379,
 	host: process.env.REDISX_CACHE_HOST || "127.0.0.1",
 	db: 0,
+	tls: isProd() ? { checkServerIdentity: () => undefined } : undefined,
 	connectionName
 });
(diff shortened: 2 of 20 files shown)


github-for-jira.sd.yml
```diff
@@ -32,12 +32,13 @@ links:
 cleanup: true
 
 resources:
-  - type: redis
-    name: bottleneck
+  - type: redisx
```
KAbakumovAtl, line 35: Do we use the same thing in our local dockers? Do we need to do any updates there?

github-for-jira.sd.yml
```diff
@@ -329,6 +330,11 @@ environmentOverrides:
           ComparisonOperator: GreaterThanOrEqualToThreshold
           Statistic: Sum
     resources:
+      - type: redisx
+        name: cache
+        attributes:
+          size: 5000
```
KAbakumovAtl, line 336: We should add an override for staging as well. It seems Micros forces size:200 for all non-prod instances unless you have an override (I'll PM you the link)

KAbakumovAtl (approved):

github-for-jira.sd.yml
```diff
@@ -32,12 +32,13 @@ links:
 cleanup: true
 
 resources:
-  - type: redis
-    name: bottleneck
```
thombergs, line 36: finally getting rid of this confusing name! :)

github-for-jira.sd.yml
mboudreau: already did.

github-for-jira.sd.yml
mboudreau: I've set it to 500 by default, which I really doubt we'll exceed that in staging for the very few people with access to it.

github-for-jira.sd.yml
mboudreau: yarp.

KAbakumovAtl (approved):

atrigueiro (approved):

src/config/redis-info.ts
```diff
@@ -1,14 +1,10 @@
...
-		port: Number(process.env.REDIS_BOTTLENECK_PORT) || 6379,
-		host: process.env.REDIS_BOTTLENECK_HOST || "127.0.0.1",
-		db: 0,
-		connectionName
-	}
+export default (connectionName: string): Redis.RedisOptions => ({
+	port: Number(process.env.REDISX_CACHE_PORT) || 6379,
+	host: process.env.REDISX_CACHE_HOST || "127.0.0.1",
+	db: 0,
+	tls: isProd() ? { checkServerIdentity: () => undefined } : undefined,
```
KAbakumovAtl, line 8: How does it work? If it is prod, then it will be a function, ortherwise undefined?

mboudreau: `isProd` should probably be renamed to `isNodeProd`, it's not specific to our deployment environments, but about if we set `NODE_ENV` to `production` which we do we ddev, stg and prod.  In our local, we do `NODE_ENV=development`.  We don't have TLS enabled on our local environments, hence we disable TLS - we only have it enabled for our deployment environments.

The `checkServerIdentity` is a workaround for ioredis to be able to communicate with AWS elasticache for Redis because AWS uses a host which redis uses IPs which would mismatch and prevent connecting to it.

thombergs (approved):

src/config/redis-info.ts
KAbakumovAtl: Ah got it now. 

>isProd should probably be renamed to isNodeProd
I'd rather suggest to isLocal, so we won't confues two types of prod (no need to do it now)

PR #718, opened 2021-09-13
ARC-468: Use ref instead of sha for getCommit

atrigueiro:
We're seeing a bunch of `Deprecation: [@octokit/rest] "sha" parameter is deprecated for ".repos.getCommit()". Use "ref" instead` warning on our logs. 

`getCommit` now expects a parameter called `ref`.

Screen Shot 2021-09-13 at 3 12 35 pm

All other places we're calling `getCommit` are already using `ref`.


mboudreau (approved):

KAbakumovAtl:
Nice!

KAbakumovAtl (approved):

PR #721, opened 2021-09-14
ARC-627 add metric and log for failed connections

rachellerathbone:
Small PR to add a log and metric to start tracking how frequently our customers run into this issue and what the causes are. Right now we know that uninstalling the app from GH results in this error but there also appear to be other causes which we currently have no visibility into.

Please also see [this PR](https://bitbucket.org/atlassian/fusion-monitoring/pull-requests/293/add-chart-to-track-number-of-failed). Just added one chart to count the number of times this occurs. From there we can use the logs to figure out what the different errors are and, later, improve the UI and experience more for our users.


src/frontend/get-jira-configuration.ts
```diff
@@ -42,18 +42,28 @@ export async function getInstallation(client, subscription) {
 
 		return response.data;
 	} catch (err) {
+		logger.error(
```
KAbakumovAtl, line 45: we need to use logger here from the request `req.log` because this logger has context information (like request id, jira hsotname, etc)

[ ] Use `req.log` instead of creating a new logger in this file

KAbakumovAtl (changes requested):

src/frontend/get-jira-configuration.ts
KAbakumovAtl: We can pass logger to getInstallation as a parameter hence it called from different places in the code.

src/frontend/get-jira-configuration.ts
```diff
@@ -38,22 +35,32 @@ export async function getInstallation(client, subscription) {
 			subscription.repoSyncState?.numberOfSyncedRepos || 0;
 		response.data.jiraHost = subscription.jiraHost;
 
-		response.data.syncStatus === "FAILED" && sendFailedStatusMetrics(id);
+		response.data.syncStatus === "FAILED" && sendFailedStatusMetrics(id, req);
 
 		return response.data;
 	} catch (err) {
+		req.log.error(
```
KAbakumovAtl, line 42: if req is optional this one might fail if it is undefined, `req?.log` should do the thing

KAbakumovAtl: It will be better to jus pass logger instead of the whole request, hence these methods don't use any other stuff from the request.

PR #728, opened 2021-09-15
Add YAML validator to Service Descriptor

atrigueiro:


package.json
```diff
@@ -21,7 +21,9 @@
     "build": "tsc -p ./tsconfig.json",
     "build:release": "tsc -p ./tsconfig.release.json",
     "build:watch": "tsc -w -p ./tsconfig.json",
-    "lint": "eslint . --ext .ts,.tsx",
+		"lint": "run-p lint:*",
+		"lint:ts": "eslint . --ext .ts,.tsx",
+		"lint:yaml": "yamllint github-for-jira.sd.yml",
```
KAbakumovAtl, line 26: That's great! 

The next step would be to run a micros validation command. (Will work only within Atlassian VPN though) 
`atlas micros service validate -f github-for-jira.sd.yml -s github-for-jira -e prod-west`  (need to be run once per each env)

KAbakumovAtl (approved):

package.json
mboudreau: yeah, that's a bit of a problem with the VPN though....

mboudreau (approved):

PR #738, opened 2021-09-16
HOT-96682 Processing push commits immediately

mboudreau:
But only if it has less than 20 commits, which is the amount we call on the github API side

### DIFF
--- a/src/github/push.ts
+++ b/src/github/push.ts
@@ -33,7 +33,7 @@ export default async (context: Context, jiraClient): Promise<void> => {
 
 	// If there's less than 20 commits (the number of commits the github API returns per call), just process it immediately
 	if(payload.commits?.length < 20 && await booleanFlag(BooleanFlags.PROCESS_PUSHES_IMMEDIATELY, true)) {
-		await processPush(context.github, createJobData(payload, jiraHost));
+		await processPush(context.github, createJobData(payload, jiraClient.baseURL));
 		return;
 	}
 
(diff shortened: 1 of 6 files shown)


src/github/push.ts
```diff
@@ -22,14 +23,19 @@ export default async (context: Context, jiraClient): Promise<void> => {
...
+	if (!payload.commits?.length) {
 		context.log(
 			{ noop: "no_commits" },
 			"Halting further execution for push since no commits were found for the payload"
 		);
 		return;
 	}
 
+	// If there's less than 20 commits (the number of commits the github API returns per call), just process it immediately
+	if(payload.commits?.length < 20 && await booleanFlag(BooleanFlags.PROCESS_PUSHES_IMMEDIATELY, true)) {
```
thombergs, line 35: Please set the default value to `false`, which is the current state. We want to explicitly set it to `true` via the feature flag.

thombergs (changes requested):

src/sync/installation.ts
```diff
@@ -240,13 +242,22 @@ export const processInstallation =
 			};
 
 			const execute = async () => {
-				for (const perPage of [20, 10, 5, 1]) {
+				if (await booleanFlag(BooleanFlags.SIMPLER_PROCESSOR, true)) {
```
thombergs, line 245: same here, please set the default to `false` to be safe

src/github/push.ts
thombergs: we can keep it like this, since LaunchDarkly will resolve the value to `false` anyways, as long as the Feature Flag isn't activated, yet

thombergs (approved):

PR #739, opened 2021-09-16
ARC-278 Add SQS for the push queue

KAbakumovAtl:

* Added support for SQS queues
* added the queue for push events

* Added 2 feature flag:
* send-push-events-to-sqs - to control if we send events to sqs
* stop-sending-push-to-redis - to control if we send push events to redis

Both redis and sqs listeners will be enabled at all times

Some things which not there yet:
* Metrics for sqs queue processing/errors
* Timeouts for sqs queue processing

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -82,6 +82,7 @@ The first time you run the app, simply run:
 npm install # installs node modules
 docker-compose up # Spin up docker containers
 npm run db # Creates DBs and initializes tables
+npm run sqs:init #initialises localstack sqs queues
 ```
 
 That's it.  Everything is ran in docker-compose, including redis, postgres, ngrok and the app (main and worker thread).
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -44,6 +44,16 @@ resources:
         - PII/IndirectConfidential # name of GitHub org
         - UGC/PrimaryIdentifier    # references to GitHub entities (commits, pull requests, etc.) and Jira issues
 
+  - type: sqs
+    name: webhooksqs
+    attributes:
+      MaxReceiveCount: 3 # If changing this attribute value, then also change the constant value in MessageRouter
+      MessageRetentionPeriod: 1209600
+      dataType:
+        - UGC/Label                # name of GitHub org / Jira site
+        - PII/IndirectConfidential # name of GitHub org
+        - UGC/PrimaryIdentifier    # references to GitHub entities (commits, pull requests, etc.) and Jira issues
+
   - name: rds
     type: dedicated-rds
     attributes:
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "db:migrate:prod": "sequelize db:migrate --env production",
     "db:dev": "run-s db:create:dev db:migrate:dev",
     "db:test": "run-s db:create:test db:migrate:test",
+		"sqs:init": "aws --endpoint-url=http://localhost:9602 sqs create-queue --queue-name webhooksqs",
     "prepare": "husky install",
     "precommit": "run-p lint build"
   },
@@ -63,6 +64,7 @@
     "@types/rate-limit-redis": "^1.7.1",
     "@types/supertest": "^2.0.11",
     "atlassian-jwt": "^2.0.0",
+    "aws-sdk": "^2.989.0",
     "axios": "^0.21.1",
     "body-parser": "^1.18.3",
     "bottleneck": "^2.19.5",
--- a/src/config/sqs.ts
+++ b/src/config/sqs.ts
@@ -1,0 +1,94 @@
+import AWS from "aws-sdk";
+import Logger from "bunyan"
+import {getLogger} from "./logger"
+import {Message} from "aws-sdk/clients/sqs";
+import envVars from "./env";
+
+// All our queues must be in the same region so we can set a global parameter here
+AWS.config.update({region: envVars.SQS_WEBHOOKSQS_QUEUE_REGION || "us-west-1"});
+
+// Create SQS service client
+const sqs = new AWS.SQS({apiVersion: "2012-11-05"});
+const logger = getLogger("config.sqs")
+
+// Replace with your account id and the queue name you setup
+
+export class SqsQueue {
+	queueUrl: string;
+	queueName: string;
+
+	constructor(queueName: string, queueUrl: string) {
+		this.queueUrl = queueUrl;
+		this.queueName = queueName;
+	}
+
+	sendMessage(payload: any, log?: Logger) {
+		log = log || logger
+		const params = {
+			MessageBody: JSON.stringify(payload),
+			QueueUrl: this.queueUrl
+		};
+		sqs.sendMessage(params, (err, data) => {
+			if (err) {
+				log.error(err, "Error sending SQS message");
+			} else {
+				log.info(`Successfully added message to sqs queue messageId: ${data.MessageId}`);
+			}
+		});
+	}
+
+	deleteMessage(message: Message) {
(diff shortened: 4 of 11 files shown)


CONTRIBUTING.md
```diff
@@ -82,6 +82,7 @@ The first time you run the app, simply run:
 npm install # installs node modules
 docker-compose up # Spin up docker containers
 npm run db # Creates DBs and initializes tables
+npm run sqs:init #initialises localstack sqs queues
```
KAbakumovAtl, line 85: We need to initialise localstack queue before using the local app. I didn't test it with ngrok (only with atlastunnel)

src/config/sqs.ts
```diff
@@ -0,0 +1,105 @@
...
+			QueueUrl: this.queueUrl,
+			MaxNumberOfMessages: 1,
+			VisibilityTimeout: 0,
+			WaitTimeSeconds: 5
+		};
+		sqs.receiveMessage(params, async (err, data) => {
+			try {
+				await this.handleSqsResponse(err, data, messageHandler)
+			} finally {
+				this.listen(messageHandler)
```
KAbakumovAtl, line 97: I need some guidance here `receiveMessage` reads messages once and I've added this call to call listen function again to query the next batch.

src/config/sqs.ts
```diff
@@ -0,0 +1,105 @@
...
+		}
+	}
+
+	listen(messageHandler) {
+		// Setup the receiveMessage parameters
+		const params = {
+			QueueUrl: this.queueUrl,
+			MaxNumberOfMessages: 1,
+			VisibilityTimeout: 0,
+			WaitTimeSeconds: 5
```
KAbakumovAtl, line 91: `WaitTimeSeconds` will prevent us from overloading sqs if there are no messages. It is a delay which receiveMessage will wait if there are no messages on the queue. So we wouldn't query it more often than once per 5 seconds if there are no messages

github-for-jira.sd.yml
```diff
@@ -44,6 +44,16 @@ resources:
         - PII/IndirectConfidential # name of GitHub org
         - UGC/PrimaryIdentifier    # references to GitHub entities (commits, pull requests, etc.) and Jira issues
 
+  - type: sqs
+    name: webhooksqs
+    attributes:
+      MaxReceiveCount: 3 # If changing this attribute value, then also change the constant value in MessageRouter
```
KAbakumovAtl, line 50: /nit I have to remove this copy-pasted comment

src/config/sqs.ts
rachellerathbone: What are we hoping the outcome will be if there's any error?

src/config/sqs.ts
```diff
@@ -0,0 +1,105 @@
...
+			}
+		});
+	}
+
+	executeMessage(message: Message, messageHandler) {
+		const messagePayload = JSON.parse(message.Body);
+		logger.info({messagePayload}, "Sqs message received");
+
+		try {
+			messageHandler({data: messagePayload})
```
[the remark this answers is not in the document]
mboudreau: And this should have an await to wait for the handler to finish since it's just ignoring the output altogether and deleting the message.

src/config/sqs.ts
```diff
@@ -0,0 +1,105 @@
...
+		sqs.deleteMessage(deleteParams, (err, data) => {
+			if (err) {
+				logger.error({err, data}, "Error deleting message from the queue");
+			} else {
+				logger.debug({data}, "Successfully deleted message from queue");
+			}
+		});
+	}
+
+	executeMessage(message: Message, messageHandler) {
```
mboudreau, line 54: You should have a type on the `messageHandler`

package.json
```diff
@@ -63,6 +64,7 @@
     "@types/rate-limit-redis": "^1.7.1",
     "@types/supertest": "^2.0.11",
     "atlassian-jwt": "^2.0.0",
+    "aws-sdk": "^2.989.0",
```
mboudreau, line 67: This is the wrong package, we should use `@aws-sdk/client-sqs`

package.json
```diff
@@ -38,6 +38,7 @@
     "db:migrate:prod": "sequelize db:migrate --env production",
     "db:dev": "run-s db:create:dev db:migrate:dev",
     "db:test": "run-s db:create:test db:migrate:test",
+		"sqs:init": "aws --endpoint-url=http://localhost:9602 sqs create-queue --queue-name webhooksqs",
```
mboudreau, line 41: should probably use AWS CLI package for this, and make it part of the start command as well.

github-for-jira.sd.yml
rachellerathbone: Do we need to address the TODO on [line 149](https://github.com/atlassian/github-for-jira/blob/cc023a76644fbf715b686a2278c46abad1ce60e9/github-for-jira.sd.yml#L149)?

src/config/sqs.ts
mboudreau: This will absolutely spam SQS constantly until messages are there.  I'd recommend we look into using long polling and using something like [sqs-consumer](https://github.com/bbc/sqs-consumer) to handle our requests.  It'll make it much simpler using a well known pattern.

src/config/sqs.ts
```diff
@@ -0,0 +1,105 @@
...
+				await this.handleSqsResponse(err, data, messageHandler)
+			} finally {
+				this.listen(messageHandler)
+			}
+		});
+	}
+}
+
+export const sqsQueues: { [key: string]: SqsQueue } = {
+	push: new SqsQueue("push", envVars.SQS_WEBHOOKSQS_QUEUE_URL || "http://localhost:9602/queue/webhooksqs")
```
mboudreau, line 104: queue should always use the environment variable as the dev should have the default one in their `.env` file and should fail if it's missing.

mboudreau (changes requested):

src/config/sqs.ts
KAbakumovAtl: I was thinking about using this library but after our experience with Bull I'd prefer to reduce amount of third party libs we use.

We already have long polling, currently set to 5 seconds, so we wouldn't spam too much until there are messages on the queue (and it is only on debug level)

src/config/sqs.ts
```diff
@@ -0,0 +1,105 @@
...
+		this.queueName = queueName;
+	}
+
+	sendMessage(payload: any, log?: Logger) {
+		log = log || logger
+		const params = {
+			MessageBody: JSON.stringify(payload),
+			QueueUrl: this.queueUrl
+		};
+		sqs.sendMessage(params, (err, data) => {
```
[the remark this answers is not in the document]
KAbakumovAtl: Can you elaborate on it?

src/config/sqs.ts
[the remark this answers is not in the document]
KAbakumovAtl: I've put `data` property to make it compatible with the handler code we already use for redis queue processing

src/config/sqs.ts
KAbakumovAtl: What change are you suggesting here?

src/config/sqs.ts
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	executeMessage(message: Message, messageHandler) {
+		const messagePayload = JSON.parse(message.Body);
+		logger.info({messagePayload}, "Sqs message received");
+
+		try {
+			messageHandler({data: messagePayload})
+			this.deleteMessage(message)
+		} catch (err) {
```
thombergs, line 61: We need to throw a specific error when hitting a rate limit and then catch it here and re-send the message to SQS with a delay from the rate limit header.

PR #740, opened 2021-09-16
HOT-96682: bump RDS instance class

bgvozdev:
https://instances.vantage.sh/rds/?compare_on=true&selected=db.m5.large,db.r5.xlarge

### DIFF
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -349,7 +349,7 @@ environmentOverrides:
         type: dedicated-rds
         attributes:
           parameters:
-            DBInstanceClass: db.m5.large
+            DBInstanceClass: db.r5.xlarge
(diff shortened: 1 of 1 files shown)


ccrolf (approved):

mboudreau:
Can we just fix the DB table instead?  Won't take too long to do and we won't lose data

bgvozdev:
Hey @mboudreau , I checked the prod database and it doesn't look we have too much huge JSON blobs:

image

Hence the problem might be something else.... Given the current state of affairs, I think it was agreed that, sadly, it was safer to have 1-2 mins of data loss than to tolerate this problem until we can find the root cause and apply the proper fix.

bgvozdev:
UPDATE: looks like our queue was stabilised for some reason... Putting on hold for now

PR #742, opened 2021-09-16
HOT-96682 Adding queue timeout with exponential backoff

mboudreau:
Also logging throttling events from octokit.

### DIFF
--- a/src/worker/main.ts
+++ b/src/worker/main.ts
@@ -30,7 +30,12 @@ function measureElapsedTime(job: Queue.Job, tags) {
 
 const queueOpts: QueueOptions = {
 	defaultJobOptions: {
-		attempts: 3,
+		attempts: 5,
+		timeout: 60000,
+		backoff: {
+			type: "exponential",
+			delay: 3 * 60 * 1000
+		},
 		removeOnComplete: true,
 		removeOnFail: true
 	},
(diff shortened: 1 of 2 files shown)


bgvozdev (approved):

src/worker/main.ts
```diff
@@ -30,7 +30,12 @@ function measureElapsedTime(job: Queue.Job, tags) {
 
 const queueOpts: QueueOptions = {
 	defaultJobOptions: {
-		attempts: 3,
+		attempts: 5,
+		timeout: 60000,
+		backoff: {
+			type: "exponential",
+			delay: 3 * 60 * 1000
+		},
```
thombergs, line 38: can you add a feature flag around this change? Better safe than sorry

thombergs (changes requested):

src/worker/main.ts
```diff
@@ -30,7 +30,12 @@ function measureElapsedTime(job: Queue.Job, tags) {
 
 const queueOpts: QueueOptions = {
 	defaultJobOptions: {
-		attempts: 3,
+		attempts: 5,
+		timeout: 60000,
```
KAbakumovAtl, line 34: Wow, that looks too easy! :) 

Let's test it by simulating a long wait in src/config/enhance-octokit.ts:18

src/worker/main.ts
KAbakumovAtl: I don't think it is feature flaggable, it can be only set on startup

KAbakumovAtl (approved):
Looks great, let's test it by simulating the long wait in the place where we have problems

thombergs (approved):

PR #748, opened 2021-09-16
Increasing DB connection limit and alert level

ccrolf:


mboudreau (approved):

KAbakumovAtl (approved):

KAbakumovAtl:
We decided to keep connection limit at the current level hence increasing it can killl DB (We had problems with CPU consumption when all connections were actively used event with current limit).

PR #752, opened 2021-09-17
Remove bottleneck

thombergs:
Removes the bottleneck configuration completely so that calls to GitHub will not wait and block the queue.

Instead, handle rate limiting errors ourselves by delaying a job until the rate limit is refreshed.

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -240,7 +240,7 @@ router.post(
 	async (request: Request, res: Response): Promise<void> => {
 
 		const queueName = request.body.queue;   // "installation", "push", "metrics", or "discovery"
-		const jobTypes = request.body.jobTypes || []; // any combination of ["active", "delayed", "waiting", "paused"]
+		const jobTypes = request.body.jobTypes || ["active", "delayed", "waiting", "paused"];
 
 		if(!jobTypes.length){
 			res.status(400);
@@ -250,7 +250,7 @@ router.post(
 
 		const queue: Queue = queues[queueName];
 
-		if (queue == undefined) {
+		if (!queue) {
 			res.status(400);
 			res.send(`queue ${queueName} does not exist (available queues: "installation", "push", "metrics", or "discovery"`);
 			return;
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -239,19 +239,26 @@ router.post(
 	elapsedTimeMetrics,
 	async (request: Request, res: Response): Promise<void> => {
 
-		const queueName = request.body.queue;
+		const queueName = request.body.queue;   // "installation", "push", "metrics", or "discovery"
+		const jobTypes = request.body.jobTypes || []; // any combination of ["active", "delayed", "waiting", "paused"]
+
+		if(!jobTypes.length){
+			res.status(400);
+			res.send("please specify the jobTypes field (available job types: [\"active\", \"delayed\", \"waiting\", \"paused\"])");
+			return;
+		}
 
 		const queue: Queue = queues[queueName];
 
 		if (queue == undefined) {
 			res.status(400);
-			res.send(`queue ${queueName} does not exist`);
+			res.send(`queue ${queueName} does not exist (available queues: "installation", "push", "metrics", or "discovery"`);
 			return;
 		}
 
 		// This remove all jobs from the queue. This way,
 		// the whole queue will be drained and all jobs will be readded.
-		const jobs = await queue.getJobs(["active", "delayed", "waiting", "paused"]);
+		const jobs = await queue.getJobs(jobTypes);
 
 		await Promise.all(jobs.map(async (job: Job) => {
 			try {
--- a/src/config/github-api.ts
+++ b/src/config/github-api.ts
@@ -1,35 +1,9 @@
 import { GitHubAPI } from "probot";
 import { getLogger } from "./logger";
-import Redis from "ioredis";
-import Bottleneck from "bottleneck";
-import getRedisInfo from "./redis-info";
-import ehanceOctokit from "./enhance-octokit";
+import enhanceOctokit from "./enhance-octokit";
 import { Options } from "probot/lib/github";
-import { isNodeTest } from "../util/isNodeEnv";
 
-// Just create one connection and share it
-const redisOptions = getRedisInfo("octokit");
-const client = new Redis(redisOptions);
-const connection = new Bottleneck.IORedisConnection({ client });
-
-const logger = getLogger("github.api");
-
-export default (options: Partial<GithubAPIOptions> = {}): GitHubAPI => {
+export default (options: Partial<Options> = {}): GitHubAPI => {
 	options.logger = options.logger || getLogger("config.github-api");
-	options.throttle = {
-		enabled: !isNodeTest(),
-		onRateLimit: (_, options) => logger.warn({ options }, "Request quota exhausted for request"),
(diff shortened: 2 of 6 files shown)


src/worker/main.ts
```diff
@@ -127,6 +128,32 @@ const sentryMiddleware = (jobHandler) => async (job) => {
...
+
+	try {
+		await jobHandler(job);
+	} catch (err) {
+
+		if (err instanceof RateLimitingError) {
+
+			const delay = err.rateLimitReset * 1000 - new Date().getTime();
+
+			logger.warn({ job }, `Rate limiting detected, setting the exponential backoff base to ${delay}`)
```
KAbakumovAtl, line 143: /nit log message and formatting (too many spaces) :(

src/api/index.ts
```diff
@@ -239,19 +239,26 @@ router.post(
...
 
-		const queueName = request.body.queue;
+		const queueName = request.body.queue;   // "installation", "push", "metrics", or "discovery"
+		const jobTypes = request.body.jobTypes || []; // any combination of ["active", "delayed", "waiting", "paused"]
+
+		if(!jobTypes.length){
+			res.status(400);
+			res.send("please specify the jobTypes field (available job types: [\"active\", \"delayed\", \"waiting\", \"paused\"])");
+			return;
+		}
```
thombergs, line 249: need this so we can reschedule only the `active` jobs in the installation queue to "free it up" once

src/config/github-api.ts
```diff
@@ -1,35 +1,9 @@
...
-	// Configure the Bottleneck Redis Client
-	options.bottleneck = options.bottleneck || Bottleneck;
-	options.connection = options.connection || connection;
-
-	return ehanceOctokit(GitHubAPI(options as Options));
-}
-
-interface GithubAPIOptions extends Options {
-	connection?: Bottleneck.IORedisConnection;
-	bottleneck?: typeof Bottleneck;
```
thombergs, line 34: throwing bottleneck over board...

src/api/index.ts
```diff
@@ -239,19 +239,26 @@ router.post(
 	elapsedTimeMetrics,
 	async (request: Request, res: Response): Promise<void> => {
 
-		const queueName = request.body.queue;
+		const queueName = request.body.queue;   // "installation", "push", "metrics", or "discovery"
+		const jobTypes = request.body.jobTypes || []; // any combination of ["active", "delayed", "waiting", "paused"]
```
mboudreau, line 243: wouldn't it be easier to have a default of all of them if none are specified?

mboudreau: You're also not validating the incoming types which might cause an error

src/api/index.ts
```diff
@@ -239,19 +239,26 @@ router.post(
...
+			res.send("please specify the jobTypes field (available job types: [\"active\", \"delayed\", \"waiting\", \"paused\"])");
+			return;
+		}
 
 		const queue: Queue = queues[queueName];
 
 		if (queue == undefined) {
 			res.status(400);
-			res.send(`queue ${queueName} does not exist`);
+			res.send(`queue ${queueName} does not exist (available queues: "installation", "push", "metrics", or "discovery"`);
```
mboudreau, line 255: I'd probably make these options dynamic `Object.keys(queues).join(", ")`

src/api/index.ts
```diff
@@ -239,19 +239,26 @@ router.post(
...
 			return;
 		}
 
 		// This remove all jobs from the queue. This way,
 		// the whole queue will be drained and all jobs will be readded.
-		const jobs = await queue.getJobs(["active", "delayed", "waiting", "paused"]);
+		const jobs = await queue.getJobs(jobTypes);
```
mboudreau, line 261: this might fail, especially if you don't validate job types

src/api/index.ts
```diff
@@ -239,19 +239,26 @@ router.post(
...
+
+		if(!jobTypes.length){
+			res.status(400);
+			res.send("please specify the jobTypes field (available job types: [\"active\", \"delayed\", \"waiting\", \"paused\"])");
+			return;
+		}
 
 		const queue: Queue = queues[queueName];
 
 		if (queue == undefined) {
```
mboudreau, line 253: just `if(!queue)` is simpler

src/sync/installation.ts
```diff
@@ -144,9 +144,7 @@ const updateJobStatus = async (
 			}
 		});
 
-		const delay = Number(process.env.LIMITER_PER_INSTALLATION) || 1000;
-
-		queues.installation.add(job.data, { delay });
+		queues.installation.add(job.data);
```
thombergs, line 147: no idea why there was a 1 second delay for each installation job, with proper rate limiting handling, this shouldn't matter

src/transforms/push.ts
```diff
@@ -73,13 +73,15 @@ export async function enqueuePush(payload: unknown, jiraHost: string, options?:
...
 	return async (job: Job): Promise<void> => {
+		let github
 		try {
-			const github = await app.auth(job.data.installationId);
-			enhanceOctokit(github);
-			await processPush(github, job.data);
+			github = await app.auth(job.data.installationId);
 		} catch (err) {
 			logger.error({ err, job }, "Could not authenticate");
+			return
```
mboudreau, line 81: This should probably throw an error of some kind, or else I think the job will be marked as 'completed' since it didn't fail.

src/worker/main.ts
```diff
@@ -127,6 +128,22 @@ const sentryMiddleware = (jobHandler) => async (job) => {
 	}
 };
 
+const setDelayOnRateLimiting = (jobHandler) => async (job) => {
+	try {
+		await jobHandler(job);
+	} catch (err) {
+		if (err instanceof RateLimitingError) {
+			const delay = err.rateLimitReset * 1000 - new Date().getTime();
```
mboudreau, line 136: I'd probably add a buffer here of a few seconds just to be safe that it doesn't call the API again at the same second that it would allow new requests and to also account for time drift between redis and node

src/worker/main.ts
```diff
@@ -127,6 +128,22 @@ const sentryMiddleware = (jobHandler) => async (job) => {
...
+			logger.warn({ job }, `Rate limiting detected, delaying job by ${delay} ms`)
+			job.opts.backoff = {
```
mboudreau, line 138: This is the correct way to set the delay on a job?  I'm not certain that this saves it to the queue properly...

mboudreau: Also, if you're doing your own retry logic, I'd probably remove the default backoff strategy and attempts and do it all here.  However, since you can't set the number of attempts made, you'll have to add it to the `job.data` property, unless you want to make it that it's impossible for it to ever fail.

mboudreau (changes requested):
few things to consider.

src/api/index.ts
thombergs: yeah, but I trust that the admin knows what they're doing :)

src/transforms/push.ts
KAbakumovAtl: I guess that was the initial intention, if we cant authorise then just skip the job

src/worker/main.ts
thombergs: it's not a nice way of doing it, but it works, and maybe it's even the intended way ...

KAbakumovAtl (approved):

thombergs:
closing in favor of https://github.com/atlassian/github-for-jira/pull/753

PR #755, opened 2021-09-17
adding a disallowlist for installations that are blocking the queues

thombergs:


src/sync/installation.ts
```diff
@@ -167,12 +167,27 @@ const updateJobStatus = async (
 const getEnhancedGitHub = async (app: Application, installationId) =>
 	enhanceOctokit(await app.auth(installationId));
 
+const isBlocked = async (installationId: number): Promise<boolean> => {
+	try {
+		const blockedInstallationsString = await stringFlag(StringFlags.BLOCKED_INSTALLATIONS, "[]");
+		const blockedInstallations: number[] = JSON.parse(blockedInstallationsString);
+		return blockedInstallations.includes(installationId);
+	} catch (e) {
```
bgvozdev, line 175: Should we log something?

src/sync/installation.ts
```diff
@@ -167,12 +167,27 @@ const updateJobStatus = async (
...
+		return false;
```
KAbakumovAtl, line 176: we need to log the error

KAbakumovAtl (dismissed):

bgvozdev (approved):

PR #756, opened 2021-09-17
HOT-96682 Graphql APIs EVERYWHERE

mboudreau:
Drastically reduce the amount of calls done and not hit the rate limit as much.


rachellerathbone (commented):
Jebus this is a huge PR!

I think it looks good but I'd say, given the size, we should definitely have a few people look at it before we approve it.

src/transforms/push.ts
```diff
@@ -133,49 +110,47 @@ export const processPush = async (github: GitHubAPI, payload) => {
...
+				url,
+				message,
+				changedFiles,
+			} = commit;
+
+			// Jira only accepts a max of 10 files for each commit, so don't send all of them
+			// const filesToSend = files.slice(0, 10);
+
+			// merge commits will have 2 or more parents, depending how many are in the sequence
+			const isMergeCommit = parentCount !== 0;
```
rachellerathbone, line 133: I think this is the issue. All commits have 1 parent (main/master). So any commit that is made is flagged as a merge commit.

rachellerathbone (changes requested):

src/frontend/get-jira-configuration.ts
```diff
@@ -64,6 +64,12 @@ export default async (
 	try {
 		const jiraHost = req.session.jiraHost;
 
+		if(!jiraHost) {
+			req.log.warn({jiraHost, req, res}, "Missing jiraHost");
+			res.status(404).send(`Missing Jira Host '${jiraHost}'`);
+			return;
+		}
```
thombergs, line 71: Adding `jiraHost` in the warn log event and the response doesn't make much sense when it's `undefined` anyways, does it?

src/frontend/validations.ts
```diff
@@ -4,8 +4,7 @@ const subdomainRegexp = /^\w(?:[\w-]{0,61}\w)?$/;
 const jiraDomains = ["atlassian.net", "jira.com"];
 
 export const validJiraDomains = (jiraSubdomain: string, jiraDomain: string): boolean =>
-	jiraSubdomain &&
-	jiraDomains.includes(jiraDomain) &&
+	jiraDomains?.includes(jiraDomain) &&
 	subdomainRegexp.test(jiraSubdomain);
```
thombergs, line 8: If `jiraSubdomain` is `undefined`, `subdomainRegexp.test(jiraSubdomain)` will evaluate to `true`, so this changes the logic!

image

src/jira/client/index.ts
```diff
@@ -243,7 +243,7 @@ export default async (
 	gitHubInstallationId: number,
 	logger?: Logger
 ) => {
-	return await getJiraClient(jiraHost, gitHubInstallationId, logger);
```
thombergs, line 246: I'm not sure I like the linting rule that makes us remove "redundant" `await`s . If someone adds some logic to this code after line 246, they might forget adding the `await` back in ...

src/jira/client/index.ts
```diff
@@ -254,13 +254,13 @@ const batchedBulkUpdate = async (
 	data,
 	instance: AxiosInstance,
 	installationId: number,
-	logger:Logger,
+	logger?: Logger,
```
thombergs, line 257: Why should the logger be optional here when it wasn't optional before?

src/frontend/get-jira-configuration.ts
mboudreau: LOL. good find. that was stupid...

src/models/sequelize.ts
```diff
@@ -14,5 +14,5 @@ config.logging = config.disable_sql_logging
 	: (query, ms) => logger.trace({ ms }, query);
 
 export const sequelize = config.use_env_variable
-	? new Sequelize.Sequelize(process.env[config.use_env_variable], config)
+	? new Sequelize.Sequelize(process.env[config.use_env_variable] || "DATABASE_URL", config)
```
thombergs, line 17: Why fall back to "DATABASE_URL"? To make a potential error message better readable?

src/frontend/validations.ts
mboudreau: if `jiraSubdomain` is undefined, then the conditional would be the same as `(undefined &&  subdomainRegexp.test(undefined))` which returns false.  What I've changed is what typescript will compile it at the end, but makes our code more succinct with optional chaining.

`jiraDomains?.includes(jiraDomain)` in TS will compile to `jiraDomains && jiraDomains.includes(jiraDomain)` in JS

src/jira/client/index.ts
mboudreau: yeah, but then the linter would complain that there's a missing await :)

Plus, if you add code after this, you'll have to change the return altogether.

src/jira/client/index.ts
mboudreau: Because the logger being passed into it from `JiraClient` is optional as well.  there were some situations where it would be undefined.

src/models/sequelize.ts
mboudreau: Whoops, that was a good catch.  the strict checks found that `process.env[config.use_env_variable]` could be undefined.  The way I've done it is wrong though, it's been fixed in another commit.

src/services/github/commit.ts
```diff
@@ -0,0 +1,136 @@
...
+
+const logger = getLogger("services.github.commits");
+
+export const getSpecificGithubCommits = async (github: GitHubAPI, params: SpecificCommitParams): Promise<GithubCommit[]> => {
+	if (!params.commitRefs?.length) {
+		return [];
+	}
+
+	try {
+		const response = (await github.graphql(
```
thombergs, line 13: nitpick: maybe put the GraphQL query into a variable like the other ones.

src/sync/queries.ts
```diff
@@ -62,9 +121,15 @@ export const getCommits = `query ($owner: String!, $repo: String!, $per_page: In
     }
   }`;
 
-export const getBranches = `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String) {
+export const getBranches = `query ($owner: String!, $repo: String!, $cursor: String) {
     repository(owner: $owner, name: $repo) {
-      refs(first: $per_page, refPrefix: "refs/heads/", after: $cursor) {
+      refs(first: 20, refPrefix: "refs/heads/", after: $cursor) {
```
thombergs, line 126: Can't we use 100 here, too, like in the other entities? Probably wouldn't do much for the overall sync time, but still ... .

I was going to propose to go even higher than 100, but I understand that's the maximum GitHub allows.

src/frontend/validations.ts
thombergs: yes, but in the original code it tests for `jiraSubdomain` and not for `jiraDomain`!

mboudreau: crap! good find.

src/services/github/commit.ts
mboudreau: In this case, we can't as we're doing string interpolation.

src/sync/queries.ts
mboudreau: I tried, but I kept getting issues because of the history of 100.  A hard graphql crash, I'm not sure why but it's definitely something internal to the API.  Reducing it to 20 * 100 made it much more stable.

src/sync/installation.ts
```diff
@@ -258,22 +262,11 @@ export const processInstallation =
...
 			const execute = async () => {
-				if (await booleanFlag(BooleanFlags.SIMPLER_PROCESSOR, true)) {
-					try {
-						return await processor(github, repository, cursor, 20);
-					} catch (err) {
-						logger.error({ err, job, github, repository, cursor, task }, "Error Executing Task");
-						handleGitHubError(err);
-					}
-				} else {
-					for (const perPage of [20, 10, 5, 1]) {
```
thombergs, line 269: We figured out the reason for this! Sometimes, the GraphQL API returns a "something went wrong with this query", with a suggestion to try again requesting a smaller sub graph. This was the solution to handle that error. When it fails with 20 items per page, try again with 10 and so on. 

This means we have to leave this logic in here (or rewrite it in some way).

This also makes apparent that there is a notable risk in switching to the GraphQL APIs everywhere, so I would very much suggest you split this PR up into multiple smaller PRs, and put the GraphQL change behind a feature flag so that we can do a slow rollout (or at least switch it back when we notice errors).

thombergs (changes requested):

src/sync/installation.ts
mboudreau: yes, already discovered that and talked about it last week, which is why some of the graphql queries has less than 100 items per call to not overload graphql.  There needs to be an investigation as to why that is since graphql rules state we can have up to 500k nodes max in one call, but it seems to struggle with like 1000 nodes?

And there would be no way to feature flag this easily since the response would be too different between to two to be able to reuse the same code.  The service that calls the API returns different data, which means the transformer needs to change too, and a bunch of other things surrounding it.  If this coded was better abstracted, it would make sense, but this would increase the work by considerable amount.  

I hadn't noticed errors with my branch while testing it with the current queries and how they're setup.

mboudreau: Hm, after some consideration, yeah, it could be a bit much in some instances.  I think the best approach is to try to keep 'chunks' at about 100 nodes each to reduce the amount of load on the github APIs.

src/services/github/pull-request-reviews.ts
```diff
@@ -0,0 +1,77 @@
...
+}
+
+export interface GithubCommitFile {
+	path: string;
+	object: {
+		commitResourcePath: string;
+	};
+}
+
+interface PullRequestReviewsParams {
```
rachellerathbone, line 72: Nice work on all the new interfaces.

src/sync/installation.ts
thombergs: It's not just the number of nodes. I think the `MAX_NODE_LIMIT_EXCEEDED` error happens only rarely. But the GitHub GraphQL API returns a bunch of other errors on some queries as well.

We need to put the for loop back in to work around these errors: `for (const perPage of [20, 10, 5, 1])`. This loop is enabled in prod currently, so if you remove the `SIMPLER_PROCESSOR` feature flag, then make the default behavior use this loop, please.

src/config/feature-flags.ts
```diff
@@ -19,7 +19,6 @@ export enum BooleanFlags {
 	PRIORITIZE_PUSHES = "prioritize-pushes",
 	EXPOSE_QUEUE_METRICS = "expose-queue-metrics",
 	PROCESS_PUSHES_IMMEDIATELY = "process-pushes-immediately",
-	SIMPLER_PROCESSOR = "simpler-processor",
```
thombergs, line 22: If we remove this feature flag, please make the default behavior the behavior **with** the for loop (see comment below, as well).

src/github/push.ts
```diff
@@ -35,8 +35,8 @@ export default async (context: Context, jiraClient): Promise<void> => {
 		return;
 	}
 
-	// If there's less than 20 commits (the number of commits the github API returns per call), just process it immediately
-	if(payload.commits?.length < 20 && await booleanFlag(BooleanFlags.PROCESS_PUSHES_IMMEDIATELY, true, jiraClient.baseURL)) {
+	// If there's less than 100 commits (the max number of commits the github API returns per call), just process it immediately
```
thombergs, line 38: This comment seems off. Why should the number of commits the GitHub API returns be relevant in this case? The commits are in the payload of the webhook, so we don't need to call the GitHub API?

src/sync/installation.ts
```diff
@@ -236,215 +233,201 @@ export const isNotFoundError = (
...
+			});
+		}
+
+		logger.info({ job, task: nextTask }, "Starting task");
+
+		const processor = tasks[task];
 
-				throw new Error(`Error processing GraphQL query: installationId=${installationId}, repositoryId=${repositoryId}, task=${task}`);
-			};
+		const handleGitHubError = (err) => {
```
thombergs, line 296: I removed this function in an earlier PR, so this must be a fragment from a merge conflict. I replaced this function with `isRetryableWithSmallerRequest()`, which returns true or false depending on whether we think it's viable to retry the request with a smaller page size.

This needs to be used in conjunction with the for loop that goes through the different page sizes. Please compare with the current `master` again.

src/sync/queries.ts
```diff
@@ -19,40 +38,80 @@ export const getPullRequests = `query ($owner: String!, $repo: String!, $per_pag
...
-                  message
-                  oid
-                  url
-                  ${includeChangedFiles ? "changedFiles" : ""}
+                  abbreviatedOid
+					        oid
+					        message
+					        authoredDate
+					        url
+					        changedFiles
```
thombergs, line 103: This field causes GitHub to return errors. What happened to the logic I built to retry without this field in a certain error (behind the feature flag `RETRY_WITHOUT_CHANGED_FILES`)? There must be a badly resolved merge conflict here.

thombergs (changes requested):

rachellerathbone (approved):

src/sync/installation.ts
mboudreau: The loop had nothing to do with graphql as most of them were using rest and not graphql.  I'd imagine they used the loop specifically to get as much data as possible from the rest API before the rate limit hits.

I'd rather we try to fix it without using the loop as it's a horrible hack that shouldn't be used at all.  For those other graphql issues, we need to log/test/work around them.  From my experience, most of the issues come from having a query that takes too long, hence reducing the amount of nodes returned seems to fix it.

src/config/feature-flags.ts
mboudreau: we want to remove the for loop though...

There's no reason to have it with graphql.

src/github/push.ts
mboudreau: graphql can return 100 nodes, so if there's more than 100 commits, we need to add it to the queue to be processed (think force push or repository move).

I figured it's easier to just use the code we already have to get all commits from said push because the graphql query has all the information we need, while the webhooks is missing several bits of data (like author), that we would need to call the API either way.

src/sync/installation.ts
mboudreau: I don't think we need this functionality at all if we code the graphql API correctly with good error handling.

src/sync/queries.ts
mboudreau: If Github returns an error on a field, which should be optional, this is a Github issue that needs to be addressed.  But we can have workaround because even if the field query does error, it still returns the data up to that point, so we should be able to continue after that point on error.

The other possibility is to not query `changedFiles` until this is fixed on the github side.

ccrolf:
To be revisited if needed

PR #758, opened 2021-09-17
run the deduplication job on an interval

thombergs:
Deduplicating the installation queue has removed the blockage in the queue. Syncs are working again 🥳 
image

However, there must still be a bug / configuration issue somewhere that causes jobs to be processed twice, which would explain why the queue explodes like this (each job adds a new job to the queue and if it gets processed twice, it will add 2 new jobs, which will add more jobs and so on).

To mitigate this issue until we found the cause for it, this PR adds code to run the deduplication job every couple of minutes.

### DIFF
--- a/src/worker/deduplicate-installations.ts
+++ b/src/worker/deduplicate-installations.ts
@@ -1,0 +1,32 @@
+import { queues } from "./main";
+import { getLogger } from "../config/logger";
+
+const logger = getLogger("deduplicate-installations");
+
+export const deduplicateInstallations = async() : Promise<number> => {
+	// This remove all jobs from the queue. This way,
+	// the whole queue will be drained and all jobs will be readded.
+	const jobs = await queues.installation.getJobs(["active", "delayed", "waiting", "paused"]);
+	const foundInstallationIds = new Set<number>();
+	const duplicateJobs = [];
+
+	// collecting duplicate jobs per installation
+	for (const job of jobs) {
+		if (!job) {
+			continue;
+		}
+		if (foundInstallationIds.has(job.data.installationId)) {
+			duplicateJobs.push(job);
+		} else {
+			foundInstallationIds.add(job.data.installationId);
+		}
+	}
+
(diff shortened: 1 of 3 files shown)


src/worker/deduplicate-installations.ts
```diff
@@ -0,0 +1,32 @@
...
+	// the whole queue will be drained and all jobs will be readded.
+	const jobs = await queues.installation.getJobs(["active", "delayed", "waiting", "paused"]);
+	const foundInstallationIds = new Set<number>();
+	const duplicateJobs = [];
+
+	// collecting duplicate jobs per installation
+	for (const job of jobs) {
+		if (!job) {
+			continue;
+		}
```
thombergs, line 17: added a null check, because if `getJobs` doesn't find any jobs, it seems to return an array with a `null` in it :/

bgvozdev (approved):

src/worker/main.ts
```diff
@@ -174,12 +175,21 @@ const sendQueueMetrics = async () => {
...
 export const start = (): void => {
 	initializeSentry();
 
 	// exposing queue metrics at a regular interval
 	setInterval(sendQueueMetrics, 60000);
 
+	// Run the deduplication job every 8 to 12 minutes (spreading it out a bit to make it less likely that two nodes are
+	// running it at exactly the same time). It should be less than 15 minutes, because if the installation queue is
+	// blocked for 15 minutes, sync jobs will begin to go into the "stalled" state.
+	setInterval(deduplicateInstallations, randomInteger(8 * 60 * 1000, 12 * 60 * 1000));
```
ccrolf, line 191: Is this run per node instance? It that's one per core per ec2 instance, that can be quite a lot. Can we e.g. add an FF in the dedupe method, making it a no-op with X% probability, meaning we can scale the risk of concurrent dedupes

thombergs:
closing. replaced by an external cron job that triggers the deduplication regularly

PR #759, opened 2021-09-18
fix null error

thombergs:


src/api/index.ts
```diff
@@ -260,7 +264,7 @@ router.post(
 			job.remove();
 		}));
 
-		res.send(`${duplicateJobs.length} duplicate jobs removed.`);
+		res.send(`${duplicateJobs.length} duplicate jobs killed with fire.`);
```
thombergs, line 267: as per @bgvozdev's request

bgvozdev (approved):

PR #761, opened 2021-09-21
ARC-653 and ARC-488 webhook metrics

rachellerathbone:
We now note the time a webhook was received to the time of completion so we have better insight into processing times for each webhook. I also included a status code for the response of each webhook. 

Also see this PR to update our metrics: https://bitbucket.org/atlassian/fusion-monitoring/pull-requests/293


src/config/enhance-octokit.ts
```diff
@@ -61,7 +61,7 @@ const instrumentRequests = (octokit: GitHubAPI) => {
 			const tags = {
 				path: extractPath(options.url),
```
KAbakumovAtl, line 62: We need to remove all path parameters from this url. Otherwise it will explode the metric cardinality.

If we can't get this information from the octokit options we need to remove this tag at all

src/worker/main.ts
```diff
@@ -82,8 +82,12 @@ Object.keys(queues).forEach((name) => {
 	});
 
 	queue.on("failed", async (job) => {
-		logger.error({ job, queue: name }, "Job failed");
+		const failedReason = job.failedReason || "failed";
+		logger.error({ job, queue: name, failedReason }, "Job failed");
 		measureElapsedTime(job, { queue: name, status: "failed" });
+
+		const tags = [`queue:${name}`, `failedReason:${failedReason}`];
```
KAbakumovAtl, line 89: Failed reason might contain parameters which potentially explode the cardinality.

We need to remove this tag, it ie enough to have this information in the log only

src/worker/main.ts
```diff
@@ -92,9 +96,8 @@ Object.keys(queues).forEach((name) => {
 		Sentry.setTag("queue", name);
 		Sentry.captureException(err);
 
-		const tags = [`name:${name}`];
-
-		statsd.increment("queue_error", tags);
+		const tags = [`queue:${name}`, `err:${err}`];
+		statsd.increment(metricError.queueError, tags);
```
KAbakumovAtl, line 100: Err might contain parameters which potentially explode the cardinality. (like installationId or jobId)

We need to remove this tag, it ie enough to have this information in the log only

src/config/enhance-octokit.ts
KAbakumovAtl: - [ ] remove path from tags

src/github/deployment.ts
```diff
@@ -1,14 +1,20 @@
...
+			{ noop: "no_jira_payload_deployment" },
+			"Halting further execution for deployment since jiraPayload is empty"
+		);
 		return;
 	}
 
-	context.log(`Sending deployment info to Jira: ${jiraClient.baseURL}`)
-	await jiraClient.deployment.submit(jiraPayload);
+	context.log(`Sending deployment info to Jira: ${jiraClient.baseURL}`);
+	const webhook = await jiraClient.deployment.submit(jiraPayload);
```
KAbakumovAtl, line 17: /nit variable naming. It is not really a webhook, it is a Jira call result

src/jira/client/index.ts
```diff
@@ -182,7 +182,7 @@ async function getJiraClient(
 						});
 					}
 
-					await batchedBulkUpdate(
+					return await batchedBulkUpdate(
```
KAbakumovAtl, line 185: Am I correctly understand you added return value so the `update` method will return a response object?

If we are adding return value here, we should add it everywhere in this file for consistency.

- [ ] make return values consistent for Jira API calls

src/transforms/push.ts
```diff
@@ -58,12 +59,14 @@ export function createJobData(payload, jiraHost: string) {
...
 	}
+
 	return {
 		repository,
 		shas,
 		jiraHost,
 		installationId: payload.installation.id,
-		webhookId: payload.webhookId || "none"
+		webhookId: payload.webhookId || "none",
+		webhookReceived: payload.webhookReceived || "none"
```
KAbakumovAtl, line 69: If webhookReceived is Date then defaulting it to "none" isn't good hence it is a "string" data type. 

@mboudreau  What would you recommend? Would it be OK to have undefined here? If we serialise it and put on redis queue would it be still undefined when we deserialise?

src/util/webhooks.ts
```diff
@@ -0,0 +1,30 @@
...
+import { metricWebhooks } from "../config/metric-names";
+
+export const getCurrentTime = () => new Date();
+
+export const calculateProcessingTimeInSeconds = (
+	webhookReceivedTime: Date,
+	webhookName: string,
+	status?: number,
+): number => {
+	const logger = getLogger("webhookProccessingTime");
```
KAbakumovAtl, line 12: We need to use logger from the context from where it called. Otherwise we are loosing context parameters
- [ ] Pass context logger and use it here (req.log or context.log)

src/util/webhooks.ts
```diff
@@ -0,0 +1,30 @@
...
+	const timeToProcessWebhookEvent =
+		(getCurrentTime().getTime() - webhookReceivedTime.getTime()) / 1000;
```
KAbakumovAtl, line 14: Milliseconds would be fine here. We use milliseconds in other metrics so better to stay consistent

- [ ] use milliseconds instead of seconds

src/transforms/push.ts
```diff
@@ -172,7 +178,10 @@ export const processPush = async (github: GitHubAPI, payload) => {
 				updateSequenceId: Date.now()
 			};
 
-			await jiraClient.devinfo.repository.update(jiraPayload);
+			const webhook = await jiraClient.devinfo.repository.update(jiraPayload);
+			const webhookName = payload.name || "none";
+
+			calculateProcessingTimeInSeconds(webhookReceived, webhookName, webhook?.status);
```
KAbakumovAtl, line 184: If webhookReceived is "none" it will fail here. We need to add a check if we have time here

src/config/enhance-octokit.ts
rachellerathbone: This was added by a GH dev 2 years ago. Do you know if it's caused any issues with our metrics?

src/worker/main.ts
```diff
@@ -82,8 +82,12 @@ Object.keys(queues).forEach((name) => {
 	});
 
 	queue.on("failed", async (job) => {
-		logger.error({ job, queue: name }, "Job failed");
+		const failedReason = job.failedReason || "failed";
+		logger.error({ job, queue: name, failedReason }, "Job failed");
 		measureElapsedTime(job, { queue: name, status: "failed" });
+
+		const tags = [`queue:${name}`];
+		statsd.increment(metricError.queueFailed, tags);
```
KAbakumovAtl, line 90: Nice, great to have these metrics here

src/config/enhance-octokit.ts
rachellerathbone: Just checked and we aren't using it in our metrics anyway. Will remove.

KAbakumovAtl:
@atrigueiro  Just made a PR for listening for repo deletions, so we need to add metrics to the delete endpoint if she merge it first 
https://github.com/atlassian/github-for-jira/pull/767

src/jira/client/index.ts
rachellerathbone: Where is it missing from? Just had a look at the file and it looks like every method either has a `return` statement or is being returned by `=>`

src/transforms/push.ts
rachellerathbone: Changed it to undefined. Added checks to see if `webhookReceived` exists before calling `calculateProcessingTimeInSeconds`

src/jira/client/index.ts
KAbakumovAtl: Yes you are right, I didn't notice that there are no brackets

KAbakumovAtl (approved):
Let's make sure that we test it thoroughly.

Particularly the `push` path. Would be great to add some unit tests as well (to test that it works well when both webhookReceived set and not)

rachellerathbone:
@KAbakumovAtl I was going to try working on tests for this today but, given the state of our app atm, think I'll need to focus on putting out fires. Created a ticket in our backlog instead.

PR #762, opened 2021-09-21
fixing null-reference that can cause a sync to fail

thombergs:
From the logs, this is the cause for some syncs to fail.

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -134,7 +134,7 @@ const updateJobStatus = async (
 		}
 	});
 
-	if (edges.length > 0) {
+	if (edges?.length) {
 		// there's more data to get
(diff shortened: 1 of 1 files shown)


src/sync/installation.ts
```diff
@@ -122,7 +122,7 @@ const updateJobStatus = async (
 		return;
 	}
 
-	const status = edges.length > 0 ? "pending" : "complete";
```
thombergs, line 125: `edges` was `null` in some cases. This would throw an error that is not caught anywhere, which means no follow-up job is queued and the sync is effectively "stalled". After 15 minutes of no updates, when the installation is loaded from the database again, the sync status will be set to `FAILED` due to inactivity.

rachellerathbone (approved):

PR #767, opened 2021-09-22
ARC-658: Delete dev info data when repo is deleted from GitHub.

atrigueiro:
https://developer.atlassian.com/cloud/jira/software/rest/api-group-development-information/#api-devinfo-0-10-repository-repositoryid-delete

### DIFF
--- a/app.yml
+++ b/app.yml
@@ -40,7 +40,7 @@ default_events:
 - pull_request_review_comment
 - push
 # - release
-# - repository
+- repository
 # - repository_import
 # - status
 # - team
--- a/src/github/index.ts
+++ b/src/github/index.ts
@@ -10,6 +10,7 @@ import webhookTimeout from "../middleware/webhook-timeout";
 import statsd from "../config/statsd";
 import { metricWebhooks } from "../config/metric-names";
 import { Application } from "probot";
+import { deleteRepository } from "./repository";
 
 export default (robot: Application) => {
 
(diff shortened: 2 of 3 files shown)


app.yml
```diff
@@ -40,7 +40,7 @@ default_events:
 - pull_request_review_comment
 - push
 # - release
-# - repository
+- repository
```
KAbakumovAtl, line 43: Would it require a new permissions from users?

KAbakumovAtl (approved):
Nice. Simple and elegant solution

src/github/index.ts
```diff
@@ -54,4 +55,6 @@ export default (robot: Application) => {
 
 	robot.on("create", middleware(createBranch));
 	robot.on("delete", middleware(deleteBranch));
+
+	robot.on("repository.deleted", middleware(deleteRepository));
```
rachellerathbone, line 59: I just tested deleting a repository and the webhook event I saw was `installation_repositories.removed`

rachellerathbone (changes requested):

KAbakumovAtl:
@rachellerathbone just made a PR for adding metrics https://github.com/atlassian/github-for-jira/pull/761 so we need to add metrics to the delete endpoint here if Rachelle merge it first

app.yml
atrigueiro: Nope, we're not adding a new permission, just subscribing to a new event.

rachellerathbone (approved):

src/github/index.ts
atrigueiro: Yeah you'll need to make your local app listen to the `repository` event now and that should work.

PR #768, opened 2021-09-22
Changing external proxy for whitelist proxy

mboudreau:
External proxy goes to the public internet before coming back to Jira.  By using the whitelist proxy, we can hit the jira API directly within our VPC which also means it's not blocked by atlassian's ip allowlist.

### DIFF
--- a/src/config/env.ts
+++ b/src/config/env.ts
@@ -26,10 +26,9 @@ if (env.error && nodeEnv !== EnvironmentEnum.production) {
 	throw env.error;
 }
 
-const getProxyFromEnvironment = (): string => {
-	const proxyHost = process.env.EXTERNAL_ONLY_PROXY_HOST;
-	const proxyPort = process.env.EXTERNAL_ONLY_PROXY_PORT;
-	return proxyHost && proxyPort ? `http://${proxyHost}:${proxyPort}` : undefined;
+const getProxyFromEnvironment = (): string | undefined => {
+	const proxy = process.env.WHITELIST_PROXY;
+	return proxy ? `http://${proxy}` : undefined;
 };
 
 // TODO: Make envvars dynamic


src/config/env.ts
```diff
@@ -26,10 +26,9 @@ if (env.error && nodeEnv !== EnvironmentEnum.production) {
 	throw env.error;
 }
 
-const getProxyFromEnvironment = (): string => {
-	const proxyHost = process.env.EXTERNAL_ONLY_PROXY_HOST;
-	const proxyPort = process.env.EXTERNAL_ONLY_PROXY_PORT;
-	return proxyHost && proxyPort ? `http://${proxyHost}:${proxyPort}` : undefined;
+const getProxyFromEnvironment = (): string | undefined => {
+	const proxy = process.env.WHITELIST_PROXY;
```
KAbakumovAtl, line 30: Can you give some details on how it works?

1. This env var is set by Micros, right? Do we need to configure anything in the Micros descriptor?

2. Are we using this proxy for all outbound calls, correct? Would it change the IPs for calls we make to GitHub? If yes, would it require any changes in the IPs whitelist on the GitHub side?

thombergs:
I'd like to test this first with a Jira site that has an IP allowlist configured. Do we even have a problem with IP allowlists in Jira currently? If yes, and this solves this, and the GitHub IP allowlist still works the same, then let's merge it.

mboudreau:
@thombergs I'd like to test that too, but not sure how to accomplish this.  I heard through the grapevine that some of our customers are using IP allowlisting on the Atlassian side, but I'm not certain who.

That being said, even without that, we should still merge it as it should make our calls to Atlassian services quicker and more stable.

mboudreau:
I've deployed this to staging and confirmed that the whitelist proxy is working as intended.

PR #769, opened 2021-09-23
Fixing repoSyncState to not redo whole JSONB object every update

mboudreau:
This should drastically reduce DB load for updating the JSONB object.  The problem stems from sequelize that can't suppose JSON objects easily: https://github.com/sequelize/sequelize/issues/4387

This is a workaround to manually set every field individually.


src/models/subscription.ts
```diff
@@ -212,10 +211,25 @@ export default class Subscription extends Sequelize.Model {
...
+	}
+
+	private setRepoSyncState(obj: Partial<Record<keyof RepoSyncState, unknown>>, parentKeys: string[], options?: SetOptions): void {
+		Object.entries(obj).forEach(([key, value]) => {
+			if (_.isPlainObject(value) && !_.isEmpty(value)) {
+				this.setRepoSyncState(value as Record<string, unknown>, parentKeys.concat([key]), options);
+			} else {
+				this.set(parentKeys.join(".") as any, value, options);
+			}
+		});
```
KAbakumovAtl, line 232: So this method recursively set values to the sync state jsonb object? So we are doing it to avoid it being parsed by postgres?

KAbakumovAtl:
I'd suggest to use a feature flag for a change like this. The change doesn't look to complicated, but definitely has performance impact on the critical path. Also in this situation it is very easy to feature flag.

mboudreau:
@KAbakumovAtl Yep, that's a good point.  Very easy to FF :)
I'll add that in.

src/models/subscription.ts
mboudreau: yeah, something in sequelize not liking setting a JSON object to be saved, but if you do it for each individual field in the JSON like `repoSyncState.repos.foo` with value of `bar`, it seems to work alright.  I think it's how it generate the SQL which is different and let postgres handle the individual field value change :)

mboudreau:
Closing this as we have a workaround of doing `jsonb_set` query directly to the DB.  Need to fix this properly by creating a new DB table.

cincodenada:
I came from Sequelize's issue here - there's a much simpler way to communicate to Sequelize that you updated the field even though they are strictly equal using `.changed()`, [as documented in the API](https://sequelize.org/master/class/lib/model.js~Model.html#instance-method-changed):

```
const mdl = await MyModel.findOne();
mdl.myJsonField.a = 1;
console.log(mdl.changed()) => false
mdl.save(); // this will not save anything
mdl.changed('myJsonField', true);
console.log(mdl.changed()) => ['myJsonField']
mdl.save(); // will save
```

PR #771, opened 2021-09-23
ARC-666: make lockDuration dependent on the timeout to avoid a job being triggered twice

thombergs:
We increased the timeout of the queue jobs, but the `lockDuration` is still the default (30 seconds). That means that after 30 seconds, a message will be "unlocked" and picked up by another worker to be processed again. Since each job adds a follow-up job to the queue, a job that is processed twice will add two follow-up-jobs to the queue, which might again add new jobs and so on, building up the waiting queue.

This PR should hopefully fix this.

### DIFF
--- a/src/worker/main.ts
+++ b/src/worker/main.ts
@@ -41,8 +41,8 @@ const getQueueOptions = (timeout: number): QueueOptions => {
 			removeOnFail: true
 		},
 		settings: {
-			// lockDuration must be smaller than the timeout, otherwise a job might be processed twice
-			lockDuration: timeout - 500
+			// lockDuration must be greater than the timeout, so that it doesn't get processed again prematurely
+			lockDuration: timeout + 500
 		},
 		redis: getRedisInfo("bull"),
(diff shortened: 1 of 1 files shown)


src/worker/main.ts
```diff
@@ -40,6 +40,10 @@ const getQueueOptions = (timeout: number): QueueOptions => {
 			removeOnComplete: true,
 			removeOnFail: true
 		},
+		settings: {
+			// lockDuration must be greater than the timeout, so that it doesn't get processed again prematurely
+			lockDuration: timeout + 500
```
KAbakumovAtl, line 45: What's the default?

KAbakumovAtl (approved):

src/worker/main.ts
thombergs: The default is 30 seconds. And we set the timeout to all queues to 60 seconds or more. That means if a message took longer than 30 seconds, it wouldn't yet time out, but it would be freed up for the next worker to execute again.

PR #773, opened 2021-09-24
ARC-671: retry loading commits on error "The changedFiles count for this commit is unavailable"

thombergs:
We're seeing a bunch of these errors in the logs, causing backfills to fail. So, we'll catch this error and retry again without asking for the `changedFiles` field.

The `changedFiles` feels can be undefined, it's only used in 2 places and those have a null-check.

### DIFF
--- a/src/sync/commits.ts
+++ b/src/sync/commits.ts
@@ -20,16 +20,20 @@ export default async (github: GitHubAPI, repository, cursor, perPage: number) =>
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any
 	let commitsData: any = {};
 
-	// TODO: fix typings for graphql
-	try {
-		commitsData = (await github.graphql(getCommitsQuery(true), {
+	const getCommits = async (includeChangedFiles: boolean) => {
+		await github.graphql(getCommitsQuery(includeChangedFiles), {
 			owner: repository.owner.login,
 			repo: repository.name,
 			per_page: perPage,
 			cursor,
 			default_ref: refName
 			// eslint-disable-next-line @typescript-eslint/no-explicit-any
-		})) as any;
+		})
+	}
+
+	// TODO: fix typings for graphql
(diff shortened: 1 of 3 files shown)


src/sync/commits.ts
```diff
@@ -13,15 +17,44 @@ export default async (github: GitHubAPI, repository, cursor, perPage: number) =>
...
+		}
+
+		// According to the logs, GraphQL queries sometimes fail because the "changedFiles" field is not available.
+		// In this case we just try again, but without asking for the changedFiles field.
+
+		logger.info("retrying without changedFiles");
+
+		const changedFilesErrors = err.errors?.filter(e => e.message?.includes("The changedFiles count for this commit is unavailable"));
+		if (changedFilesErrors.length) {
+			commitsData = (await github.graphql(getCommitsQuery(false), {
```
rachellerathbone, line 46: Is there a way to pull this out into a function so we're not repeating the same logic?

src/sync/queries.ts
```diff
@@ -52,7 +52,7 @@ export const getCommits = `query ($owner: String!, $repo: String!, $per_page: In
                   message
                   oid
                   url
-                  changedFiles
+                  ${includeChangedFiles ? "changedFiles" : ""}
```
rachellerathbone, line 55: Has this been tested? Wouldn't graphql shit itself if it tried to query target."" ?

rachellerathbone (changes requested):

src/sync/commits.ts
thombergs: done

src/sync/queries.ts
thombergs: GraphQL doesn't care about empty lines. Tested with the GraphQL explorer and with a local sync job

rachellerathbone (approved):

PR #777, opened 2021-09-27
Arc-671 fix missing return

thombergs:


rachellerathbone (approved):

src/sync/commits.ts
```diff
@@ -21,7 +21,7 @@ export default async (github: GitHubAPI, repository, cursor, perPage: number) =>
 	let commitsData: any = {};
 
 	const getCommits = async (includeChangedFiles: boolean) => {
-		await github.graphql(getCommitsQuery(includeChangedFiles), {
+		return github.graphql(getCommitsQuery(includeChangedFiles), {
```
thombergs, line 24: Missing return statement caused the value to be undefined, causing the sync error in staging :(.

rachellerathbone (approved):

KAbakumovAtl (approved):

src/sync/commits.ts
atrigueiro: We still need to await before returning, right? Guess it needs to be `return await`

thombergs: No, since this is the return value of an `async` function, we don't need an await here. But yeah, it's confusing. I don't like the linter rule that made me remove the `await` here ...

PR #779, opened 2021-09-27
Arc 675 fix sync error

thombergs:
This PR addresses the error "Something went wrong while executing your query" that GitHub sometimes returns. It includes this error into the list of "retryable errors" that are going to be retried with a smaller page size to reduce the number of nodes that the GitHub API has to return.

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -178,6 +178,27 @@ const isBlocked = async (installationId: number): Promise<boolean> => {
 	}
 }
 
+/**
+ * Determines if an an error returned by the GitHub API means that we should retry it
+ * with a smaller request (i.e. with fewer pages).
+ * @param err the error thrown by Octokit.
+ */
+export const isRetryableWithSmallerRequest = (err): boolean => {
+	if (err.errors) {
+
+		const retryableErrors = err.errors.filter(
+			(error) => {
+				return "MAX_NODE_LIMIT_EXCEEDED" == error.type
+					|| error.message?.startsWith("Something went wrong while executing your query")
+			}
+		);
+
+		return retryableErrors.length;
+	} else {
+		return false;
+	}
+};
+
 // TODO: type queues
 export const processInstallation =
 	(app: Application, queues) =>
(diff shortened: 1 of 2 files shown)


src/sync/installation.ts
```diff
@@ -178,6 +178,27 @@ const isBlocked = async (installationId: number): Promise<boolean> => {
...
+				return "MAX_NODE_LIMIT_EXCEEDED" == error.type
+					|| error.message?.startsWith("Something went wrong while executing your query")
+			}
+		);
+
+		return retryableErrors.length;
+	} else {
+		return false;
+	}
+};
```
thombergs, line 200: The function `isRetryableWithSmallerRequest()` was previously the function `handleGitHubError()`. I refactored it to be more meaningful, added a test for it, and added the "Something went wrong" to the list of retryable errors.

src/sync/installation.ts
```diff
@@ -240,38 +261,37 @@ export const processInstallation =
 
 			const processor = tasks[task];
 
-			const handleGitHubError = (err) => {
```
thombergs, line 243: Moved this method out to make it testable and renamed it to `isRetryableWithSmallerRequest()`

src/sync/installation.ts
```diff
@@ -240,38 +261,37 @@ export const processInstallation =
...
 				if (await booleanFlag(BooleanFlags.SIMPLER_PROCESSOR, true)) {
-					try {
-						return await processor(github, repository, cursor, 20);
-					} catch (err) {
-						logger.error({ err, job, github, repository, cursor, task }, "Error Executing Task");
-						handleGitHubError(err);
-					}
+
+					// just try with one page size
+					return await processor(github, repository, cursor, 20);
```
thombergs, line 268: The `SIMPLER_PROCESSOR` flag is `true` in prod, currently. I removed the `try/catch` and `handleGitHubError()` call here, because we're not iterating through different page sizes and `handleGitHubError()` would just have re-thrown the error anyways.

src/sync/installation.ts
```diff
@@ -240,38 +261,37 @@ export const processInstallation =
...
+								cursor,
+								task
+							}, `Error processing job with page size ${perPage}, retrying with next smallest page size`);
+							if (isRetryableWithSmallerRequest(err)) {
+								// error is retryable, retrying with next smaller page size
+								continue;
+							} else {
+								// error is not retryable, re-throwing it
+								throw err;
+							}
```
thombergs, line 291: This is the main change. Made the logic around the error handling a lot easier to understand, hopefully. This code is covered with the `SIMPLER_PROCESSOR` flag, so this change will only take effect, once I change the feature flag back to `false`.

src/sync/installation.ts
KAbakumovAtl, line 268: Are we going to use `isRetryableWithSmallerRequest` in `processor` function? 

It seems we did before, but now we don't

thombergs: No, it was only used here.

KAbakumovAtl (approved):

PR #780, opened 2021-09-27
ARC-648 Not Found Sync Failures

rachellerathbone:
This PR is to account for one of the causes behind failed syncs. See the `HttpError: Not Found` not entry in [this doc](https://hello.atlassian.net/wiki/spaces/PF/pages/1341394025/Investigating+causes+for+sync+failures) for further information on the 2 not found error responses.

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -124,8 +124,6 @@ const updateJobStatus = async (
 
 	const status = edges?.length ? "pending" : "complete";
 
-	logger.info("STATUS: ", status)
-
 	logger.info({ job, task, status }, "Updating job status");
 
 	await subscription.updateSyncState({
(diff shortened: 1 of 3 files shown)


src/sync/installation.ts
```diff
@@ -178,6 +178,28 @@ const isBlocked = async (installationId: number): Promise<boolean> => {
...
+// Checks if parsed error type is NOT_FOUND / status is 404 which come from 2 different sources
+// - GraphqlError: https://github.com/octokit/graphql.js/tree/master#errors
+// - RequestError: https://github.com/octokit/request.js/blob/5cef43ea4008728139686b6e542a62df28bb112a/src/fetch-wrapper.ts#L77
+export const handleNotFoundErrors = (
+	err: any,
+	job: any,
+	nextTask: Task
+): boolean | undefined => {
+	const isNotFoundErrorType =
+		err?.errors && err.errors?.filter((error) => error.type === "NOT_FOUND");
```
bgvozdev, line 190: shall we also check if err is defined?

src/sync/installation.ts
```diff
@@ -178,6 +178,28 @@ const isBlocked = async (installationId: number): Promise<boolean> => {
...
+
+	const isNotFoundError = isNotFoundErrorType?.length > 0 || err?.status === 404;
```
bgvozdev, line 192: same

bgvozdev (approved):

thombergs (approved):

src/sync/installation.ts
rachellerathbone: The ? checks if these values exist. I have tests that pass when a null value/string/empty string is passed instead of the error object 😄

PR #783, opened 2021-09-28
ARC653 metrics... take two

rachellerathbone:
Added some tests and a FF to make sure webhook received metrics works as expected. Have tested in dev and see metrics coming through for the 3 new/fixed charts [here](https://atlassian.signalfx.com/#/dashboard/E_NuSXDA0AA?groupId=E_NuSi3A0AU&configId=E_NuShaAwAA&startTime=-12h&endTime=Now&variables%5B%5D=Deployment%20Environment%20Type%3Ddeployment_env_type:&variables%5B%5D=AWS%20Environment%20Type%3Daws_tag_environment_type:&variables%5B%5D=Micros%20Environment%20Type%3Denvironment_type:%5B%22dev%22%5D). Feel free to test in dev and to confirm you can see metrics. 

Related PR: https://bitbucket.org/atlassian/fusion-monitoring/pull-requests/296


src/github/middleware.ts
```diff
@@ -62,38 +72,60 @@ export default (
...
 	return withSentry(async (context: CustomContext) => {
 		enhanceOctokit(context.github);
+		context.webhookReceived = getCurrentTime();
 
 		let webhookEvent = context.name;
 		if (context.payload.action) {
 			webhookEvent = `${webhookEvent}.${context.payload.action}`;
 		}
 
+		const webhookReceived = context.webhookReceived || undefined
```
KAbakumovAtl, line 82: We don't really need `|| undefined` here hence we assigned a value on the line 75 (And if the field webhookRecevied wasn't there it would be undefined anyway)

Maybe it is better to do 
``` 
 const webhookReceived = getCurrentTime()
 context.webhookReceived = webhookRecevied
```

src/github/middleware.ts
```diff
@@ -119,14 +154,16 @@ export default (
 			return;
 		}
 
-		context.log(`Processing event for ${jiraSubscriptionsCount} jira instances`);
+		context.log(
+			`Processing event for ${jiraSubscriptionsCount} jira instances`
+		);
```
KAbakumovAtl, line 159: For the future, It is better to avoid reformatting files, hence it makes PR review more difficult

src/transforms/push.ts
```diff
@@ -169,14 +187,28 @@ export const processPush = async (github: GitHubAPI, payload) => {
...
+				jiraPayload
+			);
+			const webhookName = payload.name || "none";
+
+			if (
+				(await booleanFlag(BooleanFlags.WEBHOOK_RECEIVED_METRICS, false)) &&
+				webhookReceived
+			) {
+				calculateProcessingTimeInSeconds(
+					webhookReceived,
```
KAbakumovAtl, line 203: So at this point `webhookReceived` was deserialised from the message payload JSON. Are we sure it has `Date` type?  Maybe we need to convert it to date explicitly? (I think that what could be the reason why it was failing before)

KAbakumovAtl: Oh I can see it was changed to Number, so should be good now

src/util/webhooks.ts
```diff
@@ -0,0 +1,59 @@
...
+			// https://hello.atlassian.net/wiki/spaces/OBSERVABILITY/pages/797646144/Reference+-+Understanding+histogram-type+metrics#So%2C-how-do-I-get-started%3F
+			statsd.histogram(
+				"webhook-event-processing.duration-ms",
+				timeToProcessWebhookEvent,
+				tags
+			);
+
+			return timeToProcessWebhookEvent;
+		} else {
+			return undefined;
```
KAbakumovAtl, line 51: I would log an error here hence such situation should be impossible

src/github/middleware.ts
rachellerathbone: I agree. Unfortunately, every time I save it autoformats. I have the same linting set up as the rest of the team and am yet to resolve this.

src/util/webhooks.ts
```diff
@@ -0,0 +1,59 @@
...
+				tags
+			);
+
+			const histogramBuckets =
+				"1000_10000_30000_60000_120000_300000_600000_3000000";
+
+			tags["gsd_histogram"] = histogramBuckets;
+			
+			// send metrics with gsd_histogram so it will be treated as a histogram-type metric
+			// https://hello.atlassian.net/wiki/spaces/OBSERVABILITY/pages/797646144/Reference+-+Understanding+histogram-type+metrics#So%2C-how-do-I-get-started%3F
```
KAbakumovAtl, line 42: I don't think we should leave links to our internal confluence in the public repo

test/unit/util/webhooks.test.ts
```diff
@@ -0,0 +1,87 @@
...
+	const mockContextLogger = {
+		info: mockInfoLog,
+	};
+
+	describe("calculateProcessingTimeInSeconds", () => {
+		it("should calculate processing time for webhook events", () => {
+			const webhookReceived = 1632726220873;
+			const webhookName = "workflow_run";
+			const status = 202;
+			const result = currentTime - webhookReceived; // 1000ms
```
KAbakumovAtl, line 28: /nit for tests it would be better to use simpler times like currentTime=2000 webhookReceived=500 and the expect the result be 500 (rather than redo calculations of the function within the test)

src/transforms/push.ts
rachellerathbone: Changed the way I was handling this since my original PR. Originally I was using new Date() to get the currentTime: `export const getCurrentTime = () => new Date();` and the error we were receiving in the logs failed here: `webhookReceivedTime.getTime()` because it was trying to call getTime() when webhookReceivedTime didn't exist. I'm not using Date.now() and only run that block of code if `webhookReceivedTime < currentTime` so if it's a number that's higher, or any other data type, it won't run. Tests confirm this behaviour.

src/util/webhooks.ts
```diff
@@ -0,0 +1,59 @@
+import statsd from "../config/statsd";
+
+export const getCurrentTime = () => Date.now();
+
+export const calculateProcessingTimeInSeconds = (
```
KAbakumovAtl, line 6: I'd rather call it `emitWebhookProcessingTimeMetrics`

Also, if we don't use the return value, we should make it Void

test/unit/util/webhooks.test.ts
```diff
@@ -0,0 +1,87 @@
...
+	// Unlock Time
+	dateNowSpy.mockRestore();
+});
+
+describe("Webhooks suite", () => {
+	const mockInfoLog = jest.fn();
+	const mockContextLogger = {
+		info: mockInfoLog,
+	};
+
```
KAbakumovAtl, line 22: We might want to spy/mock statsd here and verify that the hystogram methods were called with the expected value

test/unit/util/webhooks.test.ts
```diff
@@ -0,0 +1,87 @@
+import { calculateProcessingTimeInSeconds } from "../../../src/util/webhooks";
```
KAbakumovAtl, line 1: Nice, ++ for adding tests

src/util/webhooks.ts
rachellerathbone: Am I not already doing that?  number | undefined | void

KAbakumovAtl (approved):

PR #784, opened 2021-09-28
Arc 676 continue sync on error

thombergs:
Currently, a sync stops completely on a single error. For example, when the GitHub API times out (which it does for large requests), or if there is some other error that we don't have error handling for, the whole sync stops. All repos that have not been synced will stay unsynced.

This PR changes this behavior. If there is an unhandled error, the sync will now continue with the next task, a task being one of the following:

- syncing the commits of a repo
- syncing the PRs of a repo
- syncing the branches of a repo

I.e. if there is an error in the middle of syncing commits, the rest of the commits for this repo are skipped (and left "unsynced") and it continues with the branches and PRs.

In the future, we should add even better handling around this, so that we can continue with the next page of commits if a commit task fails, so that not all commits after the error will be left unsynced.

You can test this PR with this repository: https://github.com/thombergs/jira-test. It has some commits with a couple hundred thousand changed files, which will cause an error in the GitHub API.

### DIFF
--- a/src/models/subscription.ts
+++ b/src/models/subscription.ts
@@ -39,7 +39,7 @@ export interface RepositoryData {
 	[key: string]: unknown;
 }
 
-export type TaskStatus = "pending" | "complete";
+export type TaskStatus = "pending" | "complete" | "failed";
 
 export interface Repository {
 	id: string;
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -42,7 +42,7 @@ type TaskType = "pull" | "commit" | "branch";
 
 const taskTypes = Object.keys(tasks) as TaskType[];
 
-const completedTaskStatuses: TaskStatus[] = ["complete", "failed"];
+const completedTaskStatuses : TaskStatus[] = ["complete", "failed"];
 
 const updateNumberOfReposSynced = async (
 	repos: Repositories,
@@ -233,10 +233,10 @@ export const isNotFoundError = (
 	const isNotFoundError = isNotFoundErrorType?.length > 0 || err?.status === 404;
 
 	isNotFoundError &&
-	logger.info(
-		{ job, task: nextTask },
-		"Repository deleted after discovery, skipping initial sync"
-	);
+		logger.info(
+			{ job, task: nextTask },
+			"Repository deleted after discovery, skipping initial sync"
+		);
 
 	return isNotFoundError;
 };
@@ -425,7 +425,9 @@ export const processInstallation =
 
 				if (await booleanFlag(BooleanFlags.CONTINUE_SYNC_ON_ERROR, false, jiraHost)) {
 
-					logger.warn({ job, task: nextTask, err }, "Task failed, continuing with next task");
+					// TODO: add the jiraHost to the logger with logger.child()
+					const host = subscription.jiraHost || "none";
+					logger.warn({ job, task: nextTask, err, jiraHost: host }, "Task failed, continuing with next task");
 
 					// marking the current task as failed
 					await subscription.updateRepoSyncStateItem(nextTask.repositoryId, getStatusKey(nextTask.task as TaskType), "failed");
@@ -438,6 +440,7 @@ export const processInstallation =
 
 					await subscription.update({ syncStatus: "FAILED" });
 
+					// TODO: add the jiraHost to the logger with logger.child()
 					const host = subscription.jiraHost || "none";
 					logger.warn({ job, task: nextTask, err, jiraHost: host }, "Sync failed");
 
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -1,5 +1,11 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import SubscriptionClass, { Repositories, Repository, RepositoryData, SyncStatus } from "../models/subscription";
+import SubscriptionClass, {
+	Repositories,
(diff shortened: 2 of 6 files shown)


src/models/subscription.ts
```diff
@@ -39,7 +39,7 @@ export interface RepositoryData {
 	[key: string]: unknown;
 }
 
-export type TaskStatus = "pending" | "complete";
+export type TaskStatus = "pending" | "complete" | "failed";
```
thombergs, line 42: Currently, a task can either be "pending" (i.e. we're working on it) or "complete". Introducing the state "failed" to mark that a task has failed for some reason.

src/sync/installation.ts
```diff
@@ -49,9 +57,9 @@ const updateNumberOfReposSynced = async (
 		// all 3 statuses need to be complete for a repo to be fully synced
 		const { pullStatus, branchStatus, commitStatus } = repos[id];
 		return (
-			pullStatus === "complete" &&
-			branchStatus === "complete" &&
-			commitStatus === "complete"
+			completedTaskStatuses.includes(pullStatus)
+			&& completedTaskStatuses.includes(branchStatus)
+			&& completedTaskStatuses.includes(commitStatus)
```
thombergs, line 62: To mark a repo as "synced", we now have to consider the "complete" and "failed" statuses instead of just the "complete" status.

src/sync/installation.ts
```diff
@@ -214,7 +222,7 @@ export const isRetryableWithSmallerRequest = (err): boolean => {
 // Checks if parsed error type is NOT_FOUND / status is 404 which come from 2 different sources
 // - GraphqlError: https://github.com/octokit/graphql.js/tree/master#errors
 // - RequestError: https://github.com/octokit/request.js/blob/5cef43ea4008728139686b6e542a62df28bb112a/src/fetch-wrapper.ts#L77
-export const handleNotFoundErrors = (
+export const isNotFoundError = (
```
thombergs, line 225: Renamed this function, as it's not handling anything, but rather checking if it's a not found error.

src/sync/installation.ts
```diff
@@ -406,22 +417,36 @@ export const processInstallation =
...
+					await subscription.updateRepoSyncStateItem(nextTask.repositoryId, getStatusKey(nextTask.task as TaskType), "failed");
 
-				job.sentry.setExtra("Installation FAILED", JSON.stringify(err, null, 2));
-				job.sentry.captureException(err);
+					statsd.increment(metricTaskStatus.failed, [`type: ${nextTask.task}`]);
 
-				statsd.increment(metricSyncStatus.failed);
+					// queueing the job again to pick up the next task
+					queues.installation.add(job.data);
+				} else {
```
thombergs, line 437: This is the main change. In case of an unhandled error, we now update the status of the task as "failed" and schedule another message on the queue to continue with the next task.

src/sync/installation.ts
```diff
@@ -375,6 +383,9 @@ export const processInstallation =
 					task,
 					repositoryId
 				);
+
+				statsd.increment(metricTaskStatus.complete, [`type: ${nextTask.task}`]);
```
KAbakumovAtl, line 387: What is `nextTask.task`?  Is it type of data this task is synching? (Just want to make sure that variations of this value are bounded)

src/sync/installation.ts
```diff
@@ -406,22 +417,36 @@ export const processInstallation =
...
 
-				await subscription.update({ syncStatus: "FAILED" });
+				if (await booleanFlag(BooleanFlags.CONTINUE_SYNC_ON_ERROR, false, jiraHost)) {
+
+					logger.warn({ job, task: nextTask, err }, "Task failed, continuing with next task");
 
-				const host = subscription.jiraHost || "none";
-				logger.warn({ job, task: nextTask, err, jiraHost: host }, "Sync failed");
+					// marking the current task as failed
+					await subscription.updateRepoSyncStateItem(nextTask.repositoryId, getStatusKey(nextTask.task as TaskType), "failed");
```
KAbakumovAtl, line 433: /nit it would be great to have an enum with the task statuses rather than use strings. However not sure if it worth changing if we want to rewrite it anyway

src/sync/installation.ts
thombergs: The variable `nextTask` contains some info about the current task (the naming is terrible). Yeah, `nextTask.task` is either `commit`, `pull` or `branch` like the keys in the `tasks` variable at the top of the file. 

This code is terrible, we need to refactor it.

src/sync/installation.ts
```diff
@@ -406,22 +417,36 @@ export const processInstallation =
...
 
-				job.sentry.setExtra("Installation FAILED", JSON.stringify(err, null, 2));
-				job.sentry.captureException(err);
+					statsd.increment(metricTaskStatus.failed, [`type: ${nextTask.task}`]);
```
KAbakumovAtl, line 435: So we count here failed tasks, correct? 

Should we also count successful tasks?

If one task not equal to one repo, should we add a separate metric to count successfully sync & failed repos?

src/sync/installation.ts
thombergs: Yeah. There is a `TaskStatus` type defined as `"complete" | "pending" | "failed"`, but we should probably migrate that to a proper enum some time.

KAbakumovAtl (approved):

PR #785, opened 2021-09-28
Adding stricter tsconfig options

mboudreau:
And changing files according to the new settings.  Removed 'HYDRO' settings and files as we're not using it.


src/frontend/delete-jira-configuration.ts
```diff
@@ -9,23 +7,26 @@ import { Request, Response } from "express";
  */
 export default async (req: Request, res: Response): Promise<void> => {
 	const jiraHost = req.session.jiraHost;
+	if(!jiraHost) {
+		req.log.warn("Missing Jira Host");
+		return;
+	}
```
thombergs, line 13: Any special reason that prompted you to add this warning?

mboudreau: without it, the code would fail if the jira host is missing.  There isn't any point of moving forward without it and figured we'd want to know if this is happening consistently.

src/github/issue.ts
```diff
@@ -1,23 +1,16 @@
...
 
 	let linkifiedBody;
 	try {
 		linkifiedBody = await util.unfurl(issue.body);
 		if (!linkifiedBody) {
-			context.log(
-				{ noop: "no_linkified_body_issue" },
-				"Halting further execution for issue since linkifiedBody is empty"
-			);
+			context.log("Halting further execution for issue since linkifiedBody is empty");
```
thombergs, line 13: Why remove the `noop` field in the log event? It might be handy to filter for noop reasons. Then again, I don't care enough about this field to talk too much about it...

mboudreau: `noop` isn't a standard for us and it was a relic of github's way of doing things.  It also had no corrolation between messages and we can just as easily just filter by the message itself.

src/jira/client/index.ts
```diff
@@ -35,38 +36,40 @@ async function getJiraClient(
...
+						// Ignore any errors
+						.catch(() => undefined))
+				);
+				return responses.reduce((acc: JiraIssue[], response) => {
+					if (response?.status === 200 && !!response?.data) {
+						acc.push(response.data);
+					}
+					return acc;
+				}, []);
+			},
```
thombergs, line 58: How do we know that `getAll()` is working as before?

src/api/index.ts
```diff
@@ -153,6 +154,12 @@ router.get(
 		const githubInstallationId = Number(req.params.installationId);
 
 		try {
+			if (!req.session.jiraHost) {
+				req.log.warn({ req, res }, "Missing Jira Host");
+				res.status(500).send("Missing Jira Host");
+				return;
```
KAbakumovAtl, line 160: Do we need to add a check here? Why do we need Jira host for our admin endpoints at all?

(If we really need it the response status should be 400)

src/frontend/delete-jira-configuration.ts
KAbakumovAtl: Why not throwing error? Otherwise it will return 2xx

src/frontend/delete-github-subscription.ts
```diff
@@ -66,13 +64,12 @@ export default async (req: Request, res: Response): Promise<void> => {
 			}
 
 			const subscription = await Subscription.getSingleInstallation(req.body.jiraHost, req.body.installationId);
-			const installation = await Installation.getForHost(req.body.jiraHost);
-			const action = ActionFromSubscription(subscription, installation);
-			action.type = ActionType.DESTROYED;
-			action.actionSource = ActionSource.WEB_CONSOLE;
-
+			if(!subscription) {
```
KAbakumovAtl, line 67: It is a bit disturbing to see business logic changes in this PR, formatting changes affect many files and I am afraid I will miss something

src/frontend/get-github-setup.ts
```diff
@@ -8,16 +8,10 @@ Redirects users back to github/configuration to install their Jira instance in G
...
-
-	if (req.session.jiraHost) {
-		const { host: githubHost, session } = req;
-		const { jwt, jiraHost } = session;
-
-		const urlArgs = { githubHost, jwt, jiraHost };
-
-		return res.redirect(getGitHubConfigurationUrl(urlArgs));
+	const { host: githubHost, session: { jwt, jiraHost } } = req;
+	if (jiraHost && jwt) {
```
KAbakumovAtl, line 13: What's the impact of adding `&& jwt` condition here?

src/frontend/validations.ts
```diff
@@ -4,7 +4,7 @@ const subdomainRegexp = /^\w(?:[\w-]{0,61}\w)?$/;
 const jiraDomains = ["atlassian.net", "jira.com"];
 
 export const validJiraDomains = (jiraSubdomain: string, jiraDomain: string): boolean =>
-	jiraSubdomain &&
+	!!jiraDomains && !!jiraSubdomain &&
```
KAbakumovAtl, line 7: Wouldn't this condition be always true? `!!jiraDomains` hence it is a constant within this file? 

Maybe you wanted to do `!!jiraDomain` 

Also what does it do? Checks that it is not undefined?

KAbakumovAtl: - [ ] jiraDomain or jiraDomains

src/jira/client/index.ts
```diff
@@ -35,38 +36,40 @@ async function getJiraClient(
...
+				const responses = await Promise.all<AxiosResponse<JiraIssue> | undefined>(
+					issueIds.map((issueId) => client.issues.get(issueId, query)
+						// Ignore any errors
+						.catch(() => undefined))
+				);
+				return responses.reduce((acc: JiraIssue[], response) => {
+					if (response?.status === 200 && !!response?.data) {
+						acc.push(response.data);
+					}
+					return acc;
```
KAbakumovAtl, line 56: Why did we have to replace filter.map with reduce?

mboudreau: filter and map doesn't have the correct typings as they assume it would have the same typings as the array before it, which filter was actually changing.  Plus, a single reduce function is quicker than to do both filter & map while keeping the typings correct.

src/jira/client/index.ts
mboudreau: going to have to test in staging.

src/frontend/get-github-setup.ts
mboudreau: None as we need both to continue.  I figured might as well check for jwt as well since the typings were complaining that jwt might be undefined.

src/frontend/delete-github-subscription.ts
mboudreau: It wouldn't compile without optional chaining or a conditionals since there's no guarantee that a subscription is returned.  I could just leave it as optional chaining, but figured we'd want to log when this happens as it shouldn't.

src/frontend/validations.ts
mboudreau: whoops, yeah, suppose to be `jiraDomain`.  And yes, `!!` changes it to boolean type and not just truthy.

src/api/index.ts
mboudreau: Because it's being used to get the subscription below and the typings were erroring.  Now that I look at it though, seems like the logic of this API itself is wrong and the jiraHost should be passed in the API POST body and not part of the session.

src/sync/installation.ts
```diff
@@ -33,16 +27,19 @@ interface TaskProcessors {
...
+			perPage?: number
 		) => Promise<{ edges: any[], jiraPayload: any }>;
 }
 
 type TaskType = "pull" | "commit" | "branch";
 
 const taskTypes = Object.keys(tasks) as TaskType[];
 
-const completedTaskStatuses : TaskStatus[] = ["complete", "failed"];
+// TODO: why are we ignoring failed status as completed?
```
KAbakumovAtl, line 39: That's a change made recently by Tom. We do it to fix the case where we fail the whole sync if one repo fails. We count it as completed so we can continue through the rest of repositories (hence we are not going to do anything with the failed repo again)

src/sync/installation.ts
```diff
@@ -53,14 +50,10 @@ const updateNumberOfReposSynced = async (
...
-	const syncedRepos = repoIds.filter((id) => {
+	const syncedRepos = repoIds.filter((id: string) => {
 		// all 3 statuses need to be complete for a repo to be fully synced
 		const { pullStatus, branchStatus, commitStatus } = repos[id];
-		return (
-			completedTaskStatuses.includes(pullStatus)
-			&& completedTaskStatuses.includes(branchStatus)
-			&& completedTaskStatuses.includes(commitStatus)
-		);
+		return isTaskStatusCompleted(pullStatus, branchStatus, commitStatus);
```
KAbakumovAtl, line 56: /nit `isAllTasksStatusesCompleted` name would make more sense

test/safe/push.test.ts
```diff
@@ -25,7 +26,7 @@ describe("GitHub Actions", () => {
 				{
 					repository: event.payload.repository,
 					shas: [{ id: "test-commit-id", issueKeys: ["TEST-123"] }],
-					jiraHost: process.env.ATLASSIAN_URL,
```
KAbakumovAtl, line 28: Tests depending on env vars 😬

mboudreau: Why's that an issue if we're using the same env vars in our code?

KAbakumovAtl (approved):

KAbakumovAtl:
I tried to review it thoroughly, however the PR is huge and I am not sure how reliable my review is. Let's test it in staging before merging to prod

src/jira/client/index.ts
mboudreau: I also tested it locally and everything's working.

thombergs (approved):

PR #787, opened 2021-09-28
ARC-682: remove the "sync failed" error on the read path

thombergs:
When a subscription hasn't had its `updatedAt` field updated for the last 15 minutes, the UI would show a sync as FAILED, even if it was still running in the background.

This PR updates the `updatedAt` field more often but also removes the mapping to the FAILED state in the UI.

### DIFF
--- a/src/frontend/get-jira-configuration.ts
+++ b/src/frontend/get-jira-configuration.ts
@@ -3,29 +3,16 @@ import moment from "moment";
 import { Subscription } from "../models";
 import { NextFunction, Request, Response } from "express";
 import statsd from "../config/statsd";
-import { metricSyncStatus, metricError } from "../config/metric-names";
-import * as Sentry from "@sentry/node";
+import { metricError } from "../config/metric-names";
 
-const syncStatus = (syncStatus) =>
-	syncStatus === "ACTIVE" ? "IN PROGRESS" : syncStatus;
-
-const sendFailedStatusMetrics = (installationId: string, reqLog: any): void => {
-	const syncError = "No updates in the last 15 minutes";
-	reqLog.warn({ installationId, error: syncError }, "Sync failed");
-
-	Sentry.setExtra("Installation FAILED", syncError);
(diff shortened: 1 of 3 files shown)


rachellerathbone (approved):

src/frontend/get-jira-configuration.ts
```diff
@@ -3,29 +3,16 @@ import moment from "moment";
...
 import { NextFunction, Request, Response } from "express";
 import statsd from "../config/statsd";
-import { metricSyncStatus, metricError } from "../config/metric-names";
-import * as Sentry from "@sentry/node";
+import { metricError } from "../config/metric-names";
 
-const syncStatus = (syncStatus) =>
-	syncStatus === "ACTIVE" ? "IN PROGRESS" : syncStatus;
-
-const sendFailedStatusMetrics = (installationId: string, reqLog: any): void => {
```
thombergs, line 12: This method would get called every time someone viewed the configuration page when the subscription had its `updatedAt` field older than 15 minutes. 

This means we were logging a status change on the read path. And the status wasn't even persisted! This is wrong on so many levels...

src/models/subscription.ts
```diff
@@ -227,14 +227,15 @@ export default class Subscription extends Sequelize.Model {
 
 		this.repoSyncState.numberOfSyncedRepos = cnt;
 		await this.sequelize.query(
-			`UPDATE "Subscriptions" SET "repoSyncState" = jsonb_set("repoSyncState", '{numberOfSyncedRepos}', ':cnt', true) WHERE id = :id`,
+			`UPDATE "Subscriptions" SET "updatedAt" = NOW(), "repoSyncState" = jsonb_set("repoSyncState", '{numberOfSyncedRepos}', ':cnt', true) WHERE id = :id`,
```
thombergs, line 230: We weren't updating the `updatedAt` field in some of the steps of the installation sync, so I added them here. This way, we know from a look at the database when a sync was last active.

PR #789, opened 2021-09-30
ARC-685 Updated server side http metrics

KAbakumovAtl:
* Added requests count http metric, added server-side bucketed histogram duration metrics
* Renamed some existing metrics
* Removed actual paths from metrics tags (which was exploding our cardinality), replaced with route path or wildcard path tag where needed.


src/api/index.ts
```diff
@@ -148,7 +149,6 @@ router.get(
 	"/:installationId/repoSyncState.json",
 	check("installationId").isInt(),
 	returnOnValidationError,
-	elapsedTimeMetrics,
```
KAbakumovAtl, line 151: Removed metrics from individual calls and put them on the router level because:
* If they are here it wouldn't capture 401s and other validation error responses
* On this level It wouldn't include the validation and authentication times

src/config/metric-names.ts
```diff
@@ -17,25 +17,27 @@ export const queueMetrics = {
...
-		syncPullRequest: `${server}.http.request.sync-pull-request`,
+		duration: `${server}.http.server.requestes.duration`, //Duration of incoming HTTP requests
+		executed: `${server}.http.server.requestes`, //Count of executed incoming HTTP requests
+		github: `${server}.http.request.github`, //Octokit requests execution time histogram
+		jira: `${server}.http.request.jira`, //Jira requests execution time histogram
+		install: `${server}.http.request.install`, //Count how many installation requests we received
+		uninstall: `${server}.http.request.uninstall`, //Count how many uninstallation requests we received
+		syncPullRequest: `${server}.http.request.sync-pull-request`, //Measures retrieval time from GitHub API
 		hydroSubmission: `${server}.http.request.${metricName}`,
-		jobDuration: `${server}.http.request.job-duration`,
```
KAbakumovAtl, line 31: Changed some of these metrics because they have little to do with http requests. (Job duration counts the duration of the queue jobs)

src/config/statsd.ts
```diff
@@ -10,6 +11,7 @@ export const globalTags = {
 	region: process.env.MICROS_AWS_REGION || "localhost"
 };
 
+const RESPONSE_TIME_HISTOGRAM_BUCKETS =	"100_1000_2000_3000_5000_10000_30000_60000";
```
KAbakumovAtl, line 14: For http requests we have much smaller buckets than for overall webhooks processing because we expect our rest api reply fast. If they reply slower then few seconds, in frontend, it is considered to be a failure from the user experience point of view.

src/config/metric-names.ts
KAbakumovAtl: I'll open a pull request for our dashboards as well

KAbakumovAtl:
One downside of this solution. It adds metrics only to endpoints where we explicitly added a metrics middleware. That mean the **/github/events** endpoint won't be covered, hence it is in probot and we can't add a middleware there.

src/api/index.ts
rachellerathbone: Nice!

src/config/statsd.ts
rachellerathbone: Now that we have multiple buckets for histogram metrics do you think we should move the webhook buckets into this file?

src/config/statsd.ts
```diff
@@ -42,17 +44,38 @@ function hrtimer() {
...
+export const elapsedTimeMetrics = (path?: string) => (
 	req: Request,
 	res: Response,
 	next: NextFunction
 ) => {
 	const elapsedTimeInMs = hrtimer();
-	const { path, method } = req;
-	const tags = { path, method };
+	const method = req.method;
+	const pathTag = path || req.route?.path || "unknown";
```
rachellerathbone, line 58: I defaulted to "none" for the status property for webhook metrics in the event it wasn't in the response from GH/Jira. Do you think we should keep consistent across tags?

src/config/statsd.ts
```diff
@@ -42,17 +44,38 @@ function hrtimer() {
 	};
 }
 
-export const elapsedTimeMetrics = (
+/**
+ * Returns a middleware function which produce duration and requests count metrics
+ * @param path The value of the "path" tag. If not set then the rout path will be used
+ */
+export const elapsedTimeMetrics = (path?: string) => (
```
mboudreau, line 51: Is there a reason why you would bundle all endpoints under the same 'path'?  Wouldn't you want it to be done individually as to reliably get metrics of each one? 

I'd recommend you use `req.originalUrl` as it shows the whole path including query string if there are any.  If you don't want query string, you can parse it out or use `req.baseUrl + req.path` to get the full path without query.  BaseUrl is useful to get the router mountpoint url.

src/frontend/app.ts
```diff
@@ -133,19 +133,22 @@ export default (octokitApp: App): Express => {
 		syncStatus === "COMPLETE" ? "Connected" : "Connect"
 	);
 
-	app.use("/public", express.static(path.join(rootPath, "static")));
+	app.use("/public", elapsedTimeMetrics("/public/*"), express.static(path.join(rootPath, "static")));
 	app.use(
 		"/public/css-reset",
+		elapsedTimeMetrics("/public/*"),
```
mboudreau, line 139: There's no point of having these individual ones as the `use` above (L136) will hit all of them.

mboudreau (changes requested):
Instead of having `elapsedTimeMetrics` everywhere in the code, I'd just at it at the top level in `src/configure-robot.ts` on L93 `app.router.use(elapsedTimeMetrics);` which will capture every single request coming through.

src/config/statsd.ts
KAbakumovAtl: No, for webhook buckets we have their own. It make sense that they are different, because webhooks processing time is expected to be higher (and what normal time for webhook processing can be too slow for the frontend API)

src/frontend/app.ts
KAbakumovAtl: Oh interesting, I thought this line was overriding it. 

How does it work? Would it only override the last handler for this route( `express.static(path.join(rootPath, "static")`  will be replaced with `express.static(
			path.join(rootPath, "node_modules/@atlaskit/css-reset/dist")
		) ` ?

src/config/statsd.ts
KAbakumovAtl: So the reason for overriding path is that we can't use **req.path** as a tag. It can contain path parameters which explode metrics cardinality. (We are currently affected by this problems, we have path tag varliations like `/github/subscriptions/123` and `/github/subscriptions/345` while they should have the same tag, for example: `/github/subscruptions/<installation_id>`


2. The reason to bundle all API requests in the router - is to capture responses in case if it is 401. Hence authentication middleware is common for the router and will be executed before the individual routes middleware. So that's why we need to pass "path" as a parameter, because at that point the request won't have `route` parameter and we can't use raw request path for cardinality reason

3. Regarding this comment https://github.com/atlassian/github-for-jira/pull/789#pullrequestreview-767467777 
That's what I tried first, but it didn't work as I needed.
Unfortunately we can't put this middleware on that level. Because the request won't have `route` parameter. (request will only have it if this middleware function was executed as part of handling some route like:

````
app.get(
		"/github/subscriptions/:installationId",
		elapsedTimeMetrics(),
		csrfProtection,
		getGitHubSubscriptions
	);
```
When the middleware will be called here the req.route.path would be equal to "/github/subscriptions/:installationId"
If I put it to the top level

```
app.use(elapsedTimeMetrics())
```
the req.route will be undefined, and req.path will be "/github/subscriptions/12345" which we can't use as path tag.

So unfortunately we can't use it on the top level

src/config/statsd.ts
rachellerathbone: I wasn't saying to make them the same, I was asking if the buckets for both should live in this file so they are in a central location.

KAbakumovAtl: I think it is fine if they live where they needed. Better group code and definitions by its purpose rather than by their data type.   (For example here are http metrics, and there are webhook metrics)

src/configure-robot.ts
```diff
@@ -60,6 +60,9 @@ async function getGitHubCIDRs(logger: Logger): Promise<string[]> {
 }
 
 export default async (app: Application): Promise<Application> => {
+
+	app.router.use(elapsedTimeMetrics);
```
KAbakumovAtl, line 64: I moved it to the top level to make sure it will be executed first and capture any error if it happens in the middleware

bgvozdev: Would it make sense to move it to frontend app (src/frontend/index.ts)? It doesn't work for webhooks anyway , however this is quite confusing to have these on the top level (in my PR I'm relocating all these middlewares to frontend app)...

mboudreau: @bgvozdev you'd also want to capture github specific routes since this file calls both github and frontend routes.

mboudreau (approved):

src/config/statsd.ts
KAbakumovAtl: Changed it, but not exactly in this way, we decided to capture as much of path as we can. If we use baseUrl, we already have a part of the path if request finished in any middleware in the middle which was before the final request handler. 

So we add baseUrl + '*' when we don't know the rout path, and made it '/*' when we don't know baseUrl to make it the same format

src/configure-robot.ts
KAbakumovAtl: Yes this, however I don't think webhooks still get there. We need to find the way to extend it to probot (or solve it after we get rid of problt). Hence we'll use the same metric I'd like to leave it as is. 

Also it is important to capture here so there will be as little middleware before `elapsedTimeMetrics` as possible so we'll capture all potential 500s

KAbakumovAtl:
Tested in staging, looks good

PR #792, opened 2021-09-30
ARC-449 GitHub issue links tracking

rachellerathbone:
Added a query param so we can track the clickthrough from GitHub to Jira.


src/config/env.ts
```diff
@@ -31,6 +31,7 @@ const envVars: EnvVars = {
 	MICROS_SERVICE_VERSION: process.env.MICROS_SERVICE_VERSION,
 	NODE_ENV: nodeEnv,
 	SENTRY_DSN: process.env.SENTRY_DSN,
+	ALT_ORIGIN: process.env.ALT_ORIGIN,
```
mboudreau, line 34: What does this do?

thombergs: Let's call this `JIRA_LINK_QUERY_PARAMS` or some such and make it include the whole query (i.e. `?atlOrigin=github-for-jira` or whatever). That makes the intent clearer and reduces the code needed to add it below.

thombergs (changes requested):

src/config/env.ts
rachellerathbone: @mboudreau this will allow the growth team to track clickthroughs from GH to Jira (via the link issue generated in GH). This data will show up in their partnership tracking boards.

@thombergs I changed the param name but not sure I agree with changing the envVar property to the full query here. It doesn't really fit into the object named `envVars` if it's more than just the micros env var.

thombergs (approved):

PR #793, opened 2021-09-30
ARC-673 fix broken flow from GH

rachellerathbone:
Small PR to fix the broken flow when installing from GH. As a follow-up to this in the coming weeks there are 2 further actions I will be taking:

1. **Update the UI for the github/setup page**. This will require further design mods from Jade as I have determined that if you have a Jira host is your session, there is currently no one to target a new Jira instance unless you clear you session. This page should check if the jirsHost exists on the session. If it doesn't, just display the input field to enter a Jira site. If it does, display the input field *and* the option to click through to the Jira instance that's in the session. #ARC-619
2. **Determine if it is possible to detect if a user already has the app installed**. If so, direct to the connect an org page, otherwise automatically install the app: #ARC-620


src/util/getUrl.ts
```diff
@@ -11,7 +11,7 @@ export const getGitHubConfigurationUrl = (urlParams: UrlParams): string => {
 };
 
 export const getJiraMarketplaceUrl = (jiraHost: string): string =>
-	`https://${jiraHost}/plugins/servlet/upm/marketplace/plugins/com.github.integration.production`;
+	`${jiraHost}/plugins/servlet/ac/com.atlassian.jira.emcee/discover/app/com.github.integration.production#!/discover/app/com.github.integration.production`;
```
mboudreau, line 14: Could you use `/plugins/servlet/ac/com.atlassian.jira.emcee/discover#!/discover/app/com.github.integration.production` instead as it bring them directly to the github for jira app instead of the search screen

src/frontend/get-github-setup.ts
```diff
@@ -1,23 +1,22 @@
...
 */
-export default (req: Request, res: Response, next: NextFunction): void => {
+export default (req: Request, res: Response): void => {
 	req.log.info("Received get github setup page request");
-	const { host: githubHost, session: { jwt, jiraHost } } = req;
-	if (jiraHost && jwt) {
-		return res.redirect(getGitHubConfigurationUrl({ githubHost, jwt, jiraHost }));
+
+	if (req.session.jiraHost) {
+		res.redirect(getJiraMarketplaceUrl(req.session.jiraHost));
```
mboudreau, line 14: something that would be nice here is if they have a jiraHost (or after they've inputted it in github/setup), that there would be an API on our end to confirm if the host already has the app installed.  If they do, go directly to the manage app section, if not, bring them to the marketplace URL to install it.

mboudreau (changes requested):

src/frontend/get-github-setup.ts
rachellerathbone: As mentioned in the PR description, there is already a ticket (#620) in the backlog to follow up on this. I think we should decouple this from the fix for the broken flow and get this in as soon as possible.

src/util/getUrl.ts
```diff
@@ -11,7 +11,7 @@ export const getGitHubConfigurationUrl = (urlParams: UrlParams): string => {
 };
 
 export const getJiraMarketplaceUrl = (jiraHost: string): string =>
-	`https://${jiraHost}/plugins/servlet/upm/marketplace/plugins/com.github.integration.production`;
+	`${jiraHost}/plugins/servlet/ac/com.atlassian.jira.emcee/discover#!/discover/app/com.github.integration.productio`;
```
bgvozdev, line 14: productio?

src/frontend/get-github-setup.ts
```diff
@@ -1,23 +1,22 @@
...
+import { Request, Response } from "express";
+import { getJiraMarketplaceUrl } from '../util/getUrl';
 
 /*
 When this request is made: Installing from Jira Marketplace - GitHub org does not have Jira installed.
 Redirects users back to github/configuration to install their Jira instance in GitHub org/s.
 If the installation was done from Jira Marketplace, the app is already installed.
 */
-export default (req: Request, res: Response, next: NextFunction): void => {
+export default (req: Request, res: Response): void => {
```
bgvozdev, line 10: Is it flow completely broken now? Should we feature-flag the change?

src/util/getUrl.ts
```diff
@@ -11,7 +11,7 @@ export const getGitHubConfigurationUrl = (urlParams: UrlParams): string => {
 };
 
 export const getJiraMarketplaceUrl = (jiraHost: string): string =>
```
bgvozdev, line 13: nit: Would it make sense to use "jiraBaseUrl" as the name instead?

src/util/getUrl.ts
mboudreau, line 14: missing last character in url

mboudreau (changes requested):

bgvozdev (approved):

src/frontend/get-github-setup.ts
mboudreau: from github installation, yeah, it's fairly broken.  Good thing that most customers don't install it from the github side first :)

src/util/getUrl.ts
rachellerathbone: Ugh. Thought I fixed this last night. :sigh:

src/frontend/get-github-setup.ts
rachellerathbone: Nah, I don't think we should FF. I should've have FF this original change but, given that the install flow from GH now already broken, I think we just get this merged as is.

src/util/getUrl.ts
rachellerathbone: Yeah, you never heard of productio?...

src/util/getUrl.ts
rachellerathbone: Hmm I think we should leave as is because it directs users to our app in the Jira marketplace.

bgvozdev: Yeah, sorry, I'm not after changing the logic :) Just the name itself. Host means "blahblah.atlassian.net" ; baseurl means "https://blahblah.atlassian.net" . In this code I think we wait for the baseUrl rather than the host...

mboudreau (approved):

bgvozdev (approved):

PR #798, opened 2021-10-05
ARC-684 Metrics for webhooks reliability SLO. Fixed path tag on HTTP metrics

KAbakumovAtl:
There are 2 changes
1. I've added failure/success metrics for webhooks processing so we'll be able to count the amount of failed webhooks
2. I've changed the way we calculate "path" for HTTP metrics hence there was a bug

### DIFF
--- a/src/config/statsd.ts
+++ b/src/config/statsd.ts
@@ -57,7 +57,8 @@ export const elapsedTimeMetrics = (
 	const method = req.method;
 
 	res.once("finish", () => {
-		const pathTag = req.route?.path || ((req.baseUrl || "/") + "*" );
+
+		const pathTag = req.baseUrl ? req.baseUrl + (req.route?.path || "*") : (req.route?.path || "/*");
 		const elapsedTime = elapsedTimeInMs();
 		const statusCode  = `${res.statusCode}`;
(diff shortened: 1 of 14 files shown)


src/config/statsd.ts
```diff
@@ -57,7 +57,8 @@ export const elapsedTimeMetrics = (
 	const method = req.method;
 
 	res.once("finish", () => {
-		const pathTag = req.route?.path || ((req.baseUrl || "/") + "*" );
+
+		const pathTag = req.baseUrl ? req.baseUrl + (req.route?.path || "*") : (req.route?.path || "/*");
```
KAbakumovAtl, line 61: So the problem was that `req.route?.path` doesn't include base url. 

For example for request to `/api/resync` the `req.route.path` will be `/resync` and base url will be `/api`

src/util/webhooks.ts
```diff
@@ -3,7 +3,17 @@ import { metricWebhooks } from "../config/metric-names";
 
 export const getCurrentTime = () => Date.now();
 
-export const emitWebhookProcessingTimeMetrics = (
```
KAbakumovAtl, line 6: I've renamed the method hence I've added a counter metric here which counts how many webhooks were processed successfully

PR #799, opened 2021-10-06
ARC-646 [small] [quick karma]: fix jira hostname log for /github/callback

bgvozdev:
This should help debugging cases like JST-691980 and PCS-54514

### DIFF
--- a/src/util/getUrl.ts
+++ b/src/util/getUrl.ts
@@ -31,10 +31,10 @@ export const getJiraHostFromRedirectUrlNew = (urlOrPath: string, log: Logger): s
 	}
 	try {
 		const { host, query } = url.parse(urlOrPath, true);
-		if (query && query.xdm_e) {
+		if (query?.xdm_e) {
 			return "" + url.parse("" + query.xdm_e, false).host;
 		}
-		return host ? host : "unknown";
+		return host || "unknown";
 	} catch (err) {
 		log.error(err, "Cannot detect jiraHost from redirect URL");
 		return "error";
--- a/src/util/getUrl.ts
+++ b/src/util/getUrl.ts
@@ -1,3 +1,5 @@
+import Logger from "bunyan";
+
 interface UrlParams {
 	githubHost: string;
 	jwt: string;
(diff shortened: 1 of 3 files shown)


src/util/getUrl.ts
```diff
@@ -13,10 +15,33 @@ export const getGitHubConfigurationUrl = (urlParams: UrlParams): string => {
...
+	try {
+		const [_, jiraBaseUrlAndTail] = url.split('&xdm_e=', 2);
+		if (jiraBaseUrlAndTail) {
+			const [jiraBaseUrlEncoded] = jiraBaseUrlAndTail.split('&');
+			const jiraBaseUrl = decodeURIComponent(jiraBaseUrlEncoded);
+			if (jiraBaseUrl.startsWith("http") && jiraBaseUrl.indexOf('//') >= 0) {
+				const [_, jiraHost] = jiraBaseUrl.split('//');
+				return jiraHost;
+			}
+			return jiraBaseUrl;
```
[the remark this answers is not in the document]
bgvozdev: It is not a full URL but only path. If I use URL, it throws an exception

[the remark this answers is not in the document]
bgvozdev: Sorry, misread your comment. Yeah, we can use `URL` here, good point! Will do now

src/util/getUrl.ts
```diff
@@ -13,10 +16,27 @@ export const getGitHubConfigurationUrl = (urlParams: UrlParams): string => {
...
 	}
 };
+
+export const getJiraHostFromRedirectUrlNew = (urlOrPath: string, log: Logger): string => {
+	if (!urlOrPath) {
+		return "empty";
+	}
+	try {
+		const { host, query } = url.parse(urlOrPath, true);
+		if (query && query.xdm_e) {
```
mboudreau, line 34: can use optional chaining here `query?.xdm_e`

src/util/getUrl.ts
```diff
@@ -13,10 +16,27 @@ export const getGitHubConfigurationUrl = (urlParams: UrlParams): string => {
...
+			return "" + url.parse("" + query.xdm_e, false).host;
+		}
+		return host ? host : "unknown";
```
mboudreau, line 37: `host || "unknown"` is a bit simpler

src/util/getUrl.ts
```diff
@@ -13,10 +16,27 @@ export const getGitHubConfigurationUrl = (urlParams: UrlParams): string => {
...
 };
+
+export const getJiraHostFromRedirectUrlNew = (urlOrPath: string, log: Logger): string => {
+	if (!urlOrPath) {
+		return "empty";
+	}
+	try {
+		const { host, query } = url.parse(urlOrPath, true);
+		if (query && query.xdm_e) {
+			return "" + url.parse("" + query.xdm_e, false).host;
```
mboudreau, line 35: isn't host already a string?  then again, it would be optional right?

Can't the whole algorithm be just simplified to 
```
const { host, query } = url.parse(urlOrPath, true);
return (query?.xdm_e ? url.parse(query.xdm_e.toString(), false).host : host) || "unknown";
```

mboudreau (approved):
:+1:

src/util/getUrl.ts
bgvozdev: How come GitHub doesn't have :awesome: icon?

PR #800, opened 2021-10-06
ARC-259 Micros lifecycle events

mboudreau:
Listening for Micros lifecycle events to start/stop so that old stack don't continue processing events after the new stack is published.

Documentation for lifecycle events: https://hello.atlassian.net/wiki/spaces/MICROS/pages/169248561/Lifecycle+hooks


src/config/env.ts
```diff
@@ -61,4 +61,9 @@ export interface EnvVars {
 	JIRA_LINK_TRACKING_ID?: string,
 	PROXY?: string,
 	LAUNCHDARKLY_KEY?: string;
+
+	// Micros Lifecycle Env Vars
+	SNS_NOTIFICATION_LIFECYCLE_QUEUE_URL?:string;
+	SNS_NOTIFICATION_LIFECYCLE_QUEUE_NAME?:string;
+	SNS_NOTIFICATION_LIFECYCLE_QUEUE_REGION?:string;
```
KAbakumovAtl, line 68: How does it work? Would Micros spin up a queue per node?

mboudreau: A queue per ec2 instance as per the documentation.

src/worker.ts
```diff
@@ -1,13 +1,169 @@
...
+import { Job } from "bull";
+import { listenForClusterCommand } from "./services/cluster/listen-command";
+
+const CONCURRENT_WORKERS = process.env.CONCURRENT_WORKERS || 1;
+const logger = getLogger("worker");
+
+/**
+ * Return an async function that assigns a Sentry hub to `job.sentry` and sends exceptions.
+ */
+const sentryMiddleware = (jobHandler) => async (job) => {
```
KAbakumovAtl, line 32: Did we move it(copy it) from somewhere?

KAbakumovAtl: Ah it must be src/worker/main.ts

src/worker.ts
```diff
@@ -1,13 +1,169 @@
...
+		return;
+	}
+	logger.info("Micros Lifecycle: Stopping queue processing");
+	// TODO: change this to `probot.close()` once we update probot to latest version
+	probot.httpServer?.close();
+	await Promise.all([
+		queues.discovery.close(),
+		queues.installation.close(),
+		queues.push.close(),
+		queues.metrics.close()
```
KAbakumovAtl, line 144: Nice

src/services/cluster/listen-command.ts
```diff
@@ -0,0 +1,13 @@
+import cluster from "cluster";
```
KAbakumovAtl, line 1: Am I correctly understand that we use cluster to send our message across all processes which were spawned by BullQueues?

KAbakumovAtl:
I think I am getting the overall concept and it looks good to me. The only request would be to add some code comments explaining how it works (usage of cluster, listening for Micros events) 

Maybe add a few words about how Micros works there hence open source contributors won't be able to open confluence page link.

KAbakumovAtl (approved):

PR #804, opened 2021-10-08
ARC-585: remove logger field; add id for job executions (tracing)

bgvozdev:
I added some clarification comments to the code regarding a few minor decisions that were taken mostly because of unification and consistency.


bgvozdev:
Oops builds are failing, I'll fix and then open

PR #817, opened 2021-10-12
Changed registry to npmjs in npmrc and updated lock file

KAbakumovAtl:
Added an override to .npmrc to enforce usage of https://registry.npmjs.org/ registry. 

Also re-generated packages-lock.json file to make sure we don't use packages.atlassian.com anymore


KAbakumovAtl:
From the first attempt it didn't work. Some tests failed due to library bumps. So I've revertet packages-lock.json and then updated only libraries which used packages.atlassian.com.

PR #824, opened 2021-10-14
ARC-650 db json state

mboudreau:
Adding new table to hold state for each repository parallel to the JSON blob until we can fully switch over.


KAbakumovAtl (dismissed):

src/models/index.ts
```diff
@@ -11,34 +12,100 @@ if (!process.env.STORAGE_SECRET) {
...
-	},
-	{ sequelize }
-);
+InstallationModel.init({
+	id: {
+		type: DataTypes.INTEGER,
+		primaryKey: true,
+		allowNull: false,
+		autoIncrement: true
+	},
```
KAbakumovAtl, line 21: Are we changing Installations table somehow or we just adding id here which was missing?

KAbakumovAtl (dismissed):

src/models/index.ts
mboudreau: id was always there (look in sequelize migration scripts), but it just wasn't "registered" on the model as being there so we couldn't use it in our code.  This adds it in there and also adds validation to it so sequelize will throw and error if you try to create a new Installation with an id already set since it should be set by the DB.

KAbakumovAtl (approved):

db/migrations/20211110150400-create-sync-state-table.js
```diff
@@ -0,0 +1,120 @@
...
+module.exports = {
+	up: async (queryInterface, Sequelize) => {
+		await queryInterface.createTable(tableName, {
+			id: {
+				type: Sequelize.INTEGER,
+				allowNull: false,
+				autoIncrement: true,
+				primaryKey: true
+			},
+			subscriptionId: {
```
bgvozdev, line 14: Should it be a foreign key with ON DELETE CASCADE flag?

src/models/reposyncstate.ts
```diff
@@ -0,0 +1,127 @@
...
+
+	static async updateRepoForSubscription(subscription: Subscription, repo?: RepositoryData): Promise<RepoSyncState | undefined> {
+		const repoId = repo?.repository?.id;
+		if (!repoId) {
+			return undefined;
+		}
+		const model: RepoSyncState | undefined = await RepoSyncState.findOne({
+			where: {
+				subscriptionId: subscription.id,
+				repoId
```
bgvozdev, line 122: I find it a bit confusing that we are passing "repo" object and really only using "id" from it... Shall we pass repoId instead?

src/models/reposyncstate.ts
```diff
@@ -0,0 +1,127 @@
...
+
+	static async deleteAllFromSubscription(subscription: Subscription): Promise<number> {
+		return RepoSyncState.destroy({
+			where: {
+				subscriptionId: subscription.id
+			}
+		});
+	}
+
+	static async updateFromJson(subscription: Subscription, json: RepoSyncStateObject = {}): Promise<RepoSyncState[]> {
```
bgvozdev, line 50: nit: I feel like `json` might be not the best name here, since technically we are passing an object, not a json string...

src/models/reposyncstate.ts
bgvozdev: Ignore me, I can see it is used later 🤦

bgvozdev (approved):

db/migrations/20211110150400-create-sync-state-table.js
mboudreau: I tried to have it as a foreign key, but was having some issues with sequelize which is apparently solved in the latest version, but didn't want to take on the extra risk right now.  This can be solved at a later stage.

src/models/reposyncstate.ts
mboudreau: I was thinking more of db 'JSON' type

thombergs (approved):
Looks good. I did a test on staging and had a look in the database.

The question I'm having is: what's the migration strategy? How do we get rid of the old database column in the long run? I think we talked about this once and I have an idea, but it would be good to have that on a page somewhere, to share the context with the team.

Issue #868, opened 2021-10-27
Can't configure integration / connect to GitHub organization


fc-reus:
Can't configure integration / connect to GitHub organization.

Error message:
"Looks like something went wrong!
Session information missing - please enable all cookies in your browser settings.
We track these errors automatically, but if the problem persists feel free to contact us. In the meantime, try adding an organization again in Jira."

Activated all cookies and even tried in different browsers, still not working.

rachellerathbone:
Hey @fc-reus. Just to give you an update on this: one of my team members has an open [PR](https://github.com/atlassian/github-for-jira/pull/871) to try and resolve this issue. It's currently going through the review process and should hopefully be merged this coming week. Thanks for your patience on this.

pat-bao:
Hi @rachellerathbone, we're having the same issue as @fc-reus, however we've just seen that the PR has been closed. We're currently unable to link Jira and GitHub, is there a workaround to this issue?

pat-bao:
It seems the issue was cookie related and we've now connected it correctly.

fc-reus:
How did you manage to connect it?

mboudreau:
@fc-reus it'll work if you enable 3rd party cookies and refresh the landing page.  There was a security flaw that was fixed but made it that the user needs to have those cookies enabled for it to work until we get the proper solution out. We have one in the works, but as a workaround your just need to enable that setting in your browser.

mboudreau:
@fc-reus We've had a fix for the blocked 3rd party cookies that just went out.  You can now re-enable your cookie security policy as it was before.  Let me know if this is still affecting you or you can please close this issue if you're satisfied.  Thanks.

fc-reus:
Unfortunately I already used the workaround to connect it, so I can't test it anymore

mboudreau:
I'm going to close this issue in that case.  Open it again if you ever have more issues surrounding this.  Thanks!

PR #871, opened 2021-10-28
ARC-753 use JWT Authentication for /github/login endpoint

KAbakumovAtl:
Added JWT Authentication for /github/login endpoint and save Jira Hostname to the session.

This should solve https://github.com/atlassian/github-for-jira/issues/868 yet keeping the installation flow secure


KAbakumovAtl:
According to the review from the security team. The solution is good enough. Sean tried it in ddev and was satisfied with the results.

KAbakumovAtl:
Discarding this PR, because we figured out that we use **login** function in multiple places already, so we can't assume that we'll have Jira hostname and token in it every time it is called. So we need ot separate setting Jira Hostname from login. 

Hence fixing this PR it will require a lot of job already, we better implement a proper solution.
1. Stop using `/github/login` endpoint for "Add Github Organisation" button. Use /github/configure instead.
2. Change /github/configure to make it do the following:
  a. Check Jira JWT Token, set JiraHostname to the session. 
  b. Check if we have GitHub Auth Token in the context, call "login" function if we don't have it

pat-bao:
Hi @KAbakumovAtl, is there currently any workaround to adding a GitHub Organisation? We're attempting to link Jira to GitHub but currently we just get errors. Will we need to wait until the 'proper solution' is implemented? Thanks.

mboudreau:
@pat-bao if you enable 3rd party cookies in your browsers and refresh the landing page, it'll work

PR #873, opened 2021-10-29
ARC-710 return 401 if no req body

rachellerathbone:
This PR addresses the 500s we are seeing at around 6am AEDT every day. We were able to use the jwt in the request header to determine that the request is coming from a security scan.

When you decode the jwt seen in the req headers in [our logs ](https://splunk.paas-inf.net/en-GB/app/search/search?sid=1635473141.164145_62D9EEBE-1751-48A5-B0C6-CDABAC40F831) you see `“iss”: “csrt-fake-token-ignore”` which comes from here https://github.com/atlassian-labs/connect-security-req-tester/search?q=csrt-fake-token-ignore.

To ignore these scans, we now check for the `clientKey` in `extractInstallationFromJiraCallback`. Additionally, we swapped the order of middleware calls for the `/uninstall` endpoint so that both `/install` and `/uninstall` authenticate first, with a check to make sure either the `jiraHost` or `xdm_e` exist.


src/frontend/verify-jira-jwt-middleware.ts
```diff
@@ -7,6 +7,11 @@ const verifyJiraJwtMiddleware = (tokenType: TokenType) => async (
 	res: Response,
 	next: NextFunction
 ): Promise<void> => {
+	if (!req.query?.xdm_e && !req.body?.jiraHost) {
+		res.status(401);
+		return;
+	}
```
KAbakumovAtl, line 13: We could simply do:
```suggestion
 if(!jiraHost) {
    sendError(res, 401, "Unauthorised");
}
```

after line 15. So we don't have to make changes twice if we have to add another way of getting Jira Host.

(Also everywhere else in this file `sendError` function is used to send error response)

KAbakumovAtl (approved):

src/frontend/verify-jira-jwt-middleware.ts
```diff
@@ -1,13 +1,18 @@
...
-import { TokenType, verifyAsymmetricJwtTokenMiddleware, verifySymmetricJwtTokenMiddleware } from "../jira/util/jwt";
+import { TokenType, verifyAsymmetricJwtTokenMiddleware, verifySymmetricJwtTokenMiddleware, sendError } from "../jira/util/jwt";
 
 const verifyJiraJwtMiddleware = (tokenType: TokenType) => async (
 	req: Request,
 	res: Response,
 	next: NextFunction
 ): Promise<void> => {
 	const jiraHost = (req.query.xdm_e as string) || req.body?.jiraHost;
+
```
KAbakumovAtl, line 11: We need to add a test case for it in `verify-jira-middleware.test.ts`

tarakpurohit (approved):

PR #876, opened 2021-11-01
Remove failure metrics for 401 and 404 errors

tarakpurohit:
Ignores 401, 403 and 404s as failures for SLO.


src/github/middleware.ts
```diff
@@ -235,12 +237,20 @@ export default (
...
+					context.log.error(
+						err,
+						`Error processing the event for Jira hostname '${jiraHost}'`
+					);
+					emitWebhookFailedMetrics(webhookEvent);
+					context.sentry?.captureException(err);
+				} else {
+					context.log.info(
+						err,
+						`Error processing the event for Jira hostname '${jiraHost}'`
```
rachellerathbone, line 251: If we are breaking this out into an error and a warning should we change the messaging here to make that a little clearer? Something like: `Warning: failed to process event for the Jira hostname '${jiraHost}'`. This will just make it easier to distinguish when looking through the logs for `github.webhooks`

- [ ] update message for `github.webhooks` warning

src/github/middleware.ts
```diff
@@ -235,12 +237,20 @@ export default (
...
+				const isWarning = warnOnErrorCodes.find(code => err.message.includes(code));
+				if(!isWarning) {
+					context.log.error(
+						err,
+						`Error processing the event for Jira hostname '${jiraHost}'`
+					);
+					emitWebhookFailedMetrics(webhookEvent);
+					context.sentry?.captureException(err);
+				} else {
+					context.log.info(
```
rachellerathbone, line 249: If this is a warning we should probably change `.info` to `.warn`

- [ ] change `.info` to `.warn`

rachellerathbone (changes requested):

rachellerathbone (approved):

KAbakumovAtl (approved):

PR #878, opened 2021-11-01
Initializing chunking with empty 2d array

mboudreau:


KAbakumovAtl (approved):

src/jira/client/index.ts
```diff
@@ -281,20 +281,27 @@ const batchedBulkUpdate = async (
 	options?: { preventTransitions: boolean }
 ) => {
 	// Initialize with an empty chunk of commits so we still process the request if there are no commits in the payload
-	const commitChunks: JiraCommit[][] = _.chunk(dedupCommits(data.commits), 400);
+	let commitChunks: JiraCommit[][] = _.chunk(dedupCommits(data.commits), 400);
```
ccrolf, line 284: Let's add a feature flag, so there's no functional change when the code goes to production

ccrolf (changes requested):

src/jira/client/index.ts
mboudreau: :+1:

src/jira/client/index.ts
```diff
@@ -280,11 +281,45 @@ const batchedBulkUpdate = async (
 	logger?: Logger,
 	options?: { preventTransitions: boolean }
 ) => {
-	// Initialize with an empty chunk of commits so we still process the request if there are no commits in the payload
-	const commitChunks: JiraCommit[][] = _.chunk(dedupCommits(data.commits), 400);
+	if (booleanFlag(BooleanFlags.REMOVE_WHILE_LOOP, true)) {
```
ccrolf, line 284: Should the default value be false? Otherwise, this code would roll out if LaunchDarkly has an outage

mboudreau: but then it won't be enabled by default on your local...

rachellerathbone (approved):

KAbakumovAtl (approved):

ccrolf (approved):

PR #883, opened 2021-11-02
ARC-764: split processInstallation

bgvozdev:
Splitting processInstallation() onto two functions (feature flagged) and adding "job"  to the logger as a property.

This is needed for the future step once I add deduplication logic around the invocation of the new function.

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -416,7 +416,9 @@ export const processInstallation =
 			const {installationId, jiraHost} = job.data;
 
 			let logger: LoggerWithTarget;
-			if (await booleanFlag(BooleanFlags.SPLIT_PROCESS_INSTALLATION_FUNCTION, false, jiraHost)) {
+
+			const splitProcessInstallationFunctionFlagIsOn = await booleanFlag(BooleanFlags.SPLIT_PROCESS_INSTALLATION_FUNCTION, false, jiraHost);
+			if (splitProcessInstallationFunctionFlagIsOn) {
 				logger = rootLogger.child({job});
 			} else {
 				logger = rootLogger;
@@ -432,8 +434,9 @@ export const processInstallation =
 				jiraHost
 			});
 
-			if (await booleanFlag(BooleanFlags.SPLIT_PROCESS_INSTALLATION_FUNCTION, false, jiraHost)) {
-				return await doProcessInstallation(app, queues, job, installationId, jiraHost, logger);
+			if (splitProcessInstallationFunctionFlagIsOn) {
+				await doProcessInstallation(app, queues, job, installationId, jiraHost, logger);
+				return ;
 			}
 
 			// TODO: remove with SPLIT_PROCESS_INSTALLATION_FUNCTION feature flag cause it was moved to
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -210,13 +210,220 @@ export const isNotFoundError = (
 };
 
 // TODO: type queues
+async function doProcessInstallation(app, queues, job, installationId: number, jiraHost: string, logger: LoggerWithTarget): Promise<void> {
+	const subscription = await Subscription.getSingleInstallation(
+		jiraHost,
+		installationId
+	);
+	// TODO: should this reject instead? it's just ignoring an error
+	if (!subscription) return;
+
+	const jiraClient = await getJiraClient(
+		subscription.jiraHost,
+		installationId,
+		logger
+	);
+	const github = await getEnhancedGitHub(app, installationId);
+	const nextTask = await getNextTask(subscription);
+
+	if (!nextTask) {
+		await subscription.update({ syncStatus: "COMPLETE" });
+		statsd.increment(metricSyncStatus.complete);
+		logger.info({ job, task: nextTask }, "Sync complete");
+		return;
+	}
+
+	await subscription.update({ syncStatus: "ACTIVE" });
+
+	const { task, repositoryId, cursor } = nextTask;
+	let { repository } = nextTask;
+
+	if (!repository) {
+		// Old records don't have this info. New ones have it
+		const { data: repo } = await github.request("GET /repositories/:id", {
+			id: repositoryId
+		});
+		repository = getRepositorySummary(repo);
+		await subscription.updateSyncState({
+			repos: {
+				[repository.id]: {
+					repository
+				}
+			}
+		});
+	}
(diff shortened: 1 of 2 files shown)


src/sync/installation.ts
```diff
@@ -209,14 +209,221 @@ export const isNotFoundError = (
 	return isNotFoundError;
 };
 
+// TODO: type queues
+async function doProcessInstallation(app, queues, job, installationId: number, jiraHost: string, logger: LoggerWithTarget): Promise<void> {
+	const subscription = await Subscription.getSingleInstallation(
```
bgvozdev, line 214: Copy-paste as it is from line 441 and below

bgvozdev:
Tested in stg with FF off and ON

src/sync/installation.ts
```diff
@@ -209,14 +209,221 @@ export const isNotFoundError = (
...
+		jiraHost,
+		installationId
+	);
+	// TODO: should this reject instead? it's just ignoring an error
```
rachellerathbone, line 218: If you haven't already, can you create a ticket for this to make sure we follow up on it?

bgvozdev: Yeah, killing it here https://softwareteams.atlassian.net/browse/ARC-764 in my next PR

src/sync/installation.ts
```diff
@@ -209,14 +209,221 @@ export const isNotFoundError = (
...
 // TODO: type queues
 export const processInstallation =
 	(app: Application, queues) =>
-		async (job, logger: LoggerWithTarget): Promise<void> => {
-			const { installationId, jiraHost } = job.data;
+		async (job, rootLogger: LoggerWithTarget): Promise<void> => {
+			const {installationId, jiraHost} = job.data;
+
+			let logger: LoggerWithTarget;
+			if (await booleanFlag(BooleanFlags.SPLIT_PROCESS_INSTALLATION_FUNCTION, false, jiraHost)) {
```
rachellerathbone, line 419: The default for this flag should be `true` otherwise we'll have to manually change this if we want to hit the new code locally.

- [ ] change false to true

src/sync/installation.ts
```diff
@@ -209,14 +209,221 @@ export const isNotFoundError = (
 	return isNotFoundError;
 };
 
+// TODO: type queues
+async function doProcessInstallation(app, queues, job, installationId: number, jiraHost: string, logger: LoggerWithTarget): Promise<void> {
```
KAbakumovAtl, line 213: I don't think we need to pass `queues` as a function parameter. they are defined at the "../worker/queues" module as a global constant

src/sync/installation.ts
```diff
@@ -209,14 +209,221 @@ export const isNotFoundError = (
...
+			job.sentry.captureException(err);
+
+			statsd.increment(metricSyncStatus.failed);
+
+			throw err;
+		}
+	}
+}
+
 // TODO: type queues
```
KAbakumovAtl, line 412: Same here we don't need to pass queues as a parameter

KAbakumovAtl (approved):
I had one comment about queues parameter, but hence we just extracting the function, it might be OK to keep. (However if we are feature flagging the change anyway we should extract it I believe)

src/sync/installation.ts
```diff
@@ -225,6 +432,12 @@ export const processInstallation =
 				jiraHost
 			});
 
+			if (await booleanFlag(BooleanFlags.SPLIT_PROCESS_INSTALLATION_FUNCTION, false, jiraHost)) {
+				return await doProcessInstallation(app, queues, job, installationId, jiraHost, logger);
```
rachellerathbone, line 436: [Return await is redundant](https://eslint.org/docs/rules/no-return-await)

- [ ] update return: call await func(), followed by return;

src/sync/installation.ts
bgvozdev: If you check line 214, we are receiving it as a parameter. IMHO we should either pass it further or completely remove. Given that the purpose of this PR is to extract part of the function into a separate subfunction, I decided to pass it over, to keep PR having a single focus.

src/sync/installation.ts
```diff
@@ -225,6 +432,12 @@ export const processInstallation =
 				jiraHost
 			});
 
+			if (await booleanFlag(BooleanFlags.SPLIT_PROCESS_INSTALLATION_FUNCTION, false, jiraHost)) {
```
rachellerathbone, line 435: I noticed this line twice: `await booleanFlag(BooleanFlags.SPLIT_PROCESS_INSTALLATION_FUNCTION, false, jiraHost`. After line 416 can we do something like:

```
const splitProcessInstallationFunctionFlagIsOn = await booleanFlag(
				BooleanFlags.SPLIT_PROCESS_INSTALLATION_FUNCTION,
				true,
				jiraHost
			);
```

and then pass that variable?

- [ ] move call to FF to variable

rachellerathbone (changes requested):

src/sync/installation.ts
bgvozdev: According to yesterday's discussion it should be false: https://atlassian.slack.com/archives/C02KGV1MELE/p1635740495053400

rachellerathbone: Interesting. Michel made me set all mine to true 🤔

rachellerathbone (approved):

src/sync/installation.ts
KAbakumovAtl: We can use LaunchDarkly development environment locally, we need to copy LD key to `.env` file for that. That's how I usually test

src/sync/installation.ts
```diff
@@ -416,7 +416,9 @@ export const processInstallation =
 			const {installationId, jiraHost} = job.data;
 
 			let logger: LoggerWithTarget;
-			if (await booleanFlag(BooleanFlags.SPLIT_PROCESS_INSTALLATION_FUNCTION, false, jiraHost)) {
+
+			const splitProcessInstallationFunctionFlagIsOn = await booleanFlag(BooleanFlags.SPLIT_PROCESS_INSTALLATION_FUNCTION, false, jiraHost);
+			if (splitProcessInstallationFunctionFlagIsOn) {
 				logger = rootLogger.child({job});
```
KAbakumovAtl, line 422: Job will be added to all log entries. Is it what we really want?

KAbakumovAtl (approved):

PR #884, opened 2021-11-02
ARC-278 Added push SQS queue

KAbakumovAtl:
In this pull request:
* push queue was added
* Queue handler
* Feature flags which controls which queue to use for push events (Redis or SQS)


test/unit/sync/helpers/sync-helpers.test.ts
```diff
@@ -6,6 +6,8 @@ import {
 } from "../../../../src/sync/installation";
 import {getLogger} from "../../../../src/config/logger";
 
+jest.mock("../../../../src/models");
```
KAbakumovAtl, line 9: I have no idea why I started getting this test failure after my changes:
```
FAIL test/unit/sync/installation.test.ts
  ● Test suite failed to run

    TypeError: Cannot read property 'init' of undefined

      28 | );
      29 |
    > 30 | SubscriptionModel.init(
         |                   ^
      31 | 	{
      32 | 		gitHubInstallationId: DataTypes.INTEGER,
      33 | 		jiraHost: DataTypes.STRING,

      at Object.<anonymous> (src/models/index.ts:30:19)
      at Object.<anonymous> (src/transforms/push.ts:1:1)
      at Object.<anonymous> (src/sqs/push.ts:3:1)
      at Object.<anonymous> (src/sqs/queues.ts:4:1)
      at Object.<anonymous> (src/models/subscription.ts:7:1)
 ```
 
 But adding this mock helped mitigating it

File: src/transforms/push.ts [hunk 2]
```diff
@@ -112,7 +120,6 @@ export const processPush = async (github: GitHubAPI, payload, rootLogger: Logger
 		const webhookReceived = payload.webhookReceived || undefined;
 
 		const log = rootLogger.child({
-			name: PUSH_LOGGER_NAME, // overriding even though it is provided worker; webserver processed the webhooks synchronously, too
```
KAbakumovAtl, line 115: I removed the name override from here because:
1. It was failing when we pass logger from the sqs handler code 
2. We want to differentiate from where this code was called by logger name (From the webhook request handler, from redis message handling or sqs message handling)

KAbakumovAtl:
Currently testing it in staging. 

Ready for review, however a few things which I am still going to add in this PR:
- [ ] tests for sqs push handler
- [ ] consider if we can have a flag to disable the handler reading from sqs

src/transforms/push.ts
```diff
@@ -81,7 +84,12 @@ export async function enqueuePush(
 	jiraHost: string,
 	options?: JobOptions
 ) {
-	return queues.push.add(createJobData(payload, jiraHost), options);
+	const queueMesssagePayload = createJobData(payload, jiraHost);
+	if(await booleanFlag(BooleanFlags.SEND_PUSH_TO_SQS, false, jiraHost)) {
+		return sqsQueues.push.sendMessage(queueMesssagePayload);
+	} else {
+		return queues.push.add(createJobData(queueMesssagePayload, jiraHost), options);
```
KAbakumovAtl, line 91: Good lesson to learn here, that even the smallest refactoring can cause problems.

What I wanted to do is just to extract `const queueMesssagePayload = createJobData(payload, jiraHost);` to a variable, however I've put it in a wrong place (as an argument of createJobData function instead of createJobData function invocation itself)

If we deployed it to production it would be broken with feature flag OFF.
cc @mboudreau , @atrigueiro 

Now I'll need to figure out why there were no tests

KAbakumovAtl:
I've added tests in a separate PR https://github.com/atlassian/github-for-jira/pull/884

tarakpurohit (approved):

bgvozdev (approved):

tarakpurohit (approved):

src/sqs/push.ts
```diff
@@ -0,0 +1,43 @@
...
+	shas: { id: string, issueKeys: string[] }[],
+	jiraHost: string,
+	installationId: number,
+	webhookId: string,
+	webhookReceived?: Date,
+}
+
+export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> = {
+
+	async handle(context) {
```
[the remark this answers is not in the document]
KAbakumovAtl: I've added a bunch of stuff in the context which we might need during the processing (we can pass for example executionId, which we can use as a jobId for initial sync locks.)

I'd keep it as is for now, but we can refactor later

src/sqs/push.ts
```diff
@@ -0,0 +1,43 @@
+import { MessageHandler } from './index'
+import enhanceOctokit from "../config/enhance-octokit";
+import {processPush} from "../transforms/push";
+import app from "../worker/app";
+import {wrapLogger} from "probot/lib/wrap-logger";
+
+
```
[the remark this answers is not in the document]
KAbakumovAtl: Shame on me. Will fix in the next PR

PR #886, opened 2021-11-03
ARC-764: add Das Deduplicator (tm)

bgvozdev:
This is the idea I'm spiking: to use Redis to prevent multiple jobs running at the same time with the same installation ID.

I tried to describe everything in the comments rather than putting it here in the description (to keep it closer to the code).

This code is not ready yet to be plugged in into our sync because during the sync we are rescheduling the jobs straight away, which can be picked by other nodes BEFORE we end the execution of the current job. I'm going to refactor it in a different PR.

Keen to hear your thoughts (will also ping you in Slack)


src/sync/deduplicator.ts
```diff
@@ -0,0 +1,200 @@
...
+ * to inform other executors not to run the same job ("same"-ness is defined by a key). At the end of the execution it
+ * removes the previously raised flag.
+ *
+ * As all "very simple" ideas, it has several challenges, namely:
+ *
+ * THE FIRST CHALLENGE is: what if the executor dies (e.g. the EC2 node is rotated) during the execution? This might
+ * have a flag raised but not being deleted.
+ *
+ * The implemented solution: save the timestamp of the flag and keep updating it in the executor while it is doing its
+ * job. If it dies, the flag will be invalidated in some reasonable time.
```
KAbakumovAtl, line 15: I guess using normal TTL would do the job, hence in the unlike situation of EC2 death, we'll only delay execution for a while (Then the message will be retried and TTL expired)

If we'll be updating time as we process we close the door for having retries on the timeout hence if we recently updated the timestamp on retry the message can be discarded

src/sync/deduplicator.ts
```diff
@@ -0,0 +1,200 @@
...
+ *
+ * Always rescheduling is bad: if we have the issue with duplicates, this will keep queue increasing, not good for Redis'
+ * CPU.
+ *
+ * Always dropping is dangerous: what if the flag belongs to a "dead" executor, however it wasn't invalidated yet?
+ *
+ * The implemented solution: once the executor observes the flag of a concurrent executor, it will watch for it for
+ * some period of time and if it is moving (timestamp is being updated, which means the executor is doing "something"),
+ * we are safe to discard the job as a detected duplicate; however if it is not moving it is dangerous to discard it
+ * and should be rescheduled instead.
```
KAbakumovAtl, line 27: We have another option:
* Always discard if the flag is on
* Periodically scan the table for unprocessed syncs and then schedule a message if the flag is not there for that sync

src/sync/deduplicator.ts
```diff
@@ -0,0 +1,200 @@
...
+		await this.redis.unlink(jobKey);
+	}
+
+	async setInProgressFlag(jobKey: string, executionId: string): Promise<void> {
+		const flag: Flag = {
+			executionId,
+			timestamp: new Date().getTime()
+		};
+		// We don't want to pollute redis, autoexpire after the flag is not being updated
+		const REDIS_CLEANUP_TIMEOUT = 24 * 3600 * 1000;
```
KAbakumovAtl, line 115: 24 hours looks like a huge TTL, especially when job updates it periodically

src/sync/deduplicator.ts
```diff
@@ -0,0 +1,200 @@
...
+	 * 				 it WILL BE PROCESSED NEXT
+	 *  E_OTHER_WORKER_DOING_THIS_JOB - we have determined that the current task has been processed right now,
+	 *  																you might want to DISCARD the job to avoid duplication
+	 *  E_UNSURE_TRY_AGAIN_LATER - we are not sure if the other process is processing the job right now or not; you might like to
+	 *  					  						   RESCHEDULE the task for good (remember there's a timeout for the flag)
+	 *  <an exception is thrown> - same as E_OK with the only difference that the underlying job threw an exception; the
+	 *  													 flag was removed.
+	 */
+	async executeWithDeduplication(jobKey: string, jobRunner: () => Promise<void>): Promise<DeduplicatorResult> {
+		const executionId = 'execid' + Math.random();
```
KAbakumovAtl, line 177: Let's use `uuid` here. It is a de facto standard for random identifiers generation around the code

Also we need to add this executionId to logger parameters. Also we need to choose different name for it, hence we currently have "executionId" parameter in logs added by SQS Listener, which is message processing attempt ID

KAbakumovAtl: - [ ] Use uuid instead of number
- [ ] Rename "executionId" to something else

KAbakumovAtl: Also we can solve renaming problem if we take executionID from the SQS processing context. So we'll use the same identifier for SQS message processing attempt ID and for deduplication.

KAbakumovAtl (commented):
I have a few comments and overall I am OK with the approach. 

However I'd consider to keep it simple, hence at the moment it is a bit logic-heavy.

To me there should be just 2 methods in state
* Set flag
* Remove flag

Set flag will have its own TTL which will be equal maximum message processing time (Queue processing timeout)

Also we need to keep in mind that we are fine with occasional inconsistency here. It is not a problem if once in a while we have multiple messages running for the same job, hence they are idempotent anyway. Tolerating occasional race conditions allows us to simplify logic which will make it easier to maintain. As a result, will improve the system stability overall.

KAbakumovAtl:
> This code is not ready yet to be plugged in into our sync because during the sync we are rescheduling the jobs straight away, which can be picked by other nodes BEFORE we end the execution of the current job. I'm going to refactor it in a different PR.

That's OK, because that message will be discarded hence the job is already in progress. 

We shouldn't change anything in the existing sync code hence we are going to rewrite it anyway, there is already some code committed by Tom in place

src/sync/deduplicator.ts
bgvozdev: This TTL is for garbage collection; it is not an operational one.

src/sync/deduplicator.ts
bgvozdev: The problem with a single TTL is that it is unclear how long the job takes to execute. We might drop TTL before the job is finished, which will bring us back to duplication problem.

Having a preriodically updating flag looked more robust to me.

bgvozdev:
> > This code is not ready yet to be plugged in into our sync because during the sync we are rescheduling the jobs straight away, which can be picked by other nodes BEFORE we end the execution of the current job. I'm going to refactor it in a different PR.
> 
> That's OK, because that message will be discarded hence the job is already in progress. (Or we can add a delay for the message, if this rescheduling is in the end of the job)

If next message is discarded while the first message is in progress, we will have a staled job.

> 
> We shouldn't change anything in the existing sync code hence we are going to rewrite it anyway, there is already some code committed by Tom in place

This is a very simple drop-replacement: everywhere we call `queues.installations.add`, we will be calling `scheduleNext`, which is passed as a parameter

src/sync/deduplicator.ts
bgvozdev: > Periodically scan the table for unprocessed syncs and then schedule a message if the flag is not there for that sync

TBH I don't like this suggestion. Here all the rescheduling logic is encapsulated in a single place and we don't need the database at all. In this suggestion we will have some periodic job.

And then there's a question which job should be considered as "staled", what's the criteria? A timeout? What is this is a long-running job?

src/sync/deduplicator.ts
```diff
@@ -0,0 +1,200 @@
...
+			// Cannot tell, might have finished already (race condition)
+			return false;
+		}
+		const flagOldParsed = JSON.parse(flagOld) as Flag;
+		if (flagOldParsed.executionId !== executionId) {
+			// Cannot tell, other node might haved pick up the job and updated the flag
+			return false;
+		}
+
+		await sleep(executorFlagUpdateTimeoutMsecs * 2);
```
bgvozdev, line 131: use some pre-defined constant for sleep

src/sync/deduplicator.ts
```diff
@@ -0,0 +1,200 @@
...
+	 */
+	async executeWithDeduplication(jobKey: string, jobRunner: () => Promise<void>): Promise<DeduplicatorResult> {
+		const executionId = 'execid' + Math.random();
+
+		const inProgressFlagExecutionId = await this.inProgressStorage.hasInProgressFlag(jobKey, this.executorFlagUpdateTimeoutMsecs * 10);
+		if (inProgressFlagExecutionId) {
+			if (await this.inProgressStorage.isExecutionLive(jobKey, inProgressFlagExecutionId, this.executorFlagUpdateTimeoutMsecs)) {
+				return DeduplicatorResult.E_OTHER_WORKER_DOING_THIS_JOB;
+			}
+			return DeduplicatorResult.E_NOT_SURE_TRY_AGAIN_LATER;
```
tarakpurohit, line 184: You could use squash these line into one  like this:
`
return await this.inProgressStorage
  .isExecutionLive(jobKey, inProgressFlagExecutionId, this.executorFlagUpdateTimeoutMsecs)) ? 
      DeduplicatorResult.E_OTHER_WORKER_DOING_THIS_JOB:DeduplicatorResult.E_NOT_SURE_TRY_AGAIN_LATER;`
`

tarakpurohit (dismissed):

src/sync/deduplicator.ts
KAbakumovAtl: Didn't know! Thanks for clarifying

KAbakumovAtl (approved):

tarakpurohit (approved):

PR #887, opened 2021-11-03
ARC-771 installation duplication mitigation

rachellerathbone:
This is the first of 3 steps to resolve an issue blocking a customer from using the app. The following 2 steps will be covered in the tickets:

- ARC-772: Clean up installations table to remove duplicate entries 
- ARC-769: Add unique constraints in the installations table

This PR updates queries to the installations table as we now order installations by id. This ensures that the most recent entry is retrieved instead of the oldest (issue customer is currently having). We had discussed potentially using `updatedAt` or `createdAt` to order the entries in the installations table but determined the ID was the better option. Date values have a precision of 1 second so, while unlikely, it is not impossible that we run into an edge where a host is installed twice a near identicial timestamp. IDs, on the other hand, increment in ascending order so it is easy to reverse this order and know we can retrieve the latest entry.


src/models/installation.ts
```diff
@@ -31,11 +35,43 @@ export default class Installation extends Sequelize.Model {
...
+				order: [["id", "DESC"]]
 			}
-		});
+		} else {
+			payload = {
+				where: {
+					jiraHost: host,
+				}
+			}
+		}
```
mboudreau, line 52:
```
const payload = {
   where: {
      jiraHost: host,
   }
}
if (await sortInstallationsByIdFlagIsOn(host)) {
   payload.order = [["id", "DESC"]]
}
```

src/models/installation.ts
```diff
@@ -31,11 +35,43 @@ export default class Installation extends Sequelize.Model {
...
+				},
+				order: [["id", "DESC"]]
+			}
+		} else {
+			payload = {
+				where: {
+					jiraHost: host,
+				}
+			}
+		}
```
mboudreau, line 72: same as above here

test/unit/models/index.test.ts
```diff
@@ -133,4 +133,38 @@ describe("test installation model", () => {
 			expect(subscription.jiraHost).toBe(renamedInstallPayload.baseUrl);
 		}
 	});
+
+	it("should return the most recent entry if there are duplicate hosts", async () => {
```
mboudreau, line 137: This should be made into a describe and then separate each scenario into it's own test.  You've changed 2 functions so both needs to be tests for 0, 1 and many results.

mboudreau: You're also missing the scenario of the same host/clientKey being installed and how that would work as per the scenario of the client.  You'll have to use `Installation.create` directly instead of using `install` to add those records to the DB.

mboudreau (changes requested):

src/models/installation.ts
```diff
@@ -13,6 +14,19 @@ export const getHashedKey = (clientKey: string): string => {
...
+	booleanFlag(BooleanFlags.SORT_INSTALLATIONS_BY_ID, true, jiraHost);
+
+// interface WhereQuery {
+// 	jiraHost: string;
+// }
+// interface OrderQuer {
+// 	jiraHost: string;
+// }
+// interface QueryPayload {
+// 	where: WhereQuery;
```
mboudreau, line 27: should probably remove this :)

rachellerathbone: You think so?

test/unit/models/index.test.ts
```diff
@@ -133,4 +133,133 @@ describe("test installation model", () => {
 			expect(subscription.jiraHost).toBe(renamedInstallPayload.baseUrl);
 		}
 	});
+
+	describe("duplicate entries in Installations table", () => {
+		it("should return no installations if jiraHost is not found", async () => {
+			const jiraHost = "https://myfakejirasite.net";
```
mboudreau, line 139: there's already a global jiraHost for tests.  I suggest you just that.

test/unit/models/index.test.ts
```diff
@@ -133,4 +133,133 @@ describe("test installation model", () => {
...
+			const installations = await Installation.getAllForHost(jiraHost);
+			const installation = await Installation.getForHost(jiraHost);
+
+			expect(installations.length).toEqual(0);
+			expect(installation).toEqual(null);
+		});
+
+		it("should return a single installation if 1 host is installed", async () => {
+			await Installation.install({
```
mboudreau, line 148: Since we're testing the 2 functions you changed and not the `install` function.  I suggest you change all the `install` functions to `create` instead.  That way it's "closer to the metal" and should be better to avoid side effects.

test/unit/models/index.test.ts
```diff
@@ -133,4 +133,133 @@ describe("test installation model", () => {
...
+
+			const installations = await Installation.getAllForHost(jiraHost);
+			const installation = await Installation.getForHost(jiraHost);
+
+			expect(installations.length).toEqual(1);
+			expect(installation?.jiraHost).toEqual(jiraHost);
+		});
+
+		it("should return the most recent entry if there are 2 duplicate entries", async () => {
+			// Install jira host
```
mboudreau, line 162: isn't this tested in the scenario above?

test/unit/models/index.test.ts
```diff
@@ -133,4 +133,133 @@ describe("test installation model", () => {
...
+			await Installation.install({
+				host: jiraHost,
+				sharedSecret: "badsecret",
+				clientKey: "1234567",
+			});
+
+			const singleInstallation = await Installation.getAllForHost(jiraHost);
+			expect(singleInstallation.length).toEqual(1);
```
mboudreau, line 170: isn't this tested in the scenario above?

test/unit/models/index.test.ts
```diff
@@ -133,4 +133,133 @@ describe("test installation model", () => {
...
+			installations.forEach((installation) => {
+				// both installations should have the same host.
+				expect(installation.jiraHost).toBe(jiraHost);
+			});
+
+			// Ids of installations should decrement to ensure most recently added is retrieved first
+			expect(installations[0].id).toBeGreaterThan(installations[1].id);
+
+			const firstRetreivedInstallation = await Installation.getForHost(jiraHost);
+			expect(firstRetreivedInstallation?.id).toEqual(installations[0].id);
```
mboudreau, line 191: this is a different scenario.  Try to keep your tests to a single function that you're testing.

src/models/installation.ts
```diff
@@ -31,11 +34,31 @@ export default class Installation extends Sequelize.Model {
...
+	static async getAllForHost(host: string): Promise<Installation[]> {
+		const payload: any = {
+			where: {
+				jiraHost: host,
+			}
+		}
+
+		if (await sortInstallationsByIdFlagIsOn(host)) {
+			payload.order = [["id", "DESC"]]
+		}
```
KAbakumovAtl, line 59: /super-nit if it is a new function which never used (and we don't use it anywhere yet) we don't need to use a feature flag here. (Also ordering doesn't matter that much when we retrieve all elements anyway.

KAbakumovAtl (dismissed):
Well done! 👍

src/models/installation.ts
rachellerathbone: Just trying to play it ultra safe. Christian told me, until the app is stable, all changes should have tests and FFs. I agree, it's probably overkill here but easy to clean up.

KAbakumovAtl (approved):

mboudreau (approved):

PR #888, opened 2021-11-03
ARC-744 add commits integration test

KAbakumovAtl:
Added integration tests for commit webhooks.

- Fixed existing tests which were missing assertions
- Added new end to end test which test the push path from receiving the webhook to sending event to Jira

### DIFF
--- a/src/util/webhooks.ts
+++ b/src/util/webhooks.ts
@@ -24,7 +24,7 @@ export const emitWebhookProcessedMetrics = (
 	try {
 		// only send logs if time of webhookReceived occurred before the currentTime
 		// and if webhookReceivedTime is not null/undefined
-		if (webhookReceivedTime < currentTime) {
+		if (webhookReceivedTime <= currentTime) {
 			const timeToProcessWebhookEvent = getCurrentTime() - webhookReceivedTime;
 
 			contextLogger.info(
--- a/src/worker.ts
+++ b/src/worker.ts
@@ -1,186 +1,13 @@
 import "./config/env"; // Important to be before other dependencies
 import "./config/proxy"; // Important to be before other dependencies
 import throng from "throng";
-import { v4 as uuidv4 } from "uuid";
-import {isNodeProd, isNodeTest} from "./util/isNodeEnv";
-import { listenToMicrosLifecycle } from "./services/micros/lifecycle";
-import { ClusterCommand, sendCommandToCluster } from "./services/cluster/send-command";
-import * as Sentry from "@sentry/node";
-
-import { discovery, DISCOVERY_LOGGER_NAME } from "./sync/discovery";
-import { INSTALLATION_LOGGER_NAME, processInstallation } from "./sync/installation";
-import { processPushJob, PUSH_LOGGER_NAME } from "./transforms/push";
-import metricsJob, { METRICS_LOGGER_NAME } from "./worker/metrics-job";
-import statsd from "./config/statsd";
-import app, { probot } from "./worker/app";
-import AxiosErrorEventDecorator from "./models/axios-error-event-decorator";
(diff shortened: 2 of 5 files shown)


src/util/webhooks.ts
```diff
@@ -24,7 +24,7 @@ export const emitWebhookProcessedMetrics = (
 	try {
 		// only send logs if time of webhookReceived occurred before the currentTime
 		// and if webhookReceivedTime is not null/undefined
-		if (webhookReceivedTime < currentTime) {
+		if (webhookReceivedTime <= currentTime) {
```
KAbakumovAtl, line 27: I had to change it to `<=` here so it will be easier to use this code in tests when we mock time to a constant value

src/worker.ts
```diff
@@ -208,7 +208,9 @@ if (isNodeProd()) {
 		},
 		lifetime: Infinity
 	});
-} else {
+
+	//Don't start worker automatically in test environments to allow calling start() and stop() functions in unit tests
+} else if (!isNodeTest()) {
```
KAbakumovAtl, line 213: So, this change needs some advice from @mboudreau

I want to have an end-to-end integration test which will invovle the queue. For that purpose I had to use "Start" and "stop" functions from this file to start worker in unit tests.

Unfortunately we can't just import them, because as soon as we do it this code gets executed. So I've added a check to make sure it won't start automatically in the test environment, so we can use it in unit tests.

Another way of doing it is to extract queues startup code (start() and stop() functions to a separate file, but I wanted to avoid changing too much code in this PR. So I am doing this way.

I wonder if it can affect something? (Do we use test env when we run it in Docker?)

test/safe/push.test.ts
```diff
@@ -3,14 +3,65 @@ import nock from "nock";
...
+		}
+		nock.cleanAll();
+		jest.restoreAllMocks();
+	});
 
 	const getAtlassianUrl = () => process.env.ATLASSIAN_URL || "";
 	const createJob = (payload, jiraHost:string) => ({data: createJobData(payload, jiraHost)} as any);
 
+	//TODO
 	describe.skip("add to push queue", () => {
```
KAbakumovAtl, line 65: The tests which test that the message was send to the queue are not there. 

However there are some queue-agnostic end-to-end test below

test/safe/push.test.ts
```diff
@@ -390,4 +465,134 @@ describe("GitHub Actions", () => {
 			await expect(processPushJob(app)(job, getLogger('test'))).toResolve();
 		});
 	});
+
+	describe("end 2 end tests with queue", () => {
```
KAbakumovAtl, line 469: That's the end to end test involving queue.

For SQS pullrequest https://github.com/atlassian/github-for-jira/pull/884 I'll add more test cases with sqs feature flag on and off

bgvozdev (approved):

src/worker.ts
KAbakumovAtl: Had some chat with @tarakpurohit, we decided to move start() and stop() function to a seaparate file

KAbakumovAtl: - [ ] extract start and stop funcitons from the worker entry point code

src/worker.ts
```diff
@@ -1,186 +1,13 @@
...
-		queues.discovery.close(),
-		queues.installation.close(),
-		queues.push.close(),
-		queues.metrics.close()
-	]);
-
-	sqsQueues.stop();
-
-	running = false;
-}
```
KAbakumovAtl, line 183: Moved this functions to `src/worker/startup.ts` so we can use them without triggering the worker startup logic

tarakpurohit (approved):

bgvozdev (approved):

PR #889, opened 2021-11-04
ARC-779 Fixed worker restart and added a test

KAbakumovAtl:
In this PR i've fixed the queue re-initialisation. It turned out that our rollbacks could be broken, hence when we are trying to add listener to the queue again, it will fail if listener is already there.

I've changed the way stop and start works. Instead of shutting down the queue we just pausing it and then resuming on restart.

+Added tests which checks that worker can be restarted properly

### DIFF
--- a/src/worker/startup.ts
+++ b/src/worker/startup.ts
@@ -112,6 +112,7 @@ const sendQueueMetrics = async () => {
 };
 const commonMiddleware = (jobHandler, loggerName: string) => logMiddleware(sentryMiddleware(setDelayOnRateLimiting(jobHandler)), loggerName);
 let running = false;
+let firstInitialisation = true;
 let timer: Timeout;
 
 // Start function for Node cluster worker
@@ -125,17 +126,24 @@ export async function start() {
 	// exposing queue metrics at a regular interval
 	timer = setInterval(sendQueueMetrics, 60000);
 
-	// Start processing queues
-	queues.discovery.process(5, commonMiddleware(discovery(app, queues), DISCOVERY_LOGGER_NAME));
-	queues.installation.process(
-		Number(CONCURRENT_WORKERS),
-		commonMiddleware(processInstallation(app, queues), INSTALLATION_LOGGER_NAME)
-	);
-	queues.push.process(
-		Number(CONCURRENT_WORKERS),
-		commonMiddleware(processPushJob(app), PUSH_LOGGER_NAME)
-	);
-	queues.metrics.process(1, commonMiddleware(metricsJob, METRICS_LOGGER_NAME));
+	if (firstInitialisation) {
+		logger.info("Initialising queue listeners");
+		// Start processing queues
+		queues.discovery.process(5, commonMiddleware(discovery(app, queues), DISCOVERY_LOGGER_NAME));
+		queues.installation.process(
+			Number(CONCURRENT_WORKERS),
+			commonMiddleware(processInstallation(app, queues), INSTALLATION_LOGGER_NAME)
+		);
+		queues.push.process(
+			Number(CONCURRENT_WORKERS),
+			commonMiddleware(processPushJob(app), PUSH_LOGGER_NAME)
+		);
+		queues.metrics.process(1, commonMiddleware(metricsJob, METRICS_LOGGER_NAME));
+		firstInitialisation = false;
+	} else {
+		//resumes all Bull queues globally
+		await queues.discovery.resume(false)
+	}
 
 	sqsQueues.start()
 
@@ -154,15 +162,22 @@ export async function stop() {
 	// Stop sending metrics for queues
 	clearInterval(timer);
 
-	// Close all queues
+	// Pauses messagse processing in all queue.
+	await queues.discovery.pause(false, true)
+
+	sqsQueues.stop();
+
+	running = false;
+}
+
+/**
+ * Shuts down the worker and releases all resources. Currently used to cleanup after tests
+ */
+export async function shutdown() {
 	await Promise.all([
 		queues.discovery.close(),
 		queues.installation.close(),
 		queues.push.close(),
 		queues.metrics.close()
 	]);
-
-	sqsQueues.stop();
-
-	running = false;
 }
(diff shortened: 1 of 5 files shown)


test/unit/worker/startup.test.ts
```diff
@@ -0,0 +1,185 @@
...
+	}
+
+	it("should normally process push events sync after restart", async () => {
+
+		mockGitHubAndJiraResponsesForRequestProcessing();
+
+		await stop();
+
+		await start();
+
```
KAbakumovAtl, line 156: Currently this test is done by sending the message to the `push` queue and checking if it will be processed. We use Github and Jira rest APIs calls as indicators.

This test checks that these APIs will be called after we restarted the worker and scheduled a message.
The second test checks that we don't call these APIs after we stopped processing.

test/unit/worker/startup.test.ts
```diff
@@ -0,0 +1,185 @@
...
+	it("should not process push events when stopped", async () => {
+
+		mockGitHubAndJiraResponsesForRequestProcessing();
+
+		//await stop();
+
+		await enqueueTestPushEvent();
+
+		delay(1000)
```
KAbakumovAtl, line 175: There is no really better way to test it. We need to make sure that we NOT processing messages from the queue, which is tricky to do without waiting, hence we don't know when exactly worker is going to read.

Should be OK for now

test/utils/delay.ts
```diff
@@ -0,0 +1,3 @@
+export default function delay(time: number) {
+	return new Promise(resolve => setTimeout(resolve, time));
```
mboudreau, line 2: pretty sure jest already has this built in.

src/worker/startup.ts
```diff
@@ -154,15 +163,23 @@ export async function stop() {
...
+/**
+ * Shuts down the worker and releases all resources. Currently used to cleanup after tests
+ */
+export async function shutdown() {
 	await Promise.all([
 		queues.discovery.close(),
 		queues.installation.close(),
 		queues.push.close(),
 		queues.metrics.close()
 	]);
```
mboudreau, line 184: If we're just pausing, do we need shutdown?  There's no way of restarting the redis connection without recreating the queues from scratch.

src/worker/startup.ts
```diff
@@ -125,17 +126,25 @@ export async function start() {
...
+			Number(CONCURRENT_WORKERS),
+			commonMiddleware(processInstallation(app, queues), INSTALLATION_LOGGER_NAME)
+		);
+		queues.push.process(
+			Number(CONCURRENT_WORKERS),
+			commonMiddleware(processPushJob(app), PUSH_LOGGER_NAME)
+		);
+		queues.metrics.process(1, commonMiddleware(metricsJob, METRICS_LOGGER_NAME));
+		firstInitialisation = false;
+	} else {
```
mboudreau, line 143: Instead of this, I would set all queues to paused state when we create them in queues.ts, then add the processor to the queues at the top level of this file, then the start/stop functions just resume/pause the queues.  

Make it simpler to understand and don't have to set state.

mboudreau: extra benefit, we can get rid of the `running` var as well.

mboudreau (changes requested):

src/worker/startup.ts
KAbakumovAtl: I don't want to play with it too much given that we are going to get rid of Redis soon. It is good enough to stay in the repository for 2 more weeks

src/worker/startup.ts
KAbakumovAtl: As the function comment says, we only use it for tests cleanup. We don't re-create queues in production

test/utils/delay.ts
KAbakumovAtl: For reasons explained above, we have to wait

test/unit/worker/startup.test.ts
mboudreau: You're missing the await on this.

mboudreau: The [jest recommended way is to use fake timers and move the clock forward](https://jestjs.io/docs/timer-mocks) so that your tests doesn't actually have to wait.

test/unit/worker/startup.test.ts
```diff
@@ -0,0 +1,185 @@
...
+	afterEach(async () => {
+		try {
+			if (!nock.isDone()) {
+				// eslint-disable-next-line jest/no-jasmine-globals
+				fail("nock is not done yet");
+			}
+		} finally {
+			nock.cleanAll();
+			jest.restoreAllMocks();
+
```
mboudreau, line 59: This is already done globally for every tests.

test/unit/worker/startup.test.ts
```diff
@@ -0,0 +1,185 @@
...
+			// eslint-disable-next-line jest/no-standalone-expect
+			expect(jiraNock.pendingMocks()).toEqual([]);
+		})
+	});
+
+	it("should not process push events when stopped", async () => {
+
+		mockGitHubAndJiraResponsesForRequestProcessing();
+
+		//await stop();
```
KAbakumovAtl, line 171: oops

src/worker/startup.ts
```diff
@@ -154,15 +163,23 @@ export async function stop() {
 	// Stop sending metrics for queues
 	clearInterval(timer);
 
-	// Close all queues
+	// Pauses message processing in all queues globally.
+	// The first parameter of this function "isLocal" controls if all queues will be paused
+	await queues.discovery.pause(false, true)
```
bgvozdev, line 168: Why are we not doing same for other queues (e.g. installation)?

src/worker/startup.ts
mboudreau: It's just slightly moving code around to make it simpler and less prone to issues.  It'll be in the repo for more than 2 weeks...

src/worker/startup.ts
KAbakumovAtl: Poor and misleading bull queues documentation.

Should have done it for all the queues. But it doesn't work for some reason at all.

KAbakumovAtl:
It doesn't work. We'll have to deal with the fact we don't have rollback. Tried everything, including creating queues again.

PR #890, opened 2021-11-04
ARC-774 update logs to investigate missing prs

rachellerathbone:
Adding some extra logs & log args to make it easier to investigate failing PR webhooks. Going to follow this up with ARC-780 to validate issue keys with Jira before sending requests to the Jira REST API.


src/jira/client/index.ts
```diff
@@ -194,13 +194,15 @@ async function getJiraClient(
 						await subscription?.update({ syncWarning: issueKeyLimitWarning });
 					}
 
-					return await batchedBulkUpdate(
+					const batchBulkUpdateReq = await batchedBulkUpdate(
```
rachellerathbone, line 197: Updated as the pattern of `return await` is [redundant](https://eslint.org/docs/rules/no-return-await).

src/jira/client/index.ts
```diff
@@ -361,25 +363,31 @@ export const getTruncatedIssuekeys = (data: IssueKeyObject[] = []): IssueKeyObje
...
-		);
-		repositoryObj.branches.forEach((branch) => {
-			if (branch.lastCommit) {
-				branch.lastCommit = updateIssueKeysFor(
-					[branch.lastCommit],
-					mutatingFunc
-				)[0];
-			}
-		});
+	const logger = getLogger("jiraClient.repository.update");
```
KAbakumovAtl, line 366: Can we pass logger as a function parameter instead of creating a new one here? (Otherwise we won't be able to connect it by requestId/ jobId)

src/jira/client/index.ts
```diff
@@ -361,25 +363,31 @@ export const getTruncatedIssuekeys = (data: IssueKeyObject[] = []): IssueKeyObje
...
+
+	if (repositoryObj.commits || repositoryObj.branches) {
```
KAbakumovAtl, line 368: Let's avoid modifying logic here, let's use only as a logging condition

KAbakumovAtl, line 368:
```suggestion
	if (!repositoryObj.commits && !repositoryObj.branches) {
	  logger.warn("No branches or commits found. Cannot update.")
	}
```

KAbakumovAtl: (with no else, so we won't increase cyclomatic complexity)

KAbakumovAtl (commented):
Left some todos [ ]

Would be happy to approve after they fixed

atrigueiro (dismissed):

src/jira/client/index.ts
```diff
@@ -381,6 +384,10 @@ const updateRepositoryIssueKeys = (repositoryObj, mutatingFunc) => {
 			}
 		});
 	}
+
+	if (!repositoryObj.commits && !repositoryObj.branches) {
+		logger.warn("No branches or commits found. Cannot update.")
```
mboudreau, line 389: this will error if logger isn't set since it's optional.

KAbakumovAtl (approved):

mboudreau (approved):

PR #891, opened 2021-11-05
ARC-750 fix tests

mboudreau:
Our global `afterEach` wasn't working as intended because of the order of execution on Jest is just plain wrong.  Instead, I've combined both `afterEach` together, but this was now failing some tests that should have failed long ago.  Fixed those as well.


test/unit/sync/branch.test.ts
```diff
@@ -233,7 +233,7 @@ describe.skip("sync/branches", () => {
 				add: jest.fn()
 			}
 		};
-		await expect(processInstallation(app, queues)(job, getLogger('test'))).toResolve();
+		await expect(processInstallation(app, queues)(job, getLogger("test"))).toResolve();
```
KAbakumovAtl, line 236: Can we avoid changing format everywhere and doing changes in the same PR?
Makes it extremely difficult to read

src/config/logger.ts
```diff
@@ -30,6 +31,13 @@ const requestSerializer = (req: Request) => (!req || !req.socket) ? req : {
 	body: req.body
 };
 
+const jobSerializer = (job: Job) => job && {
+	...job,
+	// Removing potentially extremely large amount of data from logs
+	data: undefined,
+	queue: undefined
```
KAbakumovAtl, line 38: Why do we remove "queue" name from logs? 

It doesn't hold large amount of data, it is just a queue name
image

Also instead of overriding this in serialiser we should properly remove this from where it is being logged.
However it is not a scope of this PR, so should be done in a separate one

- [ ] remove unrelated logging changes from this PR

src/github/middleware.ts
```diff
@@ -138,9 +138,8 @@ export default (
 		const subscriptions = await Subscription.getAllForInstallation(
 			gitHubInstallationId
 		);
-		const jiraSubscriptionsCount = subscriptions.length;
 
-		if (!jiraSubscriptionsCount) {
+		if (!subscriptions.length) {
```
KAbakumovAtl, line 142: Why do we need to inline it?

KAbakumovAtl (commented):
This PR changes 48 files and incorporates the following changes unrelated ot its purpose:
* Refactoring of the production code
* Logging tunning
* Format changes

We should split it to smaller PR so they'll be reviewable. Now it is too huge to review

src/util/jira.ts
```diff
@@ -1,13 +1,14 @@
...
-	return author.login || author.name ? {
-		avatar: author.avatar_url || author.avatarUrl || `https://github.com/users/${author.login}.png`,
+	return author.login || author.name ? _.pickBy({
+		avatar: author.avatar_url || author.avatarUrl || (author.login ? `https://github.com/users/${author.login}.png` : undefined),
 		name: author.name || author.user?.name || author.login || author.email?.match(/^(.*)@/)?.pop() || "unknown",
 		email: author.email || `${author.login}@noreply.user.github.com`,
-		url: author.html_url || author.url || author.user?.url || `https://github.com/users/${author.login}`
-	} : {
+		url: author.html_url || author.url || author.user?.url || (author.login ? `https://github.com/users/${author.login}` : undefined)
+	}) as JiraAuthor : {
```
KAbakumovAtl, line 11: That's definirely should be done in another PR. We already faced problems in the past when doing refactoring with lodash

mboudreau: this was breaking some tests and it did find some edge cases, hence the fix.

src/github/middleware.ts
mboudreau: to make it explicit and easier to read

src/config/logger.ts
mboudreau: That's different: `logger.info({queue: queueName, queueMetrics: jobCounts}` we're still logging queue name, I'm just removing the queue object when serializing a whole job object because it also has redis information and the likes.  I could keep the queue name though.

test/safe/deployment.test.ts
```diff
@@ -1,77 +1,92 @@
...
+							email: "test-branch-author-name@github.com",
+							date: "test-branch-author-date"
+						},
+						message: "[TEST-123] test-commit-message"
+					},
+					html_url: `test-repo-url/commits/${sha}`
 				});
 
-			jiraNock.get("/rest/api/latest/issue/TEST-123?fields=summary")
+			/*jiraNock.get("/rest/api/latest/issue/TEST-123?fields=summary")
```
KAbakumovAtl, line 48: Should we remove this instead of commenting?

test/safe/pull-request.test.ts
```diff
@@ -1,63 +1,312 @@
...
+		});
+		await Subscription.create({
+			gitHubInstallationId,
+			jiraHost,
+			jiraClientKey: clientKey
+		});
 	});
 
 	afterEach(async () => {
-		if (!nock.isDone()) {
```
KAbakumovAtl, line 17: I wonder did we move it to some kind of global verification?

test/safe/workflow.test.ts
```diff
@@ -12,70 +34,54 @@ describe("GitHub Actions", () => {
...
-							id: 1234,
-							slug: "github-actions",
-						},
-					},
-				]);
-
-			jiraNock.get("/rest/api/latest/issue/TEST-123?fields=summary")
+			// TODO: need to start validating all issue keys with jira
+			/*jiraNock.get("/rest/api/latest/issue/TEST-123?fields=summary")
 				.reply(200, {
```
KAbakumovAtl, line 39: I wonder why it is commented? Don't we call this endpoint when we process webhooks?

test/setup/setup.ts
```diff
@@ -41,13 +41,13 @@ beforeEach(() => {
...
-afterEach(() => {
-	nock.cleanAll(); // removes HTTP mocks
-	jest.resetAllMocks(); // Removes jest mocks
+	try {
+		// eslint-disable-next-line jest/no-standalone-expect
+		expect(nock).toBeDone();
+	} finally {
+		nock.cleanAll(); // removes HTTP mocks
+		jest.resetAllMocks(); // Removes jest mocks
+	}
```
KAbakumovAtl, line 50: So that's the main change - to make it global, right?

KAbakumovAtl (dismissed):
Most of my questions and concerns were addressed, what left are nitpicks.

test/safe/pull-request.test.ts
mboudreau: Yep! It's in `test/setup.ts`

src/config/logger.ts
```diff
@@ -35,7 +35,7 @@ const jobSerializer = (job: Job) => job && {
 	...job,
 	// Removing potentially extremely large amount of data from logs
 	data: undefined,
-	queue: undefined
+	queueName: job?.queue?.name
```
KAbakumovAtl, line 38: Please don't change this. We'll have to get used to new splunk queries format

KAbakumovAtl: Or, hold on, is it a duplicating parameter within { job.queueName } ?

KAbakumovAtl (approved):

src/config/logger.ts
KAbakumovAtl: Ok, never mind. Hence Redis queues going away anyway. 

I am happy as long as we have an original "queue" parameter in splunk messages (which is queue name)

test/safe/workflow.test.ts
mboudreau: Yeah, it's a problem I found.  We're not validating issuekeys with jira in some instances.  Rachelle is doing a PR to fix that.

test/setup/setup.ts
mboudreau: yep.

KAbakumovAtl (approved):

src/jira/client/index.ts
```diff
@@ -194,15 +195,13 @@ async function getJiraClient(
 						await subscription?.update({ syncWarning: issueKeyLimitWarning });
 					}
 
-					const batchBulkUpdateReq = await batchedBulkUpdate(
+					return batchedBulkUpdate(
```
atrigueiro, line 198: Don't we need `return await` here?

mboudreau: Why do you think that? :)

atrigueiro: `batchedBulkUpdate` is an async function

src/util/jira.ts
```diff
@@ -1,13 +1,14 @@
 import { JiraAuthor } from "../interfaces/jira";
+import _ from "lodash";
 
 export const getJiraAuthor = (...authors: (Author | undefined)[]): JiraAuthor => {
 	const author = Object.assign({}, ...authors);
-	return author.login || author.name ? {
-		avatar: author.avatar_url || author.avatarUrl || `https://github.com/users/${author.login}.png`,
+	return author.login || author.name ? _.pickBy({
+		avatar: author.avatar_url || author.avatarUrl || (author.login ? `https://github.com/users/${author.login}.png` : undefined),
```
tarakpurohit, line 7: Should there be only one `author.avatar_url` ?

mboudreau: Nope.  Depends if the author object comes from the REST API (avatar_url) or the GraphQL API (avatarUrl).  Fun right?

src/jira/client/index.ts
mboudreau: Yup, and what does the async function returns? :)

atrigueiro: A promise. So would `return` wait until that promise is resolved before actually returning?

mboudreau: no, it won't, it just returns the promise.  Now look at the signature of the container function

tarakpurohit (approved):

PR #892, opened 2021-11-05
ARC-764: plug in the deduplicator

bgvozdev:
Plugging in the previously added deduplicator to the sync process, behind a feature flag.


src/sync/installation.ts
```diff
@@ -411,8 +450,13 @@ async function doProcessInstallation(app, queues, job, installationId: number, j
 
 // TODO: type queues
 export const processInstallation =
-	(app: Application, queues) =>
-		async (job, rootLogger: LoggerWithTarget): Promise<void> => {
+	(app: Application, queues) => {
+		const inProgressStorage = new RedisInProgressStorageWithTimeout(new Redis(getRedisInfo("installations-in-progress")));
```
KAbakumovAtl, line 454: I thought that should be incapsulated into the deduplicator

src/sync/installation.ts
```diff
@@ -427,5 +471,41 @@ export const processInstallation =
...
+					() => doProcessInstallation(app, queues, job, installationId, jiraHost, logger, (delay: number) =>
+						nextTaskDelays.push(delay)
+					));
+
+				switch (result) {
+					case DeduplicatorResult.E_OK:
+						logger.info("Job was executed by deduplicator");
+						if (nextTaskDelays.length > 1) {
+							logger.warn("Multiple next jobs were scheduled");
+						}
```
KAbakumovAtl, line 487: Should we deduplicate it here by scheduling only a single task? 

That was the source of problem which required a deduplication job. Hence all these jobs have absolutely the same payload and shared state in the database, we should schedule only a single job here.

- [ ] Schedule only a single job

KAbakumovAtl: Choosing maximum delay from the list

src/sync/installation.ts
bgvozdev: Deduplicator is storage-type agnostic.

src/sync/installation.ts
```diff
@@ -448,6 +448,23 @@ async function doProcessInstallation(app, queues, job, installationId: number, j
...
 }
 
+// Export for unit testing. TODO: consider improving encapsulation by making this logic as part of Deduplicator, if needed
+export function maybeScheduleNextTask(queue: Queue.Queue, jobData, nextTaskDelays: Array<number>, logger: LoggerWithTarget) {
+	if (nextTaskDelays.length > 0) {
+		nextTaskDelays.sort().reverse();
+		if (nextTaskDelays.length > 1) {
+			logger.warn("Multiple next jobs were scheduled, scheduling one with the highest priority");
+		}
+		const delay = nextTaskDelays.shift()!;
```
KAbakumovAtl, line 458: /super-nit nextTaskDelays[0] would read better

KAbakumovAtl (approved):

tarakpurohit (approved):

PR #893, opened 2021-11-05
ARC-783 fix repo and add tests

rachellerathbone:
The GH config page wasn't rendering due to a 500 if there were no repos when trying to map failed installations to the subscriptions to get the org name. Fixed bug and confirmed it works in both versions of the UI. Also, added tests for this helper function.

Screen Shot 2021-11-05 at 5 09 30 pm

Screen Shot 2021-11-05 at 5 08 39 pm


src/frontend/get-jira-configuration.ts
```diff
@@ -56,14 +56,14 @@ export const getFailedConnections = (installations, subscriptions) => {
...
 			...failedConnection,
 			...subscriptions.find(
-				(sub) =>
-					failedConnection.id === sub.gitHubInstallationId
+				(sub) => failedConnection.id === sub.dataValues.gitHubInstallationId
 			),
 		}))
 		.map((sub) => {
-			const repos = sub.reposSyncState || {};
+			const repos = sub?.dataValues?.repoSyncState?.repos || {};
```
mboudreau, line 63: don't use `dataValues` unless there's a reason for it.

src/frontend/get-jira-configuration.ts
```diff
@@ -56,14 +56,14 @@ export const getFailedConnections = (installations, subscriptions) => {
 		.map((failedConnection) => ({
 			...failedConnection,
 			...subscriptions.find(
-				(sub) =>
-					failedConnection.id === sub.gitHubInstallationId
+				(sub) => failedConnection.id === sub.dataValues.gitHubInstallationId
```
mboudreau, line 59: same here.

src/frontend/get-jira-configuration.ts
```diff
@@ -56,14 +56,14 @@ export const getFailedConnections = (installations, subscriptions) => {
...
 			),
 		}))
 		.map((sub) => {
-			const repos = sub.reposSyncState || {};
+			const repos = sub?.dataValues?.repoSyncState?.repos || {};
 			const repoId = Object.keys(repos);
-			const orgName = repos[repoId[0]].repository?.owner.login || undefined;
+			const orgName =
+				repos[repoId[0]] && repos[repoId[0]].repository?.owner.login || undefined;
```
mboudreau, line 66: could use optional chaining.  Also, why are we only getting the first one?  What happens if there's no repos?

src/frontend/get-jira-configuration.ts
rachellerathbone: Not using dataValues is what caused this to break the last time I addressed PR comments.

mboudreau (changes requested):
few things to change first.

src/frontend/get-jira-configuration.ts
rachellerathbone: There's a test for this https://github.com/atlassian/github-for-jira/blob/cec00d3ebf86b2cd02cf69343185f6663710b659/test/frontend/util/get-jira-configuration-helpers.test.ts#L86

KAbakumovAtl (approved):

mboudreau (approved):

atrigueiro (approved):

PR #894, opened 2021-11-08
Add webhook payload metric

tarakpurohit:


src/github/middleware.ts
```diff
@@ -77,6 +77,10 @@ export default (
 		enhanceOctokit(context.github);
 		const webhookEvent = extractWebhookEventNameFromContext(context);
 
+		// Metrics for payload size
+		emitWebhookPayloadMetrics(webhookEvent,
+			Buffer.byteLength(JSON.stringify(context.payload), "utf-8"));
```
KAbakumovAtl, line 82: Can we emit this metric before serializing the palyoad again?

We should either do it before the request was serialised or use req. parameters:
`req.get("content-length")` or `req.socket.bytesRead` if content-length is missing

src/util/webhooks.ts
```diff
@@ -87,3 +87,15 @@ export const emitWebhookFailedMetrics = (webhookName: string) => {
 	statsd.increment(metricWebhooks.webhookFailure, tags);
 
 }
+
+export const emitWebhookPayloadMetrics = (webhookName: string, size: number) => {
+	const tags = {
+		name: webhookName,
+		payloadSize: size.toString()
+	};
+	statsd.histogram(
```
KAbakumovAtl, line 96: So, hence the purpose is to identify how many webhooks are there with a pyload > 128KB 

we should use a bucketed histogram metric here (See how it is done for the latencies metrics, we hse gsd_histogram parameter where we specify buckets sizes)

src/util/webhooks.ts
```diff
@@ -87,3 +87,15 @@ export const emitWebhookFailedMetrics = (webhookName: string) => {
 	statsd.increment(metricWebhooks.webhookFailure, tags);
 
 }
+
+export const emitWebhookPayloadMetrics = (webhookName: string, size: number) => {
+	const tags = {
+		name: webhookName,
+		payloadSize: size.toString()
```
KAbakumovAtl, line 94: We must never use values with unbounded variations as tags. It will explode the metric cardinality

KAbakumovAtl (changes requested):
I have to request changes because of the metric cardinality. 

That's what blocks the PR:
- [x] Do not use payloadSize as a tag

src/github/middleware.ts
tarakpurohit: We dont have access to `req` for webhook events, hence, we have to use the payload from context

KAbakumovAtl (approved):

src/util/webhooks.ts
tarakpurohit: Added bucket sizes.

atrigueiro (approved):

thombergs (approved):

PR #898, opened 2021-11-08
Arc 776 arc 777 add timeouts for sqs processing and error handling

KAbakumovAtl:
Added timeouts and error handling to SQS Listener


test/unit/sqs/index.test.ts
```diff
@@ -22,140 +23,296 @@ describe("SqsQueue tests", () => {
...
 
 	let queue: SqsQueue<TestMessage>;
 
-	beforeEach(() => {
-		queue = createSqsQueue();
-		queue.start();
-	});
+	describe("Normal execution tests", () => {
 
-	afterEach(async () => {
```
KAbakumovAtl, line 44: Apologies for the refactoring here, but I had to group the existing tests under the `describe("Normal execution tests", ....`

Because timeouts test uses different before each/after each configuration

src/sqs/queues.ts
```diff
@@ -9,12 +9,20 @@ const sqsQueues = {
 	backfill: new SqsQueue<BackfillMessagePayload>({ queueName: "backfill",
 		queueUrl: envVars.SQS_BACKFILL_QUEUE_URL,
 		queueRegion: envVars.SQS_BACKFILL_QUEUE_REGION,
-		longPollingIntervalSec: LONG_POLLING_INTERVAL_SEC}, backfillQueueMessageHandler),
+		longPollingIntervalSec: LONG_POLLING_INTERVAL_SEC,
+		timeoutSec: 10*60,
+		maxAttempts: 3
+	},
+	backfillQueueMessageHandler,
+	{handle: () => ({retryable: true})}
```
KAbakumovAtl, line 17: For now adding trivial error handler. Will add an actual error handling for Push queue in the next PR

src/sqs/index.ts
```diff
@@ -40,24 +51,60 @@ export type QueueSettings = {
 
 	readonly longPollingIntervalSec?: number,
 
-	//TODO Add batching
+	/**
+	 * Timeout for processing a single message in seconds
+	 */
+	readonly timeoutSec: number;
```
mboudreau, line 57: Javascript works in milliseconds and it's better to keep it the standard or else you risk the chance of setting something for 2 milliseconds.

src/sqs/index.ts
```diff
@@ -258,23 +311,67 @@ export class SqsQueue<MessagePayload> {
...
+
+		log.info(`Sqs message received. Receive count: ${receiveCount}`);
 
 		try {
 			const messageProcessingStartTime = new Date().getTime();
-			await this.messageHandler.handle(context)
+
+			// Change message visibility timeout to the max processing time
+			// plus some room for error handling in case of timeout
+			await this.changeVisabilityTimeout(message, this.timeoutSec + 2, log);
```
mboudreau, line 327: `+ 2` should be a constant.  Why 2 seconds?

src/sqs/index.ts
```diff
@@ -258,23 +311,67 @@ export class SqsQueue<MessagePayload> {
...
+
+			const timeoutPromise = new Promise((_, reject) =>
+				setTimeout(() => reject(new SqsTimeoutError()), this.timeoutSec*1000)
+			);
+
+			await Promise.race([this.messageHandler.handle(context), timeoutPromise])
+
 			const messageProcessingDuration = new Date().getTime() - messageProcessingStartTime;
```
mboudreau, line 335: use `Date.now()` instead.

src/sqs/index.ts
```diff
@@ -258,23 +311,67 @@ export class SqsQueue<MessagePayload> {
...
 			this.sendProcessedMetrics(messageProcessingDuration);
 			await this.deleteMessage(message, log)
 		} catch (err) {
+
+			const errorHandlingResult = this.errorHandler.handle(err, context);
+			if(!errorHandlingResult.retryable ||
+				(errorHandlingResult.skipDlq && context.receiveCount >= this.maxAttempts)) {
+				await this.deleteMessage(message, log)
+			} else {
+				if(errorHandlingResult.retryDelaySec) {
```
mboudreau, line 345: combine conditionals

src/sqs/index.ts
```diff
@@ -258,23 +311,67 @@ export class SqsQueue<MessagePayload> {
...
+	private async changeVisabilityTimeout(message: Message, timeout: number, logger: Logger): Promise<void> {
+
+		if(!message.ReceiptHandle) {
+			logger.error("NO Receipt Handle on a message")
+			return;
+		}
+
+		const params : ChangeMessageVisibilityRequest = {
+			QueueUrl: this.queueUrl,
+			ReceiptHandle: message.ReceiptHandle || "",
```
mboudreau, line 365: don't need the `|| ""`

src/sqs/index.ts
```diff
@@ -258,23 +311,67 @@ export class SqsQueue<MessagePayload> {
...
 			//TODO Add error handling
 			statsd.increment(sqsQueueMetrics.failed, this.metricsTags)
 			log.error({err}, "error executing sqs message")
 		}
 	}
 
+	private async changeVisabilityTimeout(message: Message, timeout: number, logger: Logger): Promise<void> {
+
+		if(!message.ReceiptHandle) {
+			logger.error("NO Receipt Handle on a message")
```
mboudreau, line 359: Could probably add the message to the data log to help debug

src/sqs/index.ts
```diff
@@ -40,24 +51,60 @@ export type QueueSettings = {
...
+}
 /**
  * Handler for the queue messages
  */
 export interface MessageHandler<MessagePayload> {
 	handle(context: Context<MessagePayload>): Promise<void>;
 }
 
+export interface ErrorHandler<MessagePayload> {
+	handle(error: Error, context: Context<MessagePayload>): ErrorHandlingResult;
```
mboudreau, line 85: we know for sure that the error handler will be synchronous?

src/sqs/queues.ts
mboudreau: In that case, wouldn't it be outside of the scope of this PR and should remove all error handling from this PR?

src/sqs/index.ts
```diff
@@ -100,13 +150,16 @@ export class SqsQueue<MessagePayload> {
 	 */
 	listenerContext: ListenerContext;
 
-	public constructor(settings: QueueSettings, messageHandler: MessageHandler<MessagePayload>) {
+	public constructor(settings: QueueSettings, messageHandler: MessageHandler<MessagePayload>, errorHandler: ErrorHandler<MessagePayload>) {
```
mboudreau, line 153: handing an interface with a handler function in it seems excessive.  Couldn't it just be the function alone?

test/unit/sqs/index.test.ts
```diff
@@ -22,140 +23,296 @@ describe("SqsQueue tests", () => {
...
-			done();
+			mockRequestHandler.handle.mockImplementation((context: Context<TestMessage>) => {
+				expect(context.payload).toStrictEqual(testPayload);
+				done();
+			});
+
+			queue.stop();
+
+			//delaying to make sure all asynchronous invocations inside the queue will be finished and it will stop
+			await queue.waitUntilListenerStopped();
```
mboudreau, line 83: Wouldn't make more sense to have the stop function be async to know when it's stopped instead of calling another function?

test/unit/sqs/index.test.ts
```diff
@@ -22,140 +23,296 @@ describe("SqsQueue tests", () => {
...
 
-		queue.start();
+			mockRequestHandler.handle.mockImplementation((context: Context<TestMessage>) => {
 
-		queue.sendMessage(testPayload);
-	});
+				if(context.payload.msg !== testPayload.msg) {
+					//discard message if it is not our test message
+					return;
+				}
```
mboudreau, line 100: What's this?

test/unit/sqs/index.test.ts
```diff
@@ -22,140 +23,296 @@ describe("SqsQueue tests", () => {
...
+			//delaying to make sure all asynchronous invocations inside the queue will be finished and it will stop
+			await queue.waitUntilListenerStopped();
+
+			queue.start();
+
+			queue.sendMessage(testPayload);
 		});
 
-		queue.stop();
+		test("Message received with delay", (done: DoneCallback) => {
```
mboudreau, line 90: Don't use the `done` callback unless needed.  Async/await is the preferred way.

test/unit/sqs/index.test.ts
```diff
@@ -22,140 +23,296 @@ describe("SqsQueue tests", () => {
...
+
+			const testPayload = generatePayload();
+			const testData: { messageId: undefined | string } = {messageId: undefined};
+
+			mockRequestHandler.handle.mockImplementation((context: Context<TestMessage>) => {
+
+				if(context.payload.msg !== testPayload.msg) {
+					//discard message if it is not our test message
+					return;
+				}
```
mboudreau, line 124: again here?

test/unit/sqs/index.test.ts
```diff
@@ -22,140 +23,296 @@ describe("SqsQueue tests", () => {
...
+						testData.messageId = context.message.MessageId;
+					} else if (testData.messageId === context.message.MessageId) {
+						done.fail("Message was received more than once");
+					} else {
+						done.fail("Different message on the tests queue");
+					}
+
+				} catch (err) {
+					done(err);
+				}
```
mboudreau, line 139: isn't that was expects are for?

test/unit/sqs/index.test.ts
```diff
@@ -22,140 +23,296 @@ describe("SqsQueue tests", () => {
...
+			queue.sendMessage(testPayload);
+
+			//code before the pause
+			setTimeout(function () {
+				if (testData.messageId) {
+					done();
+				} else {
+					done.fail("No message was received");
+				}
+			}, 3000);
```
mboudreau, line 150: Use [jest timer mocks](https://jestjs.io/docs/timer-mocks) to control time instead of adding timeouts and slowing down tests.

test/unit/sqs/index.test.ts
```diff
@@ -22,140 +23,296 @@ describe("SqsQueue tests", () => {
...
+					if (receivedTime.counter == 0) {
+						receivedTime.counter++;
+						context.log.info("Delaying the message");
+						await delay(1000);
+						context.log.info("Message processed after delay");
+						return;
+					}
+
+					const currentTime = new Date().getTime();
+					expect(currentTime - receivedTime.time).toBeGreaterThanOrEqual(1000);
```
mboudreau, line 176: This will be flaky.

test/unit/sqs/index.test.ts
```diff
@@ -22,140 +23,296 @@ describe("SqsQueue tests", () => {
...
+					done();
+				} catch (err) {
+					done(err);
+				}
+			});
+			queue.sendMessage(testPayload, 1);
+		});
+
+		test("Message gets executed exactly once", (done: DoneCallback) => {
```
mboudreau, line 114: Jest/JS standard for test is to use `it` so that it's closer to language
`it("should process message only once")`

mboudreau (changes requested):

src/sqs/queues.ts
KAbakumovAtl: Because the SQSQueue constructor requires one. 

I could potentially default it in the SQS constructor, however would like to make it explicit that we have an error handler (Hence in all our queues we'll use it)

src/sqs/index.ts
KAbakumovAtl: SQS Client works in seconds, plus there are other parameters which are set in seconds. Will keep seconds for consistency.

src/sqs/index.ts
KAbakumovAtl: There is a comment above explaining it. I will add a constant

src/sqs/index.ts
KAbakumovAtl: I am getting typescript errors otherwise hence message.ReceiptHandle is optional. The "" value will never be here because in the lines above we check if it is empty and return. 

Would would you suggest to convert optional to non-optional?

src/sqs/index.ts
KAbakumovAtl: The logger will already have messageId

src/sqs/index.ts
KAbakumovAtl: Good question.

In case if we want to use a feature flag inside an error handler we'll have to make it async.

Let's do it async.

test/unit/sqs/index.test.ts
KAbakumovAtl: As the comment says: Discard message if it is not our test message.

To make sure if there are some messages from other tests, for any reason, pollute the queue, the test won't be affected.

test/unit/sqs/index.test.ts
KAbakumovAtl: There was the comment in the file saying that we need `done` here

test/unit/sqs/index.test.ts
KAbakumovAtl: We have an external dependency here - localstack sqs queue. Some things we test with delays beaucse it is faster and easier. Not the best practice, but a good tradeoff for the development speed. 

Not going to change it now (unless it become a problem)

src/sqs/index.ts
```diff
@@ -258,23 +314,65 @@ export class SqsQueue<MessagePayload> {
...
+			await Promise.race([this.messageHandler(context), timeoutPromise])
+
 			const messageProcessingDuration = new Date().getTime() - messageProcessingStartTime;
 			this.sendProcessedMetrics(messageProcessingDuration);
 			await this.deleteMessage(message, log)
 		} catch (err) {
+
+			const errorHandlingResult = await this.errorHandler(err, context);
+			if(!errorHandlingResult.retryable ||
+				(errorHandlingResult.skipDlq && context.receiveCount >= this.maxAttempts)) {
```
bgvozdev, line 345: Should we log something before we delete?

src/sqs/index.ts
```diff
@@ -258,23 +314,65 @@ export class SqsQueue<MessagePayload> {
...
+				await this.deleteMessage(message, log)
+			} else if (errorHandlingResult.retryDelaySec) {
+				await this.changeVisabilityTimeout(message, errorHandlingResult.retryDelaySec, log);
```
bgvozdev, line 348: Should we log something before we change visibility?

src/sqs/index.ts
```diff
@@ -258,23 +314,65 @@ export class SqsQueue<MessagePayload> {
...
+			}
+
 			//TODO Add error handling
```
bgvozdev, line 351: Can we remove this TODO?

src/sqs/index.ts
```diff
@@ -80,6 +128,8 @@ type ListenerContext = {
 	log: Logger;
 }
 
+const EXTRA_VISIBILITY_TIMEOUT_DELAY = 2;
```
bgvozdev, line 131: Should we have time units in name? E.g. `EXTRA_VISIBILITY_TIMEOUT_DELAY_SEC`

src/sqs/queues.ts
```diff
@@ -9,12 +9,20 @@ const sqsQueues = {
 	backfill: new SqsQueue<BackfillMessagePayload>({ queueName: "backfill",
 		queueUrl: envVars.SQS_BACKFILL_QUEUE_URL,
 		queueRegion: envVars.SQS_BACKFILL_QUEUE_REGION,
-		longPollingIntervalSec: LONG_POLLING_INTERVAL_SEC}, backfillQueueMessageHandler),
+		longPollingIntervalSec: LONG_POLLING_INTERVAL_SEC,
+		timeoutSec: 10*60,
```
bgvozdev, line 13: 10 minutes to process one message? This looks quite oof tbh... With our VisibilityTimeout set to 2 minutes (if I'm not mistaken), that's an open gate to duplicates... Am I missing something?

src/sqs/queues.ts
```diff
@@ -9,12 +9,20 @@ const sqsQueues = {
...
+	defaultErrorHandler
+	),
 
 	push: new SqsQueue<PushQueueMessagePayload>({ queueName: "push",
 		queueUrl: envVars.SQS_PUSH_QUEUE_URL,
 		queueRegion: envVars.SQS_PUSH_QUEUE_REGION,
-		longPollingIntervalSec: LONG_POLLING_INTERVAL_SEC}, pushQueueMessageHandler),
+		longPollingIntervalSec: LONG_POLLING_INTERVAL_SEC,
+		timeoutSec: 60,
+		maxAttempts: 5}, pushQueueMessageHandler,	defaultErrorHandler),
```
bgvozdev, line 25: Curious why maxAttempts/timeoutSec are different for push and backfill?

src/sqs/backfill.ts
```diff
@@ -7,13 +7,10 @@ export type BackfillMessagePayload = {
...
 
-export const backfillQueueMessageHandler : MessageHandler<BackfillMessagePayload> = {
-
-	async handle(context) {
-		if(!await booleanFlag(BooleanFlags.NEW_BACKFILL_PROCESS_ENABLED, false, context.payload.jiraHost)) {
-			context.log.info("New backfill process disabled, dropping the message ");
-			return
-		}
-		context.log.info("Porcessing backfill message");
+export const backfillQueueMessageHandler : MessageHandler<BackfillMessagePayload> = async (context) => {
```
KAbakumovAtl, line 10: I had to address some PR comments and made this as a function instead of an object. No other changes in this file

src/sqs/push.ts
```diff
@@ -22,22 +22,20 @@ export type PushQueueMessagePayload = {
 	webhookReceived?: Date,
 }
 
-export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> = {
+export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> = async (context : Context<PushQueueMessagePayload>) => {
```
KAbakumovAtl, line 25: I had to address some PR comments and made this as a function instead of an object. No other changes in this file

src/sqs/index.ts
KAbakumovAtl: Yea, good point

src/sqs/queues.ts
KAbakumovAtl: Currently it is 10 minutes for the same queue on Redis.  We can adjust the timeout when we'll be connecting Installations queue logic to SQS

src/sqs/queues.ts
KAbakumovAtl: Hence for Initial Sync queue we have many messages there anyway (duplicates), and with current logic they take much more time to process hence we make many calls to GitHub(especially when there is throttling)

So that's why it is different. For Redis queue we currently have the same configuration (except we don't retry messages)

src/sqs/queues.ts
bgvozdev: I think we need to have same or bigger VisibilityTimeout in service descriptor then...

src/sqs/index.ts
```diff
@@ -258,23 +314,66 @@ export class SqsQueue<MessagePayload> {
 			executionId: uuidv4(),
 			queue: this.queueName})
 
-		const context: Context<MessagePayload> = {message, payload, log }
+		const maybeReceivedCount: MessageAttributeValue | undefined = message.MessageAttributes ? message.MessageAttributes["ApproximateReceiveCount"] : undefined;
```
mboudreau, line 317: use optional chaining.

src/sqs/index.ts
```diff
@@ -258,23 +314,66 @@ export class SqsQueue<MessagePayload> {
...
 
-		log.info("Sqs message received");
+		const receiveCount = maybeReceivedCount ? parseInt(maybeReceivedCount?.StringValue || "1") : 1
```
mboudreau, line 319: Could just do `const receiveCount = Number(maybeReceivedCount?.StringValue) || 1`

src/sqs/index.ts
mboudreau: I'm fairly certain that if you have a check with a return beforehand, TS assumes the type is now specified and not optional and should work.  I just tried this exact code without the || and it worked fine.

test/unit/sqs/index.test.ts
mboudreau: What I'm asking is _why is it in your test and why is your state being polluted between tests?_

This is not how tests are suppose to be done and is extremely risky for getting flaky tests or false positives.

src/sqs/queues.ts
KAbakumovAtl: Currently we change the visibility timeout as soon as we read the message so whatever we put in the Micros Descriptor will be overridden.

Anyway would be good to add "just in case". We can address it in the next PR.

src/sqs/index.ts
KAbakumovAtl: I tried but it doesn't work
image

mboudreau: That's not the syntax, you're missing a dot.  `message.MessageAttributes?.["ApproximateReceiveCount"]` but you don't even need the string accessor, you can just do  `message.MessageAttributes?.ApproximateReceiveCount` directly.

test/unit/sqs/index.test.ts
mboudreau: This needs to be paid attention.  This situation has already happened at least once before with one of your previous PRs and we had to fix it because it kept failing sporadically for others.  You cannot use timers in tests, it's extremely unpredictable.  This is the reason [jest has a whole section dedicated to handling these time based issues](https://jestjs.io/docs/timer-mocks).

src/sqs/queues.ts
bgvozdev: If the value in service-descriptor overwritten, I think we should call this out explicitly in it (add a comment maybe?)...
- [ ] add comment to service descriptor explaining that we are not using VisibilityTimeout value

KAbakumovAtl:
The reason I was adding to the test:
```
if(context.payload.msg !== testPayload.msg) {
					//discard message if it is not our test message
					return;
				}
```
not because they add garbage on the queue when they work. 
It is for the case of a test failure. If a test fails on assertion then the message will be retried.

However, probably it is an overkill to optimise for that case, so I removed the checks

bgvozdev (approved):

test/unit/sqs/index.test.ts
KAbakumovAtl: Following up on the discussion we had F2F. That's the problem you are talking about, which we address in the past
https://github.com/atlassian/github-for-jira/pull/853/files

Flakiness had nothing to do with timeouts

mboudreau (approved):

PR #901, opened 2021-11-10
FD-26662: clean up feature flag

thombergs:


src/sync/commits.ts
```diff
@@ -37,10 +36,6 @@ export default async (github: GitHubAPI, repository: Repository, cursor?: string
 		commitsData = await getCommits(true);
 	} catch (err) {
 
-		if (!await booleanFlag(BooleanFlags.RETRY_WITHOUT_CHANGED_FILES, false)) {
-			throw err;
-		}
```
thombergs, line 42: The FF is currently enabled on all environments, which means we can just remove these three lines.

KAbakumovAtl: Do we have any tests from where we need to remove it?

KAbakumovAtl (dismissed):

src/sync/commits.ts
thombergs: no, those were the wild west days... :(

thombergs: no, those were the wild west days... :(

mboudreau (approved):

mboudreau (approved):

bgvozdev (approved):

PR #902, opened 2021-11-10
FD-26660: clean up feature flag

thombergs:
Cleaning up a feature flag. Please view the diff with "ignore whitespace" enabled, because the formatter formatted some whitespaces.

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -6,20 +6,20 @@ import SubscriptionClass, {
 	SyncStatus,
 	TaskStatus
 } from "../models/subscription";
-import {Subscription} from "../models";
+import { Subscription } from "../models";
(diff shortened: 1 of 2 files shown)


src/sync/installation.ts
```diff
@@ -413,38 +413,23 @@ async function doProcessInstallation(app, queues, job, installationId: number, j
 			return;
 		}
 
-		if (await booleanFlag(BooleanFlags.CONTINUE_SYNC_ON_ERROR, false, jiraHost)) {
```
thombergs, line 416: The feature flag is currently enabled on all environments, so we can just remove the `else` branch.

KAbakumovAtl (approved):

mboudreau (approved):

PR #903, opened 2021-11-10
FD-26653: clean up feature flag

thombergs:
Please view the diff with "ignore whitespaces" because the formatter formatted some whitespaces properly.

### DIFF
--- a/src/worker/startup.ts
+++ b/src/worker/startup.ts
@@ -1,22 +1,21 @@
 import Timeout = NodeJS.Timeout;
-import {Job} from "bull";
-import {LoggerWithTarget} from "probot/lib/wrap-logger";
-import {RateLimitingError} from "../config/enhance-octokit";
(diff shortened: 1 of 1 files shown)


src/worker/startup.ts
```diff
@@ -91,25 +90,24 @@ const setDelayOnRateLimiting = (jobHandler) => async (job: Job, logger: LoggerWi
 		throw err;
 	}
 };
-const sendQueueMetrics = async () => {
-	if (await booleanFlag(BooleanFlags.EXPOSE_QUEUE_METRICS, false)) {
```
thombergs, line 95: The [feature flag](https://hello.atlassian.net/browse/FD-26653) is currently enabled on all environments so we can just remove the if statement.

mboudreau (approved):

yvonnevictoria (approved):

PR #907, opened 2021-11-11
ARC-793 Disable backfill button after click

rachellerathbone:
To prevent users from firing a call to sync multiple times, I've disabled the button after it has been clicked.

Example: I'm clicking on 'Restart backfill' multiple times but only one request is made.

https://user-images.githubusercontent.com/37155488/141381146-6b56e469-2129-4530-8cde-19044577cc41.mov


static/css/jira-configuration.css
```diff
@@ -234,6 +234,15 @@
   color: #455570 !important;
 }
 
+.restart-backfill-button {
+   /* override aui styling */
+  padding: 0 !important;
```
yvonnevictoria, line 239: I usually try and avoid using `!important`. Is there another way to do this?

yvonnevictoria: There are ways to target elements more specifically and override items through this. Can do a pair programming session to discuss this.

rachellerathbone: I usually try to avoid using `!important` as well (I actually have a blog post about the use of it being something you should avoid). Unfortunately, we are using Atlassian's AUI components and the default styling comes from those. Using `!important` is the only way to override these styles here, hence the comment `/* override aui styling */`

rachellerathbone: Updated and created a ticket to remove the other `!important` entries from css. Used those initially when setting up the new UI because the AUI components were giving me a headache.  That should probably be cleaned up now though.

yvonnevictoria: Awesome, thank you :)

yvonnevictoria (approved):

mboudreau (approved):

PR #908, opened 2021-11-12
ARC-763 follow up items - part 1

rachellerathbone:
Addressing 1 of 2 follow-up items from PR: add typings to getFailedConnections. Will address the other follow-up item in a separate PR.


src/frontend/get-jira-configuration.ts
```diff
@@ -50,18 +51,26 @@ const formatDate = function (date) {
 	};
 };
 
-export const getFailedConnections = (installations, subscriptions) => {
-	return installations
+export const getFailedConnections = (
+	installations: FailedInstallations[] | typeof Installation[],
```
mboudreau, line 55: This seems weird.  Fairly sure the only thing that should be coming in is `FailedInstallations[]`

src/frontend/get-jira-configuration.ts
```diff
@@ -50,18 +51,26 @@ const formatDate = function (date) {
...
+	subscriptions: Subscriptions[]
+) => {
+	return (installations as FailedInstallations[])
 		.filter((response) => !!response.error)
 		.map((failedConnection) => {
 			const sub = subscriptions.find(
-				(sub) => failedConnection.id === sub.gitHubInstallationId
-			)
+				(subscription: Subscriptions) =>
+					Number(failedConnection.id) === subscription.gitHubInstallationId
```
mboudreau, line 63: pretty sure this is already a number at runtime?

src/frontend/get-jira-configuration.ts
```diff
@@ -91,7 +100,7 @@ export default async (
 
 		const failedConnections = getFailedConnections(
 			installations,
```
mboudreau, line 102: This is where the real problem lies: the `getInstallation()` doesn't return a type and it's also return 2 different types in one function instead of having one resolved type and another rejected type.  Fixing `getInstallation` would make life much easier as to find the failed installations, you just need to find those that were rejected.

mboudreau (changes requested):
I think we can simplify this a bit.  Let me know if you want to pair on this.

src/config/interfaces.ts
```diff
@@ -0,0 +1,73 @@
+import { RepoSyncState } from "../models/subscription";
+
+export interface Subscriptions {
```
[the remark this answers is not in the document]
rachellerathbone: Already tried that and I get the error: `'Subscription' refers to a value, but is being used as a type here. Did you mean 'typeof Subscription'?ts(2749)`. If I then update `Subscription[]` to `typeof Subscription[]`, like I did for Installation, I then get:
```
No overload matches this call.
  Overload 1 of 2, '(predicate: (this: void, value: typeof Subscription, index: number, obj: (typeof Subscription)[]) => value is typeof Subscription, thisArg?: any): typeof Subscription | undefined', gave the following error.
    Argument of type '(subscription: Subscriptions) => boolean' is not assignable to parameter of type '(this: void, value: typeof Subscription, index: number, obj: (typeof Subscription)[]) => value is typeof Subscription'.
      Types of parameters 'subscription' and 'value' are incompatible.
        Type 'typeof Subscription' is missing the following properties from type 'Subscriptions': gitHubInstallationId, jiraHost, jiraClientKey, createdAt, updatedAt
  Overload 2 of 2, '(predicate: (value: typeof Subscription, index: number, obj: (typeof Subscription)[]) => unknown, thisArg?: any): typeof Subscription | undefined', gave the following error.
    Argument of type '(subscription: Subscriptions) => boolean' is not assignable to parameter of type '(value: typeof Subscription, index: number, obj: (typeof Subscription)[]) => unknown'.
      Types of parameters 'subscription' and 'value' are incompatible.
        Type 'typeof Subscription' is not assignable to type 'Subscriptions'.ts(2769)
```

Do you know how to resolve that?

src/config/interfaces.ts
```diff
@@ -0,0 +1,73 @@
...
+	"referrer-policy": string;
+	server: string;
+	"strict-transport-security": string;
+	"transfer-encoding": string;
+	vary: string;
+	"x-content-type-options": string;
+	"x-frame-options": string;
+	"x-github-media-type": string;
+	"x-github-request-id": string;
+	"x-xss-protection": string;
```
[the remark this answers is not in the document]
rachellerathbone: They were for me when I looked at the response.

src/frontend/get-jira-configuration.ts
rachellerathbone: Not at this point. We pass in all installations and then filter out the ones without errors on line 59.

src/frontend/get-jira-configuration.ts
rachellerathbone: Shit, you're right. Thought I saw it as a number in the response when I checked. Let me fix the type.

src/frontend/get-jira-configuration.ts
rachellerathbone: I think this be done as a separate ticket. My follow-up item was to add typings to `getFailedConnections`, not refactor a function that was a part of the original codebase when we took over.

src/config/interfaces.ts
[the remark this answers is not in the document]
mboudreau: yeah, it's something I've been meaning to fix.  There's kinda 2 different "Subscription", one is a constant value in `src/models/index.ts` where we initialize sequelize models with it, and the other is the Subscription Class, which is the typings.  What you want is the class for types.  You can accomplish this by doing `import SubscriptionClass from "../models/subscription";`

src/config/interfaces.ts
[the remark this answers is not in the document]
mboudreau: are we're using all of them?  Headers are normally just optional.  Heck, looking at Axios type definition, they don't even try to type them because each server can send different headers, even with different capitalizations rules.  So Axios just have headers as `any`.  This might be overdoing it just a tad :)

src/config/interfaces.ts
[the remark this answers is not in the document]
rachellerathbone: Sounds easy enough :)

src/frontend/get-jira-configuration.ts
mboudreau: I think the type should be `installations: (FailedInstallations | InstallationClass)[]` as they're mixed within the same array and not 2 separate arrays of different types.  Also, do `import InstallationClass from "../models/installation"` instead of `typeof Installation`.  This is the same issue as mentioned before with Subscription.

src/frontend/get-jira-configuration.ts
```diff
@@ -50,18 +52,26 @@ const formatDate = function (date) {
 	};
 };
 
-export const getFailedConnections = (installations, subscriptions) => {
-	return installations
+export const getFailedConnections = (
+	installations: FailedInstallations[] | typeof Installation[],
+	subscriptions: SubscriptionClass[]
+) => {
+	return (installations as FailedInstallations[])
```
mboudreau, line 59: I wouldn't cast it here since right now, they're not that type until they get filtered.

src/frontend/get-jira-configuration.ts
```diff
@@ -50,18 +52,26 @@ const formatDate = function (date) {
...
 		.filter((response) => !!response.error)
 		.map((failedConnection) => {
```
mboudreau, line 61: making this `(failedConnection:FailedInstallation)` should cast it automatically from this point on.

src/frontend/get-jira-configuration.ts
```diff
@@ -50,18 +52,26 @@ const formatDate = function (date) {
 	};
 };
 
-export const getFailedConnections = (installations, subscriptions) => {
-	return installations
+export const getFailedConnections = (
+	installations: FailedInstallations[] | typeof Installation[],
+	subscriptions: SubscriptionClass[]
+) => {
```
mboudreau, line 58: Missing return type of function :)

src/frontend/get-jira-configuration.ts
mboudreau: yeah, fair.

src/frontend/get-jira-configuration.ts
rachellerathbone: If I don't do it before line 60 I get the error `Property 'error' does not exist on type 'FailedInstallations | Installation'.
  Property 'error' does not exist on type 'Installation'.`

mboudreau: ah right that's because it doesn't know which ones is which at that point, which is fair.  You could cast it as `unknown` in the filter then `FailedInstallation` in the map

mboudreau (approved):

src/config/interfaces.ts
```diff
@@ -0,0 +1,38 @@
+interface FailledInstallationsRequestHeaders {
```
atrigueiro, line 1: `Failed` instead of `Failled`?

atrigueiro (approved):

mboudreau (approved):

PR #909, opened 2021-11-12
Use new github client under a feature flag for PRs

tarakpurohit:
Starts using the Github Client for processing PR backfill.

### DIFF
--- a/src/github/client/app-token-holder.ts
+++ b/src/github/client/app-token-holder.ts
@@ -1,6 +1,10 @@
 import { AsymmetricAlgorithm, encodeAsymmetric } from "atlassian-jwt";
 import AuthToken, { ONE_MINUTE, TEN_MINUTES } from "./auth-token";
 
+//TODO: Remove Probot dependency to find privateKey
+import * as PrivateKey from "probot/lib/private-key";
+import envVars from "../../config/env";
+
 /**
  * Holds the GitHub app's token to authenticate as a GitHub app and refreshes it as needed.
  *
@@ -15,9 +19,9 @@ export default class AppTokenHolder {
 	private readonly appId: string;
 	private currentToken: AuthToken;
 
-	constructor(key: string, appId: string) {
-		this.key = key;
-		this.appId = appId;
+	constructor() {
+		this.key = PrivateKey.findPrivateKey() || "";
+		this.appId = envVars.APP_ID;
 	}
 
 	/**
--- a/src/github/client/app-token-holder.ts
+++ b/src/github/client/app-token-holder.ts
@@ -19,10 +19,20 @@ export default class AppTokenHolder {
 	private readonly appId: string;
 	private currentToken: AuthToken;
 
+	private static instance: AppTokenHolder;
+
 	constructor() {
 		this.key = PrivateKey.findPrivateKey() || "";
 		this.appId = envVars.APP_ID;
 	}
+
+	public static getInstance(): AppTokenHolder {
+		if (!AppTokenHolder.instance) {
+			AppTokenHolder.instance = new AppTokenHolder();
+		}
+		return AppTokenHolder.instance;
+	}
+
 
 	/**
 	 * Gets the current app token or creates a new one if the old is about to expire.
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -1,5 +1,5 @@
 import { Octokit } from "@octokit/rest";
-import axios, { AxiosInstance, AxiosRequestConfig } from "axios";
+import axios, { AxiosInstance, AxiosRequestConfig, AxiosResponse } from "axios";
 import AppTokenHolder from "./app-token-holder";
 import InstallationTokenCache from "./installation-token-cache";
 import AuthToken from "./auth-token";
@@ -76,14 +76,35 @@ export default class GitHubClient {
 	public async getPullRequests(
 		owner: string,
 		repo: string,
-		pageSize: number,
-		page: number): Promise<Octokit.PullsListResponseItem[]> {
+		pageSize?: number,
+		page?: number,
+		options?: any
+	): Promise<AxiosResponse<Octokit.PullsListResponseItem[]>> {
 		const response = await this.axios.get<Octokit.PullsListResponseItem[]>(`/repos/${owner}/${repo}/pulls`, {
 			...await this.installationAuthenticationHeaders(),
 			params: {
 				installationId: this.githubInstallationId,
 				per_page: pageSize,
-				page: page
+				page: page,
+				...options
+			}
+		});
+
+		return response;
+	}
+
+	/**
+	 * Get a single pull request for the given repository.
+	 */
+	public async getPullRequest(
+		owner: string,
+		repo: string,
+		pullNumber: string,
+	): Promise<Octokit.PullsGetResponse> {
+		const response = await this.axios.get<Octokit.PullsGetResponse>(`/repos/${owner}/${repo}/pulls/${pullNumber}`, {
+			...await this.installationAuthenticationHeaders(),
+			params: {
+				installationId: this.githubInstallationId,
 			}
 		});
 
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -3,6 +3,7 @@ import axios, { AxiosInstance, AxiosRequestConfig, AxiosResponse } from "axios";
 import AppTokenHolder from "./app-token-holder";
 import InstallationTokenCache from "./installation-token-cache";
 import AuthToken from "./auth-token";
+import { GetPullRequestParams } from "./types";
 
 /**
  * A GitHub client that supports authentication as a GitHub app.
@@ -17,16 +18,14 @@ export default class GitHubClient {
 	private readonly githubInstallationId: number;
 
 	constructor(
-		appTokenHolder: AppTokenHolder,
-		installationTokenCache: InstallationTokenCache,
 		githubInstallationId: number,
 		baseURL = "https://api.github.com"
 	) {
 		this.axios = axios.create({
 			baseURL
 		});
-		this.appTokenHolder = appTokenHolder;
-		this.installationTokenCache = installationTokenCache;
+		this.appTokenHolder = new AppTokenHolder();
+		this.installationTokenCache = new InstallationTokenCache(1000);
 		this.githubInstallationId = githubInstallationId;
 	}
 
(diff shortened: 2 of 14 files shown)


src/github/client/app-token-holder.ts
```diff
@@ -1,6 +1,10 @@
 import { AsymmetricAlgorithm, encodeAsymmetric } from "atlassian-jwt";
 import AuthToken, { ONE_MINUTE, TEN_MINUTES } from "./auth-token";
 
+//TODO: Remove Probot dependency to find privateKey
+import * as PrivateKey from "probot/lib/private-key";
```
KAbakumovAtl, line 5: /nit we could just import a function to avoid star imports

src/github/client/github-client.ts
```diff
@@ -76,14 +76,35 @@ export default class GitHubClient {
 	public async getPullRequests(
 		owner: string,
 		repo: string,
-		pageSize: number,
-		page: number): Promise<Octokit.PullsListResponseItem[]> {
+		pageSize?: number,
+		page?: number,
+		options?: any
```
KAbakumovAtl, line 81: I wonder how do we use these options? Would it better to specify a concrete parameters here to remove ambiguity?

src/sync/branches.ts
```diff
@@ -2,9 +2,10 @@ import transformBranches from "./transforms/branch";
 import { getBranches as getBranchesQuery } from "./queries";
 import { GitHubAPI } from "probot";
 import { Repository } from "../models/subscription";
+import GitHubClient from "../github/client/github-client";
 
 // TODO: better typings
-export default async (github: GitHubAPI, repository:Repository, cursor?:string | number, perPage?:number) => {
+export default async (github: GitHubAPI, _newGithub: GitHubClient, _useNewGHClient: boolean, repository:Repository, cursor?:string | number, perPage?:number) => {
```
KAbakumovAtl, line 8: Do we change any branches logic here as well?

src/sync/installation.ts
```diff
@@ -245,7 +250,14 @@ async function doProcessInstallation(app, queues, job, installationId: number, j
 		installationId,
 		logger
 	);
+	const useNewGHClient = await booleanFlag(BooleanFlags.USE_NEW_GITHUB_CLIENT__FOR_PR, false, jiraHost);
```
KAbakumovAtl, line 253: So would we have one feature flag for all? 

If yes then it will be better to call it differently USE_NEW_GITHUB_CLIENT

Otherwise we should evaluate within each processor

src/sync/branches.ts
mboudreau, line 8: Can you explain what these 2 new attributes are doing?

src/sync/pull-request.ts
```diff
@@ -57,18 +61,25 @@ export default async function(
...
-				direction: "desc"
-				// Retry up to 6 times pausing for 10s, for *very* large repos we need to wait a while for the result to succeed in dotcom
-			}));
+			(
+
+				{
+					data: edges,
+					status,
+					headers
+				} = useNewGHClient?
```
KAbakumovAtl, line 72: If `useNewGHClient` we'll use an old one here

src/sync/pull-request.ts
```diff
@@ -87,9 +98,19 @@ export default async function(
...
+						owner: repository.owner.login,
+						repo: repository.name,
+						pull_number: pull.number
+					})).data;
+
+				//TODO: Add getGithubUser call to new GH client
+				const ghUser = await getGithubUser(github, prDetails.user?.login);
 				const data = await transformPullRequest(
 					{ pullRequest: pull, repository },
-					github
```
KAbakumovAtl, line 92: So did we pass a client into transformPullRequest before? and now we extract that part here?

src/sync/pull-request.ts
```diff
@@ -87,9 +98,19 @@ export default async function(
...
 		await Promise.all(
 			edges.map(async (pull) => {
+				const prDetails = useNewGHClient ? await newGithub.getPullRequest(repository.owner.login, repository.name, pull.number) :
+					(await github?.pulls?.get({
+						owner: repository.owner.login,
+						repo: repository.name,
+						pull_number: pull.number
+					})).data;
+
+				//TODO: Add getGithubUser call to new GH client
```
KAbakumovAtl, line 108: I believe we should do it in this PR, otherwise we'll be using both clients on Pull Reqeusts path

src/sync/pull-request.ts
```diff
@@ -57,18 +61,25 @@ export default async function(
...
+				{
+					data: edges,
+					status,
+					headers
+				} = useNewGHClient?
+					await github.pulls.list({ ...vars, state: "all", sort: "created", direction: "desc"})
+					:
+					await newGithub
+						.getPullRequests(repository.owner.login, repository.name, perPage, Number(cursor),
+							{
```
KAbakumovAtl, line 77: I have one idea, hence we still have to pass all old parameters here, should we keep old and new client APIs the same, so we can simply replace the GitHub Client object and keep the rest of the code in these handlers as is?

test/github/client/app-token-holder.test.ts
```diff
@@ -1,11 +1,7 @@
 import AppTokenHolder from "../../../src/github/client/app-token-holder";
-import fs from "fs";
 
 describe("AppTokenHolder", () => {
 
-	const privateKeyBase64 = fs.readFileSync("./test/github/client/dummy.key", "utf-8");
-	const privateKey = Buffer.from(privateKeyBase64, "base64").toString("ascii");
```
KAbakumovAtl, line 7: so it will be handled by ProbotKey, right?

src/sync/branches.ts
tarakpurohit: Nothing right now. Next PRs will start using these similar to `pull-request` processor. 
The processors are overloaded so I couldnt get away with adding it only to that file hence the `_` infront to avoid linting errors.

src/github/client/app-token-holder.ts
tarakpurohit: The package doesnt export it and in the long run we will want to remove this dependency and have our own solution.

src/sync/pull-request.ts
tarakpurohit: Will add this call to new client

src/github/client/github-client.ts
thombergs: Yeah, I would very much like to see typed parameters here. Or maybe a type like `GetPullRequestsOptions` with all the options.

src/sync/branches.ts
thombergs: then maybe remove this change from this PR for now?

src/sync/commits.ts
```diff
@@ -3,11 +3,12 @@ import { getCommits as getCommitsQuery, getDefaultRef } from "./queries";
 import { GitHubAPI } from "probot";
 import { getLogger } from "../config/logger";
 import { Repository } from "../models/subscription";
+import GitHubClient from "../github/client/github-client";
 
 const logger = getLogger("sync.commits");
 
 // TODO: better typings
-export default async (github: GitHubAPI, repository: Repository, cursor?: string | number, perPage?: number) => {
+export default async (github: GitHubAPI, _newGithub: GitHubClient, _useNewGHClient: boolean, repository: Repository, cursor?: string | number, perPage?: number) => {
```
thombergs, line 11: same here, doesn't belong into this PR.

src/sync/installation.ts
```diff
@@ -245,7 +250,14 @@ async function doProcessInstallation(app, queues, job, installationId: number, j
...
+
+	const appTokenCache = new AppTokenHolder();
```
thombergs, line 255: nitpick: variable name should be `appTokenHolder`

src/sync/pull-request.ts
```diff
@@ -57,18 +61,25 @@ export default async function(
...
+								...vars,
```
thombergs, line 78: the `...vars` is passing the `owner` and `repo` options into the function that are already defined as specific parameters. We shouldn't pass an untyped `options` object here, imo (see comment above).

src/sync/pull-request.ts
thombergs: It is the same API in this case. But if the old API is crap, I wouldn't make the new API equally crap and instead add some mapping in this code here.

thombergs (changes requested):

src/sync/branches.ts
tarakpurohit: Not that straight forward. Need to have these in the parameter list as they are required for a processor. Will try to refactor it but they will soon be needed when I work on `branch` and `commits`

mboudreau: I'm just not a fan that we're polluting the function signature with a bunch of stuff that's inconsequential to it.

Personally, I would have preferred if we made a Github service abstraction so we can control the client at the service level, so long as the return signature is the same...

src/sync/pull-request.ts
mboudreau: I'm not a fan of this syntax.  I'd rather you separate the tertiary operator result into it's own const before doing an object deconstruction on it

src/github/client/github-client.ts
```diff
@@ -70,24 +69,37 @@ export default class GitHubClient {
...
 			...await this.installationAuthenticationHeaders(),
-			params: {
+			params: { 
 				installationId: this.githubInstallationId,
-				per_page: pageSize,
-				page: page
+				...params
 			}
 		});
+		//TODO: error handling
```
KAbakumovAtl, line 80: We can remove this comment from here hence error handling is not the responsibility of the HTTP Client

src/github/client/installation-token-cache.ts
```diff
@@ -21,6 +22,14 @@ export default class InstallationTokenCache {
 		this.installationTokenCache = new LRUCache<number, AuthToken>({ max: maxTokens });
 	}
 
+	public static getInstance(): InstallationTokenCache {
+		if (!InstallationTokenCache.instance) {
+			InstallationTokenCache.instance = new InstallationTokenCache(1000);
```
KAbakumovAtl, line 27: /nit we can extract 1000 as a constant here

src/github/client/types.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export type GetPullRequestParams = {
+	state?: string;
+	head?: string;
+	base?: string;
+	sort?: string;
+	direction?: string;
+	per_page?: number;
+	page?: number;
+}
```
KAbakumovAtl, line 25: Nice!

KAbakumovAtl (approved):

src/sync/pull-request.ts
tarakpurohit: So most likely, this syntax will go away when the featureFlag is cleaned up

thombergs (approved):

PR #912, opened 2021-11-15
ARC-791 Removed large payloads from being logged on 404s from Jira

KAbakumovAtl:
In this Pull Request I did the following:

* Wrapped AxiosError in JiraClientError for more convenient error handling. So we'll have a response status code as error parameter (which we need for implementing "push" events error handling). And removed "config" from the AxiosError cause to prevent it from getting into logs (it was generating enormous payloads)
* Removed extra logging on the push request failure. (Now they conditionally logged in the axios client)


src/jira/client/axios.ts
```diff
@@ -230,10 +262,9 @@ const instrumentRequest = (response) => {
  * @param {import("axios").AxiosError} error - The Axios error response object.
  * @returns {Promise<Error>} a rejected promise with the error inside.
  */
-const instrumentFailedRequest = (logger) => {
+const instrumentFailedRequest = () => {
 	return (error) => {
 		instrumentRequest(error?.response);
-		logger.error(error, "Error during Axios request");
```
KAbakumovAtl, line 236: Removing extra logging from here. This log will be duplicated in the error handler

src/jira/client/index.ts
```diff
@@ -299,10 +297,7 @@ const batchedBulkUpdate = async (
 				installationId
 			}
 		};
-		return instance.post("/rest/devinfo/0.10/bulk", body).catch((err) => {
-			logger?.error({ ...err, body, data }, "Jira Client Error: Cannot update Repository");
-			return Promise.reject(err);
```
KAbakumovAtl, line 304: Removed log from here, because it will be logged in the Jira Client

src/transforms/push.ts
```diff
@@ -207,7 +208,7 @@ export const processPush = async (github: GitHubAPI, payload, rootLogger: Logger
 			);
 		}
 	} catch (error) {
-		rootLogger.error(error, "Failed to process push");
+		log.error({err: error}, "Failed to process push");
```
bgvozdev, line 211: nit: we could rename "error" to "err" and then it would've been `{ err }`

src/jira/client/axios.ts
```diff
@@ -84,16 +100,32 @@ function getErrorMiddleware(logger: Logger) {
 		 * @returns {Promise<Error>} The rejected promise
 		 */
 		(error: AxiosError): Promise<Error> => {
-			if (error?.response) {
-				const status = error.response.status;
+
```
bgvozdev, line 103: Should we feature-flag the change?

bgvozdev (dismissed):

src/jira/client/axios.ts
```diff
@@ -55,6 +55,22 @@ function getAuthMiddleware(secret: string) {
...
+export class JiraClientError extends Error {
+	status?: number;
+	cause: AxiosError
+
+	constructor(message: string, cause: AxiosError, status?:number) {
+		super(message);
+		this.status = status
+
+		//Remove config from the cause to prevent large payloads from being logged
+		this.cause = {...cause, config: {}};
```
bgvozdev, line 70: Should this have a test?

src/jira/client/axios.ts
bgvozdev: On the other hand, this is error path, probably doesn't make much sense

src/jira/client/axios.ts
```diff
@@ -84,16 +100,32 @@ function getErrorMiddleware(logger: Logger) {
...
+			const status = error?.response?.status;
+
+			if (status) {
 
 				// truncating the detail message returned from Jira to 200 characters
-				const errorMessage = getJiraErrorMessages(status);
-				// Creating an object that isn't of type Error as bunyan handles it differently
-				// Log appropriate level depending on status - WARN: 300-499, ERROR: everything else
-				(status >= 300 && status < 500 ? logger.warn : logger.error)(error, errorMessage);
+				errorMessage = getJiraErrorMessages(status);
```
tarakpurohit, line 112: Move this statement out of the `if` block. It will never print the default error message if the status/response is undefined

src/jira/client/axios.ts
```diff
@@ -55,6 +55,22 @@ function getAuthMiddleware(secret: string) {
 	);
 }
 
+/**
+ * Wrapper for AxiosError, which includes error status
+ */
+export class JiraClientError extends Error {
```
mboudreau, line 61: Instead of doing a custom error, you can modify the `errorSerializer` in logger.ts to fix this issue globally.  But you'll have to revert the `{ err }` change above.

mboudreau (changes requested):
Please check out global error serializer in logger.ts

src/jira/client/axios.ts
KAbakumovAtl: That's the way I've chosen exactly to avoid doing any modifications in errorSerializer. 

Firstly I've implemented with errorSerializer, but that didn't seem clean to me for the following reasons.

We can't detect if it is AxiosError in the serialiser because it is interface, so we can't do instanceof. 
AxiosError implements its own "toJSON" function, which prevents us from adding fields to it (they won't be logged unless we override toJSON there. 

Another reason to use our own exception type is to enrich it with a "status" parameter which we'll further use for error handling

mboudreau: We can detect it and we can use `instanceof AxiosError` since the error coming into the serializer is not typed at all.  You can even user `err.name` instead if you wanted to for using a switch statement.

The logger doesn't have to use the `.toJSON` function and it's a non-standard function anyways.  You could override it, but frankly, this is the whole point of the serializer in the first place.

AxiosError already has a code/status value.  No need to add an extra class.

mboudreau: right, AxiosError is a type not a class, which is why all axios errors has a boolean field of `isAxiosError` to know that the error is specific to axios.

mboudreau: also, added benefit of having this in the global serializer is that we could test it more thoroughly, which wouldn't be a bad idea for this.

bgvozdev: Having status in the exception is very handy @mboudreau because based on it we could do different strategies for SQS retries. Also, we already have status field in Oktokit's client exception, thus this change would make things more uniform.

KAbakumovAtl: We just brainstormed it with Tarak and we concluded that:
* It is beneficial to remap AxiosError to our internal error model to have stricter typing. It will allow to separate Axios http client domain model from our App domain model. 
* We'll have more control over the error (Aadding status without adding a mess to existing error object)
* Unification. We'll have to map some of errors in GitHub client to RateLimitingError (like we do now in octokit). So using our own errors everywhere will make the approach more uniformed.

It will significantly simplify error handling

src/jira/client/axios.ts
KAbakumovAtl: Yes this. It is error handling path, so should be relatively safe without feature flagging. (In worst case we'll loose some metrics) 

Also was tested in ddev with real failure (404 from deleted Jira instance)

bgvozdev (approved):

tarakpurohit (approved):

thombergs (approved):

bgvozdev (approved):

PR #913, opened 2021-11-15
Adding a workaround to connect to the DB directly for migrations

mboudreau:
Pgbouncer sidecar seems to change the permission model on the DB, setting it as the user itself instead of the role.  Micros data team needs to look further into this, but in the meantime I've found a workaround of using `PG_DATABASE_URL` to connect directly to the repo - circumventing pgbouncer - just for the sequelize migration.


bgvozdev (approved):

db/config.json
```diff
@@ -20,5 +20,18 @@
 			"min": 0,
 			"idle": 10000
     }
-  }
+  },
+	"production-migrate": {
+		"dialect": "postgres",
+		"use_env_variable": "PG_DATABASE_URL",
```
KAbakumovAtl, line 26: So the difference here is in the database url, right? DATABASE_URL goes through the PG Bouncer and PG_DATABASE_URL goes directly to postgres, correct?

mboudreau: Correct.

mboudreau: This is just a workaround for now until pgbouncer permissions are fixed with the sidecar.

thombergs (approved):

PR #915, opened 2021-11-16
Adding jiraHost for Feature Flag

mboudreau:
Forgot to add jiraHost to be able to target specific instances, like arc-team ones.

### DIFF
--- a/src/models/subscription.ts
+++ b/src/models/subscription.ts
@@ -247,7 +247,7 @@ export default class Subscription extends Sequelize.Model {
 		this.changed("repoSyncState", true);
 		await this.save();
 
-		if (await booleanFlag(BooleanFlags.NEW_REPO_SYNC_STATE, true, this.jiraHost)) {
+		if (await booleanFlag(BooleanFlags.NEW_REPO_SYNC_STATE, false, this.jiraHost)) {
 			await RepoSyncState.updateFromJson(this, this.repoSyncState);
 		}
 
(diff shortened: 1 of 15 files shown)


src/models/subscription.ts
```diff
@@ -247,7 +247,7 @@ export default class Subscription extends Sequelize.Model {
 		this.changed("repoSyncState", true);
 		await this.save();
 
-		if (await booleanFlag(BooleanFlags.NEW_REPO_SYNC_STATE, true)) {
+		if (await booleanFlag(BooleanFlags.NEW_REPO_SYNC_STATE, true, this.jiraHost)) {
```
bgvozdev, line 250: AFAIK we should fallback to false, shouldn't we? 🌶️

bgvozdev (approved):

src/models/subscription.ts
mboudreau: yeah, probably.  We really need to fix this for local dev though :/

KAbakumovAtl: I use dev launch darkly for local devloop, there is an instruction on how to do it in confluence go/arc-ops -> howto feature flags

KAbakumovAtl (dismissed):

bgvozdev (dismissed):

bgvozdev (approved):

thombergs (approved):

KAbakumovAtl (approved):

PR #916, opened 2021-11-16
Enabling progressive deployments

KAbakumovAtl:
Trying to enable progressive deployment, as an attempt to fix latency alarms on deployments.


KAbakumovAtl:
Progressive rollout step should be ~5 minutes, and we don't need that much. It is too heavy solution to mitigate a noisy alert

PR #918, opened 2021-11-17
noissue: Add missing required vars.

atrigueiro:


src/config/env.ts
```diff
@@ -17,6 +17,8 @@ const requiredEnvVars = [
 	"SQS_PUSH_QUEUE_URL",
 	"SQS_PUSH_QUEUE_REGION",
 	"MICROS_AWS_REGION",
+	"AWS_ACCESS_KEY_ID",
+	"AWS_SECRET_ACCESS_KEY"
```
mboudreau, line 21: I've made this mistake before.  You only need this on your development flow, but not on prod and this will cause a deployment error.

I have a ticket to separate the .env files further for common/default, dev and test or maybe to just add those 2 as required for dev.

mboudreau (changes requested):
This PR isn't needed and should be closed.

PR #919, opened 2021-11-17
ARC-788 Fix approximate receive counter

KAbakumovAtl:
* Added `lastAttempt` parameter to the queue execution context
* Fixed receiveCount calcluation
* Added a test

### DIFF
--- a/src/sqs/index.ts
+++ b/src/sqs/index.ts
@@ -39,6 +39,12 @@ export type Context<MessagePayload> = {
 	 * How many times this messages attempted to be processed, including the current attempt
 	 */
 	receiveCount: number;
+
+
+	/**
+	 * Indicates if it is the last attempt to process this message
+	 */
+	lastAttempt: boolean;
 }
 
 export type QueueSettings = {
@@ -265,7 +271,8 @@ export class SqsQueue<MessagePayload> {
 		const params = {
 			QueueUrl: this.queueUrl,
 			MaxNumberOfMessages: 1,
-			WaitTimeSeconds: this.longPollingIntervalSec
+			WaitTimeSeconds: this.longPollingIntervalSec,
+			AttributeNames: ["ApproximateReceiveCount"]
 		};
 
 		try {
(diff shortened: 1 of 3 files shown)


test/unit/sqs/index.test.ts
```diff
@@ -244,16 +254,56 @@ describe("SqsQueue tests", () => {
 
 			mockErrorHandler.mockImplementation((error: Error, context: Context<TestMessage>) => {
 
-				expect(context.payload.msg).toBe(testPayload.msg);
-				expect(error).toBeInstanceOf(SqsTimeoutError);
+				try {
+					expect(context.payload.msg).toBe(testPayload.msg);
+					expect(error).toBeInstanceOf(SqsTimeoutError);
```
KAbakumovAtl, line 259: Turns out if expect fails here it doesn't fail the test (cuz exception doesn't reach the top level). So we have to handle it and call fail explicitly

thombergs (approved):

src/sqs/index.ts
```diff
@@ -318,9 +325,9 @@ export class SqsQueue<MessagePayload> {
 			executionId: uuidv4(),
 			queue: this.queueName})
 
-		const receiveCount = Number(message.MessageAttributes?.ApproximateReceiveCount?.StringValue || "1");
+		const receiveCount = Number(message.Attributes?.ApproximateReceiveCount || "1");
 
-		const context: Context<MessagePayload> = {message, payload, log, receiveCount: receiveCount}
+		const context: Context<MessagePayload> = {message, payload, log, receiveCount: receiveCount, lastAttempt: receiveCount===this.maxAttempts}
```
bgvozdev, line 330: nit: Shall we maybe use `>=`, just in case? :)

src/sqs/index.ts
```diff
@@ -351,7 +358,7 @@ export class SqsQueue<MessagePayload> {
 					(errorHandlingResult.skipDlq && context.receiveCount >= this.maxAttempts)) {
 					log.info("Deleting the message hence it reached the maximum amount of retries")
 					await this.deleteMessage(message, log)
-				} else if (errorHandlingResult.retryDelaySec) {
+				} else if (errorHandlingResult.retryDelaySec !== undefined /*zero delay supported*/) {
```
bgvozdev, line 361: nit: it took me 15-20 seconds to digest the comment. Might something like `"zero seconds delay is also supported" ` be more clear?

bgvozdev (approved):

src/sqs/index.ts
KAbakumovAtl: Oh yes, exactly! In case of DLQ replay

bgvozdev (approved):

thombergs (approved):

PR #921, opened 2021-11-17
Fixing redis queue for tests by adding prefix

mboudreau:
Fixes flakiness in tests that used the Redis bull queues by creating separate queues for the tests compared to development so the running app doesn't consume the test messages.


src/worker/queues.ts
```diff
@@ -26,6 +27,7 @@ const getQueueOptions = (timeout: number): QueueOptions => {
 			removeOnComplete: true,
 			removeOnFail: true
 		},
+		prefix: envVars.BULL_QUEUE_PREFIX,
```
KAbakumovAtl, line 30: Looks like a very simple and elegant solution.

just want to make sure that if we don't have this env var in prod we won't add any prefixes. (Cuz I am afraid if we'll start using another "queue" in prod and loose messages from the one we were using before.)

Is there any way to test it in ddev/staging? Would be great to do before merging

mboudreau: Well, the env var isn't set anywhere else but in `.env.test` and will just return `undefined` for all other environments, which will then default it.

However, that being said, I think I found a better option which is to set the key prefix for _all redis keys_ and not just bull, so that no tests ever affects development.

Happy to deploy to staging just to be 100% sure though. One way to test this would be to start a big sync before, deploy this to staging, then make sure the sync continues afterwards.

mboudreau: actually, nvm, using redis `keyPrefix` didn't seem to work well with bull queues for some reason.

bgvozdev (approved):

mboudreau:
@KAbakumovAtl I've deployed to staging while in the middle of a sync to be 100% sure that the new setting doesn't affect anything by testing and it worked fine.

thombergs (approved):

KAbakumovAtl (approved):

PR #922, opened 2021-11-17
ARC-788 error handling for the push queue (SQS)

KAbakumovAtl:
In this PR:
* Metrics emission from processPush function was removed
* Added push queue error handler & tests

### DIFF
--- a/src/sqs/push.ts
+++ b/src/sqs/push.ts
@@ -41,24 +41,30 @@ export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> =
 	await processPush(github, payload, wrapLogger(context.log));
 }
 
+/**
+ * Sometimes we can get errors from Jira and GitHub which does not indicate a failured webhook. For example:
+ *  Jira site is gone, we'll get 404
+ *  GitHub App is not installed anymore we'll get 401
+ *  etc.
+ *
+ *  In such cases webhook processing doesn't make sense anymore and we need to silently discard these errors
+ */
+const UNRETRYABLE_STATUS_CODES = [401, 404, 403];
 
-const NON_ERRONEOUS_STATUSES = [401, 404, 403];
-const RATE_LIMITING_DELAY_BUFFER = 10;
-const EXPONENTIAL_BACKOFF_BASE = 60;
+const RATE_LIMITING_DELAY_BUFFER_SEC = 10;
+const EXPONENTIAL_BACKOFF_BASE_SEC = 60;
+const EXPONENTIAL_BACKOFF_MULTIPLIER = 3;
 export const pushQueueErrorHandler : ErrorHandler<PushQueueMessagePayload> = async (error: JiraClientError | Octokit.HookError | RateLimitingError | Error,
 	context: Context<PushQueueMessagePayload>) : Promise<ErrorHandlingResult> => {
 
-	//Handle non-failure cases
 	const maybeResult = maybeHandleNonFailureCase(error, context);
-	if(maybeResult) {
+	if (maybeResult) {
 		return maybeResult;
 	}
 
-	//Handle failure cases
 	const errorHandlingResult = handleFailureCase(error, context);
 
-	if(!errorHandlingResult.retryable || context.lastAttempt ) {
-		//Emit metrics if the failure is not retryable
+	if (!errorHandlingResult.retryable || context.lastAttempt ) {
 		context.log.error({error}, "Webhook push processing failed and won't be retried anymore");
 		emitWebhookFailedMetrics("push")
 	}
@@ -67,9 +73,9 @@ export const pushQueueErrorHandler : ErrorHandler<PushQueueMessagePayload> = asy
 }
 
 function maybeHandleNonFailureCase(error: Error, context: Context<PushQueueMessagePayload>): ErrorHandlingResult | undefined {
-	if(error instanceof JiraClientError &&
+	if (error instanceof JiraClientError &&
 		error.status &&
-		NON_ERRONEOUS_STATUSES.includes(error.status)) {
+		UNRETRYABLE_STATUS_CODES.includes(error.status)) {
 		context.log.warn(`Received ${error.status} from Jira. Unretryable. Discarding the message`);
 		return {retryable: false}
 	}
@@ -78,23 +84,21 @@ function maybeHandleNonFailureCase(error: Error, context: Context<PushQueueMessa
 	//Unfortunately we can't check if error is instance of Octokit.HookError because it is not a calss, so we'll just rely on status
 	//TODO Add error handling for the new GitHub client when it will be done
 	const maybeErrorWithStatus : any = error;
-	if(maybeErrorWithStatus.status && NON_ERRONEOUS_STATUSES.includes(maybeErrorWithStatus.status)) {
+	if (maybeErrorWithStatus.status && UNRETRYABLE_STATUS_CODES.includes(maybeErrorWithStatus.status)) {
 		context.log.warn({err: maybeErrorWithStatus}, `Received error with ${maybeErrorWithStatus.status} status. Unretryable. Discarding the message`);
 		return {retryable: false}
 	}
 
 	return undefined;
-
 }
 
 function handleFailureCase(error: Error, context: Context<PushQueueMessagePayload>): ErrorHandlingResult {
 	if (error instanceof RateLimitingError) {
-		// delaying until the rate limit is reset (plus a buffer of a couple seconds)
-		const delaySec = error.rateLimitReset + RATE_LIMITING_DELAY_BUFFER - (new Date().getTime() / 1000);
+		const delaySec = error.rateLimitReset + RATE_LIMITING_DELAY_BUFFER_SEC - (new Date().getTime() / 1000);
 		return {retryable: true, retryDelaySec: delaySec}
 	}
 
-	//Use exponential backoff in any other case
-	const delaySec = EXPONENTIAL_BACKOFF_BASE * Math.pow(3, context.receiveCount);
+	//In case if error is unknown we should use exponential backoff
+	const delaySec = EXPONENTIAL_BACKOFF_BASE_SEC * Math.pow(EXPONENTIAL_BACKOFF_MULTIPLIER, context.receiveCount);
 	return {retryable: true, retryDelaySec: delaySec}
 }
--- a/src/sqs/push.ts
+++ b/src/sqs/push.ts
@@ -1,9 +1,11 @@
-import {Context, MessageHandler} from "./index"
-import enhanceOctokit from "../config/enhance-octokit";
+import {Context, ErrorHandler, ErrorHandlingResult, MessageHandler} from "./index"
+import enhanceOctokit, {RateLimitingError} from "../config/enhance-octokit";
 import {processPush} from "../transforms/push";
 import app from "../worker/app";
 import {wrapLogger} from "probot/lib/wrap-logger";
-
+import {JiraClientError} from "../jira/client/axios";
+import { Octokit } from "@octokit/rest";
+import {emitWebhookFailedMetrics} from "../util/webhooks";
 
 export type PayloadRepository = {
 	id: number,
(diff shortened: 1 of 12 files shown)


src/sqs/queues.ts
```diff
@@ -22,7 +22,7 @@ const sqsQueues = {
 		queueRegion: envVars.SQS_PUSH_QUEUE_REGION,
 		longPollingIntervalSec: LONG_POLLING_INTERVAL_SEC,
 		timeoutSec: 60,
-		maxAttempts: 5}, pushQueueMessageHandler,	defaultErrorHandler),
+		maxAttempts: 5}, pushQueueMessageHandler,	pushQueueErrorHandler),
```
KAbakumovAtl, line 25: We don't need a feature flag for this change because if something goes wrong we can always use existing feature flag to fallback to the Redis queue.

src/transforms/push.ts
```diff
@@ -106,110 +106,103 @@ export function processPushJob(app: Application) {
 }
 
 export const processPush = async (github: GitHubAPI, payload, rootLogger: LoggerWithTarget) => {
```
KAbakumovAtl, line 108: The only thing I did here is removing try/catch block
```
} catch (err) {
		log.error({ err }, "Failed to process push");
		emitWebhookFailedMetrics("push");
		throw err;
}
```

We don't need to do it here because:
1. If this function fail in the middleware during webhook processing, we already emit metrics in the middleware error handling (currently we emitting failure metrics twice)
2. For SQS we'll have its own error handling. (Otherwise we'll emit a failure on each retry)
3. We don't use Redis for pushes anymore, so we won't loosing its metrics

src/sqs/push.ts
```diff
@@ -39,3 +41,60 @@ export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> =
...
+
+const NON_ERRONEOUS_STATUSES = [401, 404, 403];
+const RATE_LIMITING_DELAY_BUFFER = 10;
+const EXPONENTIAL_BACKOFF_BASE = 60;
+export const pushQueueErrorHandler : ErrorHandler<PushQueueMessagePayload> = async (error: JiraClientError | Octokit.HookError | RateLimitingError | Error,
+	context: Context<PushQueueMessagePayload>) : Promise<ErrorHandlingResult> => {
+
+	//Handle non-failure cases
+	const maybeResult = maybeHandleNonFailureCase(error, context);
+	if(maybeResult) {
```
bgvozdev, line 53: supernit: here and in other cases: there should either be a space between `if` and `(` everywhere or nowhere. I see both in different parts of your PR

src/sqs/push.ts
```diff
@@ -39,3 +41,60 @@ export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> =
...
 }
 
+
+const NON_ERRONEOUS_STATUSES = [401, 404, 403];
+const RATE_LIMITING_DELAY_BUFFER = 10;
+const EXPONENTIAL_BACKOFF_BASE = 60;
+export const pushQueueErrorHandler : ErrorHandler<PushQueueMessagePayload> = async (error: JiraClientError | Octokit.HookError | RateLimitingError | Error,
+	context: Context<PushQueueMessagePayload>) : Promise<ErrorHandlingResult> => {
+
+	//Handle non-failure cases
```
bgvozdev, line 51: nit: the comment repeats the code, I suggest to remove it

src/sqs/push.ts
```diff
@@ -39,3 +41,60 @@ export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> =
...
+	const maybeResult = maybeHandleNonFailureCase(error, context);
+	if(maybeResult) {
+		return maybeResult;
```
bgvozdev, line 54: nit: Would `return !maybeResult` make it clearer?

src/sqs/push.ts
```diff
@@ -39,3 +41,60 @@ export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> =
...
+	}
+
+	//Handle failure cases
```
bgvozdev, line 57: nit: same, IMHO the commend should tell "why?", not "what?". I'd suggest to remove the comment

src/sqs/push.ts
```diff
@@ -39,3 +41,60 @@ export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> =
...
+	const errorHandlingResult = handleFailureCase(error, context);
+
+	if(!errorHandlingResult.retryable || context.lastAttempt ) {
+		//Emit metrics if the failure is not retryable
```
bgvozdev, line 61: nit: same, IMHO the commend should tell "why?", not "what?". I'd suggest to remove the comment

src/sqs/push.ts
```diff
@@ -39,3 +41,60 @@ export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> =
...
+		emitWebhookFailedMetrics("push")
+	}
+
+	return errorHandlingResult;
+}
+
+function maybeHandleNonFailureCase(error: Error, context: Context<PushQueueMessagePayload>): ErrorHandlingResult | undefined {
+	if(error instanceof JiraClientError &&
+		error.status &&
+		NON_ERRONEOUS_STATUSES.includes(error.status)) {
```
bgvozdev, line 72: nit: I'd suggest `NON_FAILURE_STATUSES`, because you've used word "failure" for this situation before (less notions => easier to follow IMHO)

src/sqs/push.ts
```diff
@@ -39,3 +41,60 @@ export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> =
...
+		context.log.warn(`Received ${error.status} from Jira. Unretryable. Discarding the message`);
+		return {retryable: false}
+	}
+
+	//If error is Octokit.HookError, then we need to check the response status
+	//Unfortunately we can't check if error is instance of Octokit.HookError because it is not a calss, so we'll just rely on status
+	//TODO Add error handling for the new GitHub client when it will be done
+	const maybeErrorWithStatus : any = error;
```
bgvozdev, line 80: nit: Could it be `const { status } = error;` instead?

src/sqs/push.ts
```diff
@@ -39,3 +41,60 @@ export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> =
...
+		return {retryable: false}
+	}
+
+	return undefined;
+
+}
+
+function handleFailureCase(error: Error, context: Context<PushQueueMessagePayload>): ErrorHandlingResult {
+	if (error instanceof RateLimitingError) {
+		// delaying until the rate limit is reset (plus a buffer of a couple seconds)
```
bgvozdev, line 92: nit: same, IMHO the commend should tell "why?", not "what?". I'd suggest to remove the comment

src/sqs/push.ts
```diff
@@ -39,3 +41,60 @@ export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> =
 	await processPush(github, payload, wrapLogger(context.log));
 }
 
+
+const NON_ERRONEOUS_STATUSES = [401, 404, 403];
+const RATE_LIMITING_DELAY_BUFFER = 10;
```
bgvozdev, line 46: Should we maybe also include measurement units? Is it seconds? Minutes?

src/sqs/push.ts
```diff
@@ -39,3 +41,60 @@ export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> =
...
+const EXPONENTIAL_BACKOFF_BASE = 60;
```
bgvozdev, line 47: Should we maybe also include measurement units? Is it seconds? Minutes?

src/sqs/push.ts
bgvozdev: Actually, should it be something like `RETRYABLE_STATUSES`? I feel like "retryable/non-retryable" is more clear and explicit than "error/failure"....

bgvozdev: And the same goes to the names of functions (e.g. `maybeHandleRetryableCase`)...

src/sqs/push.ts
```diff
@@ -39,3 +41,60 @@ export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> =
...
+}
+
+function handleFailureCase(error: Error, context: Context<PushQueueMessagePayload>): ErrorHandlingResult {
+	if (error instanceof RateLimitingError) {
+		// delaying until the rate limit is reset (plus a buffer of a couple seconds)
+		const delaySec = error.rateLimitReset + RATE_LIMITING_DELAY_BUFFER - (new Date().getTime() / 1000);
+		return {retryable: true, retryDelaySec: delaySec}
+	}
+
+	//Use exponential backoff in any other case
```
bgvozdev, line 97: nit: same, IMHO the commend should tell "why?", not "what?". I'd suggest to remove the comment

src/sqs/push.ts
```diff
@@ -39,3 +41,60 @@ export const pushQueueMessageHandler : MessageHandler<PushQueueMessagePayload> =
...
+	const delaySec = EXPONENTIAL_BACKOFF_BASE * Math.pow(3, context.receiveCount);
```
bgvozdev, line 98: Should 3 be a constant, too?

test/unit/sqs/push.test.ts
```diff
@@ -0,0 +1,134 @@
...
+import {getLogger} from "../../../src/config/logger";
+import {JiraClientError} from "../../../src/jira/client/axios";
+import {RateLimitingError} from "../../../src/config/enhance-octokit";
+import {Octokit} from "probot";
+
+
+describe("Push Queue Error Handler", () => {
+
+
+	let statsdIncrementSpy = jest.spyOn(statsd, "histogram");
```
bgvozdev, line 13: supernit: formatting (empty lines)

test/unit/sqs/push.test.ts
```diff
@@ -0,0 +1,134 @@
...
+
+	it("Sends metrics when it was last attempt", async () => {
+
+		await pushQueueErrorHandler(new Error(), createContext(1, true));
+
+		expect(statsdIncrementSpy).toBeCalledTimes(1);
+	});
+
+
+	function getJiraClientError(code: number) {
```
bgvozdev, line 73: supernit: too many empty lines

bgvozdev:
Great work, @KAbakumovAtl ! In general it looks good to me, the only sensible comment I have is about the notions that you are using. I see that you've used 3 different words: failure, error and non-retryable. I'd suggest to stick to a single one and rename everything else to use it. My personal perefence is "retryable/non-retryable" cause I think it is pretty explicit and describes the behaviour of the system best.

- [ ] consider unifying the used notions describing "retryable/non-retryable" errors

src/sqs/push.ts
KAbakumovAtl: But if there is a result we don't need to return

bgvozdev: Ah, wait, sorry, my bad! I meant `return maybeResult!` 🤦  🤣  Explicitly show in the code that we are returning a value, not "maybe" value - this is what I wanted

src/sqs/push.ts
KAbakumovAtl: Looks like we need to add some liner rules so it will be auto-formatted

src/sqs/push.ts
KAbakumovAtl: Ok, I guess log message explains it enough

src/sqs/push.ts
KAbakumovAtl: So the thing is when it is not a failure, we won't send a faulre metric. 

So I wanted to name it to reflect taht it is not a failure

src/sqs/push.ts
KAbakumovAtl: That was done to avoid typescript errors. If I do `const { status } = error;` It will say that the error doesnt' have "status" field. However its subclasses can. So I had to do this to avoid ignoring typescript&linter errors

src/sqs/push.ts
bgvozdev: "Non-failure = we retry", is my understanding correct?

KAbakumovAtl:
I fixed what I could. Ready for re-review

bgvozdev (dismissed):

bgvozdev (approved):

tarakpurohit (approved):

PR #925, opened 2021-11-18
Add webhook payload metric behind a FF

tarakpurohit:
This pull request adds code to record payload size in bytes.


src/github/middleware.ts
```diff
@@ -77,6 +77,12 @@ export default (
 		enhanceOctokit(context.github);
 		const webhookEvent = extractWebhookEventNameFromContext(context);
 
+		// Metrics for webhook payload size
+		if (await booleanFlag(BooleanFlags.PAYLOAD_SIZE_METRIC, false, jiraHost)) {
+			emitWebhookPayloadMetrics(webhookEvent,
+				Buffer.byteLength(JSON.stringify(context.payload), "utf-8"));
+		}
```
KAbakumovAtl, line 84: Nice!

KAbakumovAtl (dismissed):

mboudreau (approved):

src/github/middleware.ts
```diff
@@ -173,6 +167,12 @@ export default (
 			context.log = context.log.child({ jiraHost });
 			context.log("Processing event for Jira Host");
 
+			// Metrics for webhook payload size
+			if (await booleanFlag(BooleanFlags.PAYLOAD_SIZE_METRIC, false, jiraHost)) {
+				emitWebhookPayloadMetrics(webhookEvent,
+					Buffer.byteLength(JSON.stringify(context.payload), "utf-8"));
+			}
+
```
KAbakumovAtl, line 175: If we put it here the metric won't be accurate anymore, because it will be published per each Jira Instance subscription. So it could be multiple times per webhook.

What if we'll use this flag without Jira Host? We don't really need the progressive rollout here, so we don't really need Jira Host.

tarakpurohit: Thats correct. I will remove the `jiraHost` value and move it outside to where it was.

mboudreau (approved):

bgvozdev (approved):

PR #928, opened 2021-11-22
ARC-814 RepoSyncState table as source of truth

mboudreau:
Adding new FF and making all 'reads' for repo sync state come from the new table.  We're still writing to the JSON blob column so we can switch back in case of issues.  Next step after fully deployed is to remove the repoSyncState column and all references to it.


src/api/index.ts
```diff
@@ -171,7 +172,11 @@ router.get(
 				return;
 			}
 
-			res.json(subscription.repoSyncState);
+			if(await booleanFlag(BooleanFlags.REPO_SYNC_STATE_AS_SOURCE, false)) {
```
thombergs, line 175: We should pass the `jiraHost` into the feature flag evaluation, otherwise we can't do a progressive rollout.

test/frontend/util/get-jira-configuration-helpers.test.ts
```diff
@@ -9,88 +16,214 @@ describe("getFailedConnections", () => {
...
+		await Subscription.destroy({ truncate: true });
+		await RepoSyncState.destroy({ truncate: true });
+	});
+
+	it("Should get installation from repoSyncState", async () => {
+		const result = await getInstallation(GitHubAPI(), subscription);
+		expect(result).toMatchObject(match);
+	});
+
+	it("Should get installation from Repo Sync State table", async () => {
```
thombergs, line 192: Nice to see that the tests for the old and new logic both assert against the same expected object.

thombergs (commented):
The `jiraHost` is missing on one feature flag evaluation, otherwise, looks good!

src/api/index.ts
mboudreau: this is for the admin API, so I wasn't sure how to approach this.  I can add it, but that means that as an admin API, we might have different functionality based on which customer is do the action for.

thombergs: yeah, this is alright, then, I guess. However, we have to be aware that we have to excplicitly enable a feature flag for the `global` user when we're not passing in a jira host URL, here. I just added the `global` user to our team segment so that these features are automatically enabled for us in production.

thombergs (approved):

package-lock.json
```diff
@@ -8092,7 +8101,7 @@
       "dependencies": {
         "clone": {
           "version": "2.1.2",
-          "resolved": "https://registry.npmjs.org/clone/-/clone-2.1.2.tgz",
+          "resolved": "https://packages.atlassian.com/api/npm/npm-remote/clone/-/clone-2.1.2.tgz",
```
thombergs, line 8104: I believe this should point to `npmjs.org` instead of `packages.atlassian.com`?

package-lock.json
```diff
@@ -10024,7 +10033,7 @@
     },
     "semver-compare": {
       "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/semver-compare/-/semver-compare-1.0.0.tgz",
+      "resolved": "https://packages.atlassian.com/api/npm/npm-remote/semver-compare/-/semver-compare-1.0.0.tgz",
```
thombergs, line 10036: see above

src/frontend/get-jira-configuration.ts
```diff
@@ -104,7 +112,7 @@ export default async (
 			)
 		);
 
-		const failedConnections = getFailedConnections(
+		const failedConnections = await getFailedConnections(
```
thombergs, line 115: does this mean `failedConnections` has always been an unresolved promise?

src/models/reposyncstate.ts
```diff
@@ -98,9 +181,23 @@ export default class RepoSyncState extends Sequelize.Model {
...
+				repoId
+			}
+		});
+		return model?.setFromRepositoryData(repo)?.save();
+	}
+
+	setFromRepositoryData(repo?: RepositoryData): RepoSyncState {
 		if (repo) {
-			this.repoUpdatedAt = new Date(repo.repository?.updated_at || Date.now());
+			this.repoUpdatedAt = new Date(repo.repository?.updated_at ?? Date.now());
```
thombergs, line 200: Why use `??` instead of `||` here?

thombergs (approved):

src/frontend/get-jira-configuration.ts
mboudreau: No, it used to be a sync function which I changed to async for the `booleanFlag` call

src/models/reposyncstate.ts
mboudreau: `??` is null coalescing since we're looking for if it's `undefined` or `null`.  A test that we use is using `new Date(0)` because we wanted to compare the specific 8601 date and instead of using a random number, 0 seemed easy, but 0 is falsy, hence it would return `Date.now()` and fail the test.

thombergs (approved):

KAbakumovAtl:
Please link the filled up FD ticket

mboudreau:
@KAbakumovAtl https://hello.atlassian.net/browse/FD-27945

src/frontend/get-jira-configuration.ts
```diff
@@ -1,13 +1,15 @@
...
+import { RepoSyncState, Subscription } from "../models";
 import { NextFunction, Request, Response } from "express";
 import statsd from "../config/statsd";
 import { metricError } from "../config/metric-names";
 import { FailedInstallations } from "../config/interfaces";
+import { booleanFlag, BooleanFlags } from "../config/feature-flags";
+import Logger from "bunyan";
 
-function mapSyncStatus(syncStatus: string): string {
+function mapSyncStatus(syncStatus: SyncStatus = SyncStatus.PENDING): string {
```
KAbakumovAtl, line 12: In which case this method will have to use the default parameter?

mboudreau: `Subscription.syncStatus` is optional, we needed a way to handle in the off chance that it wasn't set yet, but almost certainly would be by this point as this is for current subscriptions which any new subscriptions is automatically synced and set to `PENDING`.  https://github.com/atlassian/github-for-jira/blob/main/src/models/subscription.ts#L166

src/frontend/get-jira-configuration.ts
```diff
@@ -57,27 +62,30 @@ interface FailedConnections {
...
-			const repoId = Object.keys(repos);
-			const orgName = repos[repoId[0]]?.repository?.owner.login || undefined;
+		.map(async (failedConnection: FailedInstallations) => {
+			const sub = subscriptions.find(sub => failedConnection.id === sub.gitHubInstallationId);
+			let orgName;
+			if (sub && await booleanFlag(BooleanFlags.REPO_SYNC_STATE_AS_SOURCE, false, sub.jiraHost)) {
+				const repo = await RepoSyncState.findOneFromSubscription(sub);
+				orgName = repo.repoOwner;
+			} else {
+				const repos = sub?.repoSyncState?.repos || {};
```
KAbakumovAtl, line 78: At this point would sub already have repoSyncState retrieved from DB? 

If yes, how does our flag prevents us from retrieving it from the old table?

src/models/installation.ts
```diff
@@ -94,7 +95,7 @@ export default class Installation extends Sequelize.Model {
 		await this.destroy();
 	}
 
-	async subscriptions(): Promise<Subscription[]> {
+	async subscriptions(): Promise<SubscriptionClass[]> {
```
KAbakumovAtl, line 98: I wonder what's the difference between Subscription and SubscriptionClass? And how does this type change affect things?

src/frontend/get-jira-configuration.ts
mboudreau: it doesn't.  I'm not sure I understand why it shouldn't.  We should be able to flip between the feature flag on/off without any data loss or issues.  The next PR, after this one is 100%, will remove the `repoSyncState` column altogether which will now reduce the DB load considerably.

src/models/installation.ts
mboudreau: yeah, we need to fix that at some point.  It's because we're using an export default class called `Subscription` at `src/models/subscription` but we also have a export const value called `Subscription` which is an instance of the class that's being built by Sequelize and provided by `src/models/index`.  Values and Class types are different and you cannot place a value as a type of a function.  If you look at the imports at the top of the file, I had to import both the class type and the value as you need to refer to the value if you want to do operations on the DB with sequelize or else it'll just error at runtime.

KAbakumovAtl:
> @KAbakumovAtl https://hello.atlassian.net/browse/FD-27945
Thanks, 

Just a few things:

* The following fields are missing:
  * Roll-out Plan
  * Due date
  (Please refer to go/arc-flags for the fields guide)

* For "Support Readiness" we should say that the change potentially affects Backfill process
* Optionally for monitoring we can also add a Splunk query to errors in the Backfill queue

src/frontend/get-jira-configuration.ts
KAbakumovAtl: Cool, so this PR is to make a flag which changes the logic to not use `repoSyncState` anymore and then in another PR we'll stop fetching it when the flag is rolled out.

Am I understand it correctly?

mboudreau: Correct :)

mboudreau:
@KAbakumovAtl done.

thombergs (approved):

KAbakumovAtl (approved):

PR #929, opened 2021-11-22
ARC-819: remove rate-limiting logic on blanket 403 and instead log an…

thombergs:


src/config/enhance-octokit.ts
```diff
@@ -29,11 +29,11 @@ const instrumentRequests = (octokit: GitHubAPI) => {
 			throw new RateLimitingError(rateLimitReset);
 		}
 
-		if (error.status === 403) {
-			// delaying for an hour
-			logger.warn({ err: error }, "GitHub returned 403");
-			throw new RateLimitingError(new Date().getTime() / 1000 + 60 * 60);
```
thombergs, line 35: previously, we would throw a `RateLimitingError` on every 403 from GitHub, which is wrong

src/config/enhance-octokit.ts
```diff
@@ -29,11 +29,11 @@ const instrumentRequests = (octokit: GitHubAPI) => {
...
+		if (error.status === 403 && error.stack?.includes("has an IP allow list enabled")) {
+			logger.warn({ err: error }, "blocked by GitHub allowlist!");
```
mboudreau, line 33: I'd probably make that an error, not a warn

src/config/enhance-octokit.ts
```diff
@@ -29,11 +29,11 @@ const instrumentRequests = (octokit: GitHubAPI) => {
...
+			statsd.increment(metricError.blockedByGitHubAllowlist);
 		}
```
thombergs, line 35: Now, we're checking the stack trace for a special error message from GitHub and emit a metric to make IP allowlist errors more prominent (I'll later add a low-prio alert on that metric)

src/config/enhance-octokit.ts
thombergs: done

src/config/enhance-octokit.ts
thombergs: Tested the metric and new log event in staging.

rachellerathbone (approved):

atrigueiro (approved):

PR #930, opened 2021-11-22
[INSTANT KARMA] Issue/none clean up deduplicator feature flag

bgvozdev:


src/sync/installation.ts
```diff
@@ -146,11 +143,7 @@ const updateJobStatus = async (
 		// there's more data to get
 		await subscription.updateRepoSyncStateItem(repositoryId, getCursorKey(task), edges[edges.length - 1].cursor);
 
-		if (useScheduleNextTask) {
-			scheduleNextTask(0);
-		} else {
-			queues.installation.add(job.data);
-		}
+		scheduleNextTask(0);
```
mboudreau, line 146: wouldn't it make more sense to have the delay in `scheduleNextTask` default to 0 so you don't have to specify it everywhere?

thombergs (approved):

src/sync/installation.ts
bgvozdev: It would, but I'd say that 1. that's a "nit" and 2. that's a FF cleanup, which means it would've diverted from the purpose of this PR, so I'm not too keen of doing this change here...

atrigueiro (approved):

PR #931, opened 2021-11-23
Arc 808 error handling

tarakpurohit:
Adds RateLimiting error and error handler for new Github client requests.


src/github/client/interceptors.ts
```diff
@@ -0,0 +1,103 @@
...
+export const handleFailedRequest = (logger: Logger) =>
+	(error) => {
+		const response = error.response;
+		if(response) {
+			const status = response?.status;
+			const errorMessage = `Error executing Axios Request ` + error.message;
+
+			const rateLimitResetHeaderValue:string = response.headers?.["x-ratelimit-reset"];
+			if(response.headers?.["x-ratelimit-remaining"] == "0" && rateLimitResetHeaderValue)
+			{
```
thombergs, line 85: /nitpick newline?

src/github/client/interceptors.ts
```diff
@@ -0,0 +1,103 @@
...
+				logger.warn({ err: error }, "Rate limiting error");
+				const rateLimitReset: number = parseInt(rateLimitResetHeaderValue);
+				return Promise.reject(new RateLimitingError(rateLimitReset, error, status));
```
thombergs, line 88: The `RateLimitingError` should also contain a field `rateLimitRemaining`, because the new backfilling mechanism may use that to decide to delay messages. We could use that field to implement a "reserve" of requests so we don't exhaust the rate limit completely.
cc @mboudreau

src/github/client/github-client.ts
```diff
@@ -1,29 +1,46 @@
...
 	) {
+		this.logger = logger || getLogger("github.client.axios");
 		this.axios = axios.create({
 			baseURL
 		});
+		this.axios.interceptors.request.use(urlParamsMiddleware)
+		this.axios.interceptors.request.use(setRequestStartTime);
+		this.axios.interceptors.response.use(
+			undefined,
+			handleFailedRequest(this.logger)
```
thombergs, line 38: Is there no better way to add an error handler with Axios? I remember stumbling across this when building the initial version of the `GitHubClient`, but I couldn't find a better solution, either. Passing `undefined` is a code smell for me.

src/github/client/interceptors.ts
```diff
@@ -0,0 +1,103 @@
...
+			}
+
+			if (status === 403)
+			{
+				logger.warn({ err: error }, "Github returned 403 without ratelimit header");
+				return Promise.reject(new RateLimitingError(Date.now() / 1000 + 60 * 60, error, status));
+			}
```
thombergs, line 95: This is the part that I modified in the old `enhance-octokit.ts` file. Please replace this if statement with the following code:

```
if (error.status === 403 && error.stack?.includes("has an IP allow list enabled")) {
			logger.error({ err: error }, "blocked by GitHub allowlist!");
			statsd.increment(metricError.blockedByGitHubAllowlist);
}
```
(you can copy it from `enhance-octokit.ts`

src/github/client/interceptors.ts
tarakpurohit: That makes sense. I will add that in 👍

src/github/client/github-client.ts
tarakpurohit: I couldnt find any better way then this. I hoped that they would have taken an array of interceptors but could find any version that did that.

bgvozdev (approved):

thombergs (approved):

src/github/client/github-client.ts
```diff
@@ -85,21 +103,30 @@ export default class GitHubClient {
...
+			repo
+		});
 	}
 
 	/**
 	 * Get a single pull request for the given repository.
 	 */
 	public async getPullRequest(owner: string, repo: string, pullNumber: string): Promise<AxiosResponse<Octokit.PullsGetResponse>> {
-		return await this.get<Octokit.PullsGetResponse>(`/repos/${owner}/${repo}/pulls/${pullNumber}`);
+		return await this.get<Octokit.PullsGetResponse>(`/repos/:owner/:repo/pulls/:pullNumber`, {
```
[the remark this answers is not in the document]
mboudreau: I think using string interpolation is fine with Github as all their orgs, repo names and IDs are URL safe.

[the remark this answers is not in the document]
tarakpurohit: Also, this was changed because the `path` is passed in as a tag on metrics. This clusters them together instead of unqiue paths.

PR #933, opened 2021-11-24
Issue/ARC-707 making sync queue agnostic

bgvozdev:
Once we remove the FF, the sync code will become queue-agnostic which we could use with both SQS and Redis queues (another step towards the migration)

### DIFF
--- a/src/models/subscription.ts
+++ b/src/models/subscription.ts
@@ -190,7 +190,7 @@ export default class Subscription extends Sequelize.Model {
 	static async startNewSyncProcess(subscription: Subscription) {
 		//TODO Add a sync start logic
 		await sqsQueues.backfill.sendMessage({
-			gitHubInstallationId: subscription.gitHubInstallationId,
+			installationId: subscription.gitHubInstallationId,
 			jiraHost: subscription.jiraHost
 		});
 	}
--- a/src/worker/startup.ts
+++ b/src/worker/startup.ts
@@ -127,7 +127,18 @@ export async function start() {
 	queues.discovery.process(5, commonMiddleware(discovery(app, queues), DISCOVERY_LOGGER_NAME));
 	queues.installation.process(
 		Number(CONCURRENT_WORKERS),
-		commonMiddleware(processInstallation(app, queues), INSTALLATION_LOGGER_NAME)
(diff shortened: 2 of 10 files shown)


src/models/subscription.ts
```diff
@@ -190,7 +190,7 @@ export default class Subscription extends Sequelize.Model {
 	static async startNewSyncProcess(subscription: Subscription) {
 		//TODO Add a sync start logic
 		await sqsQueues.backfill.sendMessage({
-			gitHubInstallationId: subscription.gitHubInstallationId,
+			installationId: subscription.gitHubInstallationId,
```
mboudreau, line 193: this could be confusing as we do have an 'Installation' model with an id

src/worker/startup.ts
```diff
@@ -127,7 +127,18 @@ export async function start() {
 	queues.discovery.process(5, commonMiddleware(discovery(app, queues), DISCOVERY_LOGGER_NAME));
 	queues.installation.process(
 		Number(CONCURRENT_WORKERS),
-		commonMiddleware(processInstallation(app, queues), INSTALLATION_LOGGER_NAME)
+		commonMiddleware(processInstallation(app, queues, () => {
+			return Promise.resolve({
+				schedule: async (jobData, delayMsecs) => {
+					// TBD: switch to SQS with a FF
+					if ((delayMsecs || 0) > 0) {
```
mboudreau, line 134: this can just be `if(delayMsecs)`  Or a simpler way of doing it is not having the conditional at all `await queues.installation.add(jobData, _.pickBy({delay: delayMsecs}));`

src/models/subscription.ts
bgvozdev: This is our current payload that we store in Redis. We have to have the same for an easy migration, unless I miss something..

bgvozdev: Plz check http://github.com/atlassian/github-for-jira/blob/a807b2195065a48f68a89ec3f37a09ec3bc2fa3e/src/sync/installation.ts#L454-L454

mboudreau: ah right, we really should fix that at some point before someone mistaken it

mboudreau (approved):

thombergs (approved):

Issue #855, opened 2021-10-25
Support Deployments with a "branch & merge" strategy


thombergs:
When using a git flow like branching strategy with `feature` branches and a `main` (or `release`) branch in which they get merged, Deployments won't show up in Jira after the merge.

If I push a commit with issue key `KEY-1` into the `feature` branch, the deployment shows up connected with the `KEY-1` issue in Jira.

When I merge the `feature` branch into `main`, the deployment that is triggered from the `main` branch, doesn't show up in Jira (i.e. it's not being connected with the `KEY-1` issue.

I guess we need to implement some smart logic that crawls all commits that have been made on the `feature` branch, but are not yet in the `main` branch, and check those commits for issue keys and then add those issue keys to the deployment

eialanjones:
+1

jakila:
I came across this issue after reading this line https://github.com/atlassian/github-for-jira/blob/075c4339968fdaafc781ff581acf5a32398f4da7/src/transforms/deployment.ts#L67
If i am reading this correctly, then only the last commit is considered when processing deployments, isn't it?

I want to highlight another use case were this is quite unfortunate:
We have some environments that can be updated only during certain maintenance windows (i.e. once a week). Usually there are five to ten commits on our "main" branch between each deployment to these environments.  Each commit has an jira issue key.
However, as only the last commit is parsed for jira issue keys, at most one jira issue gets the right deployment state. And all others issues show as "not deployed".

rachellerathbone:
Hey @jakila and @eialanjones. Good news, we've just rolled an update for this out to all customers. Please let us know if this now works as expected.

jakila:
That's great news! 
I just did a quick check and it seems to be working for us. Will take a few more days (and a few more deployments) to verify it is working as expected. 
But already now the deployment view in our Jira looks more complete than ever before 🚀 

A big "Thank You" to atrigueiro, rachellerathbone and everyone else working on the topic!

rachellerathbone:
That's great to hear @jakila 🎉

palexcast:
Hi, we are experiencing a similar issue.
We have a single branch (main) which we deploy to, and we mark each commit with the task id in the name.
On every commit to the main branch we deploy to our development environment. The workflow will then wait for a review /approval before deploying to staging, and then the same for production.

All our tasks are displayed as deployed to development as expected, but only the last one to go staging/prod will be updated as released to the correct environment.

pdyck:
[...]

PR #936, opened 2021-11-24
ARC-702: Support branch and merge workflow for deployments

atrigueiro:
Issue explained here: https://github.com/atlassian/github-for-jira/issues/855
This PR solves the workflow for deployments. Similar work still needs to be done for Builds.
More details about the solution are here: https://hello.atlassian.net/wiki/spaces/PF/pages/1411118410/Investigating+how+to+support+branch+merge+workflows


src/transforms/deployment.ts
```diff
@@ -1,7 +1,79 @@
...
+async function getLastSuccessfulDeployCommitSha(
+	owner: string,
+	repoName: string,
+	github: GitHubAPI,
+	deployments: Octokit.ReposListDeploymentsResponseItem[]
+): Promise<string> {
+
+	for (const deployment of deployments) {
+		// Get each deployment status for this environment so we can have their statuses' ids
+		const listDeploymentStatusResponse: Octokit.Response<Octokit.ReposListDeploymentStatusesResponse> = await github.repos.listDeploymentStatuses({
```
mboudreau, line 19: This will only get the last 30 statuses, you can request up to 100 if you want (same rate limit)

src/transforms/deployment.ts
```diff
@@ -1,7 +1,79 @@
...
+			owner: owner,
+			repo: repoName,
+			deployment_id: deployment.id
+		});
+
+		// Filter only the successful ones
+		const lastSuccessful: Octokit.ReposListDeploymentStatusesResponseItem[] = listDeploymentStatusResponse.data.filter(deployment => deployment.state === "success");
+		if (lastSuccessful) {
```
mboudreau, line 27: this will always return true unless the `listDeploymentStatusResponse.data` is undefined which will then break your code.  Are you sure you didn't mean to do `find` instead of `filter`?

src/transforms/deployment.ts
```diff
@@ -1,7 +1,79 @@
...
+			return deployment.sha;
+		}
+	}
+
+	// If there's no successful deployment on the list of deployments that GitHub returned us (max. 250) then we'll return the last one from the array, even if it's a failed one.
+	return deployments[deployments.length - 1].sha;
```
mboudreau, line 33: why the last one?  Wouldn't the first one be the 'latest'?  If not, I'd recommend you set the order when you call the API since you're for loop is wrong in that case.

src/transforms/deployment.ts
```diff
@@ -1,7 +1,79 @@
...
+
+	// Grab all deployments for this repo
+	const deployments: Octokit.Response<Octokit.ReposListDeploymentsResponse> = await github.repos.listDeployments({
+		owner: owner,
+		repo: repoName,
+	})
+
+	// Filter per current environment and exclude itself
+	const filteredDeployments = deployments.data
+		.filter(deployment => deployment.environment === currentDeployEnv)
```
mboudreau, line 53: You can filter environments and other things from the API itself: https://docs.github.com/en/rest/reference/repos#list-deployments

src/transforms/deployment.ts
```diff
@@ -1,7 +1,79 @@
...
+async function getCommitMessagesSinceLastSuccessfulDeployment(
+	owner: string,
+	repoName: string,
+	currentDeploySha: string,
+	currentDeployId: number,
+	currentDeployEnv: string,
+	github: GitHubAPI
+): Promise<string> {
+
+	// Grab all deployments for this repo
```
mboudreau, line 45: this won't grab all deployments, only the last 30.

src/transforms/deployment.ts
```diff
@@ -1,7 +1,79 @@
...
+	})
+
+	// Filter per current environment and exclude itself
+	const filteredDeployments = deployments.data
+		.filter(deployment => deployment.environment === currentDeployEnv)
+		.filter(deployment => deployment.id !== currentDeployId);
+
+	// If this is the very first successful deployment ever, return nothing because we won't have any commit sha to compare with the current one.
+	if (!filteredDeployments.length) {
+		return "";
```
mboudreau, line 58: you can return `undefined` here.

src/transforms/deployment.ts
```diff
@@ -1,7 +1,79 @@
...
+import {booleanFlag, BooleanFlags} from "../config/feature-flags";
+
+async function getLastSuccessfulDeployCommitSha(
+	owner: string,
+	repoName: string,
+	github: GitHubAPI,
+	deployments: Octokit.ReposListDeploymentsResponseItem[]
+): Promise<string> {
+
+	for (const deployment of deployments) {
```
mboudreau, line 17: you might want to have a try/catch in this loop in case something happens when calling the github API

src/transforms/deployment.ts
```diff
@@ -1,7 +1,79 @@
...
+		base: lastSuccessfullyDeployedCommit,
+		head: currentDeploySha
+	})
+
+	let allCommitMessages = "";
+	for (const commit of commitsDiff.data.commits) {
+		allCommitMessages = allCommitMessages + " " + commit.commit.message
+	}
+
+	return allCommitMessages;
```
mboudreau, line 75: here's a little something handy for ya ;)

`return commitsDiff.data?.commits?.map(c => c.message).join(" ");`

src/transforms/deployment.ts
thombergs: Does `if(lastSuccessful)` resolve to `true` if the array is empty?

Yeah, I guess using `find` instead of `filter` is more straightforward.

src/transforms/deployment.ts
atrigueiro: Ha that looks way better! Thanks :)

src/transforms/deployment.ts
atrigueiro: Yes, the first one is the latest but if we return the latest when we do the commits diff, later on, we won't have any commits in between deployments - because they'll be the same. Grabbing the last of the array means grabbing the oldest one. AFAIK, the API returns deployments from newest to oldest.

src/transforms/deployment.ts
atrigueiro: Changed to use `find` instead :)

src/interfaces/jira.ts
```diff
@@ -83,7 +83,7 @@ export interface JiraDeployment {
 	displayName: string;
 	url: string;
 	description: string;
-	lastUpdated: number;
+	lastUpdated: string;
```
mboudreau, line 86: I think this should actually be a `Date` since Date will stringify to UTC string - `JSON.stringify({date:new Date()})` becomes `'{"date":"2021-12-02T03:20:46.554Z"}'`

src/config/feature-flags.ts
```diff
@@ -26,6 +26,8 @@ export enum BooleanFlags {
 	PRIORITIZE_PUSHES = "prioritize-pushes",
 	USE_NEW_GITHUB_CLIENT__FOR_PR = "git-hub-client-for-pullrequests",
 	NEW_REPO_SYNC_STATE = "new-repo-sync-state",
+	SUPPORT_BRANCH_AND_MERGE_WORKFLOWS_FOR_DEPLOYMENTS = "support-branch-and-merge-workflows-for-deployments",
+	INCLUDE_JIRA_HOSTNAME_TO_DEDUP_KEY = "include-jira-hostname-to-dedup-key",
```
thombergs, line 30: I think this feature flag has been removed before (probably came in due to a merge conflict).

src/transforms/deployment.ts
```diff
@@ -1,7 +1,84 @@
...
+	github: GitHubAPI,
+	logger?: LoggerWithTarget
+): Promise<string | undefined> {
+
+	// Grab the last 100 deployments for this repo
+	const deployments: Octokit.Response<Octokit.ReposListDeploymentsResponse> = await github.repos.listDeployments({
+		owner: owner,
+		repo: repoName,
+		environment: currentDeployEnv,
+		per_page: 100 // Default is 30, max we can request is 100
```
thombergs, line 59: This should be lower (say, 10), because we're making a call to get deployment status _for each of those deployments_. We don't want to make 100 calls (that count into the rate limit) for each deployment status event that we receive.

src/transforms/deployment.ts
```diff
@@ -1,7 +1,84 @@
...
+import {Octokit} from "@octokit/rest";
+import {booleanFlag, BooleanFlags} from "../config/feature-flags";
+
+// https://docs.github.com/en/rest/reference/repos#list-deployments
+async function getLastSuccessfulDeployCommitSha(
+	owner: string,
+	repoName: string,
+	github: GitHubAPI,
+	deployments: Octokit.ReposListDeploymentsResponseItem[],
+	logger
```
thombergs, line 16: Why isn't the logger typed?

src/transforms/workflow.ts
```diff
@@ -32,6 +32,10 @@ function mapStatus(status: string, conclusion?: string): string {
 			return "unknown";
 	}
 }
+//
+// function findJiraIssueKeysFromPR() {
+//
+// }
```
thombergs, line 38: This can be trashed, I guess :)

src/transforms/workflow.ts
```diff
@@ -50,6 +54,16 @@ function mapPullRequests(pull_requests: GitHubPullRequest[]): JiraPullRequest[]
...
+
+	/*
+	* 1. Grab the pr
+	* 2. Get the commits from that PR max 250
+	* 3. Get the message for every commit
+	* 4. Check if there's a Jira issue key on that message
+	* 5. Add to issueKeys const
+	*
+	* */
+
```
thombergs, line 66: This comment doesn't make sense in this file, does it?

thombergs (commented):
Some minor changes (see comments)

src/transforms/deployment.ts
atrigueiro: Because I forgot :coneofshame:

test/safe/deployment.test.ts
```diff
@@ -84,6 +85,8 @@ describe("Deployment Webhook", () => {
 			Date.now = jest.fn(() => 12345678);
 
 			await expect(app.receive(fixture)).toResolve();
+			// Clean up all nock mocks
+			nock.cleanAll();
```
mboudreau, line 89: You shouldn't need to do this.  It can give false positives.

test/unit/transforms/deployments.test.ts
```diff
@@ -67,3 +76,156 @@ describe("deployment environment mapping", () => {
...
+		when(booleanFlag).calledWith(
+			BooleanFlags.SUPPORT_BRANCH_AND_MERGE_WORKFLOWS_FOR_DEPLOYMENTS,
+			expect.anything(),
+			expect.anything()
+		).mockResolvedValue(true);
+
+		const jiraPayload = await transformDeployment(GitHubAPI(), deployment_status.payload, "testing.atlassian.net", getLogger("deploymentLogger"))
+
+		// Clean up all nock mocks
+		nock.cleanAll();
```
mboudreau, line 202: See above.

test/safe/deployment.test.ts
mboudreau: also, we clean all the mocks in a global afterEach

test/unit/transforms/deployments.test.ts
```diff
@@ -67,3 +76,156 @@ describe("deployment environment mapping", () => {
...
+					url: "test-repo-url/commit/885bee1-commit-id-1c458/checks",
+				},
+				environment: {
+					id: "Production",
+					displayName: "Production",
+					type: "production",
+				},
+			}]
+		})
+		expect(githubNock.pendingMocks()).toEqual([]);
```
mboudreau, line 227: This is also done in global afterEach

src/transforms/deployment.ts
```diff
@@ -1,7 +1,84 @@
...
+		.filter(deployment => deployment.id !== currentDeployId);
+
+	// If this is the very first successful deployment ever, return nothing because we won't have any commit sha to compare with the current one.
+	if (!filteredDeployments.length) {
+		return undefined;
+	}
+
+	const lastSuccessfullyDeployedCommit = await getLastSuccessfulDeployCommitSha(owner, repoName, github, filteredDeployments, logger);
+
+	const commitsDiff = await github.repos.compareCommits({
```
rachellerathbone, line 73: Could we extract this out into a function so it could be shared between files? I need to use the exact same function in the solution for builds. No point having the same block on code in both builds and deployments. We can just share 😄

rachellerathbone: Alternatively, I can put it in a shared location in my PR and just update deployments after you get this bad boy merged.

atrigueiro: The latter sounds like less trouble aka avoiding conflicts :)

mboudreau (dismissed):

src/transforms/deployment.ts
rachellerathbone: The PR for builds has been merged. We should update this now to use `compareCommitsBetweenBaseAndHeadBranches` located in `src/transforms/util/githubApiRequests.ts`

rachellerathbone (changes requested):

thombergs (dismissed):

src/transforms/deployment.ts
rachellerathbone: Think you missed this. We should keep the code dry and reuse where we can. Same logic for this exists in another location. Just need to import it and pass in these same args.

rachellerathbone (changes requested):

src/transforms/deployment.ts
atrigueiro: Just refactored it and solved conflicts. Next time let's merge older PRs first to avoid conflicts :)

rachellerathbone (approved):

rachellerathbone (approved):

thombergs (approved):

PR #938, opened 2021-11-28
Edge case tests

tarakpurohit:


test/jira/multiple-jira.test.ts
```diff
@@ -0,0 +1,496 @@
...
+			jiraHost: jira2Host,
+			jiraClientKey: clientKey
+		});
+	});
+
+	afterEach(async () => {
+		await Subscription.destroy({ truncate: true });
+		await Subscription.destroy({ truncate: true });
+		await Installation.destroy({ truncate: true });
+		await Installation.destroy({ truncate: true });
```
mboudreau, line 39: You only need 1 of each.

test/setup/setup.ts
```diff
@@ -19,13 +19,17 @@ function resetEnvVars() {
 
 declare global {
 	let jiraHost: string;
+	let jira2Host: string;
```
mboudreau, line 22: Please don't put this here as it's only specific to one test suite.  Please move it to the test suite file

test/setup/setup.ts
```diff
@@ -34,7 +38,9 @@ declare global {
 beforeEach(() => {
 	resetEnvVars();
 	global.jiraHost = process.env.ATLASSIAN_URL || "";
+	global.jira2Host = process.env.ALTERNATE_ATLASSIAN_URL || "https://test2-atlassian-instance.net";
```
mboudreau, line 41: Dont' use an env var for that, just set one in the test file.

mboudreau (changes requested):
Great test to have.  A few comments to address.  Also, I'd probably place a lot of the nock JSON into it's own separate JSON file as it's kinda hard to read the tests with so many.

mboudreau (approved):
Much better :+1:

bgvozdev (approved):

PR #939, opened 2021-11-29
ARC-753 blocked third party cookies fix

mboudreau:
This is a solution to work around 3rd party cookies being blocked by many enterprise customers from with in the iframe of the jira app.


src/util/getUrl.ts
```diff
@@ -1,13 +1,11 @@
...
 import url from "url";
+import envVars from "../config/env";
 
 export const getGitHubConfigurationUrl = (
-	githubHost,
-	jwt,
 	jiraHost
 ): string =>
-	`https://${githubHost}/github/configuration?jwt=${jwt}&xdm_e=${jiraHost}`;
+	`https://${jiraHost}/plugins/servlet/ac/com.github.integration.${envVars.INSTANCE_NAME}/github-post-install-page`;
```
thombergs, line 8: I feel that this is the crux, but I don't understand why. Is this where the `jiraHost` was lost? So now, we're redirecting to the Jira page, which "inject" the `jiraHost` into the iFrame automatically?

thombergs:
It would help to have an updated version of the sequence diagram at https://hello.atlassian.net/wiki/spaces/PF/pages/1195931611/GitHub+App+Organisation+Connection+and+Authentication.+Technical+details to better understand the changes.

src/util/getUrl.ts
mboudreau: Not at all.  This is just a function that returns the URL of where to go after we hit `/github/setup` when we install the app from the github side.  I change this because there's no way to get the JWT token from the `/github/setup` page, so instead of redirecting to a url that won't work, I'm going to redirect them to the jiraHost plugin directly since that injects the jwt token into the iframe.

This isn't related to 3rd party cookies, this is just something I noticed along the way that's adjacent.  Plus, we don't store the jwt token in the session anymore, so this was broken.

mboudreau:
@thombergs I'd love to update the sequence diagram, but it was just made in some kind of app and then the image pasted to confluence.  There's no way for me to update it without redoing the whole thing, which might be a good idea...

db/config.json
```diff
@@ -1,7 +1,12 @@
 {
   "development": {
     "dialect": "postgres",
-    "use_env_variable": "DATABASE_URL"
+    "use_env_variable": "DATABASE_URL",
```
mboudreau, line 4: for some reason, I was getting some local connection issues with sequelize and this seems to fix that up.  I think there's code somewhere that's running and never resolving properly?  We are going to need to update sequelize soon.

package.json
```diff
@@ -73,6 +73,7 @@
     "bull": "^3.12.1",
     "bunyan": "^1.8.15",
     "bunyan-format": "^0.2.1",
+    "cookie-parser": "^1.4.6",
```
mboudreau, line 76: Needed to parse/read/set raw cookies.

src/config/statsd.ts
```diff
@@ -55,27 +55,33 @@ export const elapsedTimeMetrics = (
 ) => {
 	const elapsedTimeInMs = hrtimer();
 	const method = req.method;
+	(req.log || logger).debug({
+		method: req.method,
+		path: req.path,
+		body: JSON.stringify(req.body),
+		query: req.query
+	}, `${method} request initialized for path ${req.path}`);
```
mboudreau, line 63: Added this as an easier way to debug what's getting hit.  It's been very useful to finding issues quickly.

src/frontend/app.ts
```diff
@@ -102,7 +103,8 @@ export default (octokitApp: App): Express => {
 			maxAge: 30 * 24 * 60 * 60 * 1000, // 30 days
 			signed: true,
 			sameSite: "none",
-			secure: true
+			secure: true,
+			httpOnly: false
```
mboudreau, line 107: By default, session cookies can only be modified by the server, but we needed the frontend to be able to set the `jiraHost` cookie to be saved in session.

src/frontend/app.ts
```diff
@@ -140,6 +142,35 @@ export default (octokitApp: App): Express => {
 		)
 	);
 
+	app.get(["/session", "/session/*"], (req: Request, res: Response, next:NextFunction) => {
```
mboudreau, line 145: New session redirect page where it saves the `jiraHost` as a cookie from `localStorage` value saved.

src/frontend/app.ts
```diff
@@ -140,6 +142,35 @@ export default (octokitApp: App): Express => {
...
+			title: "Redirecting...",
+			APP_URL: process.env.APP_URL,
+			redirectUrl: new URL(req.params[0], process.env.APP_URL).href,
+			nonce: res.locals.nonce
+		});
+	});
+
+	// Saves the jiraHost cookie to the secure session if available
+	app.use((req: Request, res: Response, next: NextFunction) => {
+		const jiraHost = req.cookies.jiraHost;
```
mboudreau, line 159: Save the jiraHost cookie into session and remove cookie.

src/frontend/app.ts
```diff
@@ -140,6 +142,35 @@ export default (octokitApp: App): Express => {
...
+		if (jiraHost) {
+			// Save jirahost to secure session
+			req.session.jiraHost = jiraHost;
+			// delete jirahost from cookies.
+			res.clearCookie("jiraHost");
+		}
+
+		// Only save xdm_e query when on the post install url (iframe url)
+		if (req.path == postInstallUrl && req.method == "GET" && req.query.xdm_e) {
+			req.session.jiraHost = req.query.xdm_e as string;
```
mboudreau, line 169: Only save the `xdm_e` query as `jiraHost` for the post install url

src/frontend/delete-jira-configuration.ts
```diff
@@ -6,26 +6,28 @@ import { Request, Response } from "express";
  * Handle the when a user deletes an entry in the UI
  */
 export default async (req: Request, res: Response): Promise<void> => {
-	const jiraHost = req.session.jiraHost;
+	const jiraHost = req.body.jiraHost;
```
mboudreau, line 9: Any frontend API calls needs to pass in `jiraHost` as part of body since it wouldn't be set in the session yet.

src/frontend/verify-jira-jwt-middleware.ts
```diff
@@ -7,10 +7,11 @@ const verifyJiraJwtMiddleware = (tokenType: TokenType) => async (
 	res: Response,
 	next: NextFunction
 ): Promise<void> => {
-	const jiraHost = (req.query.xdm_e as string) || req.body?.jiraHost;
+	const jiraHost = req.body?.jiraHost || req.query.xdm_e as string;
 
 	if(!jiraHost) {
 		sendError(res, 401, "Unauthorised");
+		return;
```
mboudreau, line 14: Apparently, we were continuing on even if `jiraHost` was missing.

src/worker/startup.ts
```diff
@@ -93,6 +94,11 @@ const setDelayOnRateLimiting = (jobHandler) => async (job: Job, logger: LoggerWi
 };
 
 const sendQueueMetrics = async () => {
+	// Only send queue metrics in prod
+	if (!isNodeProd()) {
+		return;
+	}
+
```
mboudreau, line 101: I kept getting "sending queue metrics" logs in my dev, so we now only send them in production mode.

test/unit/models/axios-error-event-decorator.test.ts
```diff
@@ -35,14 +35,13 @@ describe("AxiosErrorDecorator", () => {
...
 
-			expect(decoratedEvent.extra.request).toEqual({
+			expect(decoratedEvent.extra.request).toMatchObject({
 				method: "GET",
 				path: "/foo/bar",
 				host: "www.example.com",
 				headers: {
 					accept: "application/json, text/plain, */*",
 					host: "www.example.com",
-					"user-agent": "axios/0.21.1"
```
mboudreau, line 45: This test would fail if we updated axios.

thombergs:
> @thombergs I'd love to update the sequence diagram, but it was just made in some kind of app and then the image pasted to confluence. There's no way for me to update it without redoing the whole thing, which might be a good idea...

Yeah, we should redo it in plantUML and then add the plantUML code in an expand below the diagram. IntelliJ has a plantUML plugin that works quite well and you can get a sequence diagram up in no time

src/frontend/app.ts
```diff
@@ -140,6 +135,35 @@ export default (octokitApp: App): Express => {
 		)
 	);
 
+	app.get(["/session", "/session/*"], (req: Request, res: Response, next:NextFunction) => {
```
KAbakumovAtl, line 138: Should it be authenticated with the context token?

src/frontend/app.ts
```diff
@@ -140,6 +135,35 @@ export default (octokitApp: App): Express => {
...
+		const jiraHost = req.cookies.jiraHost;
+		if (jiraHost) {
+			// Save jirahost to secure session
+			req.session.jiraHost = jiraHost;
+			// delete jirahost from cookies.
+			res.clearCookie("jiraHost");
+		}
+
+		// Only save xdm_e query when on the post install url (iframe url)
+		if (req.path == postInstallUrl && req.method == "GET" && req.query.xdm_e) {
```
KAbakumovAtl, line 161: /nit we could have it as a middleware function for the postInstall endpoint rather than checking it here.

src/frontend/delete-jira-configuration.ts
KAbakumovAtl: Wouldn't it?

src/frontend/verify-jira-jwt-middleware.ts
```diff
@@ -7,10 +7,11 @@ const verifyJiraJwtMiddleware = (tokenType: TokenType) => async (
 	res: Response,
 	next: NextFunction
 ): Promise<void> => {
-	const jiraHost = (req.query.xdm_e as string) || req.body?.jiraHost;
+	const jiraHost = req.body?.jiraHost || req.session.jiraHost;
```
KAbakumovAtl, line 10: Why can't we use `req.query.xdm_e` here? If it doesn't match JWT token there will be 401. So should be secure to keep it. (Also, wouldn't it break the Jira iframe page?)

src/util/getUrl.ts
KAbakumovAtl: Do we have jiraHost here if we installed the app from GitHub?

mboudreau: We may, but more often than not, no.  Which is why there's logic before it to either automatically redirect if `jiraHost` is known or ask for the jira domain if it's not.

KAbakumovAtl:
That looks great. 

I have a few comments:
* We need to have authentication for the /session endpoint. Adding context JWT will be easy. Otherwise it is possible to trick a user to redirect to arbitrary URL by sending them the /session link.

* We should update the diagram updated on the page with the sequence diagrams ("GitHub App Organisation Connection and Authentication. Technical details.") 

I am still trying to understand how it works, so having a quick session would be good as well

src/frontend/app.ts
mboudreau: hm, might be a good idea to prevent any other MitM attacks.

mboudreau: Actually, sorry, we don't need to.  The page only checks for a window variable of `jiraHost` if it's available, sets the cookie and redirects to the correct page.  The variable is injects from the parent window (the iframe in jira) which cannot be done from anywhere else because of cross domain issues which stops javascript injections from another domain.

src/frontend/app.ts
mboudreau: We could, but problem with that is you have your session logic scattered across. I'd rather have it in a single location as it makes it easier to modify later on.

src/frontend/delete-jira-configuration.ts
mboudreau: No.  this is called from AJAX in the iframe, which cannot set the cookie because of 3rd party cookies being blocked.  This is a more explicit way of doing it.

src/frontend/verify-jira-jwt-middleware.ts
mboudreau: We can, but that means scattering the logic of `xdm_e` and `jiraHost` across multiple files.  I find it simpler to have it in a single location.  It doesn't break it because we set it in the session from the middleware in app.ts, where the rest of the logic resides.  

That being said, I've been playing with the idea that we should be using `res.locals.jiraHost` instead so we can truly 'centralize' the logic into a single variable and then we just need a single middleware that can do all the logic.

src/frontend/app.ts
KAbakumovAtl: But it is also about redirecting a user to any page specified in a path parameters. (For example an attacker can pretend that the link is for github.atlassian.com domain, but there is a dodgy redirect link which leads to a scam web-site)

What I mean - if we can protect it we should do it. Even if we don't know how it can be exploited yet, we might find in the future

mboudreau: If the redirect url is the concern, there are ways to prevent redirecting to any URL, but even if it did, there's no security concerns here, it just redirects to a random url

KAbakumovAtl: It called 'Open Redirect' vulnerability https://cwe.mitre.org/data/definitions/601.html

KAbakumovAtl (dismissed):

src/frontend/app.ts
KAbakumovAtl: Discussed in person. Should be secure enough because we prepend our domain name before the redirect.

src/frontend/verify-jira-jwt-middleware.ts
mboudreau: I've improved this by using `res.locals.jiraHost` as the main entrypoint throughout the codebase for jiraHost.  Now we only have a single location that does all the logic for everything, let it be `xdm_e` query, included in body or from session.

KAbakumovAtl (approved):

rachellerathbone (approved):

PR #949, opened 2021-12-03
Issue/HOT-97500 new GitHub client for pushes

bgvozdev:


src/github/client/github-client.ts
```diff
@@ -112,8 +112,9 @@ export default class GitHubClient {
 	/**
 	 * Get a single pull request for the given repository.
 	 */
+	// TODO: add a unit test
 	public async getPullRequest(owner: string, repo: string, pullNumber: string): Promise<AxiosResponse<Octokit.PullsGetResponse>> {
-		return await this.get<Octokit.PullsGetResponse>(`/repos/:owner/:repo/pulls/:pullNumber`, {
+		return await this.get<Octokit.PullsGetResponse>(`/repos/:owner/:repo/pulls/:pullNumber`, {},  {
```
thombergs, line 117: context: we added `{}` as the second parameter here, because we forgot to do that in the new `getCommit()` function that we added and got an error in the test. So we added it in these two functions, as well, but we skipped writing a test for those functions for now (there wasn't a test, yet)

src/transforms/push.ts
```diff
@@ -152,15 +155,18 @@ export const processPush = async (github: GitHubAPI, payload, rootLogger: Logger
...
-						repo,
-						ref: sha.id,
-					});
+					} = useNewGithubClient
+						? await github.getCommit(owner.login, repo, sha.id)
+						: await githubOld.repos.getCommit({
+							owner: owner.login,
+							repo,
+							ref: sha.id,
+						});
```
thombergs, line 169: This is the main change. Using the new GitHub client to get a commit on the push path. This is what caused the queue to clog up last night.

thombergs (approved):

tarakpurohit (approved):

PR #950, opened 2021-12-05
Fix the rate limit error handling

thombergs:


src/github/client/interceptors.ts
```diff
@@ -84,10 +85,10 @@ export const handleFailedRequest = (logger: Logger) =>
 			const errorMessage = `Error executing Axios Request ` + error.message;
 
 			const rateLimitResetHeaderValue:string = response.headers?.["x-ratelimit-reset"];
-			const rateLimitRemainingHeaderValue:string = response.headers?.["x-ratelimit-reset"];
-			if(response.headers?.["x-ratelimit-remaining"] == "0" && rateLimitResetHeaderValue) {
+			const rateLimitRemainingHeaderValue:string = response.headers?.["x-ratelimit-remaining"];
+			if(status === 403 && rateLimitRemainingHeaderValue == "0") {
```
thombergs, line 89: We should rely on GitHub's 403 status to throw a `RateLimitError`. Other requests with ratelimit-remaining==0 may go through fine (it's just the last request that will go through).

Also, there was a copy&paste error with the `x-ratelimit-remaining` header name

src/github/client/interceptors.ts
```diff
@@ -84,10 +85,10 @@ export const handleFailedRequest = (logger: Logger) =>
...
 				logger.warn({ err: error }, "Rate limiting error");
-				const rateLimitReset: number = parseInt(rateLimitResetHeaderValue);
+				const rateLimitReset: number = parseInt(rateLimitResetHeaderValue) || Date.now() / 1000 + ONE_HOUR_IN_SECONDS;
```
thombergs, line 91: Falling back to rate limit reset one hour from now if it hasn't been specified for some reason.

src/github/client/interceptors.ts
```diff
@@ -96,12 +97,6 @@ export const handleFailedRequest = (logger: Logger) =>
 				return Promise.reject(new BlockedIpError(error, status));
 			}
 
-			if (status === 403) {
-				const rateLimitRemaining: number = parseInt(rateLimitRemainingHeaderValue);
-				logger.warn({ err: error }, "Github returned 403 without ratelimit header");
-				return Promise.reject(new RateLimitingError(Date.now() / 1000 + 60 * 60, rateLimitRemaining, error, status));
-			}
-
```
thombergs, line 104: Removed this block, because we don't want to handle all 403s as rate limit events. It may be a permissions issue and this would hide it behind a rate limit error (we had that case in the old code, this is probably where this code here comes from)

src/github/client/interceptors.ts
KAbakumovAtl: Was there a case where we had 403 but no x-ratelimit-remaining header, but it was rate limiting?

src/github/client/interceptors.ts
KAbakumovAtl: Would it be an exception from parseInt function if rateLimitResetHeaderValue is undefined?

tarakpurohit (approved):

KAbakumovAtl (approved):

src/github/client/interceptors.ts
KAbakumovAtl: I would suggest to add some extra logging for that case. I remember in the past we had some sneaky rate limit errors which were masked as 403s but didn't have enough info on them to decide that it is a rate limit.

src/github/client/interceptors.ts
thombergs: no, `parseInt()` returns `NaN` in that case, and that is `falsy`, so we fall back to the value on the right side of the `||`.

PR #951, opened 2021-12-06
ARC-826 updating workflows to support multiple issue keys

rachellerathbone:
Issue explained here: #855
This PR solves the workflow for builds. Related to work done for [deployments](https://github.com/atlassian/github-for-jira/pull/936).

More details about the issue & solutions are here (#1 was the solution that worked for builds): https://hello.atlassian.net/wiki/spaces/PF/pages/1411118410/Investigating+how+to+support+branch+merge+workflows

Created a builds doc so customers are aware of how this functionality works and what is required from them: https://github.com/atlassian/github-for-jira/pull/952


src/transforms/util/githubApiRequests.ts
```diff
@@ -0,0 +1,31 @@
...
+	base: string;
+	head: string;
+}
+
+// Used to compare commits for builds and deployments so we can
+// obtain all issue keys referenced in commit messages.
+export const compareCommitsBetweenBaseAndHeadBranches = async (
+	payload: CompareCommitsPayload,
+	github: GitHubAPI,
+	logger?: LoggerWithTarget
```
bgvozdev, line 16: Would it make sense to make it mandatory? Seems it is passed in all calls... (For tests we could always pass `getLogger('test')`

src/transforms/workflow.ts
```diff
@@ -33,42 +37,91 @@ function mapStatus(status: string, conclusion?: string): string {
...
 }
 
-export default (context: Context): JiraBuildData | undefined => {
-	const { workflow_run, workflow } = context.payload;
-	const issueKeys = issueKeyParser().parse(`${workflow_run.head_branch}\n${workflow_run.head_commit.message}`);
+export default async (
+	githubClient: GitHubAPI,
+	payload: WorkflowPayload,
+	jiraHost: string,
+	logger?: LoggerWithTarget
```
bgvozdev, line 59: Would it make sense to make it mandatory? Seems it is passed in all calls... (For tests we could always pass getLogger('test')

src/transforms/workflow.ts
```diff
@@ -33,42 +37,91 @@ function mapStatus(status: string, conclusion?: string): string {
...
+		repository,
+		run_number,
+		status,
+		updated_at,
+	} = workflow_run;
+
+	let issueKeys;
+
+	const supportBranchAndMergeWorkflowForBuildsFlagIsOn = await booleanFlag(BooleanFlags.SUPPORT_BRANCH_AND_MERGE_WORKFLOWS_FOR_BUILDS, false, jiraHost);
+	const workflowHasPullRequest = pull_requests.length > 0;
```
bgvozdev, line 79: Should it be under feature flag, too? (e.g. undefined `pull_requests` might blow the thing up)

src/transforms/workflow.ts
```diff
@@ -33,42 +37,91 @@ function mapStatus(status: string, conclusion?: string): string {
...
+				pipelineId: workflow.id,
+				buildNumber: run_number,
+				updateSequenceNumber: Date.now(),
+				displayName: name,
+				url: html_url,
+				state: mapStatus(status, conclusion),
+				lastUpdated: updated_at,
+				issueKeys,
+				references: workflowHasPullRequest
+					? mapPullRequests(pull_requests).slice(0, 5)
```
bgvozdev, line 122: What's 5? Would it make sense to make it into a constant with a descriptive name?

bgvozdev (dismissed):

src/transforms/workflow.ts
rachellerathbone: It shouldn't. `pull_requests` is always in the workflow run object. It's either an empty array if no pull request has been opened, or an array that contains 2 objects (head and base branches in the pr) if one has.

src/transforms/workflow.ts
rachellerathbone: To be honest, I have no idea. I think this was added when we introduced builds and deployments in the app. I just changed `workflow_run.pull_requests` to pull_requests`

src/transforms/workflow.ts
rachellerathbone: And I'm already doing that 😄. Will get rid of the ?

src/transforms/workflow.ts
rachellerathbone: Just realised there is already a comment there that says `// Optional information that links PRs. Min items: 1, Max items: 5`

src/transforms/workflow.ts
rachellerathbone: Just realised there is already a comment there that says `// Optional information that links PRs. Min items: 1, Max items: 5`

bgvozdev: I'd suggest to make it into a constant e.g. `MAX_PULLREQUEST_REFERENCES` and then get rid of it

src/transforms/workflow.ts
bgvozdev: Yeah, I understand it shouldn't, but IMHO it is best to keep all the new code under a feature flag, especially if this is easy to achieve...

bgvozdev (approved):

src/transforms/workflow.ts
rachellerathbone: I agree. However, in this case, pull_requests.length was already being used for the `references` in the payload. I just extracted it to a variable because it's now being used in 2 places in the code. If I only enable it if the FF is enabled, it means it will not be set to pass to the payload.

bgvozdev (approved):

src/transforms/util/githubApiRequests.ts
```diff
@@ -0,0 +1,31 @@
...
+// obtain all issue keys referenced in commit messages.
+export const compareCommitsBetweenBaseAndHeadBranches = async (
+	payload: CompareCommitsPayload,
+	github: GitHubAPI,
+	logger: LoggerWithTarget
+): Promise<string | undefined | void> => {
+	try {
+		const commitsDiff = await github.repos.compareCommits(payload);
+		const allCommitMessagesFromPullRequest = commitsDiff.data?.commits
+			?.map((c) => c.commit.message)
```
yvonnevictoria, line 21: nit: move question mark to line above

test/unit/transforms/util/githubApiRequests.test.ts
```diff
@@ -0,0 +1,88 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+import GitHubAPI from "../../../../src/config/github-api";
+import { compareCommitsBetweenBaseAndHeadBranches } from "../../../../src/transforms/util/githubApiRequests";
+import { getLogger } from "../../../../src/config/logger";
+
```
yvonnevictoria, line 5: nit: the indents in this and the json files are different to the others (2 spaces vs 4 space tab)

src/transforms/util/githubApiRequests.ts
rachellerathbone: It autoformats when I save.

test/unit/transforms/util/githubApiRequests.test.ts
```diff
@@ -0,0 +1,88 @@
...
+				base: pull_requests[0].base.ref,
+				head: pull_requests[0].head.ref,
+			};
+
+			const pullRequestCommits = Object.assign(
+				{},
+				require("../../../fixtures/api/pull-request-multiple-commits-diff.json")
+			);
+
+			const data = pullRequestCommits.data;
```
yvonnevictoria, line 29: nit: consistency between destructuring above/in other files and here
`const { data } = pullRequestCommits;`

test/unit/transforms/util/githubApiRequests.test.ts
rachellerathbone: Again, auto formatting on save.

yvonnevictoria (approved):

bgvozdev (approved):

PR #958, opened 2021-12-07
ARC-839 remove error pg FF and unused files for UI

rachellerathbone:
Last of the FF removals for the UI changes.


yvonnevictoria (approved):

views/maintenance.hbs
```diff
@@ -30,7 +30,7 @@
 </head>
 <body class="aui-page-hybrid">
 <div class="ac-content">
-  <div class="page-content">
+  <div class="pageContent">
```
mboudreau, line 33: Why this change?  Also, rule of thumb for css classes is to use kebab case, not camel case.

views/maintenance.hbs
```diff
@@ -8,7 +8,7 @@
   <title>{{title}}</title>
   <link rel="stylesheet" href="/public/css-reset/bundle.css" media="all" />
   <link rel="stylesheet" href="/public/atlassian-ui-kit/bundle.css" media="all" />
-  <link rel="stylesheet" href="/public/css/styles.css" media="all" />
+  <link rel="stylesheet" href="/public/css/global.css" media="all" />
```
mboudreau, line 11: was this meant to be?

views/maintenance.hbs
rachellerathbone: As part of the UI updates everything that was shared went into `global.css`. With plain CSS, yes, but when using BEM conventions, you can choose to use camel case. https://en.bem.info/methodology/naming-convention/

views/maintenance.hbs
rachellerathbone: Yep. As part of the UI changes I renamed styles as a file with globally shared styling, it makes more sense for the file to be named 'global'.

views/maintenance.hbs
KAbakumovAtl: Is it how it named in global.css? If not, should we rename it in the css file as well? 🤔

views/maintenance.hbs
mboudreau: agreed :)

views/maintenance.hbs
mboudreau: It's fine for now, I think we'll need to discuss in the future so that we use an agreed upon standard across the board.

mboudreau (approved):

PR #971, opened 2021-12-13
ARC-862 Set autoscaling rules for queue sizes and fix dev environment

KAbakumovAtl:

* Added scaling rules based on the queue sizes:
  * Scale out if push queue has more than 200 messages for 3 minutes
  * Scale out if backfill queue has more than 64 messages for 3 minutes

I've also made some fixes for ddev:
* Added globaledge proxy
* Removed automatic service cleanup


github-for-jira.sd.yml
```diff
@@ -43,7 +43,7 @@ links:
   source:
     url: git@bitbucket.org:atlassian/github-jira-integration.git
 
-cleanup: true
+cleanup: false
```
KAbakumovAtl, line 46: Had to set to `false` so our service won't be automatically undeployed if we didn't deploy it for a few days.

github-for-jira.sd.yml
```diff
@@ -203,8 +203,55 @@ workers:
...
-      #TODO Replace with queue size metrics when we switch to SQS
-      metrics: *CpuMemScalingRules
+      metrics: &CpuMemAndQueuesScalingRules
+        complexScalingRule:
+          EvaluationPeriods: 3
+          Threshold:
+            Lower: 30
+            Upper: 80
+          Metrics:
+            - Expression: "MAX([cpuUsage, memoryUsage, (pushQueueMessages/250)*100, (backfillQueueMessages/80)*100])"
```
KAbakumovAtl, line 213: Here we calculate queues conditions along with memory and CPU. We evaluate every condition as a percent of how much it met. For CPU and memory it is a simply consumption. For queues is how much queue is filled up to the Threshold. 

For push queue we set threshold to 250, so when it is met on 80% (200 messages) we'll scale up.

For the backfill queue we use lower threshold because it tend to have less messages piling up (due to the difference in how this queue is used)

github-for-jira.sd.yml
```diff
@@ -240,7 +286,20 @@ environmentOverrides:
           instance: t2.small
           min: 1
           max: 5
-          metrics: *CpuMemScalingRules
+          metrics: *CpuMemAndQueuesScalingRules
+    resources:
```
KAbakumovAtl, line 290: I had to add this resource because without it we are loosing DNS name which we used for dev environment.

github-for-jira.sd.yml
mboudreau: :scream:

github-for-jira.sd.yml
```diff
@@ -331,7 +390,7 @@ environmentOverrides:
           instance: t3.medium
           min: 1
           max: 5
-          metrics: *CpuMemScalingRules
+          metrics: *CpuMemAndQueuesScalingRules
```
KAbakumovAtl, line 393: Use same scaling rules for worker in staging and prod

github-for-jira.sd.yml
```diff
@@ -240,7 +286,20 @@ environmentOverrides:
...
+        attributes:
+          ip_whitelist:
+            - public
+          routes:
+            - match: # Blackhole IP that keeps spamming us
+                prefix: /
+                external_address_header_match:
+                  - 94.156.174.137
+              route:
+                cluster: blackhole # Deny by sending traffic to blackhole
```
mboudreau, line 302: wasn't the reason to not have this was because we didn't want it to be publicly available and only accessible through the VPN?

mboudreau (changes requested):
Need to revise ddev public accessibility

github-for-jira.sd.yml
KAbakumovAtl: How would we install it into prod Jira then?

KAbakumovAtl: I'll remove it from this PR then, so this conversation won't slow down the scaling rules change.

mboudreau: :+1:

mboudreau (approved):

github-for-jira.sd.yml
thombergs: I'm pretty sure this is not enforced in production environments. Right? Right???? :)

thombergs (approved):

github-for-jira.sd.yml
KAbakumovAtl: No, this is only for ddev, as far as I remember. Definitely not a thing in prod

PR #974, opened 2021-12-15
ARC-896 restored globaledge for ddev, removed 5xx alert from prod

KAbakumovAtl:
Restoring Global Edge for ddev,

Removed 5xx alerts from prod

### DIFF
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -287,6 +287,19 @@ environmentOverrides:
           min: 1
           max: 5
           metrics: *CpuMemAndQueuesScalingRules
+    resources:
+      - type: globaledge
+        name: proxy
+        attributes:
+          ip_whitelist:
+            - public
+          routes:
+            - match: # Blackhole IP that keeps spamming us
+                prefix: /
+                external_address_header_match:
+                  - 94.156.174.137
+              route:
+                cluster: blackhole # Deny by sending traffic to blackhole
     alarms:
       overrides:
         LatencyHigh: null
@@ -297,7 +310,6 @@ environmentOverrides:
         WebServerInstanceVolumeSpaceUtilizationAlarmHigh: null
         WebServerMemoryAlarmHigh: null
         WebServerServiceRespawnAlarm: null
-        HighSeverityAlarmWhenTooManyBackend5xxErrors: null
 
   staging:
     config:
@@ -353,7 +365,6 @@ environmentOverrides:
               ReturnData: false
     alarms:
       overrides:
-        HighSeverityAlarmWhenTooManyBackend5xxErrors: null
         LatencyHigh: null
 #TODO Uncomment when we stop doing hourly deployments
 #        ElbResponseTimeAlert:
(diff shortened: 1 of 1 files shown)


github-for-jira.sd.yml
```diff
@@ -297,7 +310,6 @@ environmentOverrides:
         WebServerInstanceVolumeSpaceUtilizationAlarmHigh: null
         WebServerMemoryAlarmHigh: null
         WebServerServiceRespawnAlarm: null
-        HighSeverityAlarmWhenTooManyBackend5xxErrors: null
```
KAbakumovAtl, line 300: This is already overridden above for all envs globally, so redundant here

github-for-jira.sd.yml
```diff
@@ -442,15 +453,6 @@ environmentOverrides:
 #          Period: 180
 #          ComparisonOperator: GreaterThanThreshold
 #          Statistic: Average
-        HighSeverityAlarmWhenTooManyBackend5xxErrors:
```
KAbakumovAtl, line 445: Removing this per-env override, should remove the alert itself, because it already was removed globally

github-for-jira.sd.yml
```diff
@@ -287,6 +287,19 @@ environmentOverrides:
...
+        attributes:
+          ip_whitelist:
+            - public
+          routes:
+            - match: # Blackhole IP that keeps spamming us
+                prefix: /
+                external_address_header_match:
+                  - 94.156.174.137
+              route:
+                cluster: blackhole # Deny by sending traffic to blackhole
```
mboudreau, line 302: Make this global then just set the overrides specific to the env :)

This way we don't have to duplicate the blackhole.

mboudreau (changes requested):

github-for-jira.sd.yml
KAbakumovAtl: Micros don't recommend us make this config global. However I've added a variable to reduce configs copy-pasting

mboudreau (approved):

thombergs (approved):

github-for-jira.sd.yml
KAbakumovAtl: And it turned out to be wrong!

I've cab see HighSeverityAlarmWhenTooManyBackend5xxErrors appeared after deploying to ddev. Surprisingly I don't see LP 5xx alerts which were also overridden in global config but not here.

So going to add explicit overrides for all of these alerts

KAbakumovAtl: I was wrong again! Just realised there was an alert HighSeverityAlarmWhenTooManyBackend5xxErrors which wasn't overridden in the global config

github-for-jira.sd.yml
```diff
@@ -169,6 +169,7 @@ alarms:
       Period: 120
       ComparisonOperator: GreaterThanThreshold
       Statistic: Average
+    HighSeverityAlarmWhenTooManyBackend5xxErrors: null
```
KAbakumovAtl, line 172: Added missing override.

KAbakumovAtl:
Tested it works in ddev

bgvozdev (approved):

mboudreau (approved):

thombergs (approved):

PR #975, opened 2021-12-15
New github client for branch backfill

tarakpurohit:


src/github/client/github-client.ts
```diff
@@ -98,17 +98,19 @@ export default class GitHubClient {
 		return response;
 	}
 
-	private async graphql<T>(query: string): Promise<AxiosResponse> {
-		return await this.axios.post<T>("https://api.github.com/graphql",
+	public async graphql<T>(query: string, variables: any): Promise<AxiosResponse<T>> {
```
mboudreau, line 101: Variables should be optional or have a default.  Also, don't use 'any', should probably be something like `Record<string, string | number>`

mboudreau: This should stay private as we don't want people to use the graphql querying directly, we want them to use a 'getBranch' function which calls the graphql function.

mboudreau (changes requested):
We need to create a 'getBranch' function that properly returns a typed response.

src/github/client/github-client.ts
tarakpurohit: I think keeping it `private` defeats the purpose of having a graphql option. Keeping it private, everytime we would want add new code to client even if there is a small difference in query.

KAbakumovAtl: I don't agree that we shouldn't expose graphql option. GraphQL queries can be complex and tied directly to a business logic. Some GraphQL queries are not generic hence should be the responsibility of the business logic to control them, not the client.

src/sync/branches.ts
```diff
@@ -3,18 +3,23 @@ import { getBranches as getBranchesQuery } from "./queries";
...
 
 // TODO: better typings
-export default async (github: GitHubAPI, _newGithub: GitHubClient, _jiraHost: string, repository:Repository, cursor?:string | number, perPage?:number) => {
+export default async (github: GitHubAPI, newGithub: GitHubClient, jiraHost: string, repository:Repository, cursor?:string | number, perPage?:number) => {
 	// TODO: fix typings for graphql
-	const { edges } = ((await github.graphql(getBranchesQuery, {
+	
+	const useNewGHClient = await booleanFlag(BooleanFlags.USE_NEW_GITHUB_CLIENT_FOR_BRANCHES, false, jiraHost);
+	const client = useNewGHClient ? newGithub : github;
+	const results = ((await client.graphql(getBranchesQuery, {
```
KAbakumovAtl, line 14: Do we have a unit test for this? Would be good to add tests for the flag

KAbakumovAtl (dismissed):

KAbakumovAtl:
Great Job, I like the code itself. Just one comment about unit tests. 

For approve, Please link FD ticket filled according to instructions

src/github/client/github-client.ts
mboudreau: It's not a client, it's a service.  Axios is the client.  This already has business logic to it.

The whole point of having this service is to centralize common functionality, like retrieving branches, and make sure their type safe.  If you expose the graphql function, you lose both.

KAbakumovAtl: Aren't the name says 'github-client'?

So, for normal APIs what you suggested was working and we were doing it in this way. For GraphQL it is not really the case for the reason explained above. 

Right now the change is simple, elegant and easily feature flaggable. Should we make it complicated and increase the risk of breaking Initial Sync?

thombergs: I agree that the `graphql()` function should stay private and we instead should expose use-case-specific functions like `getBranch()` etc that use it under the hood. Code should be as specific as possible and only as generic as it has to be.

src/sync/branches.ts
thombergs: I would like to see the `getBranchesQuery` encapsulated in the `GitHubClient` (see my other comment) and instead call a function `getBranches()` here. This would also make it possible to add a unit test for the `getBranches()` function in the existing test for the GitHub client.

The `GitHubClient` should abstract away the fact that it's using GraphQL or REST under the hood.

src/github/client/github-client.ts
tarakpurohit: Here's what the trade off will be, the current query is specific for things you need for initial sync. In case, you want to use `getBranches()` somewhere else, and it doesnt have the specific items in the query, the query will need to be updated or we would have to add another function like `getBranchDetailsForSomething()`

I think there is no right/wrong way here its more like what are we ready to change the query and test everything when something changes ?

mboudreau: Yeah, I'd go for multiple `getBranch` functions as needed if they're that different.  The whole point of the service is to abstract out what's happening behind the covers; is it using REST or GraphQL? The code that calls it doesn't know and shouldn't care.  What it should care about is what are the inputs needed for this function and the type that it spits out afterwards.  

It makes the whole thing much more testable, consistent and stable.

tarakpurohit: No worries, will make the change 👍

src/github/client/github-queries.ts
```diff
@@ -6,3 +6,123 @@ query {
 		}
 	}
 }`
+
+export const getPullRequests = `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String) {
```
KAbakumovAtl, line 10: I've moved everything from src/sync/queries.ts here, so our GraphQL queries will be in a single place

KAbakumovAtl:
Closing this PR, because it was replaced with https://github.com/atlassian/github-for-jira/pull/1033

PR #976, opened 2021-12-15
ARC-903: add support to call on-premise GHE API

thombergs:
This PR modifies the GitHub client so that it can make requests to (on-premise) GitHub Enterprise instances as well as to the cloud instance.

Tested successfully on local, will do another test in staging before merging. Backfill and webhooks are being processed nicely, so the client is still working. Given the good test coverage of the GitHub Client, I'm not going to put this behind a feature flag to save a lot of effort.


src/github/client/app-token-holder.ts
```diff
@@ -37,17 +47,21 @@ export default class AppTokenHolder {
 	/**
 	 * Gets the current app token or creates a new one if the old is about to expire.
 	 */
-	public getAppToken(): AuthToken {
-		if (!this.currentToken || this.currentToken.isAboutToExpire()) {
-			this.currentToken = AppTokenHolder.createAppJwt(this.key, this.appId);
+	public getAppToken(appId: InstallationId): AuthToken {
```
thombergs, line 50: This is the main change in `AppTokenHolder`: previously it held a single app token, because we had a single app. With each GHE instance having its own app, it now needs to hold a token for each of those apps. So we now have to pass in an installation ID to get the right app token.

src/github/client/github-client.ts
```diff
@@ -19,17 +20,17 @@ export default class GitHubClient {
...
-	private readonly githubInstallationId: number;
-	private logger: Logger;
+	private readonly githubInstallationId: InstallationId;
+	private readonly logger: Logger;
 
 	constructor(
-		githubInstallationId: number,
+		githubInstallationId: InstallationId,
 		logger: Logger,
-		baseURL = "https://api.github.com"
```
thombergs, line 28: The url is now encapsulated in the `InstallationId` type, which contains all "coordinates" we need to identify an app installation across cloud or on-premise GitHub instances.

src/sync/installation.ts
```diff
@@ -227,7 +228,7 @@ async function doProcessInstallation(app, job, installationId: number, jiraHost:
 		logger
 	);
 
-	const newGithub = new GitHubClient(installationId, logger);
+	const newGithub = new GitHubClient(getCloudInstallationId(installationId), logger);
```
thombergs, line 231: The only changes in the code that's using the GitHub client is that we now pass in an `InstallationId` object instead of a numeric installation ID, so that the client has all the information it needs to call the cloud or on-premise GitHub API.

test/github/client/github-client.test.ts
```diff
@@ -272,6 +278,53 @@ describe("GitHub Client", () => {
...
+					return "cloud private key";
+				case "http://github.mydomain.com":
+					return "GHE private key";
+				default:
+					throw new Error("unknown github instance!");
+			}
+		});
+
+		const client = new GitHubClient(
+			new InstallationId("http://github.mydomain.com", 4711, githubInstallationId),
```
thombergs, line 316: @rachellerathbone this is how we will need to instantiate a `GitHubClient` in the code that's handling a webhook from GHE. All values in the `InstallationId` (URL, appId, installationId) we can hard-code for the spike.

rachellerathbone (dismissed):

src/github/client/installation-id.ts
```diff
@@ -0,0 +1,45 @@
+import envVars from "../../config/env";
+
+/**
+ * An installation ID uniquely identifies an installation of a GitHub app across the (single) cloud instance
+ * and (potentially many) GHE instances.
+ */
+export class InstallationId {
```
bgvozdev, line 7: TBH I find it pretty weird to have a class called "Id" which contains some URL... not saying it collides with the current installationIds we have around. I was sure that this was some wrapper thin type/alias before actually reading this code....

bgvozdev (approved):

src/github/client/installation-id.ts
thombergs: Yeah, I struggled with the name a bit, too. 

We'll need a new domain object `GitHubApp` when introducing GHE, which will contain the URL and app ID. Once we have that, we should probably split this up into two objects: a (numeric) `installationId` and a `GitHubApp` that contains the `url` and the `appId`

from https://hello.atlassian.net/wiki/spaces/PF/pages/1446606916/GHE+Task+Breakdown: 

image

PR #977, opened 2021-12-15
Arc 818 add new queue and fix error handling

KAbakumovAtl:
This is a first part of the discovery queue works. 

Decided to split it to miltiple PR so they'll be smaller and easier to review.

Here I do the following:

* Adding a new queue as a Micros resource and adding infrastructure for the listener
* Fixed the way we handle errors for backfill queue. Error handler had a code to send metrics about "push" queue failure. I've extracted it to an optional wrapper


test/unit/sqs/error-handlers.test.ts
```diff
@@ -1,15 +1,14 @@
...
 import {getLogger} from "../../../src/config/logger";
 import {JiraClientError} from "../../../src/jira/client/axios";
 import {RateLimitingError as OldRateLimitingError} from "../../../src/config/enhance-octokit";
 import {Octokit} from "probot";
 import {RateLimitingError} from "../../../src/github/client/errors";
 import {AxiosError} from "axios";
 
-
-describe("jiraOktokitErrorHandler", () => {
+describe("error-handlers", () => {
```
KAbakumovAtl, line 11: Sorry for too many line changes in these file. I had to add a top-level describe, because not we test 2 things from that file, each of them having its own "describe"

src/sqs/index.ts
```diff
@@ -101,6 +101,11 @@ export type ErrorHandlingResult = {
 	 */
 	retryable: boolean;
 
+	/**
+	 * Indicates if it was a failure
+	 */
+	isError: boolean;
```
bgvozdev, line 107: I find this one very confusing. What does "isError" means... This is a ErrorHandlingResult, means that there was a error in ErrorHandling, right? So I'd expect it to be true in some "catch" clause and false everywhere else... but then I see it is populated as true in quite some places... Could you please clarify?

KAbakumovAtl: There can be 2 different cases:
1) The exception we got due to a faulty situation, like 500 from GitHub or`Cannot read property '<prop-name>' of undefined`
2) The situation where exception was expected and despite of it we can consider that the message is processed (retryable: false). For example: 401 from GitHub

For case 1) we set isError = true, for 2) we set isError = false.

Then we use isError to figure out if we need to send a metric that we have failed to process a webhook.  


So  to remove some  confusion I can rename it to , for example, isMessageFailed and add comment. Good?

KAbakumovAtl: Changed the name and the comment + some logic around the non-failure results processing

src/sqs/index.ts
```diff
@@ -346,13 +352,18 @@ export class SqsQueue<MessagePayload> {
...
 			try {
 				const errorHandlingResult = await this.errorHandler(err, context);
-				if (!errorHandlingResult.retryable) {
-					log.info("Deleting the message because it is not retryable")
+
+				if(errorHandlingResult.isFailure) {
+					log.error({err}, "Error while executing SQS message")
+					statsd.increment(sqsQueueMetrics.failed, this.metricsTags)
+				} else {
+					log.info({err}, "Expected exception while executing SQS message. Not an error, deleting the message.")
```
KAbakumovAtl, line 362: Changed the logic here a bit.

For non-failure cases we will:
* Log the error in Info level instead of Error
* Won't send a failure metric

src/sqs/error-handlers.ts
```diff
@@ -28,22 +28,32 @@ export const jiraOctokitErrorHandler : ErrorHandler<any> = async (error: JiraCli
 		return maybeResult;
 	}
 
-	const errorHandlingResult = handleFailureCase(error, context);
+	return handleFailureCase(error, context);
+}
 
-	if (!errorHandlingResult.retryable || context.lastAttempt ) {
-		context.log.error({error}, "Webhook push processing failed and won't be retried anymore");
-		emitWebhookFailedMetrics("push")
```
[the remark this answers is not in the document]
thombergs: good catch!

src/sqs/index.ts
```diff
@@ -346,13 +352,18 @@ export class SqsQueue<MessagePayload> {
...
+				}
+
+				if (!errorHandlingResult.retryable || !errorHandlingResult.isFailure) {
```
thombergs, line 365: It would be nice if all these conditions would be encapsulated in functions with well-readable names. Something like `isRetryableError()`, `hasReachedMaximumRetries()`, and `shouldBeDelayed()`

thombergs (approved):

src/sqs/index.ts
```diff
@@ -345,28 +359,50 @@ export class SqsQueue<MessagePayload> {
...
+				statsd.increment(sqsQueueMetrics.failed, this.metricsTags)
+			} else {
+				log.info({err}, "Expected exception while executing SQS message. Not an error, deleting the message.")
 			}
+
+			if (isNotAFailure(errorHandlingResult)) {
+				log.info("Deleting the message because the error is not a failure")
+				await this.deleteMessage(message, log)
+			} else if (isNotRetryable(errorHandlingResult)) {
+				log.info("Deleting the message because the error is not retryable")
```
bgvozdev, line 381: Should it be a warning? That's a error scenario, right?...

src/sqs/index.ts
```diff
@@ -345,28 +359,50 @@ export class SqsQueue<MessagePayload> {
...
+				await this.deleteMessage(message, log)
+			} else if (errorHandlingResult.skipDlq && this.isMessageReachedRetryLimit(context)) {
+				log.info("Deleting the message because it has reached the maximum amount of retries")
```
bgvozdev, line 384: Same here, I think this should be a warn, too... this doesn't look like a happy path at all

bgvozdev (approved):

src/sqs/index.ts
KAbakumovAtl: Technically, it is kind of expected error which we treat as a "success", but yea I can make it warn

src/sqs/index.ts
KAbakumovAtl: Sure, I'll change in the following PR

PR #987, opened 2021-12-17
ARC-908 Adding github token check

mboudreau:
We've noticed that there are some errors that are happening because a person's Github token may have expired.  Github doesn't explicitly tell us the expiry time of the token because it can be revoked for a number of reasons (mostly security), so an expiry alone isn't going to cut it.  

Instead, when we need the github token, we will call a github API that doesn't add to the rate limit cost to validate the token if we have one.  If it's valid, we move on, if not, we redirect to the github oauth flow to get a new one :)


src/config/env.ts
```diff
@@ -44,7 +44,8 @@ const envVars: EnvVars = {
 	SENTRY_DSN: process.env.SENTRY_DSN,
 	JIRA_LINK_TRACKING_ID: process.env.JIRA_LINK_TRACKING_ID,
 	PROXY: getProxyFromEnvironment(),
-	GITHUB_REPO_URL: "https://github.com/atlassian/github-for-jira"
+	GITHUB_REPO_URL: "https://github.com/atlassian/github-for-jira",
+	GITHUB_DOMAIN: process.env.GHE_HOST || "github.com"
```
KAbakumovAtl, line 48: How do we switch between GitHub and GHE? are we going to run a separate app?

src/frontend/delete-github-subscription.ts
```diff
@@ -1,16 +1,14 @@
 import { Subscription } from "../models";
 import { Request, Response } from "express";
 
-/**
- * Handle the when a user adds a repo to this installation
- */
 export default async (req: Request, res: Response): Promise<void> => {
-	if (!req.session.githubToken) {
+	const { github, githubToken, jiraHost } = res.locals;
+	if (!githubToken) {
```
KAbakumovAtl, line 6: Can it be that wen the user first opens /jira/configuration page and wants to remove an org, we will return 401 because they don't have GitHub token yet?

src/config/env.ts
mboudreau: Don't know.  This is just what was already in the code but made it global.  I'm guessing at some point we'll have a value in the DB or something?

test/unit/frontend/github-configuration.test.ts
```diff
@@ -183,6 +187,7 @@ describe("Frontend", () => {
 					.set(
 						"Cookie",
 						getCookieHeader({
+							githubTokenExpiry: Date.now() + 60000,
```
KAbakumovAtl, line 190: How do we know the expiration time is 60000 ? 

(+ We need to make a constant for it)

src/frontend/delete-github-subscription.ts
mboudreau: `/github/subscription` is called from github configuration page: https://github.com/atlassian/github-for-jira/blob/main/static/js/github-configuration.js#L20

Since you're on the github configuration page, you will have a github token by then.

test/unit/frontend/github-configuration.test.ts
mboudreau: shoot, forgot to push it, this isn't used anymore.

src/frontend/app.ts
```diff
@@ -198,7 +198,6 @@ export default (octokitApp: App): Express => {
 	app.get(
 		"/github/setup",
 		csrfProtection,
-		oauth.checkGithubAuth,
```
KAbakumovAtl, line 201: I wonder how did it work before? 

Am I correctly understand that we don't check this token anymore to support the case when we install the app from GitHub (without opening Jira) and get redirected here without GitHub Token?

src/config/env.ts
KAbakumovAtl: Looks like it was some experimental code. For GHE it shouldn't be global parameter, because each GHE host can have its own domain name. However I don't mind leaving it as long as we don't use it. (cc @thombergs Do you know anything about these GHE params?)

src/frontend/post-github-setup.ts
```diff
@@ -41,11 +41,7 @@ const validateJiraSite = (
 		.then((response) => {
 			// If Jira site is valid, response returns a state of RUNNING
 			if (response?.data?.state === "RUNNING") {
-				res.redirect(
-					req.session.githubToken
-						? getJiraMarketplaceUrl(jiraSiteUrl)
-						: "/github/login"
-				);
+				res.redirect(getJiraMarketplaceUrl(jiraSiteUrl));
```
KAbakumovAtl, line 44: Am I correctly understand that now this redirect will happen in token validation middleware?

src/jira/client/index.ts
```diff
@@ -43,7 +43,7 @@ async function getJiraClient(
 
 	// TODO: need to create actual class for this
 	const client = {
-		baseURL: instance.defaults.baseURL,
+		baseURL: installation.jiraHost,
```
KAbakumovAtl, line 46: Is it safe to do this change?

src/config/env.ts
mboudreau: yeah, good point.  I can add a comment to it.  It could be that the idea is we have an app instance for each GHE customer?  I dunno...

src/frontend/github-client-middleware.ts
```diff
@@ -6,23 +6,20 @@ import Logger from "bunyan";
...
-		res.locals.github = GithubAPI({
-			auth: req.session.githubToken
-		});
-	} else {
-		res.locals.github = GithubAPI();
-	}
+	const { jiraHost } = res.locals;
+
+	// If githubToken isn't set, this GithubAPI will be unauthed
+	res.locals.github = GithubAPI();
```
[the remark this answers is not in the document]
mboudreau: Oh shoot, good catch, I forgot to include this `get-github-configuration`

src/frontend/app.ts
mboudreau: It worked before because it would go to `/github/setup`, see that it was missing the github token, go through the oauth flow to get it, redirect back to `/github/setup` then go through the process normally.  Nothing was broken per say, but we just don't need the github token at all for the github setup which is used after installing the app on the github side and needing to redirect back to jira.

src/jira/client/index.ts
mboudreau: Yeah, I noticed today in the logs that in some cases `instance.defaults.baseURL` is undefined which isn't good.  `installation.jiraHost` is used a few lines up to set the BaseURL for the axios instance.

src/frontend/post-github-setup.ts
mboudreau: Correct, but not for this route at it's not needed at all for the github setup page/post, hence the removal.

docker-compose.yml
```diff
@@ -50,4 +50,5 @@ services:
       DATABASE_URL: postgres://postgres:postgres@postgres:5432/jira-dev
       SQS_BACKFILL_QUEUE_URL: http://localstack:4566/000000000000/backfill
       SQS_PUSH_QUEUE_URL: http://localstack:4566/000000000000/push
+      SQS_DISCOVERY_QUEUE_URL: http://localstack:4566/000000000000/discovery
```
mboudreau, line 53: Adding this as it broke in my local `docker-compose up`

src/frontend/github-oauth.ts
```diff
@@ -148,14 +151,39 @@ export default (opts: OAuthOptions): GithubOAuth => {
 		checkGithubAuth: async (req: Request, res: Response, next: NextFunction) => {
 			const traceLogsEnabled = await booleanFlag(BooleanFlags.TRACE_LOGGING, false);
 			const tracer = new Tracer(logger.child(opts), "checkGithubAuth", traceLogsEnabled);
+			try {
+				const { githubToken } = req.session;
```
mboudreau, line 155: This is the only place that's acceptable to access the `githubToken` from the session.  Everything else needs to call `res.locals.githubToken`

src/frontend/github-oauth.ts
```diff
@@ -148,14 +151,39 @@ export default (opts: OAuthOptions): GithubOAuth => {
...
+					tracer.trace("github token missing, calling login()");
+					throw "Missing github token";
+				}
+				tracer.trace("found github token in session. validating token with API.");
 
-			if (!req.session.githubToken) {
-				tracer.trace("found github token in session, calling login()");
-				res.locals.redirect = req.originalUrl;
-				return login(req, res);
+				await axios.get(`https://api.${envVars.GITHUB_DOMAIN}`, {
```
mboudreau, line 162: This is the call that will validate the token.  Returns 200 if good, 401 if not.

src/frontend/github-oauth.ts
```diff
@@ -148,14 +151,39 @@ export default (opts: OAuthOptions): GithubOAuth => {
...
+						Authorization: `Bearer ${githubToken}`
+					}
+				});
+
+				req.log.debug(`Github token is valid, continuing...`);
+
+				// Everything's good, set it to res.locals
+				res.locals.githubToken = githubToken;
+				// TODO: Not a great place to put this, but it'll do for now
+				res.locals.github = GithubApi({ auth: githubToken });
```
mboudreau, line 173: If the token is good, we replace the Github API local with one including the githubToken for auth.

src/frontend/github-oauth.ts
```diff
@@ -148,14 +151,39 @@ export default (opts: OAuthOptions): GithubOAuth => {
...
+				return next();
+			} catch (e) {
+				req.log.debug(`Github token is not valid.`);
+				// If its a GET request, we can redirect to login and try again
+				if (req.method == "GET") {
```
mboudreau, line 178: We can only redirect to the github login flow if it's a GET request, for any other requests, we need to fail the request and handle that failure on the front-end side.  This is not implemented here as this was always an issue, but we should have a better way of handling API failures in the UI.

src/frontend/post-github-setup.ts
```diff
@@ -25,53 +26,51 @@ const renderGitHubSetupPageVersion = async (
...
-			// If Jira site is not valid, it returns a 404
-			req.log.error({ error }, "Invalid Jira site entered.");
-			renderGitHubSetupPageVersion(domain, req, res);
+): Promise<void> => {
+
+	// Check to see if we already have an installation of the app
+	if(await Installation.getForHost(jiraHost)) {
+		// If so, redirect to the app itself
+		return res.redirect(getJiraAppUrl(jiraHost));
+	}
```
mboudreau, line 40: Seemed easier to find the install in our DB before doing a status check on jira.

mboudreau: and if already installed, you can go directly to the app, no need to install it.

src/frontend/validations.ts
```diff
@@ -3,10 +3,17 @@
...
-export const validJiraDomains = (jiraDomain: string, jiraTopleveldomain: string): boolean =>
-	!!jiraDomain && !!jiraTopleveldomain &&
-	domainRegexp.test(jiraDomain) &&
-	jiraTopleveldomains.includes(jiraTopleveldomain);
+export const validJiraDomains = (jiraDomain: string, jiraTopleveldomain: string): boolean => {
+	try {
+		return !!jiraDomain && !!jiraTopleveldomain &&
+			domainRegexp.test(jiraDomain) &&
+			jiraTopleveldomains.includes(jiraTopleveldomain) &&
+			!!new URL(jiraDomain);
```
mboudreau, line 11: We needed to make sure it was a properly formatted URL as well, and not just have `atlassian.net` in it.

src/config/env.ts
bgvozdev: Shouldn't it be `GITHUB_HOSTNAME`?  https://superuser.com/a/59094

thombergs (dismissed):
I would like to do an impromptu blitz test for this change before merging. I.e. deploy it to staging and have 2-3 people do some exploratory testing.

src/frontend/get-github-configuration.ts
```diff
@@ -133,8 +133,7 @@ const removeFailedConnectionsFromDb = async (req: Request, installations: any, j
 };
 
 export default async (req: Request, res: Response, next: NextFunction): Promise<void> => {
-	const { githubToken } = req.session;
-	const { jiraHost } = res.locals;
+	const { jiraHost, githubToken } = res.locals;
```
bgvozdev, line 136: Was `githubToken` moved from `req.session` to `res.locals`? Should we have a feature flag here?

src/frontend/delete-github-subscription.ts
```diff
@@ -1,16 +1,14 @@
 import { Subscription } from "../models";
 import { Request, Response } from "express";
 
-/**
- * Handle the when a user adds a repo to this installation
- */
 export default async (req: Request, res: Response): Promise<void> => {
-	if (!req.session.githubToken) {
+	const { github, githubToken, jiraHost } = res.locals;
```
bgvozdev, line 5: I think this should be feature flagged, shouldn't it?

src/frontend/get-github-subscriptions.ts
```diff
@@ -2,11 +2,11 @@ import { Subscription } from "../models";
 import { NextFunction, Request, Response } from "express";
 
 export default async (req: Request, res: Response, next: NextFunction): Promise<void> => {
-	if (!req.session.githubToken) {
+	const { github, client, isAdmin, githubToken } = res.locals;
```
bgvozdev, line 5: I think this should be feature flagged so we could keep reading from session if something is wrong... Am I missing something?

src/frontend/github-client-middleware.ts
```diff
@@ -6,23 +6,20 @@ import Logger from "bunyan";
 import { booleanFlag, BooleanFlags } from "../config/feature-flags";
 
 export default (octokitApp: App): RequestHandler => async (req: Request, res: Response, next: NextFunction): Promise<void> => {
-	if (req.session.githubToken) {
```
bgvozdev, line 9: Feature flag?

src/frontend/github-oauth.ts
```diff
@@ -148,14 +151,39 @@ export default (opts: OAuthOptions): GithubOAuth => {
 		checkGithubAuth: async (req: Request, res: Response, next: NextFunction) => {
 			const traceLogsEnabled = await booleanFlag(BooleanFlags.TRACE_LOGGING, false);
 			const tracer = new Tracer(logger.child(opts), "checkGithubAuth", traceLogsEnabled);
+			try {
+				const { githubToken } = req.session;
+				if (!githubToken) {
+					tracer.trace("github token missing, calling login()");
+					throw "Missing github token";
```
bgvozdev, line 158: Feature flag?

src/frontend/get-github-configuration.ts
mboudreau: Yes.  It's the only way to centralize the check if the github token is valid or not.  We can feature flag it, but it affects many different areas.  I did manually test these in staging as well as had to updated the tests surrounding it to make sure that the res.locals var is set  or else they would fail.

src/frontend/delete-github-subscription.ts
mboudreau: we can, but it's a lot of trouble.  And currently, this is a bug stopping some people from configuring their github orgs

src/frontend/github-client-middleware.ts
```diff
@@ -6,23 +6,20 @@ import Logger from "bunyan";
...
-		res.locals.github = GithubAPI({
-			auth: req.session.githubToken
-		});
-	} else {
-		res.locals.github = GithubAPI();
-	}
+	const { jiraHost } = res.locals;
+
+	// If githubToken isn't set, this GithubAPI will be unauthed
```
bgvozdev, line 11: Hmm.... we are not passing it to GithubAPI below, does it mean it is always unauthed, or it will be set up later somewhere?

src/frontend/github-client-middleware.ts
```diff
@@ -6,23 +6,20 @@ import Logger from "bunyan";
...
+	res.locals.github = GithubAPI();
 
 	res.locals.client = GithubAPI({
 		auth: octokitApp.getSignedJsonWebToken()
 	});
 
-	if (res.locals.jiraHost && await booleanFlag(BooleanFlags.CALL_IS_ADMIN_AS_APP, true, res.locals.jiraHost)){
```
bgvozdev, line 21: Does booleanFlag play nicely with undefined/null jiraHost?

src/frontend/github-oauth.ts
```diff
@@ -148,14 +151,39 @@ export default (opts: OAuthOptions): GithubOAuth => {
 		checkGithubAuth: async (req: Request, res: Response, next: NextFunction) => {
 			const traceLogsEnabled = await booleanFlag(BooleanFlags.TRACE_LOGGING, false);
 			const tracer = new Tracer(logger.child(opts), "checkGithubAuth", traceLogsEnabled);
+			try {
+				const { githubToken } = req.session;
+				if (!githubToken) {
+					tracer.trace("github token missing, calling login()");
```
bgvozdev, line 157: I'd rather have it as "info", might be helpful for debugging

src/frontend/github-oauth.ts
```diff
@@ -148,14 +151,39 @@ export default (opts: OAuthOptions): GithubOAuth => {
...
-				return login(req, res);
+				await axios.get(`https://api.${envVars.GITHUB_DOMAIN}`, {
+					headers: {
+						Authorization: `Bearer ${githubToken}`
+					}
+				});
+
+				req.log.debug(`Github token is valid, continuing...`);
+
+				// Everything's good, set it to res.locals
```
bgvozdev, line 170: nit: The comment is duplicating code and IMHO not needed.

src/frontend/github-oauth.ts
```diff
@@ -148,14 +151,39 @@ export default (opts: OAuthOptions): GithubOAuth => {
...
+				res.locals.githubToken = githubToken;
+				// TODO: Not a great place to put this, but it'll do for now
+				res.locals.github = GithubApi({ auth: githubToken });
+				return next();
+			} catch (e) {
+				req.log.debug(`Github token is not valid.`);
+				// If its a GET request, we can redirect to login and try again
```
bgvozdev, line 177: supernit: its -> it's

src/frontend/github-oauth.ts
```diff
@@ -148,14 +151,39 @@ export default (opts: OAuthOptions): GithubOAuth => {
...
+				if (req.method == "GET") {
+					req.log.debug(`Trying to get new Github token...`);
```
bgvozdev, line 179: also I'd  vote for info level

src/frontend/post-github-setup.ts
bgvozdev: Looks like a corner-case, should we maybe add a log to have visibility how often it happens?

src/frontend/post-github-setup.ts
```diff
@@ -25,53 +26,51 @@ const renderGitHubSetupPageVersion = async (
...
-	// Check that the entered domain is valid by making a request to the status endpoint
-	axios(`${jiraSiteUrl}/status`, {
-		method: "GET",
-		headers: {
-			"content-type": "application/json",
-		},
-	})
-		.then((response) => {
-			// If Jira site is valid, response returns a state of RUNNING
-			if (response?.data?.state === "RUNNING") {
```
bgvozdev, line 43: Wondering why we have removed this check?

src/frontend/validations.ts
```diff
@@ -3,10 +3,17 @@
 const domainRegexp = /^\w(?:[\w-]{0,61}\w)?$/;
 const jiraTopleveldomains = ["atlassian.net", "jira.com"];
 
-export const validJiraDomains = (jiraDomain: string, jiraTopleveldomain: string): boolean =>
-	!!jiraDomain && !!jiraTopleveldomain &&
-	domainRegexp.test(jiraDomain) &&
-	jiraTopleveldomains.includes(jiraTopleveldomain);
+export const validJiraDomains = (jiraDomain: string, jiraTopleveldomain: string): boolean => {
```
bgvozdev, line 6: Is it "domain" or "hostname"?

bgvozdev:
I'm really surprised to see so little changes in tests for such a big PR. Together with the fact that we are not feature-flagging the change, this change looks a bit risky to me...

src/frontend/github-client-middleware.ts
bgvozdev: @mboudreau ?

src/frontend/get-github-subscriptions.ts
bgvozdev: Discussed in slack, we are not feature flagging this change

src/config/env.ts
bgvozdev: @mboudreau ?

docker-compose.yml
KAbakumovAtl: Thanks!

src/frontend/github-oauth.ts
bgvozdev: @mboudreau ?

src/frontend/get-github-setup.ts
```diff
@@ -1,5 +1,7 @@
 import { Request, Response } from "express";
-import { getJiraAppUrl, getJiraMarketplaceUrl } from "../util/get-url";
+import { getJiraAppUrl, getJiraMarketplaceUrl, jiraSiteExists } from "../util/jira-utils";
+import { Installation } from "../models";
+// import { Installation } from "../models";
```
KAbakumovAtl, line 4: /nit remove the commented line

src/frontend/github-oauth.ts
bgvozdev, line 158: Shouldn't it be `throw new Error("Missing GitHub token")`?

src/frontend/github-oauth.ts
bgvozdev: @mboudreau ?

src/frontend/post-github-setup.ts
bgvozdev: @mboudreau ?

static/js/github-setup.js
```diff
@@ -18,6 +18,29 @@ const hasProtocol = (str) =>
...
+				if (body.error) {
+					throw body;
+				}
+
+				// Redirecting the user to the correct jira instance
+				// Should probably redo this with query strings and just redoing the GET request
+				document.location.href = body.redirect;
+			})
+			.fail(function(response) {
+				// Handle failure
```
bgvozdev, line 39: What's the plan here?

test/unit/frontend/github-configuration.test.ts
```diff
@@ -39,163 +30,299 @@ describe("Frontend", () => {
...
+						jiraHost,
+						githubToken: "token"
+					})
+				)
+				.expect(401);
+		});
+	});
+
+	// TODO: try to figure out why this keeps giving a JSON web token decode error
+	describe.skip("#GET", () => {
```
bgvozdev, line 130: Am I understand this correctly that we don't have a test for the happy path?

mboudreau: Never did, no.  The scope of this PR kept getting bigger and all the while is bug is still in production.  We can create another PR later to fix the tests.

static/js/github-setup.js
mboudreau: Dunno.  It was never able to deal with an error in the first place. I created a ticket for assessing the github/setup flow.

src/frontend/github-oauth.ts
mboudreau: Doesn't need to in JS.  In this case, we don't need the stacktrace as it's only used for flow control.

src/frontend/github-client-middleware.ts
mboudreau: `locals.github` is unauthed until we hit the `checkGithubOauth` function which validates the token and sets a new `locals.github` with the token.  If it's not a valid token, gets a new one from github.

This whole way of setting the github client in the locals is not great and needs to be changed, and this was a workaround to centralize the token validation.

src/frontend/post-github-setup.ts
mboudreau: First, because the false conditional did nothing and just returned a blank page to the user, and second, we don't need to as if we call `/status` on site that doesn't exist, it returns a 404.  I couldn't find much documentation on this endpoint, but I'm guessing that it returns a `running: false` only if it's been disabled or in maintenance mode of some kind.  Not sure if/how we want to handle that edge case.

src/config/env.ts
rachellerathbone: @mboudreau that is something that we'll be decided upon during the investigations for GHE. We may find that GHE customers have one GitHub app that is installed on every org and/or we may have GHE customers that ask each org admin to create their own GH app. In the latter scenario, we could have multiple instances per customer. Still need to think through how we want to handle this.

bgvozdev (approved):

thombergs (approved):

PR #991, opened 2021-12-21
ARC-742: adapts tests to allow provenance of PR issue keys

ccrolf:
This is a non-functional change to allow provenance of issue keys in PR association. This means a smaller diff when opening the functional change PR, which follow after.


yvonnevictoria (approved):

thombergs (approved):
I had to look up the word "provenance" :)

PR #992, opened 2021-12-21
ARC-913: add deployment queue (without functionality, yet)

thombergs:
Adds a queue that I want to use for incoming `deployment_status` webhooks. This PR only adds the queue without any functionality, yet.

The listener for the queue is not yet active, so nothing changes code-wise. Only the SQS queue will be created on deployment.

Successfully deployed to dev and nothing broke 😬


github-for-jira.sd.yml
```diff
@@ -108,6 +108,17 @@ resources:
         - PII/IndirectConfidential # name of GitHub org
         - UGC/PrimaryIdentifier    # references to GitHub entities (commits, pull requests, etc.) and Jira issues
 
+  - type: sqs
+    name: deployment
+    attributes:
+      MaxReceiveCount: 5
+      MessageRetentionPeriod: 1209600
+      VisibilityTimeout: 62 #Visibility timeout will be overridden by the SQS Listener when we read message from the queue. We set it here anyway in case if that override fails
```
thombergs, line 116: No idea where the 62 comes from, I copied it. Trusting @KAbakumovAtl here :)

src/sqs/deployment.ts
```diff
@@ -0,0 +1,9 @@
+import { WebhookPayloadDeploymentStatus } from "@octokit/webhooks";
+
+export type DeploymentMessagePayload = {
+	jiraHost: string,
+
+	// The original webhook payload from GitHub. We don't need to worry about the SQS size limit because metrics show
+	// that payload size for deployment_status webhooks maxes out at 13KB.
+	webhookPayload: WebhookPayloadDeploymentStatus
+}
```
thombergs, line 9: The SQS message payload will just contain the plain webhook payload since it's small enough.

bgvozdev (approved):

rachellerathbone (approved):

PR #993, opened 2021-12-21
ARC-742 Associate PRs to issue keys found in the PR body

ccrolf:
This adds identification of issue keys in the PR body. Adding the commit messages may come in a later PR.

The change is feature-flagged so we can merge and roll out after the xmas period.


src/transforms/pull-request.ts
```diff
@@ -48,11 +49,11 @@ function mapReviews(reviews: Octokit.PullsListReviewsResponse = []) {
...
 // TODO: define arguments and return
 export default async (github: GitHubAPI, pullRequest: Octokit.PullsGetResponse, reviews?: Octokit.PullsListReviewsResponse, log?: LoggerWithTarget) => {
-	const { title: prTitle, head } = pullRequest;
+	const { title: prTitle, head, body } = pullRequest;
+
 	// This is the same thing we do in sync, concatenating these values
-	const issueKeys = issueKeyParser().parse(
-		`${prTitle}\n${pullRequest.head.ref}`
-	);
+	const textToSearch = await booleanFlag(BooleanFlags.ASSOCIATE_PR_TO_ISSUES_IN_BODY, true) ? `${prTitle}\n${head.ref}\n${body}}` : `${prTitle}\n${head.ref}`
```
ccrolf, line 55: I changed the old one to use `head.ref` instead, as that was already extracted before

ccrolf:
Please also review the FD: http://go/j/FD-28756 to check that I filled in everything correctly

src/transforms/pull-request.ts
KAbakumovAtl: Do we need to change old code if we are going to roll out the feature flag anyway? Would be safer to keep it unchanged

KAbakumovAtl (approved):

src/transforms/pull-request.ts
rachellerathbone: I agree. Better to leave old code unchanged and create a follow-up ticket to implement this way once FF is completely rolled out.

KAbakumovAtl (approved):

rachellerathbone (approved):

lance0428:
I am not totally sure I have traced this all back to the correct place but I believe this change is causing some issues for our org.  We have recently started noticing that related jira tickets mentioned in PR body are having the PR being associated incorrectly w/ the PR.  This change in behavior is having a fallout effect of then moving the status of the incorrectly associated ticket in Jira to a state it shouldn't.

Example scenario:
- Working PROJ-2 jira ticket (`IN PROGRESS`) that is a follow up to already `DONE` PROJ-1 ticket
- Commit final changes to github for PROJ-2 ticket
- create and submit PR for PROJ-2 and include summary w/ PROJ-2 and add a reference to PROJ-**1** in body that is a link like `[PROJ-1](https://myorg.atlassian.net/browse/PROJ-1)` so it is easily clickable.

expected behavior: PROJ-1 that was already `DONE` remains in `DONE` state.  We would also prefer the PR wasn't associated w/ the Jira ticket PROJ-1 as it isn't the PR for that ticket but that isn't a 'huge' problem for us.

actual behavior: PROJ-1 is auto-moved to `IN REVIEW` even though the PR is actually for PROJ-2

Is there a way to still allow links in PR body so devs can easily click and not trigger this auto-move behavior?

PR #994, opened 2021-12-22
ARC-913: add processing logic for the deployment queue

thombergs:
Adds the actual processing of `deployment_status` events from the `deployment` queue.

Tested successfully locally and in staging.

Feature ticket: https://hello.atlassian.net/browse/FD-28779


src/github/middleware.ts
```diff
@@ -71,7 +71,7 @@ function extractWebhookEventNameFromContext(context: CustomContext<any>): string
 
 // TODO: fix typings
 export default (
-	callback: (context: CustomContext, jiraClient: JiraClient, util: any) => Promise<void>
+	callback: (context: CustomContext, jiraClient: JiraClient, util: any, githubInstallationId: number) => Promise<void>
```
thombergs, line 74: @ccrolf this is where I pass the `installationId` into the callback function that is wrapped by the middleware

src/github/deployment.ts
```diff
@@ -1,30 +1,97 @@
...
+import getJiraClient, { DeploymentsResult } from "../jira/client";
+import { booleanFlag, BooleanFlags } from "../config/feature-flags";
+import sqsQueues from "../sqs/queues";
+import { GitHubAPI } from "probot";
+import { WebhookPayloadDeploymentStatus } from "@octokit/webhooks";
+import { LoggerWithTarget } from "probot/lib/wrap-logger";
 
-export default async (context: CustomContext, jiraClient): Promise<void> => {
-	const jiraPayload = await transformDeployment(context.github, context.payload, jiraClient.baseURL, context.log);
+export default async (context: CustomContext, jiraClient, githubInstallationId: number): Promise<void> => {
```
thombergs, line 11: @ccrolf here I added the installationId as a parameter. The `middleware` will pass that parameter in (see comment below).

docker-compose.yml
```diff
@@ -50,4 +50,6 @@ services:
       DATABASE_URL: postgres://postgres:postgres@postgres:5432/jira-dev
       SQS_BACKFILL_QUEUE_URL: http://localstack:4566/000000000000/backfill
       SQS_PUSH_QUEUE_URL: http://localstack:4566/000000000000/push
+      SQS_DEPLOYMENT_QUEUE_URL: http://localstack:4566/000000000000/deployment
+      SQS_DISCOVERY_QUEUE_URL: http://localstack:4566/000000000000/discovery
```
thombergs, line 54: Had to add the discovery queue here because I had errors when testing locally.

src/github/deployment.ts
```diff
@@ -1,30 +1,97 @@
...
+
+	if (await booleanFlag(BooleanFlags.USE_SQS_FOR_DEPLOYMENT, false, jiraClient.baseURL)) {
+		await sqsQueues.deployment.sendMessage({
+			jiraHost: jiraClient.baseURL,
+			installationId: githubInstallationId,
+			webhookPayload: context.payload,
+			webhookReceived: new Date().getTime(),
+			webhookId: context.id
+		});
+	} else {
```
thombergs, line 21: The code in this `else` block is actually not new, it's 1:1 the code that was there before.

src/github/deployment.ts
```diff
@@ -1,30 +1,97 @@
...
+		webhookReceived && emitWebhookProcessedMetrics(
+			webhookReceived,
+			name,
+			log,
+			result?.status
+		);
+	}
+};
+
+export const processDeployment = async (
```
thombergs, line 52: This function is a copy of the old processing logic (in the `else` block above), polished and tuned a bit so that it can be called from the SQS message handler

src/github/middleware.ts
```diff
@@ -208,7 +208,7 @@ export default (
 			const util = getJiraUtil(jiraClient);
 
 			try {
-				await callback(context, jiraClient, util);
+				await callback(context, jiraClient, util, gitHubInstallationId);
```
thombergs, line 211: I passed the installation ID into the webhook processing function so that I could add it to the SQS message without doing another database call.

src/sqs/deployment.ts
```diff
@@ -1,9 +1,33 @@
...
 export type DeploymentMessagePayload = {
 	jiraHost: string,
+	installationId: number,
+	webhookReceived: number,
+	webhookId: string,
 
 	// The original webhook payload from GitHub. We don't need to worry about the SQS size limit because metrics show
 	// that payload size for deployment_status webhooks maxes out at 13KB.
-	webhookPayload: WebhookPayloadDeploymentStatus
+	webhookPayload: WebhookPayloadDeploymentStatus,
```
thombergs, line 14: In some cases (like pullrequests with a big description), we have to be conscious of message size, because SQS messages have a limit of 256KB. But `deployment_status` events are very small.

src/sqs/deployment.ts
```diff
@@ -1,9 +1,33 @@
...
+}
+
+export const deploymentQueueMessageHandler: MessageHandler<DeploymentMessagePayload> = async (context: Context<DeploymentMessagePayload>) => {
+
+	context.log.info("Handling deployment message from the SQS queue")
+
+	const messagePayload: DeploymentMessagePayload = context.payload;
+
+	const github = await app.auth(messagePayload.installationId);
```
thombergs, line 23: I thought about replacing the old GitHub client with the new GitHub client while at it, but decided to do one step at a time.

rachellerathbone (approved):

docker-compose.yml
KAbakumovAtl: Good catch! Didn't notice that because I never run the app in docker for local testing

KAbakumovAtl:
- [ ] We need to fill "Rolllout Plan", "Support Readiness", "Tidying Up" and "Change Visibility" fields on the FD ticket

src/github/deployment.ts
```diff
@@ -1,30 +1,97 @@
...
 
-export default async (context: CustomContext, jiraClient): Promise<void> => {
-	const jiraPayload = await transformDeployment(context.github, context.payload, jiraClient.baseURL, context.log);
+export default async (context: CustomContext, jiraClient, _util, githubInstallationId: number): Promise<void> => {
+
+	if (await booleanFlag(BooleanFlags.USE_SQS_FOR_DEPLOYMENT, false, jiraClient.baseURL)) {
+		await sqsQueues.deployment.sendMessage({
+			jiraHost: jiraClient.baseURL,
+			installationId: githubInstallationId,
+			webhookPayload: context.payload,
```
KAbakumovAtl, line 17: So, we are confident that these payloads are < 256kb?

src/sqs/queues.ts
```diff
@@ -52,22 +54,22 @@ const sqsQueues = {
 		timeoutSec: 60,
 		maxAttempts: 5
 	},
-	async () => {
-		//TODO Implement
-	},
-	async (): Promise<ErrorHandlingResult> => ({ retryable: true, isFailure: true })
+	deploymentQueueMessageHandler,
+	webhookMetricWrapper(jiraOctokitErrorHandler, "deployment")
```
KAbakumovAtl, line 58: Nice

src/github/deployment.ts
thombergs: yes. The metrics show that a deployment_status events is max 13 KB (there is no unbounded content in it like in PR descriptions or issue descriptions)

KAbakumovAtl:
Do we have any tests for deployments? (Like integration test?)

If yes, we should test them with the feature flag turned ON. If there are none, we should add one.

- [ ] add tests

src/github/deployment.ts
```diff
@@ -11,6 +11,7 @@ import { LoggerWithTarget } from "probot/lib/wrap-logger";
 export default async (context: CustomContext, jiraClient, _util, githubInstallationId: number): Promise<void> => {
 
 	if (await booleanFlag(BooleanFlags.USE_SQS_FOR_DEPLOYMENT, false, jiraClient.baseURL)) {
+		console.log("SCHEDULING MESSAGE");
```
KAbakumovAtl, line 14: Console.log 😮

src/github/deployment.ts
```diff
@@ -88,6 +91,8 @@ export const processDeployment = async (
 		}, "Jira API rejected deployment!");
 	}
 
+	console.log(`WEBHOOK RECEIVED: ${webhookReceivedDate.getTime()}`)
```
KAbakumovAtl, line 94: Another debug log

src/util/webhooks.ts
```diff
@@ -25,7 +25,7 @@ export const emitWebhookProcessedMetrics = (
 		// only send logs if time of webhookReceived occurred before the currentTime
 		// and if webhookReceivedTime is not null/undefined
 		if (webhookReceivedTime <= currentTime) {
-			const timeToProcessWebhookEvent = getCurrentTime() - webhookReceivedTime;
+			const timeToProcessWebhookEvent = currentTime - webhookReceivedTime;
```
KAbakumovAtl, line 28: Good finding

test/safe/deployment.test.ts
```diff
@@ -28,7 +47,95 @@ describe("Deployment Webhook", () => {
 	});
 
 	describe("deployment_status", () => {
-		it("should update the Jira issue with the linked GitHub deployment", async () => {
+
+		it("should queue and process a deployment event", async () => {
```
KAbakumovAtl, line 51: Nice!

src/github/deployment.ts
thombergs: at least that proves that I've run the code  :)

KAbakumovAtl (dismissed):

bgvozdev (approved):

src/github/deployment.ts
```diff
@@ -1,30 +1,99 @@
...
+import getJiraClient, { DeploymentsResult } from "../jira/client";
+import { booleanFlag, BooleanFlags } from "../config/feature-flags";
+import sqsQueues from "../sqs/queues";
+import { GitHubAPI } from "probot";
+import { WebhookPayloadDeploymentStatus } from "@octokit/webhooks";
+import { LoggerWithTarget } from "probot/lib/wrap-logger";
 
-export default async (context: CustomContext, jiraClient): Promise<void> => {
-	const jiraPayload = await transformDeployment(context.github, context.payload, jiraClient.baseURL, context.log);
+export default async (context: CustomContext, jiraClient, _util, githubInstallationId: number): Promise<void> => {
```
mboudreau, line 11: If we're not using util, just remove it?

mboudreau (approved):

src/github/deployment.ts
thombergs: it is part of the "interface" that is used in the other webhook functions, so I couldn't just remove that param here

PR #995, opened 2021-12-22
Cleaning up use-sqs-for-backfill flag

KAbakumovAtl:
Cleaning up the flag.

As part of the cleanup we had to extract `findOrStartSync` function from `Subscription` class to a separate module. The reason for that is a cyclic dependency it was creating which was breaking tests.

It was moved to `sync-utils.ts`, also it makes more sense there, because this function relates more to sync than to subscriptions data model.


src/frontend/post-github-configuration.ts
```diff
@@ -52,12 +53,15 @@ export default async (req: Request, res: Response): Promise<void> => {
...
 		}
 
-		await Subscription.install({
+		const subscription = await Subscription.install({
 			clientKey: getHashedKey(req.body.clientKey),
 			installationId: req.body.installationId,
 			host: res.locals.jiraHost
 		});
 
+		await findOrStartSync(subscription);
```
KAbakumovAtl, line 62: Moved the logic of starting the sync outside the install function. This is to remove the cyclic dependency. Subscription class should be merely for data manipulation, and sync should be started externally.

src/sync/sync-utils.ts
```diff
@@ -0,0 +1,39 @@
...
+		logger.info("Starting Jira sync");
+		await queues.discovery.add({ installationId, jiraHost });
+		return;
+	}
+
+
+	// Otherwise, just add a job to the queue for this installation
+	// This will automatically pick back up from where it left off
+	// if something got stuck
+	await sqsQueues.backfill.sendMessage({installationId, jiraHost}, 0, logger);
```
KAbakumovAtl, line 38: The function itself was moved from Subscription module here. 

Only this part of the function was changed by removing the feature flag.

KAbakumovAtl:
I've separated queues which are used for sqs unit test and backfill tests. It helped to pass the build. However it is still a mystery to me why it was failing (and passing locally!).

It feels like that some sqs tests didn't wait to done() to be called. I'll add some logs to these tests to be prepared to analyse it further

bgvozdev (approved):

KAbakumovAtl:
Just tested in staging, all good

.env.test
```diff
@@ -14,6 +14,8 @@ SQS_DISCOVERY_QUEUE_URL=http://127.0.0.1:4566/000000000000/test-discovery
 SQS_DISCOVERY_QUEUE_REGION=us-west-1
 SQS_DEPLOYMENT_QUEUE_URL=http://127.0.0.1:4566/000000000000/test-deployment
 SQS_DEPLOYMENT_QUEUE_REGION=us-west-1
+SQS_TEST_QUEUE_URL=http://127.0.0.1:4566/000000000000/test-sqs-client
+SQS_TEST_QUEUE_REGION=us-west-1
```
thombergs, line 18: What is this test queue for?

src/sync/installation.ts
```diff
@@ -404,30 +405,27 @@ async function doProcessInstallation(app, job, installationId: number, jiraHost:
...
-	nextTaskDelays: Array<number>,
+	nextTaskDelaysMs: Array<number>,
 	logger: LoggerWithTarget
 ) {
-	if (nextTaskDelays.length > 0) {
-		nextTaskDelays.sort().reverse();
-		if (nextTaskDelays.length > 1) {
+	if (nextTaskDelaysMs.length > 0) {
+		nextTaskDelaysMs.sort().reverse();
+		if (nextTaskDelaysMs.length > 1) {
```
thombergs, line 414: 👏🏼

src/sync/sync-utils.ts
thombergs: Why not call the file `start-sync` instead of `sync-utils`?

.env.test
thombergs: Ah got it. It's used in integration tests.

thombergs (approved):

PR #998, opened 2021-12-23
ARC-818 add ff to switch from redis to sqs

KAbakumovAtl:
* Added Discovery queue processing logic for SQS
* Added a feature flag to choose between Discovery queue and SQS for the queue

my ff will be here go/flag-ticket/FD-28819

- [x] TODO - fill up the flag fields


KAbakumovAtl:
- [ ] TODO Add a test for Discovery queue

We never had this tests, so it is a good opportunity to add one

src/sqs/error-handlers.ts
```diff
@@ -20,7 +20,7 @@ const RATE_LIMITING_DELAY_BUFFER_SEC = 10;
 const EXPONENTIAL_BACKOFF_BASE_SEC = 60;
 const EXPONENTIAL_BACKOFF_MULTIPLIER = 3;
 
-export const jiraOctokitErrorHandler : ErrorHandler<any> = async (error: JiraClientError | Octokit.HookError | OldRateLimitingError | RateLimitingError | Error,
+export const jiraAndGitHubErrorsHandler : ErrorHandler<any> = async (error: JiraClientError | Octokit.HookError | OldRateLimitingError | RateLimitingError | Error,
```
KAbakumovAtl, line 23: I've renamed this one so it will make more sense

src/sqs/index.ts
```diff
@@ -371,17 +371,17 @@ export class SqsQueue<MessagePayload> {
 				log.error({err}, "Error while executing SQS message")
 				statsd.increment(sqsQueueMetrics.failed, this.metricsTags)
 			} else {
-				log.info({err}, "Expected exception while executing SQS message. Not an error, deleting the message.")
+				log.warn({err}, "Expected exception while executing SQS message. Not an error, deleting the message.")
```
KAbakumovAtl, line 374: The debt from the previous PR, cc @bgvozdev

src/sqs/discovery.ts
```diff
@@ -1,4 +1,12 @@
...
+import app from "../worker/app";
+import {discovery} from "../sync/discovery";
+
 export type DiscoveryMessagePayload = {
 	installationId: number,
 	jiraHost: string
 }
+
+export const discoveryQueueMessageHandler : MessageHandler<DiscoveryMessagePayload> = async (context) => {
+	await discovery(app)({data: context.payload}, context.log);
```
KAbakumovAtl, line 11: The only thing this queue is doing is calling the existing handler for Discovery queue.

We are not using any of its middleware, because:
1) We don't use Sentry anyway
2) We handle timeouts in our error handler

thombergs (approved):

src/sync/sync-utils.ts
```diff
@@ -1,15 +1,19 @@
...
 import {queues} from "../worker/queues";
 import sqsQueues from "../sqs/queues";
 import Subscription from "../models/subscription";
+import {LoggerWithTarget} from "probot/lib/wrap-logger";
+import Logger from 'bunyan';
 
 
 export async function findOrStartSync(
 	subscription: Subscription,
+	logger: LoggerWithTarget | Logger,
```
mboudreau, line 13: If you're adding a new optional param, add it to the end.  You don't need to do the `logger = logger || getLogger("sync-utils");` below, you can just do `logger = getLogger("sync-utils")`

KAbakumovAtl: Good catch, it is actually not optional, it is required. So I'll just remove the line

KAbakumovAtl:
Finally added the test

thombergs (approved):

mboudreau (approved):

PR #1001, opened 2021-12-24
ARC-944: add branch queue processor

thombergs:
This adds the processing logic for the `create` webhooks.

It has soaked on staging (with FF toggled on) for a couple hours and the Pollinator is green, which means branches are still working.

Feature ticket: https://hello.atlassian.net/browse/FD-28829


src/github/branch.ts
```diff
@@ -3,38 +3,102 @@ import issueKeyParser from "jira-issue-key-parser";
...
-	if (!jiraPayload) {
+	if (await booleanFlag(BooleanFlags.USE_SQS_FOR_BRANCH, false, jiraClient.baseURL)) {
+		await sqsQueues.branch.sendMessage({
+			jiraHost: jiraClient.baseURL,
+			installationId: githubInstallationId,
+			webhookReceived: new Date().getTime(),
+			webhookId: context.id,
+			webhookPayload
+		})
+	} else {
```
thombergs, line 30: This `else` block is the original processing code - untouched.

src/github/branch.ts
```diff
@@ -3,38 +3,102 @@ import issueKeyParser from "jira-issue-key-parser";
...
+		webhookReceived && emitWebhookProcessedMetrics(
+			webhookReceived,
+			name,
+			log,
+			jiraResponse?.status
+		);
+	}
+};
+
+export const processBranch = async (
```
thombergs, line 58: This function is a copy of the old processing logic (in the `else` block above), polished and modified to abstract away from Probot.

src/sqs/queues.ts
```diff
@@ -56,7 +56,7 @@ const sqsQueues = {
 		maxAttempts: 5
 	},
 	deploymentQueueMessageHandler,
-	webhookMetricWrapper(jiraOctokitErrorHandler, "deployment")
+	webhookMetricWrapper(jiraOctokitErrorHandler, "deployment_status")
```
thombergs, line 59: That was an issue from a previous PR with the deployment queue. It would count the failing events against the wrong metric name 😬

mboudreau (approved):

PR #1002, opened 2021-12-24
ARC-816 Remove all Subscription.repoSyncState references

mboudreau:
And included flags.


src/config/feature-flags.ts
```diff
@@ -18,14 +18,13 @@ export enum BooleanFlags {
 	// When cleaning up the SEND_PUSH_TO_SQS feature flag, please also clean up the PRIORITIZE_PUSHES
 	// feature flag, because it doesn't make sense with SQS any more.
 	USE_NEW_GITHUB_CLIENT__FOR_PR = "git-hub-client-for-pullrequests",
-	NEW_REPO_SYNC_STATE = "new-repo-sync-state",
 	SUPPORT_BRANCH_AND_MERGE_WORKFLOWS_FOR_DEPLOYMENTS = "support-branch-and-merge-workflows-for-deployments",
 	TRACE_LOGGING = "trace-logging",
 	SUPPORT_BRANCH_AND_MERGE_WORKFLOWS_FOR_BUILDS = "support-branch-and-merge-workflows-for-builds",
 	USE_NEW_GITHUB_CLIENT_FOR_PUSH = "use-new-github-client-for-push",
-	REPO_SYNC_STATE_AS_SOURCE = "repo-sync-state-as-source",
```
thombergs, line 26: I presume that both feature flags are rolled out to 100% already?

thombergs (approved):

src/config/feature-flags.ts
```diff
@@ -18,14 +18,13 @@ export enum BooleanFlags {
...
 	USE_SQS_FOR_DEPLOYMENT = "use-sqs-for-deployment",
 	USE_SQS_FOR_BRANCH = "use-sqs-for-branch",
+	USE_NEW_GITHUB_CLIENT_TO_COUNT_REPOS = "use-new-github-client-to-count-repos",
```
KAbakumovAtl, line 27: I wonder why this feature flag is back?

test/unit/frontend/github-configuration.test.ts
```diff
@@ -126,9 +131,14 @@ describe("Github Configuration", () => {
...
 	});
 
-	// TODO: try to figure out why this keeps giving a JSON web token decode error
-	describe.skip("#GET", () => {
+	describe("#GET", () => {
 		it("should return 200 when calling with valid Github Token", async () => {
+
+			// Don't use new github client for now - need better way to test with it
+			when(booleanFlag)
+				.calledWith(BooleanFlags.USE_NEW_GITHUB_CLIENT_TO_COUNT_REPOS, true, jiraHost)
```
KAbakumovAtl, line 139: It looks like this flag was mistakenly added back to the feature-flags.js

This one was cleaned up last week

 - [ ] remove USE_NEW_GITHUB_CLIENT_TO_COUNT_REPOS flag

src/config/feature-flags.ts
mboudreau: correct!

src/config/feature-flags.ts
mboudreau: because a test was still using it and was breaking?

test/unit/frontend/github-configuration.test.ts
mboudreau: cool, bad merge then.  I'll fix.

src/config/feature-flags.ts
KAbakumovAtl: Commented there in the test

src/models/index.ts
```diff
@@ -44,7 +44,6 @@ SubscriptionModel.init({
 	gitHubInstallationId: DataTypes.INTEGER,
 	jiraHost: DataTypes.STRING,
 	selectedRepositories: DataTypes.ARRAY(DataTypes.INTEGER),
-	repoSyncState: DataTypes.JSONB,
```
KAbakumovAtl, line 47: So after we remove this the repoSyncState column won't be used anymore, correct?

test/frontend/get-jira-configuration.test.ts
```diff
@@ -1,74 +1,46 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import { mocked } from "ts-jest/utils";
 import getJiraConfiguration from "../../src/frontend/get-jira-configuration";
-import { Installation, Subscription } from "../../src/models";
-
-jest.mock("../../src/models");
+import { Installation, RepoSyncState, Subscription } from "../../src/models";
+// import InstallationClass from "../../src/models/installation";
```
KAbakumovAtl, line 4: /nit remove commented line?

test/frontend/get-jira-configuration.test.ts
```diff
@@ -1,74 +1,46 @@
...
+import SubscriptionClass from "../../src/models/subscription";
 
 describe("Jira Configuration Suite", () => {
-	let consoleSpy: jest.SpyInstance;
-	let subscriptions;
-	let installation;
+	let subscription: SubscriptionClass;
+	// let installation: InstallationClass;
```
KAbakumovAtl, line 9: /nit remove commented code

test/frontend/util/get-jira-configuration-helpers.test.ts
```diff
@@ -220,7 +201,7 @@ describe("getInstallation", () => {
 			repoUrl: "github.com/atlassian/github-for-jira",
 			branchStatus: "failed",
 			repoUpdatedAt: new Date(0)
-		});
+		});*/
```
KAbakumovAtl, line 204: /nit commented code?

test/safe/branch.test.ts
```diff
@@ -171,6 +173,8 @@ describe("Branch Webhook", () => {
 			const ref = encodeURIComponent("heads/TES-123-test-ref");
 			const sha = "test-branch-ref-sha";
 
+			// mockGitHubAuthRequest();
```
KAbakumovAtl, line 176: We need to mock it in any test which uses GitHub Client. 

Otherwise we rely on the test execution order. Because the reason is that this test is passing without mocking the token endpoint - the other test ("should queue and process a create webhook") gets  executed first and token stays in the Client and this test don't request it because token is already there.

This so far was the source of flakyness

test/safe/branch.test.ts
```diff
@@ -14,6 +14,16 @@ describe("Branch Webhook", () => {
 	let app: Application;
 	const gitHubInstallationId = 1234;
 
+	const mockGitHubAuthRequest = () => {
+		githubNock.post(`/app/installations/${gitHubInstallationId}/access_tokens`)
```
KAbakumovAtl, line 18: Optionally is missing

test/safe/branch.test.ts
```diff
@@ -234,6 +238,11 @@ describe("Branch Webhook", () => {
 			Date.now = jest.fn(() => 12345678);
 
 			await expect(app.receive(fixture)).toResolve();
+
+			await waitUntil(async () => {
+				expect(githubNock).toBeDone();
+				expect(jiraNock).toBeDone();
+			});
```
KAbakumovAtl, line 245: Good finding!

test/unit/configure-robot.test.ts
```diff
@@ -1,52 +0,0 @@
-import supertest from "supertest";
-import { createProbot } from "probot";
-import * as PrivateKey from "probot/lib/private-key";
-import configureAndLoadApp from "../../src/configure-robot";
-import {Writable} from "stream";
-
-describe('configure-robot', () => {
```
KAbakumovAtl, line 7: Why is this deleted?

src/models/index.ts
mboudreau: That's correct :)  I have another ticket to remove it altogether from the DB after this is merged.

test/safe/branch.test.ts
mboudreau: I did that on purpose.  Optionally is frowned upon with nock unless it's outside the scope of your tests.  This particular call is important and we want our test to fail if it's missing from the code.

test/safe/branch.test.ts
mboudreau: From what I tried, the sounds of the flakiness was the missing `waitUntil` for one the tests, which means the next test would run without finishing.  When I added this function in here, the tests failed because the auth request was never called.  I'm guessing this is because the Github client is never actually called in this test.

test/unit/configure-robot.test.ts
mboudreau: It wasn't really testing anything that we aren't already testing in other places.  This was a remnant of the old code and didn't think it was needed anymore.

thombergs (dismissed):

thombergs (approved):

test/safe/branch.test.ts
```diff
@@ -163,8 +161,6 @@ describe("Branch Webhook", () => {
 	describe("Create Branch (with disabled FF - delete this test with FF cleanup)", () => {
 		it("should update Jira issue with link to a branch on GitHub", async () => {
 
-			mockGitHubAuthRequest();
```
KAbakumovAtl, line 169: So, we need to have it here too, otherwise if this test gets executed first it will fail

test/unit/frontend/github-configuration.test.ts
```diff
@@ -133,12 +131,6 @@ describe("Github Configuration", () => {
 
 	describe("#GET", () => {
 		it("should return 200 when calling with valid Github Token", async () => {
-
-			// Don't use new github client for now - need better way to test with it
-			when(booleanFlag)
-				.calledWith(BooleanFlags.USE_NEW_GITHUB_CLIENT_TO_COUNT_REPOS, true, jiraHost)
```
KAbakumovAtl: Nice

KAbakumovAtl (approved):

test/safe/branch.test.ts
```diff
@@ -15,17 +15,15 @@ describe("Branch Webhook", () => {
 	let app: Application;
 	const gitHubInstallationId = 1234;
 
-	const mockGitHubAuthRequest = () => {
-		githubNock.post(`/app/installations/${gitHubInstallationId}/access_tokens`)
-			.optionally()
+	const mockGitHubAuthRequest = () =>
+		githubNock
+			.post(`/app/installations/${gitHubInstallationId}/access_tokens`)
 			.reply(200, {
```
KAbakumovAtl, line 21: We need to have optionally here

test/safe/branch.test.ts
```diff
@@ -160,14 +160,17 @@ describe("Branch Webhook", () => {
...
+				expect(githubNock).toBeDone();
+				expect(jiraNock).toBeDone();
+			});
 		});
 	});
 
 	describe("Create Branch (with disabled FF - delete this test with FF cleanup)", () => {
 		it("should update Jira issue with link to a branch on GitHub", async () => {
 
-			mockGitHubAuthRequest();
```
KAbakumovAtl, line 169: It is flaky because you removed `mockGitHubAuthRequest();` from here

KAbakumovAtl (changes requested):
Sorry i clicked "approve" button too quickly and there is no other way to undo it. We need to sort the test out

test/safe/branch.test.ts
mboudreau: if this is flaky, it means the code is flaky as well, having potentially 2 separate code paths for the same event.  If this is not an optional nock (which it shouldn't be), this test will fail as there's an extra nock that never gets called (the auth request). 

The flakiness can also be attributed to the `waitUntil` if it takes longer than the test timeout.

KAbakumovAtl:
Now deployment test fails :( 

Seems we need to apply the same trick with cache cleanup there

mboudreau:
Added it globally, seems to work now :)

KAbakumovAtl (dismissed):
Well done!

KAbakumovAtl:
Still not sure how it works.

We call `mockGitHubAuthRequest` only one of the branches tests. 

However there are 2 tests which use GitHub Client. 

That could mean that cache reset doesn't really do the job and we still get token for the second test from cache

KAbakumovAtl (approved):

PR #1006, opened 2022-01-03
Fix 2 typos in ip-allowlist.md

nightpool:
These typos led to broken links in the docs, meaning that users had to manually fix
them before being able to navigate to the referenced pages

### DIFF
--- a/docs/ip-allowlist.md
+++ b/docs/ip-allowlist.md
@@ -1,7 +1,6 @@
 # GitHub IP Allow List Configuration
 
-If your organization is using [GitHub's Organization IP Allow List](https://docs.github.
-com/en/organizations/keeping-your-organization-secure/managing-allowed-ip-addresses-for-your-organization), it needs to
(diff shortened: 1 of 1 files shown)


thombergs:
Hey @nightpool thanks a lot for fixing those links!

Since you're obviously working with IP allowlists, here is a recent info that's not in the docs, yet: GitHub has an issue in the evaluation of IP allowlists which means that some requests are blocked by GitHub even though the IP allowlist contains the correct IP addresses. This should be fixed by adding the IP addresses  `13.52.5.96` through `13.52.5.111` to your organization's IP allow list (each separately, not as a CIDR range).

I will add this information with some background to the docs.

thombergs (approved):

bgvozdev (approved):

nightpool:
@thombergs haha, that's good to know! actually, I was just trying to get a sense of the overall feature set that the GitHub for Jira plugin contained, and I was simply curious about the IP allowlists. I probably won't be implementing them, but it's good to know they're a possibility!

PR #1015, opened 2022-01-06
Removed Redis queues code

KAbakumovAtl:
This is the first PR which completely removes Redis queues code.

In the next PR I am going to get rid of `bull` queues dependency. 

It is quite hard to do now, because it will require some minor refactoring of the `src/sync/installation.ts` 

I decide to do it in a separate PR so it will be easier to review


src/api/index.ts
```diff
@@ -246,80 +242,16 @@ router.post(
...
-
-		// removing duplicate jobs
-		await Promise.all(duplicateJobs.map((job: Job) => {
-			logger.info({ job }, "removing duplicate job");
-			job.remove();
-		}));
-
-		res.send(`${duplicateJobs.length} duplicate jobs killed with fire.`);
+	async (_: Request, res: Response): Promise<void> => {
+		res.send(`This endpoint doesn't do anything useful anymore`);
```
bgvozdev, line 246: Maybe we could just kill the whole endpoint?

bgvozdev (approved):

src/api/index.ts
thombergs: yes, kill the whole endpoint! It's not being called anymore, anyways.

src/api/index.ts
```diff
@@ -246,80 +242,16 @@ router.post(
...
 	}
 );
 
 router.post(
 	"/requeue",
 	bodyParser,
-	async (request: Request, res: Response): Promise<void> => {
```
thombergs, line 289: same here, kill the whole endpoint!

src/worker/startup.ts
```diff
@@ -1,146 +1,17 @@
...
-// Start function for Node cluster worker
 export async function start() {
 	if (running) {
 		logger.debug("Worker instance already running, skipping.");
 		return;
 	}
 
 	logger.info("Micros Lifecycle: Starting queue processing");
-	// exposing queue metrics at a regular interval
-	timer = setInterval(sendQueueMetrics, 60000);
```
thombergs, line 130: do we still have charts with these metrics? Or are we emitting the same metrics with the new queues?

thombergs (approved):

PR #1016, opened 2022-01-06
Removed use-new-github-client-for-push feature flag, removed `bull` dependency from push processor

KAbakumovAtl:
Second PR, here I did the following:

* Removed `use-new-github-client-for-push` feature flag
* Removed `bull` library dependency from the `push` queue processing code
* Changed integration test to use SQS handler instead of old redis job processor
* Removed installationID parameter from GitHub Client, which we were passing wrongly

### DIFF
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -91,7 +91,6 @@ export default class GitHubClient {
 		const response = await this.axios.get<T>(url, {
 			...await this.installationAuthenticationHeaders(),
 			params: {
-				installationId: this.githubInstallationId,
 				...params,
 			},
 			urlParams,
@@ -106,9 +105,6 @@ export default class GitHubClient {
 			},
 			{
 				...await this.installationAuthenticationHeaders(),
-				params: {
-					installationId: this.githubInstallationId,
-				}
 			});
 	}
 
--- a/src/sqs/push.ts
+++ b/src/sqs/push.ts
@@ -1,7 +1,5 @@
 import { Context, MessageHandler } from "./index"
-import enhanceOctokit from "../config/enhance-octokit";
 import { processPush } from "../transforms/push";
-import app from "../worker/app";
 import { wrapLogger } from "probot/lib/wrap-logger";
 import GitHubClient from "../github/client/github-client";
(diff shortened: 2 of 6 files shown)


src/sqs/push.ts
```diff
@@ -20,7 +18,7 @@ export type PushQueueMessagePayload = {
 	jiraHost: string,
 	installationId: number,
 	webhookId: string,
-	webhookReceived?: Date,
+	webhookReceived?: number,
```
KAbakumovAtl, line 21: Some typing here was wrong, and surprisingly Typescript didn't catch it

test/safe/branch.test.ts
```diff
@@ -59,14 +68,6 @@ describe("Branch Webhook", () => {
 			const ref = encodeURIComponent("heads/TES-123-test-ref");
 			const sha = "test-branch-ref-sha";
 
-			githubNock.post(`/app/installations/1234/access_tokens`)
-				.reply(200, {
-					expires_at: Date.now() + 3600,
-					permissions: {},
-					repositories: {},
-					token: "token"
-				})
```
KAbakumovAtl, line 68: This was wrong. Firstly this was only in a single test, so it was relying that the first test will be executed first and cache the token. 

Secondly it should be Optional, because if token is already cached then we won't request it again.

src/github/client/github-client.ts
```diff
@@ -91,7 +91,6 @@ export default class GitHubClient {
 		const response = await this.axios.get<T>(url, {
 			...await this.installationAuthenticationHeaders(),
 			params: {
-				installationId: this.githubInstallationId,
```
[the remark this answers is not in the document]
thombergs: This parameter has been used internally, only, to get access to the installation ID in an Axios interceptor for authentication. But authentication is not done with Axios interceptors anymore, so this parameter can go away.

src/sqs/push.ts
thombergs: I had wondered about that when I implemented the other queues, but was afraid to touch it 👍🏼

thombergs (approved):

PR #1028, opened 2022-01-11
Removed bull queue types dependency from installation.ts Got rid of b…

KAbakumovAtl:
Third and the last PR for bull queues cleanup.

This time

1. I've removed bull queue dependency from installation.ts
2. Removed all job and task objects logging (job logging didn't make sense it only logged Sentry object of the job) Task logging is not reasonable in every logline (I've logged it once when we start execute the task, we can look it up by message id if needed).
3. Removed Jobs logger adapter, because we don't log them anymore
4. Removed 'bull' npm package dependency 🥳


src/sync/installation.ts
```diff
@@ -176,16 +171,13 @@ export const isNotFoundError = (
...
-		"Repository deleted after discovery, skipping initial sync"
-	);
+	logger.info("Repository deleted after discovery, skipping initial sync");
 
 	return isNotFoundError;
 };
 
 // TODO: type queues
-async function doProcessInstallation(app, job, installationId: number, jiraHost: string, logger: LoggerWithTarget, scheduleNextTask: (delayMs) => void): Promise<void> {
+async function doProcessInstallation(app, data: BackfillMessagePayload, sentry: Hub, installationId: number, jiraHost: string, logger: LoggerWithTarget, scheduleNextTask: (delayMs) => void): Promise<void> {
```
KAbakumovAtl, line 180: That's the whole thing which was done in this file: 

Getting rid of Job type parameter, and replacing it with "data" and "sentry" - only those properties of Job which are used in this method.

Also got rid of "task" parameter in logs everywhere except "Starting task" log

test/unit/sync/branch.test.ts
```diff
@@ -5,10 +5,11 @@ import { mocked } from "ts-jest/utils";
...
 import { Application } from "probot";
 import { createWebhookApp } from "../../utils/probot";
-import createJob from "../../setup/create-job";
 import { processInstallation } from "../../../src/sync/installation";
 import nock from "nock";
 import {getLogger} from "../../../src/config/logger";
+import {Hub} from "@sentry/types/dist/hub";
+import {BackfillMessagePayload} from "../../../src/sqs/backfill";
 
 jest.mock("../../../src/models");
```
KAbakumovAtl, line 15: Commit, branch and pullrequest tests are skipped here btw,
We need to think if we want to keep them or delete them.

mboudreau (approved):
Hurray!  Next up, removing redis ;)

PR #1034, opened 2022-01-13
ARC-880 blocked IP per organization

mboudreau:
Right now, when trying to add a new organization, if one of them uses the IP whitelist and is currently blocking our IP for some reason (like manual input), then the whole page fails outright not allowing the user to add any organizations.

This is to make it fail gracefully per org and show an appropriate message to the user about that specific org.


thombergs (approved):

src/frontend/get-jira-configuration.ts
```diff
@@ -19,60 +19,58 @@ function mapSyncStatus(syncStatus: SyncStatus = SyncStatus.PENDING): string {
 	}
 }
 
-export async function getInstallation(client, subscription: SubscriptionClass, reqLog?: Logger) {
+export interface InstallationResults {
```
KAbakumovAtl, line 22: Should we use type instead of interface?

src/frontend/get-github-configuration.ts
```diff
@@ -1,59 +1,63 @@
 import { Installation, Subscription } from "../models";
 import { NextFunction, Request, Response } from "express";
-import { getInstallation } from "./get-jira-configuration";
+import { getInstallations, InstallationResults } from "./get-jira-configuration";
```
KAbakumovAtl, line 3: /nit Should we move getInstallations and InstallationResults somewhere in util class (or closer to the model) so our APIs code won't depend on each other?

src/frontend/get-github-configuration.ts
```diff
@@ -65,41 +69,33 @@ async function getInstallationsWithAdmin(log: Logger,
...
-			numberOfRepos: numberOfRepos || 0,
-			admin
-		});
+			installationsWithAdmin.push({
+				...installation,
+				numberOfRepos: numberOfRepos || 0,
+				admin
+			});
+		} catch (err) {
+			log.warn({err, installationId: installation.id, org: installation.account.login}, "Cannot check admin or get number of repos per org")
```
KAbakumovAtl, line 85: Am I correctly understand that adding this try-catch block it is the main change in this PR and everything else is merely a refactoring?

KAbakumovAtl: How would it look in UI? We won't display any message aren't we?

KAbakumovAtl (dismissed):

KAbakumovAtl:
Overall looks good, I've left a few nits and questions

src/frontend/get-github-configuration.ts
mboudreau: Agreed.  I'm working on another PR to fix those kinds of issues as well as some type stuff.

src/frontend/get-github-configuration.ts
mboudreau: That's correct.  The refactoring was needed though as we needed a way to separate installations that worked vs those that didn't, and then decide if we want to uninstall those that didn't.

No UI, at this point we're just preventing the page from crashing altogether.  I'll have another PR where we can show the ones that failed with a message to the user that I need to talk to Jade about.

src/frontend/get-jira-configuration.ts
mboudreau: Why?  Type seems like more work to implement?

src/frontend/get-github-configuration.ts
mboudreau: agreed, I'll have another PR to address this.

src/frontend/get-jira-configuration.ts
KAbakumovAtl: For consistency, because everywhere else we use types for data objects

```
export type InstallationResults = {
	fulfilled: AppInstallation[],
	rejected: FailedAppInstallation[],
	total: number
}
```

Seems that's all we need to do (If I am not missing something)

src/frontend/get-github-configuration.ts
```diff
@@ -90,6 +90,8 @@ const getInstallationsWithAdmin = async (
 
 const removeFailedConnectionsFromDb = async (req: Request, installations: InstallationResults, jiraHost: string): Promise<void> => {
 	await Promise.all(installations.rejected
+		// Only uninstall deleted installations
+		.filter(failedInstallation => failedInstallation.deleted)
```
KAbakumovAtl, line 94: That's an important change. Was it caught by tests? (If not, should we add some tests?)

KAbakumovAtl (approved):

src/frontend/get-jira-configuration.ts
mboudreau: I see 38 matches for `interface` use and 35 for `type` use in the codebase.  It's definitely not the standard.

I normally prefer `interface` for objects as it's easier to read when you're doing `extends` or using utils like `Partial`, and you can't used  `implements` with types.  For example:

```
type Foo = { blah: string; };
type Bar = { weee:boolean };
type FooBarType = foo & bar & { something: string; };

interface FooBar extends Foo, Bar {
   something: string;
}
```

Personally, I find interface to be cleaner and there's some nifty tricks you can do with it too.  Types are good for unions and merges of types.

KAbakumovAtl (approved):

PR #1036, opened 2022-01-18
ARC-809 Added error handling for GraphQL errors

KAbakumovAtl:
This PR fixes a few issues:

1. It turned out that x-ratelimit-reset was never used to calculate delay for backfill (There was a bug). 
2. We absolutely ignored all GraphQL errors. I've added error handling, similar to what Octokit does
3. Added handling for the new error format for bakcfill

Sorry in advance I had to refactor one thing along the way. I'll mention in code comments what exactly was changed.


src/github/client/errors.ts
```diff
@@ -1,13 +1,16 @@
-import { AxiosError } from "axios";
+import {AxiosError, AxiosResponse} from "axios";
 
 export class GithubClientError extends Error {
 	status?: number;
-	cause: AxiosError;
-	constructor(message: string, cause: AxiosError, status?: number) {
+	cause?: AxiosError;
+
+	constructor(message: string, status?: number, cause?: AxiosError) {
```
KAbakumovAtl, line 7: I've swapped "status" and "cause", because for GraphQL errors don't have axios exceptions, but there is status code (200)

src/github/client/github-client.ts
```diff
@@ -1,38 +1,21 @@
...
+import {getLogger} from "../../config/logger";
+import {urlParamsMiddleware} from "../../util/axios/common-middleware";
+import {InstallationId} from "./installation-id";
 import {GetBranchesQuery, GetBranchesResponse, ViewerRepositoryCountQuery} from "./github-queries";
+import {GraphQLError, GraphQLErrors, rateLimitErrorFromResponse} from "./errors";
 
-/**
- * The response type for GitHub GraphQL calls.
- * Was copied from @octokit/graphql to avoid adding a dependency on Octokit Graphql client
- */
```
KAbakumovAtl, line 18: This comment doesn't make sense anymore.

I've extracted GraphQLErrors as a separate type and moved it to errors.ts, also added "type" parameter, because it was missing.

src/github/client/interceptors.ts
```diff
@@ -86,12 +85,11 @@ export const handleFailedRequest = (logger: Logger) =>
 			const status = response?.status;
 			const errorMessage = `Error executing Axios Request ` + error.message;
 
-			const rateLimitResetHeaderValue:string = response.headers?.["x-ratelimit-reset"];
 			const rateLimitRemainingHeaderValue:string = response.headers?.["x-ratelimit-remaining"];
 			if(status === 403 && rateLimitRemainingHeaderValue == "0") {
 				logger.warn({ err: error }, "Rate limiting error");
-				const rateLimitReset: number = parseInt(rateLimitResetHeaderValue) || Date.now() / 1000 + ONE_HOUR_IN_SECONDS;
-				return Promise.reject(new RateLimitingError(rateLimitReset, 0, error, status));
+				const rateLimitingError = rateLimitErrorFromResponse(response, error);
```
KAbakumovAtl, line 91: I've extracted exception creation to a separate function rateLimitErrorFromResponse, because the same logic is used for graphql error handling

src/sync/installation.ts
```diff
@@ -314,56 +316,66 @@ async function doProcessInstallation(app, data: BackfillMessagePayload, sentry:
...
-		const rateLimit = Number(err?.headers?.["x-ratelimit-reset"]);
-		const delay = Math.max(Date.now() - rateLimit * 1000, 0);
-
-		if (delay) {
-			// if not NaN or 0
-			logger.info({ delay }, `Delaying job for ${delay}ms`);
-			scheduleNextTask(delay);
-			return;
-		}
+		await handleBackfillError(err, logger, scheduleNextTask, data, task, repositoryId, subscription, nextTask);
```
KAbakumovAtl, line 319: I've moved it to a separate method, this one was too huge

src/sync/installation.ts
```diff
@@ -314,56 +316,66 @@ async function doProcessInstallation(app, data: BackfillMessagePayload, sentry:
...
-				"You have triggered an abuse detection mechanism"
-			)
-		) {
-			// Too much server processing time, wait 60 seconds and try again
-			logger.warn("Abuse detection triggered. Retrying in 60 seconds");
-			scheduleNextTask(60_000);
-			return;
-		}
+	const rateLimit = (err instanceof RateLimitingError || err instanceof OldRateLimitingError) ? err.rateLimitReset : Number(err?.headers?.["x-ratelimit-reset"]);
+	const delay = Math.max(rateLimit * 1000 - Date.now(), 0);
```
KAbakumovAtl, line 332: A few things were changed in error handling.

Before:
```
	const rateLimit = Number(err?.headers?.["x-ratelimit-reset"]);
	const delay = Math.max(Date.now() - rateLimit * 1000, 0);
```

Delay calculation was wrong, because it would be always zero (rate-limit-reset time is always greater than the current time)

I've added error type check and getting rate limit from our rate limiting exceptions

src/sync/installation.ts
```diff
@@ -380,7 +392,7 @@ export async function maybeScheduleNextTask(
 		const delayMs = nextTaskDelaysMs.shift()!;
 		logger.info("Scheduling next job with a delay = " + delayMs);
 
-		await sqsQueues.backfill.sendMessage(jobData, (delayMs || 0) / 1000, logger);
+		await sqsQueues.backfill.sendMessage(jobData, Math.ceil((delayMs || 0) / 1000), logger);
```
KAbakumovAtl, line 395: I also had to change the delay time here. Because sqsQueue has the delay in seconds, so we had to round it.

test/unit/sync/branch.test.ts
```diff
@@ -102,6 +102,23 @@ describe("sync/branches", () => {
 		};
 	}
 
+	function nockGitHubGraphQlRateLimit(rateLimitReset: string) {
```
KAbakumovAtl, line 105: Testing graphql error exactly in the same format I've got it from Postman

test/unit/sync/branch.test.ts
```diff
@@ -269,6 +287,13 @@ describe("sync/branches", () => {
 			expect(scope).not.toBeDone();
 			nock.removeInterceptor(interceptor);
 		});
+
+		it("should reschedule message with delay if there is rate limit", async () => {
```
KAbakumovAtl, line 291: Adding test to the branches test, because it is the only test which covers error handling part of the backfill logic.

It is on my list to restore PRs and Commits tests too

src/github/client/interceptors.ts
mboudreau: Wouldn't it make more sense to keep it in here and use the same logic of this function for both REST and GraphQL calls?

KAbakumovAtl: We can't unfortunately, because the responses for normal API and graphql are different. Normal API relies with 403 error, while GraphQL returns 200 and the error in the response body.

However, I can incapsulate some logic in RateLimitingError constructor to avoid code duplication

src/github/client/errors.ts
```diff
@@ -1,13 +1,16 @@
-import { AxiosError } from "axios";
+import {AxiosError, AxiosResponse} from "axios";
 
 export class GithubClientError extends Error {
```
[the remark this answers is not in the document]
KAbakumovAtl: So GithubClientError is that type

src/github/client/errors.ts
```diff
@@ -17,14 +20,57 @@ export class RateLimitingError extends GithubClientError {
...
 	constructor(error: AxiosError, status?: number) {
-		super("Blocked by GitHub allowlist", error, status);
+		super("Blocked by GitHub allowlist", status, error);
+	}
+}
+
+/**
+ * Type for errors section in GraphQL response
+ */
+export type GraphQLErrors = [
```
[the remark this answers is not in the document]
KAbakumovAtl: Yea, makes sense

src/github/client/github-client.ts
```diff
@@ -124,14 +107,26 @@ export default class GitHubClient {
...
 				variables
 			},
 			{
 				...await this.installationAuthenticationHeaders(),
 			});
+
+		const graphqlErrors = response.data.errors;
+		if(graphqlErrors && graphqlErrors.length > 0) {
+
+			if (graphqlErrors[0].type == "RATE_LIMITED") {
```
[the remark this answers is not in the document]
KAbakumovAtl: so usually if there is rate limiting there are no other errors. (As I found experimentally). But yea, totally makes sense, I'll change it

src/github/client/github-client.ts
```diff
@@ -124,14 +107,26 @@ export default class GitHubClient {
...
+				return Promise.reject(rateLimitErrorFromResponse(response));
+			}
+
+			return Promise.reject(new GraphQLError(graphqlErrors[0].message, graphqlErrors));
```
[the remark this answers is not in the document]
KAbakumovAtl: It does, I passing them as a second parameter. The problem is the message will be too big if there are many errors, so I truncated.

But you are right, I'll make it more explicit by telling in the error message that there are more errors.

src/sqs/error-handlers.ts
```diff
@@ -24,12 +24,17 @@ const EXPONENTIAL_BACKOFF_MULTIPLIER = 3;
 export const jiraAndGitHubErrorsHandler : ErrorHandler<any> = async (error: JiraClientError | Octokit.HookError | OldRateLimitingError | RateLimitingError | Error,
 	context: Context<any>) : Promise<ErrorHandlingResult> => {
 
-	const maybeResult = maybeHandleNonFailureCase(error, context);
+	const maybeResult = maybeHandleNonFailureCase(error, context)
+		|| maybeHandleRateLimitingError(error)
```
KAbakumovAtl, line 28: I had to change it here, because of test failure.

`maybeHandleNonFailureCase` was returning the unretryable result if the response code was 403. 

I had to move it after the rate limiting check, because rate limiting errors have "status" field which can be 403.

src/github/client/github-client.ts
```diff
@@ -124,14 +107,28 @@ export default class GitHubClient {
 	}
 
 	private async graphql<T>(query: string, variables?: Record<string, string | number | undefined>): Promise<AxiosResponse<GraphQlQueryResponse<T>>> {
-		return  await this.axios.post<GraphQlQueryResponse<T>>("https://api.github.com/graphql",
+		const response = await this.axios.post<GraphQlQueryResponse<T>>("https://api.github.com/graphql",
```
thombergs, line 110: Not really due to this change, but: 

This url shouldn't be hard-coded, but instead something like`this.installationId.githubBaseUrl + "/graphql"` (or create a field `githubGraphqlBaseUrl` in the `InstallationId`.

Otherwise, we can't disambiguate between GHE and GH Cloud lateron.

thombergs (dismissed):

thombergs (approved):

PR #1057, opened 2022-01-31
ARC-969 fix test consistency

mboudreau:
Fixing a lot of issues with flaky tests and testing in general so that the test can pass consistently and not give false positives.


src/github/client/installation-token-cache.ts
```diff
@@ -18,13 +18,13 @@ export default class InstallationTokenCache {
 	 * @param maxTokens the max number of tokens that should be in the cache at any time. If the size of the cache is about to go above this
 	 * number, the least recently used tokens are evicted from the cache.
 	 */
-	constructor(maxTokens: number) {
-		this.installationTokenCache = new LRUCache<number, AuthToken>({ max: maxTokens });
+	constructor() {
+		this.installationTokenCache = new LRUCache<number, AuthToken>({ max: 1000 });
```
KAbakumovAtl, line 22: Why did we set it to constant here?

src/jira/client/axios.ts
```diff
@@ -6,18 +6,11 @@ import statsd from "../../config/statsd";
...
 const instance = process.env.INSTANCE_NAME;
 const iss = `com.github.integration${instance ? `.${instance}` : ""}`;
 
-// TODO: type hack to fix custom implementation of URL templating vars
-declare module "axios" {
-	interface AxiosRequestConfig {
-		urlParams?: Record<string, string>;
-	}
-}
-
```
KAbakumovAtl, line 20: Was there a reason to change it now?

KAbakumovAtl:
That's great that we fixing tests consistency.

However this PR is 66 files big and 1000+ lines got changed. 

After reviewing the first 10 files I found tons of unrelated refactoring, and some changes related to fixing flaky tests. 

I am not against refactoring, but we can't squash everything into a single PR. It is risky. 

I can't proceed with the review. We need to split it to smaller PR.

src/github/client/installation-token-cache.ts
mboudreau: Because it never changed anywhere.  No point of having it as a variable.

src/jira/client/axios.ts
mboudreau: URL params were not being encoded, which is a potential security issue and good practice to do so.  It also had some very weird behavior that was creating bad usage patterns with axios.

src/sqs/queues.ts
```diff
@@ -68,20 +70,46 @@ const sqsQueues = {
...
+			sqsQueues.push.stop(),
+			sqsQueues.discovery.stop(),
+			sqsQueues.deployment.stop(),
+			sqsQueues.branch.stop()
+		]);
+		logger.info("All queues stopped");
+	},
 
-export default sqsQueues
+	purge: async () => {
```
thombergs, line 94: As far as I can tell, the `purge()` function isn't used anywhere, is it? It's a dangerous function to have around!

src/util/handlebars/partials.ts
```diff
@@ -1,15 +1,16 @@
...
+	fs.readdirSync(partialPath)
+		?.filter(file => file.endsWith(".hbs"))
+		.forEach(file => {
+			hbs.registerPartial(
+				file.replace(/\.hbs$/, ""), // removes extension of file
+				fs.readFileSync(path.resolve(partialPath, file), {
+					encoding: "utf8"
+				})
+			);
+		});
```
thombergs, line 15: I think this may answer my question about the path from above.

thombergs (dismissed):

KAbakumovAtl (changes requested):
I am agains merging this PR because it is too risky to merge large PRs

src/sqs/queues.ts
mboudreau: That's a very good point.  I had it originally for an `afterEach` for all SQS tests, but decided to be more granular.  I'll remove it.

KAbakumovAtl: So the Purge function doesn't guarantee consistency. Messages can still be on the queue after we called it

thombergs (dismissed):

src/sqs/queues.ts
```diff
@@ -68,20 +70,46 @@ const sqsQueues = {
...
+			sqsQueues.backfill.purgeQueue(),
+			sqsQueues.push.purgeQueue(),
+			sqsQueues.discovery.purgeQueue(),
+			sqsQueues.deployment.purgeQueue(),
+			sqsQueues.branch.purgeQueue()
+		]);
+		logger.info("All queues purged");
+	},
+
+	getMessageCount: async () => {
```
KAbakumovAtl: Can we add comments that it is test-only?

mboudreau: I removed them

src/github/middleware.ts
```diff
@@ -173,7 +172,8 @@ export default (
 
 			if (await booleanFlag(BooleanFlags.MAINTENANCE_MODE, false, jiraHost)) {
 				context.log(
-					`Maintenance mode ENABLED for jira host ${jiraHost} - Ignoring event of type ${webhookEvent}`
+					{jiraHost, webhookEvent},
```
yvonnevictoria, line 175: I've noticed you've done this in a couple of places. What's the advantage of moving away from string interpolation?

test/safe/branch.test.ts
```diff
@@ -256,8 +236,9 @@ describe("Branch Webhook", () => {
 			});
 		});
 
-		it("should not update Jira issue if there are no issue Keys in the branch name", async () => {
+		it.skip("should not update Jira issue if there are no issue Keys in the branch name", async () => {
```
yvonnevictoria, line 239: Why are we skipping rather than deleting these? Or fixing them?

test/setup/setup.ts
```diff
@@ -3,57 +3,98 @@ import env from "../../src/config/env";
 import "./matchers/to-have-sent-metrics";
 import "./matchers/nock";
 import "./matchers/to-promise";
-import statsd from "../../src/config/statsd";
 import { sequelize } from "../../src/models/sequelize";
-import InstallationTokenCache from "../../src/github/client/installation-token-cache";
-import AppTokenHolder from "../../src/github/client/app-token-holder";
+import { mocked } from "ts-jest/utils";
+// WARNING: Be very careful what you import here as it might affect test
+// in other others because of dependency tree.  Keep imports to a minimum.
```
yvonnevictoria, line 9: `in other others` hahaha. Not sure what the word was meant to be here

test/setup/setup.ts
```diff
@@ -64,5 +105,5 @@ afterAll(async () => {
 	// stop only if setup did run. If using jest --watch and no tests are matched
 	// we need to not execute the require() because it will fail
 	// TODO: fix wrong typing for statsd
-	statsd.close(() => undefined);
+	// statsd.close(() => undefined);
```
yvonnevictoria, line 108: Why are we commenting this out rather than deleting? Also there is a TODO above here that should be fixed or deleted.

yvonnevictoria (dismissed):

test/setup/setup.ts
KAbakumovAtl: It is a strong no go. We shuldn't skip working tests

KAbakumovAtl: We went through this once before and caused an incident

test/safe/branch.test.ts
KAbakumovAtl: It is a strong no go. We shuldn't skip working tests.

We went through this once before and caused an incident

src/github/middleware.ts
mboudreau: In splunk or if you're using a log aggregate tool locally, it can group together errors with the same message and show you how often an issue actually happens.  This cannot happen when you use string interpolation since every message is now somewhat unique depending on the variable.  The message should be consistent across the board and the added benefit that if you put the variable in the 'data' of the log object, you can now filter by that variable.  No data is lost, but your logs improve drastically and makes it much easier to diagnose issues particularly at scale.

test/safe/branch.test.ts
mboudreau: I didn't write them originally and this PR is big enough (and my enough of my time used) without having to fix other people's tests.  We could delete them, but I think the idea was that this was a test to be implemented by someone but never finished.

test/setup/setup.ts
mboudreau: :|

test/setup/setup.ts
mboudreau: Yeah, probably should delete this.

src/github/client/installation-token-cache.ts
KAbakumovAtl: /nit we can extract it into constant to improve clarity

src/github/middleware.ts
KAbakumovAtl: I believe we should add jiraHost into context.log, when we know it, so it will be added to all logs automatically. (However this is out of scope of this PR)

src/sqs/index.ts
```diff
@@ -213,46 +208,61 @@ export class SqsQueue<MessagePayload> {
...
-			this.log.error("Queue is already stopped")
+	public async stop() {
+		if (!this.listenerContext || this.listenerContext.stopped) {
+			this.log.warn("Queue is already stopped");
 			return;
 		}
 		this.listenerContext.log.info("Stopping the queue");
 		this.listenerContext.stopped = true;
+
+		return this.waitUntilListenerStopped();
```
KAbakumovAtl, line 238: 👍

test/safe/branch.test.ts
KAbakumovAtl: I wonder why did we have to skip it? Was it failing?

The test-case makes sense to me, we are trying to send a branch without issue keys and make sure we don't call Jira.


However, I think you are right that the test wouldn't work. Now when we have a branch queue it can give false-negatives, because we don't wait until the queue message will be processed. We should make it as a unit test for the queue listener. Which is out of scope of this PR.

KAbakumovAtl: Let's leave a not about it (to move them to unit tests and reason why)

test/setup/setup.ts
```diff
@@ -3,66 +3,104 @@ import env from "../../src/config/env";
...
+	// Assign defaults to process.env, but don't override existing values if they
+	// are already set in the environment.
+	process.env = {
+		...process.env,
+		...env
+	};
+};
+
+const clearState = async () => Promise.all([
+	sequelize.truncate({ truncate: true })
```
KAbakumovAtl, line 50: Nice

test/frontend/get-jira-configuration.test.ts
```diff
@@ -34,12 +34,6 @@ describe("Jira Configuration Suite", () => {
 		});
 	});
 
-	afterEach(async () => {
-		await Subscription.destroy({ truncate: true });
-		await Installation.destroy({ truncate: true });
-		await RepoSyncState.destroy({ truncate: true });
```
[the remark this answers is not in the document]
KAbakumovAtl: I found the answer on my question

test/unit/sync/branch.test.ts
```diff
@@ -1,29 +1,26 @@
...
+import { booleanFlag, BooleanFlags } from "../../../src/config/feature-flags";
+
+jest.mock("../../../src/sqs/queues");
 jest.mock("../../../src/config/feature-flags");
 
-describe("sync/branches", () => {
+
+describe.each([true, false])("sync/branches - New GH Client feature flag is '%s'", () => {
 	const installationId = 1234;
+	const useNewGithubClient = true;
```
KAbakumovAtl, line 23: It will be constantly true now. Should we set it to the describe parameter value?

```suggestion
	describe.each([true, false])("sync/branches - New GH Client feature flag is '%s'", (useNewGithubClient) => {
	const installationId = 1234;
```

KAbakumovAtl: - [ ] fix feature flag mocking

KAbakumovAtl (commented):
I've left some comments. 

I've left 2 tasks in places which are potentially broken. Would be happy to approve after it addressed

src/sqs/index.ts
```diff
@@ -368,69 +381,73 @@ export class SqsQueue<MessagePayload> {
...
+	private async changeVisibilityTimeout(message: Message, timeout: number, logger: Logger): Promise<void> {
+		if (!message.ReceiptHandle) {
+			logger.error(`No ReceiptHandle in message with ID = ${message.MessageId}`);
+			return;
+		}
 
-		if(!message.ReceiptHandle) {
-			logger.error(`No ReceiptHandle in message with ID = ${message.MessageId}`)
+		if(!Number.isInteger(timeout) || timeout < 0) {
+			logger.error(`Timeout needs to be a positive integer.`);
```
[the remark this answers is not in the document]
mboudreau: ...yeah, and we're checking for if it's a negative (under zero)?

src/sqs/queues.ts
```diff
@@ -6,10 +6,13 @@ import { jiraAndGitHubErrorsHandler, webhookMetricWrapper } from "./error-handle
 import { DiscoveryMessagePayload, discoveryQueueMessageHandler } from "./discovery";
 import { DeploymentMessagePayload, deploymentQueueMessageHandler } from "./deployment";
 import { BranchMessagePayload, branchQueueMessageHandler } from "./branch";
+import { getLogger } from "../config/logger";
 
 const LONG_POLLING_INTERVAL_SEC = 3;
+const logger = getLogger("sqs-queues");
 
-const sqsQueues = {
+// TODO: Make this a class
```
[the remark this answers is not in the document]
mboudreau: coding conventions says to use a class where possible.  There are some benefits to this like making the queues `readonly` so nobody changes it by mistake

test/frontend/get-jira-configuration.test.ts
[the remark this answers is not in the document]
mboudreau: I figured it was simpler to do it globally for every test as part of state cleanup.  I also added the same cleanup on `beforeAll` as it's happened to me a few times if I force exit a test in the middle of debugging and the state is still in the DB

test/github/client/installation-token-cache.test.ts
```diff
@@ -1,5 +1,6 @@
 import InstallationTokenCache from "../../../src/github/client/installation-token-cache";
 import AuthToken, { ONE_MINUTE, TEN_MINUTES } from "../../../src/github/client/auth-token";
+jest.unmock("lru-cache");
```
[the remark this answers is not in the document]
mboudreau: we mock it globally in `setup.ts` so that we have to be explicit in our nocks what is being called.  This removes a lot of flakiness when it comes to running all tests vs a single test, or if the tests are not run in a specific order.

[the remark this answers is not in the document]
mboudreau: in this case though, since we're trying to test the caching itself, we unmock it so lru-cache works as it should.

test/safe/deployment.test.ts
```diff
@@ -37,11 +29,12 @@ describe("Deployment Webhook", () => {
...
 		});
 
+		await sqsQueues.deployment.start();
 	});
 
 	afterEach(async () => {
-		await Installation.destroy({ truncate: true });
-		await Subscription.destroy({ truncate: true });
+		await sqsQueues.deployment.stop();
+		await sqsQueues.deployment.purgeQueue();
```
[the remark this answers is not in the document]
mboudreau: Starting/stopping every listener for every single tests is inefficient and redundant.  If you're trying to test that scenario, you're welcome to start multiple listeners in that test suite.

test/unit/sync/branch.test.ts
mboudreau: whoops, forgot to change that back after debugging.

test/safe/deployment.test.ts
```diff
@@ -13,14 +12,7 @@ describe("Deployment Webhook", () => {
...
 	beforeAll(async () => {
-		//Start worker node for queues processing
-		await start();
-	});
-
-	afterAll(async () => {
-		//Stop worker node
-		await stop();
-		await sqsQueues.deployment.waitUntilListenerStopped();
+		await sqsQueues.branch.purgeQueue();
```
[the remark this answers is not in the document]
mboudreau: yes.

src/github/middleware.ts
yvonnevictoria: Oh that makes sense. Thanks for explaining @mboudreau

test/safe/branch.test.ts
mboudreau: I'm skipping it because it literally does nothing.  It creates a mock function that can't ever be called, then asserting that it's never called.  We're not asserting anything of value, and as such, this test will pass no matter what happens.  I'm skipping it because it's a reminder that this test needs to be fixed.

src/sqs/index.ts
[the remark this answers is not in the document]
KAbakumovAtl: Oh, this one will be false `!Number.isInteger(timeout)` when timeout is not number. Got confused for a moment

test/safe/branch.test.ts
KAbakumovAtl: It could happen when we were migrating to branches queue. Before this test would work, because it would fail if we try making request to Jira (and we also try to fetch the latest commit in the handler, if I am not wrong)

KAbakumovAtl (approved):

PR #1060, opened 2022-02-01
Arc 1047

joshkay10:
**ARC-1047 INNOVATION PROJECT CONFIG-AS-CODE**

[Config as code](https://hello.atlassian.net/wiki/spaces/~713490667/pages/1508946202/GitHub+for+Jira+-+Config+as+Code)

As part of the config-as-code innovation project I have created a new service **repo-config.service.ts**. The is intention to read the contents of a repos .jira/config.yml, transform it in JSON and save the data into a new repoConfig database.

Private function **getRepoConfig** uses the new githubclient function **getRepositoryContent** (refer to bottom of this file) to request the contents of .jira/config.yml and converts it from a base64 to a string.
- currently it has a generic catch if it does not find the file/fails, this will need to be enhanced for 404 vs actual failed request etc.

Private function **convertYamlToRepoConfig** is helper function that can convert yaml to JSON (we required stringified JSON to save to the database), could be easily refactored into a generic util function... but have not at this stage.

Private function **saveRepoConfigToDB** is a  wrapper function to call the create/update function on the new repoConfig database... 

Lastly included an exposed function called **processRepoConfig** which will run all of the previous functions to achieve saving a repos ./jira/config.yml data into a database repoConfig.

CHANGES TO THE GITHUB CLIENT

Created a new function to call the [Get repository content API](https://docs.github.com/en/rest/reference/repos#get-repository-content)

### DIFF
--- a/src/config-as-code/repo-config-service.ts
+++ b/src/config-as-code/repo-config-service.ts
@@ -6,16 +6,14 @@ import { RepoConfig } from "./repo-config";
 import RepoConfigDatabaseModel from "./repo-config-database-model";
 
 const logger = getLogger("services.config-as-code");
-const CONFIG_PATH = "config.yml";
+const CONFIG_PATH = "aconfig.yml";
 
 const getRepoConfig = async (githubInstallationId: number, owner: string, repo: string): Promise<string | null> => {
 	const client = new GitHubClient(getCloudInstallationId(githubInstallationId), logger);
-
 	try {
 		const response = await client.getRepositoryContent(owner, repo, CONFIG_PATH)
 		const contentString = Buffer.from(response.data.content, "base64").toString("ascii"); 
 		logger.debug(`Converted ${response.data.content} to ${contentString}`);
-
 		return contentString;
 	} catch (err) {
 		logger.error(err, "Get repo failed.");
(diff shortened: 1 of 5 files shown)


mboudreau:
Could you please describe the PR in a bit more detail?  Also, please add a short title to the PR with the ticket number so it's easier to identify in the commits

src/config-as-code/repo-config-service.ts
```diff
@@ -0,0 +1,55 @@
...
+const CONFIG_PATH = "aconfig.yml";
+
+const getRepoConfig = async (githubInstallationId: number, owner: string, repo: string): Promise<string | null> => {
+	const client = new GitHubClient(getCloudInstallationId(githubInstallationId), logger);
+	try {
+		const response = await client.getRepositoryContent(owner, repo, CONFIG_PATH)
+		const contentString = Buffer.from(response.data.content, "base64").toString("ascii"); 
+		logger.debug(`Converted ${response.data.content} to ${contentString}`);
+		return contentString;
+	} catch (err) {
```
thombergs, line 18: I think we should just let this error bubble up in the call stack. The caller should care about the proper error handling. Same with other catch blocks in the other functions.

src/config-as-code/repo-config-service.ts
```diff
@@ -0,0 +1,55 @@
...
+import GitHubClient from "../github/client/github-client";
+import { getCloudInstallationId } from "../github/client/installation-id";
+import { getLogger } from "../config/logger";
+import { RepoConfig } from "./repo-config";
+import RepoConfigDatabaseModel from "./repo-config-database-model";
+
+const logger = getLogger("services.config-as-code");
+const CONFIG_PATH = "aconfig.yml";
+
+const getRepoConfig = async (githubInstallationId: number, owner: string, repo: string): Promise<string | null> => {
```
thombergs, line 11: Maybe rename to `getRepoConfigFromGitHub()` to make it unmistakenly clear where we get it from? :)

src/config-as-code/repo-config-service.ts
```diff
@@ -0,0 +1,55 @@
+import YAML from "yaml";
+import GitHubClient from "../github/client/github-client";
+import { getCloudInstallationId } from "../github/client/installation-id";
+import { getLogger } from "../config/logger";
+import { RepoConfig } from "./repo-config";
+import RepoConfigDatabaseModel from "./repo-config-database-model";
+
+const logger = getLogger("services.config-as-code");
+const CONFIG_PATH = "aconfig.yml";
```
thombergs, line 9: This should probably be `.jira/config.yml`?

src/transforms/branch.ts
```diff
@@ -50,14 +56,38 @@ export default async (github: GitHubAPI, webhookPayload: WebhookPayloadCreate) =
 
 	const lastCommit = await getLastCommit(github, webhookPayload, issueKeys);
 
+	console.log("SHOW ME THE RETURN");
```
thombergs, line 59: Is this from another change? Doesn't seem to have to do with the config as code stuff.

src/github/client/github-client.ts
```diff
@@ -192,4 +190,14 @@ export default class GitHubClient {
 		return response?.data?.data;
 	}
 
-}
+	/**
+	 * Get the repositiry contents from PATH
+	 */
+	public async getRepositoryContent(owner: string, repo: string, path: string) : Promise<AxiosResponse<Octokit.GitGetBlobResponse>>  {
```
thombergs, line 196: From the GitHub docs it looks more like `Octokit.ReposGetContentsResponse` is the expected return type here. That also contains the size of the file so we can throw an error in the service of the size is bigger than x.

src/config-as-code/repo-config-service.ts
```diff
@@ -0,0 +1,55 @@
...
+		logger.debug(`Converted ${response.data.content} to ${contentString}`);
+		return contentString;
+	} catch (err) {
+		logger.error(err, "Get repo failed.");
+		return null;
+	}
+}
+
+const convertYamlToRepoConfig = (input: string): RepoConfig => {
+	return YAML.parse(input)
```
thombergs, line 25: Can we make sure here somehow that the returned JSON only contains the fields that we expect? We don't want to store garbage in the database when someone decides to put (valid) garbage YAML in their config file.

thombergs (approved):
I left some comments, but feel free to merge it into the `config-as-code` branch any time. We'll do a proper review of the whole branch when we merge it into `main` anyways.

src/config-as-code/repo-config-service.ts
```diff
@@ -0,0 +1,55 @@
+import YAML from "yaml";
```
thombergs, line 1: Once the structure has settled a bit, let's also create a test for this file (and add a test case for the new function in the GitHub client)

PR #1066, opened 2022-02-02
Adding optional chaining to view helpers to prevent null assertions

mboudreau:
Found a few issues in the logs while I was looking at something else.  It seems to happen after you've just restarted a backfill and `syncStatus` is null while the page refreshes, making the `replace` fail.  This is to prevent that from happening.  Since this is a helper function, there's no types to prevent this from happening.


src/util/handlebars/helpers.ts
```diff
@@ -1,9 +1,9 @@
 import hbs from "hbs";
 
 export const registerHandlebarsHelpers = () => {
-	hbs.registerHelper("toLowerCase", (str) => str.toLowerCase());
+	hbs.registerHelper("toLowerCase", (str?: string) => str?.toString().toLowerCase());
 
-	hbs.registerHelper("replaceSpaceWithHyphen", (str) => str.replace(/ /g, "-"));
+	hbs.registerHelper("replaceSpaceWithHyphen", (str?: string) => str?.toString().replace(/ /g, "-"));
```
KAbakumovAtl, line 6: Where its being used and how can we test it?

If str is undefined, we'll get "Undefined', should we do an empty string instead?

mboudreau: In `jira-configuration.hbs`.  We can add tests to it.  Good point on the empty string.

mboudreau: Tests added.

rachellerathbone (approved):

PR #1067, opened 2022-02-03
ARC-1059 url params revert and JWT token fix

mboudreau:
Fixed the JWT token generation which wasn't using the correct path/query to generate the token from, causing a 401 http status from the jira API.

As part of this changed, I've also changed the url variable syntax to be more specific by using `{var}` to prevent some edge cases with replacements.


KAbakumovAtl:
Could you point to the places in the code which were causing issues and how it was fixed?

Also, can we add tests for it?

src/jira/client/axios.ts
```diff
@@ -1,74 +1,60 @@
...
-					Authorization: `JWT ${jwtToken}`
-				}
-			};
-		}
-	);
-}
+const getAuthMiddleware = (secret: string, instance: AxiosInstance) =>
+	(config: AxiosRequestConfig): AxiosRequestConfig => {
+		// Generate full URI based on current config
+		const uri = instance.getUri(config);
```
mboudreau, line 22: @KAbakumovAtl This is specifically where the problem lied.  It was using `config.url` to parse the path and querystring, but that's an optional parameter and most would use the `params` option instead for queries instead of adding it inline.  By doing this, the path/query could be wrong depending on the situation.

Now we can actually get the full URI that will be generated by Axios before the request and parse that instead since there is an encoding step that happens for all params.

As for tests, there's probably a few we can add around the axios client since we'd need to test it in conjunction with the other middleware which can interfere.  I'll add some in.

KAbakumovAtl: That sounds like it could affect Push API as well. Is it because we didn't use query parameters and the url didnt' change?

src/github/client/github-client.ts
```diff
@@ -132,7 +132,7 @@ export default class GitHubClient {
 	 * Lists pull requests for the given repository.
 	 */
 	public async getPullRequests(owner: string, repo: string, pullRequestParams: GetPullRequestParams): Promise<AxiosResponse<Octokit.PullsListResponseItem[]>> {
-		return await this.get<Octokit.PullsListResponseItem[]>(`/repos/:owner/:repo/pulls`, pullRequestParams, {
+		return await this.get<Octokit.PullsListResponseItem[]>(`/repos/{owner}/{repo}/pulls`, pullRequestParams, {
```
KAbakumovAtl, line 135: This is GitHub Client. How would this format change impact requests?

src/jira/client/axios.ts
mboudreau: yep.

test/unit/util/axios/url-params-middleware.test.ts
```diff
@@ -0,0 +1,110 @@
+import { urlParamsMiddleware } from "../../../../src/util/axios/url-params-middleware";
+import { AxiosRequestConfig } from "axios";
+
+describe("URL Params Axios Middleware", () => {
```
KAbakumovAtl, line 4: Nice

yvonnevictoria (approved):

test/unit/util/axios/jira-auth-middleware.test.ts
```diff
@@ -0,0 +1,69 @@
+import axios, { AxiosRequestConfig } from "axios";
+import fs from "fs";
+import { jiraAuthMiddleware } from "../../../../src/util/axios/jira-auth-middleware";
+import path from "path";
+
+describe("Jira Auth Axios Middleware", () => {
```
KAbakumovAtl, line 6: Nice

KAbakumovAtl (approved):
Looks good.

I'd only ask to test changes in staging before merging

src/github/client/github-client.ts
mboudreau: It won't.  Github client uses Axios with URL Param middleware, so it'll do the correct replacements.

PR #1077, opened 2022-02-09
Use url params for GitHub installation Id

KAbakumovAtl:
Because we use path of the original URL as a metric tag, doing this:
```
this.axios.post<Octokit.AppsCreateInstallationTokenResponse>(`/app/installations/${githubInstallationId}/access_tokens`
```
exploded our metrics cardinality.

I've reeplaced it with urlParams

### DIFF
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -88,8 +88,11 @@ export default class GitHubClient {
 	 * API in the name of an installation of that app (to access the users' data).
 	 */
 	private async createInstallationToken(githubInstallationId: number): Promise<AuthToken> {
-		const response = await this.axios.post<Octokit.AppsCreateInstallationTokenResponse>(`/app/installations/${githubInstallationId}/access_tokens`, {}, {
-			...this.appAuthenticationHeaders()
+		const response = await this.axios.post<Octokit.AppsCreateInstallationTokenResponse>(`/app/installations/{githubInstallationId}/access_tokens`, {}, {
+			...this.appAuthenticationHeaders(),
+			urlParams: {
+				githubInstallationId
+			}
 		});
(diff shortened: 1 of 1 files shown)


src/github/client/github-client.ts
```diff
@@ -104,7 +107,7 @@ export default class GitHubClient {
 	}
 
 	private async graphql<T>(query: string, variables?: Record<string, string | number | undefined>): Promise<AxiosResponse<GraphQlQueryResponse<T>>> {
-		const response = await this.axios.post<GraphQlQueryResponse<T>>(this.githubInstallationId.githubBaseUrl +"/graphql",
+		const response = await this.axios.post<GraphQlQueryResponse<T>>("/graphql",
```
KAbakumovAtl, line 110: I've realised that we set the same base URL to the client itself already.

We have integration tests which cover this cahnge

src/github/client/github-client.ts
```diff
@@ -88,8 +88,11 @@ export default class GitHubClient {
 	 * API in the name of an installation of that app (to access the users' data).
 	 */
 	private async createInstallationToken(githubInstallationId: number): Promise<AuthToken> {
-		const response = await this.axios.post<Octokit.AppsCreateInstallationTokenResponse>(`/app/installations/${githubInstallationId}/access_tokens`, {}, {
-			...this.appAuthenticationHeaders()
+		const response = await this.axios.post<Octokit.AppsCreateInstallationTokenResponse>(`/app/installations/{githubInstallationId}/access_tokens`, {}, {
+			...this.appAuthenticationHeaders(),
+			urlParams: {
+				githubInstallationId
+			}
```
KAbakumovAtl, line 95: That's the main fix. Thankfully we have tons of tests which already cover this path

mboudreau (approved):
:clap:

PR #1079, opened 2022-02-09
Arc 778 - Change generated create pull request link to new format to allow title overrides with query params

joshkay10:
Change PR URL to new format to include title override query param

e.g. from:
https://github.com/atlassian/github-for-jira/pull/new/branch-x
to
https://github.com/atlassian/github-for-jira/compare/branch-x?title=ARC-778-branch-x

This URL was created in 3 seperate locations for the attribute **createPullRequestUrl**
- github-for-jira/src/transforms/pull-request.ts
- github-for-jira/src/transforms/branch.ts
- github-for-jira/src/sync/transforms/branch.ts

So I created a single Util function to import into each of the above and call function **generateCreatePullRequestUrl** which will generate the URL and ensure that the issue keys are attached appropriately.

The above function has unit tests for with single, multiple and none issue keys here
-  github-for-jira/test/unit/transforms/util/pullRequestLinkGenerator.test.ts


src/transforms/util/pullRequestLinkGenerator.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Generates a create pull request link for GH in the format [URL]/compare/[name]?title=[...keys]-[name]
+ * This format allows the use of title query param to set the PR name
+ */
+export const generateCreatePullRequestUrl = (baseUrl: string, name: string, issueKeys: string[] | null) => {
```
KAbakumovAtl, line 5: /nit should we do `issueKeys?: string[]` instead?

test/fixtures/api/graphql/branch-ref-nodes.json
```diff
@@ -13,7 +13,7 @@
                   }
                 ]
               },
-              "name": "TES-321-branch-name",
+              "name": "test-branch-name",
```
KAbakumovAtl, line 16: I wonder in which tests it is used? I am afraid it can affect some tests.

(/nit Should we name it explicitly `branch-with-issue-key-in-the-last-commit` to improve tests readability?)

test/unit/transforms/util/pullRequestLinkGenerator.test.ts
```diff
@@ -0,0 +1,33 @@
+import { generateCreatePullRequestUrl } from "../../../../src/transforms/util/pullRequestLinkGenerator";
+
+describe("pullRequestLinkGenerator()", () => {
+
+	it("should structure the url correctly with issue keys in title - for single issue key", async () => {
+		const BASE_URL = "github";
+		const NAME = "branch-7";
+		const ISSUE_KEYS = ["TEST-222"];
+		const ISSUE_KEYS_FLAT = "TEST-222"
+		const EXPECTED_RESULT = `github/compare/${NAME}?title=${ISSUE_KEYS_FLAT}%20-%20${NAME}`;
```
KAbakumovAtl, line 10: Let's use the exact string in the Expected Result. 

Otherwise this test repeats the implementation

src/transforms/util/pullRequestLinkGenerator.ts
```diff
@@ -0,0 +1,17 @@
...
+	// Remove duplicate keys
+	const uniqueIssueKeys: string[] = [...new Set(issueKeys)];
+	let title: string | undefined = uniqueIssueKeys?.join(" ");
+
+	// Only prepend issue keys if they exist
+	if (title) {
+		title = encodeURIComponent(`${title} - ${name}`);
+	} else {
+		title = encodeURIComponent(`${name}`);
+	}
```
mboudreau, line 15: I think this whole thing could be simplified to 
```
const keys = issueKeys.length ? `${issueKeys.join(" ")} - ` : "";
const title = encodeURIComponent(keys + name);
```

src/transforms/util/pullRequestLinkGenerator.ts
```diff
@@ -0,0 +1,17 @@
...
+	return  `${baseUrl}/compare/${name}?title=${title}`;
```
mboudreau, line 16: You'll need to `encodeURIComponent` of name here.

An idea I had for this since you're doing url params is you could use the `URLParamsMiddleware` function to do what you want:

```
return urlParamsMiddleware({
   url: "{baseUrl}/compare/{name}?title={title}"
   urlParams: {
      baseUrl,
      name,
      title
   }
}).url;
```

Might be overkill though.  I think we might need to create a shared util from the logic in URLParamsMiddleware that does this for any url.

mboudreau (changes requested):

src/transforms/util/pullRequestLinkGenerator.ts
joshkay10: After the changes from your previous comments the function is quite succinct, so will omit the urlParamsMiddleware at this stage.

src/transforms/pull-request.ts
```diff
@@ -82,7 +82,7 @@ export default async (github: GitHubAPI, pullRequest: Octokit.PullsGetResponse,
 				? []
 				: [
 					{
-						createPullRequestUrl: generateCreatePullRequestUrl(pullRequest?.head?.repo?.html_url, pullRequest?.head?.ref, issueKeys), 
+						createPullRequestUrl: generateCreatePullRequestUrl(pullRequest?.head?.repo?.html_url, pullRequest?.head?.ref, issueKeys || undefined),
```
KAbakumovAtl, line 85: I wonder if we should explicitly do issueKeys || undefined ? 

Looks like an excessive check. (Otherwise we'd better move it inside the function)

KAbakumovAtl (dismissed):

src/transforms/util/pullRequestLinkGenerator.ts
```diff
@@ -0,0 +1,10 @@
+/**
+ * Generates a create pull request link for GH in the format [URL]/compare/[name]?title=[...keys]-[name]
+ * This format allows the use of title query param to set the PR name
+ */
+export const generateCreatePullRequestUrl = (baseUrl: string, name: string, issueKeys: null | string[] = []) => {
```
ccrolf, line 5: Is it possible to use a feature-flag in this method, so we can do a gradual rollout (ie, reduce risk)?

joshkay10: Can do, created flag use-new-github-pull-request-url-format, will include.

src/sync/transforms/branch.ts
```diff
@@ -13,16 +14,13 @@ import _ from "lodash";
...
 	const branchKeys = issueKeyParser().parse(branch.name) || [];
 	const pullRequestKeys = issueKeyParser().parse(
 		branch.associatedPullRequests.nodes.length ? branch.associatedPullRequests.nodes[0].title : ""
 	) || [];
 	const commitKeys = issueKeyParser().parse(branch.target.message) || [];
-
-	const allKeys = branchKeys
-		.concat(pullRequestKeys)
-		.concat(commitKeys)
+	const allKeys = _.union(branchKeys, pullRequestKeys, commitKeys)
```
joshkay10, line 23: Not sure what the feeling is around _ here, it was available so I used it.... can use straight JS if thats preferred?

This line is to remove duplicates.

src/sync/branches.ts
```diff
@@ -13,6 +13,7 @@ export default async (logger: LoggerWithTarget, github: GitHubAPI, newGithub: Gi
 	logger.info("Syncing branches: started");
 
 	const useNewGHClient = await booleanFlag(BooleanFlags.USE_NEW_GITHUB_CLIENT_FOR_BRANCHES, false, jiraHost);
+	const useNewGHPrUrl = await booleanFlag(BooleanFlags.USE_NEW_GITHUB_PULL_REQUEST_URL_FORMAT, false, jiraHost);
```
joshkay10, line 16: This was placed here instead of inside the transform/branches.ts files so I had access to Jira host, but also there was a bunch of extra conversions down the chain to async just to use the flag..... 
TLDR; was trying to keep the tidy up simpler.

src/sync/transforms/branch.ts
KAbakumovAtl: I don't have an opinion about it. Maybe @mboudreau can suggest?

KAbakumovAtl (approved):

src/transforms/util/pullRequestLinkGenerator.ts
mboudreau: You still need to `encodeURIComponent` of name since a branch name can have `/` in it.

mboudreau: also, don't you need `quick_pull=1` in the querystring to make it a PR right away?  because the first page here is just a compare.  By adding this in the query, you go directly to the create PR page.

src/transforms/util/pullRequestLinkGenerator.ts
mboudreau, line 5: issueKeys should never be `null` if  you're using a default value for it: `issueKeys: string[] = []`

src/sync/branches.ts
```diff
@@ -36,6 +37,6 @@ export default async (logger: LoggerWithTarget, github: GitHubAPI, newGithub: Gi
 
 	return {
 		edges,
-		jiraPayload: transformBranches({ branches, repository })
+		jiraPayload: await transformBranches({ branches, repository, useNewGHPrUrl })
```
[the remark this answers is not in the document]
mboudreau: hm, actually, you could separate each value into it's own attribute with typings to make it more succint.

mboudreau (changes requested):
Noticed something missing in the generated URL

src/transforms/util/pullRequestLinkGenerator.ts
joshkay10: it was to appease typescript for the function issueKeyParser() in src/transforms/pull-request.ts which returns a null | string issuekey set,  but I can handle earlier before this functions called.

src/transforms/util/pullRequestLinkGenerator.ts
joshkay10: I have the return encoding the entire string 
	return  ``${baseUrl}/compare/${name}?title=${title}`;` that should cover it?

joshkay10: appended quick_pull=1 too, can confirm it goes direct to PR.

KAbakumovAtl: But if we encode the entire string we also encode `/` and '?' so the URL will become invalid.

We need to encode each parameter individually. (The reason to encode the string to prevent injection attacks, so nobody will be able to put special characters in URL and make it pointing to another path)

KAbakumovAtl: So it should be
`${ encodeURIComponent(baseUrl)}/compare/${ encodeURIComponent(name)}?title=${ encodeURIComponent(title)}`

joshkay10: I accept, i have encoded each individually

	const title = encodeURIComponent(keys + name);
	const branchName = encodeURIComponent(name);
	const base = encodeURIComponent(baseUrl);
	const url = `${base}/compare/${branchName}?title=${title}&quick_pull=1`

KAbakumovAtl (dismissed):

mboudreau (dismissed):
Looking good

KAbakumovAtl (approved):

PR #1080, opened 2022-02-09
Added histogram metric for Jira client

KAbakumovAtl:
Currently Jira Client sends only old histogram metric. I've added another one with Histogram buckets like we do in other places.

### DIFF
--- a/src/jira/client/axios.ts
+++ b/src/jira/client/axios.ts
@@ -114,6 +114,7 @@ const setRequestStartTime = (config) => {
 export const extractPath = (someUrl = ""): string =>
 	decodeURIComponent(url.parse(someUrl).pathname || "");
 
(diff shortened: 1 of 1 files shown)


src/jira/client/axios.ts
```diff
@@ -133,6 +134,8 @@ const instrumentRequest = (response) => {
 		status: response.status
 	};
 
+	statsd.histogram(metricHttpRequest.jira, requestDurationMs, tags);
+	tags["gsd_histogram"] = RESPONSE_TIME_HISTOGRAM_BUCKETS;
 	statsd.histogram(metricHttpRequest.jira, requestDurationMs, tags);
```
KAbakumovAtl, line 139: Posting the same metric with `gsd_histogram` to produce a bucketed histogram metric. (Like we do for GH client and server side request metrics)

joshkay10 (approved):

PR #1081, opened 2022-02-09
Arc 1042 Added GitHub client timeout

KAbakumovAtl:
In this PR I've added a feature flag to control GitHub Client Timeout

I've made it a number feature flag so we can test which timeout better (30sec vs 10sec)

### DIFF
--- a/src/config/feature-flags.ts
+++ b/src/config/feature-flags.ts
@@ -68,8 +68,6 @@ export const stringFlag = async (flag: StringFlags, defaultValue: string, jiraHo
 export const numberFlag = async (flag: NumberFlags, defaultValue: number, jiraHost?: string): Promise<number> =>
 	Number(await getLaunchDarklyValue(flag, defaultValue, jiraHost));
 
-
-
 export const isBlocked = async (installationId: number, logger: LoggerWithTarget): Promise<boolean> => {
 	try {
 		const blockedInstallationsString = await stringFlag(StringFlags.BLOCKED_INSTALLATIONS, "[]");
--- a/src/config/feature-flags.ts
+++ b/src/config/feature-flags.ts
@@ -28,6 +28,10 @@ export enum StringFlags {
 	BLOCKED_INSTALLATIONS = "blocked-installations"
 }
 
+export enum NumberFlags {
+	GITHUB_CLIENT_TIMEOUT = "github-client-timeout"
+}
+
 const createLaunchdarklyUser = (jiraHost?: string): LDUser => {
 	if (!jiraHost) {
 		return {
@@ -43,7 +47,7 @@ const createLaunchdarklyUser = (jiraHost?: string): LDUser => {
 	};
 };
 
-const getLaunchDarklyValue = async (flag: BooleanFlags | StringFlags, defaultValue: boolean | string, jiraHost?: string): Promise<boolean | string> => {
+const getLaunchDarklyValue = async (flag: BooleanFlags | StringFlags | NumberFlags, defaultValue: boolean | string | number, jiraHost?: string): Promise<boolean | string | number> => {
 	try {
 		await launchdarklyClient.waitForInitialization();
 		const user = createLaunchdarklyUser(jiraHost);
@@ -60,6 +64,11 @@ export const booleanFlag = async (flag: BooleanFlags, defaultValue: boolean, jir
 
 export const stringFlag = async (flag: StringFlags, defaultValue: string, jiraHost?: string): Promise<string> =>
 	String(await getLaunchDarklyValue(flag, defaultValue, jiraHost));
+
+export const numberFlag = async (flag: NumberFlags, defaultValue: number, jiraHost?: string): Promise<number> =>
+	Number(await getLaunchDarklyValue(flag, defaultValue, jiraHost));
+
+
 
 export const isBlocked = async (installationId: number, logger: LoggerWithTarget): Promise<boolean> => {
 	try {
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -5,14 +5,19 @@ import AppTokenHolder from "./app-token-holder";
 import InstallationTokenCache from "./installation-token-cache";
 import AuthToken from "./auth-token";
 import { GetPullRequestParams } from "./types";
-import { handleFailedRequest, instrumentFailedRequest, instrumentRequest, setRequestStartTime } from "./interceptors";
+import {
+	handleFailedRequest,
+	instrumentFailedRequest,
+	instrumentRequest,
+	setRequestStartTime,
+	setRequestTimeout
+} from "./interceptors";
 import { metricHttpRequest } from "../../config/metric-names";
 import { getLogger } from "../../config/logger";
 import { urlParamsMiddleware } from "../../util/axios/url-params-middleware";
 import { InstallationId } from "./installation-id";
 import { GetBranchesQuery, GetBranchesResponse, ViewerRepositoryCountQuery } from "./github-queries";
 import {GithubClientGraphQLError, GraphQLError, RateLimitingError} from "./errors";
-import {numberFlag, NumberFlags} from "../../config/feature-flags";
 
 type GraphQlQueryResponse<ResponseData> = {
 	data: ResponseData;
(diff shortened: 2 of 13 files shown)


src/github/client/github-client.ts
```diff
@@ -31,15 +32,32 @@ export default class GitHubClient {
 	private readonly githubInstallationId: InstallationId;
 	private readonly logger: Logger;
 
-	constructor(
+	public static async build(githubInstallationId: InstallationId,
+		logger: Logger,
+		appTokenHolder: AppTokenHolder = AppTokenHolder.getInstance()) : Promise<GitHubClient> {
+		const timeout = await numberFlag(NumberFlags.GITHUB_CLIENT_TIMEOUT, 30000);
+		return new GitHubClient(githubInstallationId, logger, appTokenHolder, timeout);
```
KAbakumovAtl, line 39: I had to make a builder method here, and made the constructor private.

Because constructor can't be async

src/github/client/github-client.ts
```diff
@@ -31,15 +32,32 @@ export default class GitHubClient {
...
-		appTokenHolder: AppTokenHolder = AppTokenHolder.getInstance()
+		appTokenHolder: AppTokenHolder = AppTokenHolder.getInstance(),
+		timeout: number
 	) {
 		this.logger = logger || getLogger("github.client.axios");
-		this.axios = axios.create({
-			baseURL: githubInstallationId.githubBaseUrl
-		});
+
+
```
KAbakumovAtl, line 50: /nit empty lines

src/config/feature-flags.ts
```diff
@@ -61,6 +65,11 @@ export const booleanFlag = async (flag: BooleanFlags, defaultValue: boolean, jir
 export const stringFlag = async (flag: StringFlags, defaultValue: string, jiraHost?: string): Promise<string> =>
 	String(await getLaunchDarklyValue(flag, defaultValue, jiraHost));
 
+export const numberFlag = async (flag: NumberFlags, defaultValue: number, jiraHost?: string): Promise<number> =>
+	Number(await getLaunchDarklyValue(flag, defaultValue, jiraHost));
+
+
+
```
KAbakumovAtl, line 72: /nit empty lines

src/github/client/interceptors.ts
```diff
@@ -75,6 +77,12 @@ export const instrumentFailedRequest = (metricName) =>
 export const handleFailedRequest = (logger: Logger) =>
 	(error) => {
 		const response = error.response as AxiosResponse;
+
+		if (response?.status === 408 || error.code === "ECONNABORTED") {
```
KAbakumovAtl, line 81: That's the only way to determine if it is a timeout with Axios client (See https://github.com/axios/axios/issues/1174)

src/jira/client/axios.ts
```diff
@@ -133,6 +134,8 @@ const instrumentRequest = (response) => {
 		status: response.status
 	};
 
+	statsd.histogram(metricHttpRequest.jira, requestDurationMs, tags);
+	tags["gsd_histogram"] = RESPONSE_TIME_HISTOGRAM_BUCKETS;
 	statsd.histogram(metricHttpRequest.jira, requestDurationMs, tags);
```
KAbakumovAtl, line 140: That's from another PR. It will go away from here after it merged

KAbakumovAtl:
- [ ] FD Ticket to fill up go/flag-ticket/FD-29853

src/github/client/interceptors.ts
mboudreau: Yes, but no.  This can also be the status/error code for aborting a request from our client side.

You should use the axios specific timeout code of `ETIMEDOUT`.  Upgrade Axios to the latest version (0.25.0) and add a transitional object to the config:
```
transitional: {
    clarifyTimeoutError: true
  },
  ```

This will make axios differentiate between the two.  I believe this will be the default moving forward.

**Review by mboudreau — CHANGES_REQUESTED** [review]
Coding pattern needs to be improved or else we can't change timeout dynamically as well as error handling

File: src/github/client/interceptors.ts (hunk 4 above)
Comment (mboudreau) [reply src/github/client/interceptors.ts:81]: More on this here: https://github.com/axios/axios/blob/master/lib/adapters/xhr.js#L133

There's a lot of talk about this on the project and by using this pattern will help future proof our code for new axios versions.

File: src/github/client/github-client.ts [hunk 6]
```diff
@@ -31,15 +32,32 @@ export default class GitHubClient {
 	private readonly githubInstallationId: InstallationId;
 	private readonly logger: Logger;
 
-	constructor(
+	public static async build(githubInstallationId: InstallationId,
```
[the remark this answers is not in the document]
KAbakumovAtl: Our client is not a long-living entity

We create a new client instance when:
1. We process SQS Message
2. Cann get-github-configuration REST API

That mean the timeout will be active only within a scope of a small operation, which usually takes less than a minute
Hence we don't care if our feature flag change won't be applied immediately.

src/github/client/interceptors.ts
KAbakumovAtl: Oh great, I thought they didn't implement it.

src/github/client/github-client.ts
```diff
@@ -31,15 +32,32 @@ export default class GitHubClient {
...
+		logger: Logger,
+		appTokenHolder: AppTokenHolder = AppTokenHolder.getInstance()) : Promise<GitHubClient> {
+		const timeout = await numberFlag(NumberFlags.GITHUB_CLIENT_TIMEOUT, 30000);
```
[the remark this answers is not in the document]
KAbakumovAtl: In unlikely case if LD is down we want to keep it as if it was rolled out

src/github/client/interceptors.ts
```diff
@@ -18,6 +19,17 @@ export const setRequestStartTime = (config) => {
 	return config;
 };
 
+/**
+ * Sets the timeout to the request based on the github-client-timeout feature flag
+ */
+export const setRequestTimeout = async (config: AxiosRequestConfig): Promise<AxiosRequestConfig> => {
+	const timeout = await numberFlag(NumberFlags.GITHUB_CLIENT_TIMEOUT, 30000);
+	if(timeout) {
```
mboudreau, line 27: This will always return true?

src/sync/pull-request.ts
```diff
@@ -46,7 +46,7 @@ export default async function(
 ) {
 	logger.info("Syncing PRs: started");
 	let status: number;
-	let headers: Headers = {};
```
mboudreau, line 49: Why this change?

src/github/client/github-client.ts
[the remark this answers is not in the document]
mboudreau: @KAbakumovAtl Oh.  That's not good.  The whole point of having a client is that it's long lived.  It can better manage connections or have the ability to abort old connections.  It could also potentially create a memory leak if you just keep creating it or at the very least the GC will keep getting hammered.

src/github/client/github-client.ts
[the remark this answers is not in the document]
mboudreau: Shouldn't we keep it to 60s?  That's usually the default HTTP timeout.

src/github/client/interceptors.ts
mboudreau: unless you return `0` as a number flag, which I really doubt would happen

KAbakumovAtl: Nope, timeout can be 0

src/sync/pull-request.ts
KAbakumovAtl: Forgot to mention. With update of the library the Headers interface become incompatible with the type of the headers from the AxiosResponse.

Do you have any ideas on how to make it better?

src/github/client/interceptors.ts
mboudreau: Wouldn't it just return the defaultValue when it's off?

mboudreau: You're also not checking if the config has a value already that's been explicitly set.  I'd probably go with the approach of `config.timeout = config.timeout || await numberFlag(NumberFlags.GITHUB_CLIENT_TIMEOUT, 60000);`

src/sync/pull-request.ts
mboudreau: Change the `Headers` interface to extend `AxiosResponseHeaders`?

src/github/client/interceptors.ts
KAbakumovAtl: No the defaultValue will be returned when LaunchDarkly is offline. Otherwise LD will return the value which we set in LD when targeting is OFF. 

>You're also not checking if the config has a value already that's been explicitly set.
We don't set it explicitly in requests so far. Doesn't make sense to add this check now. We'll get rid of the interceptor as soon as we clean up the flag.

mboudreau (approved):

PR #1085, opened 2022-02-11
ARC-672 routes folder structure

mboudreau:
This is the first PR of many to improve the folder structure of the project.  First up is the express routes to make it much easier to navigate through code and the paths.


src/jira/util/jwt.ts
```diff
@@ -75,7 +75,7 @@ function decodeAsymmetricToken(token: string, publicKey: string, noVerify: boole
 
 
 function verifyQsh(qsh: string, req: Request): boolean {
-	const requestInAtlassianJwtFormat = {...req, pathname: req.path}
```
mboudreau, line 78: This completely broke because `req.path` is not the correct variable to use as it's only the router specific path, not the full path.

package.json
```diff
@@ -68,18 +68,18 @@
     "atlassian-jwt": "^2.0.0",
     "aws-sdk": "^2.999.0",
     "axios": "^0.25.0",
-    "body-parser": "^1.18.3",
+    "body-parser": "^1.19.1",
```
KAbakumovAtl, line 71: I wonder how does the library version bump help the folder structure?

mboudreau: I was getting eslint resolver issues and there was an issue with the express-validator that I needed to upgrade to the latest to use it appropriately for the new routing. While I was here, I saw that there were a few things related to routing/express that needed to upgrade and figured if we're testing all these routes, it would make sense to update these packages instead of updating them and retesting all the routes anyways.

package.json
```diff
@@ -26,7 +26,7 @@
     "lint": "run-p lint:*",
     "lint:ts": "eslint . --ext .ts,.tsx",
     "lint:yaml": "yamllint github-for-jira.sd.yml",
-    "pretest": "run-p setup:test db:test",
+    "pretest": "run-p clean setup:test db:test",
```
mboudreau, line 29: Should clean built files before doing tests

src/api/index.ts
```diff
@@ -1,424 +0,0 @@
-import express, { NextFunction, Request, Response } from "express";
```
mboudreau, line 1: This file was broken into several under src/routes/api.

src/app.ts
```diff
@@ -1,17 +1,44 @@
...
+	app.set("trust proxy", true);
+	app.set("case sensitive routing", false);
+	app.set("view engine", "hbs");
+	const viewPath = path.resolve(process.cwd(), "views");
+	app.set("views", viewPath);
+	registerHandlebarsPartials(path.resolve(viewPath, "partials"));
+	registerHandlebarsHelpers();
+
+	// Add github client middleware which uses the octokit app
+	app.use(getGithubClientMiddleware(octokitApp));
```
mboudreau, line 33: We have to have this here for now as it uses the octokit app to get github clients and such.  This will change in the future as we more away from probot/ocotkit.

src/app.ts
```diff
@@ -62,19 +89,17 @@ function secureHeaders(router: Router, frontendApp: Express) {
 	router.use(helmet.hidePoweredBy());
 }
 
-export default (app: Application): void => {
+export const setupFrontend = (app: Application): void => {
```
mboudreau, line 92: using named export as part of coding convention.

src/frontend/app.ts
```diff
@@ -1,343 +0,0 @@
-import bodyParser from "body-parser";
```
mboudreau, line 1: This file used to be the top level router, which is now multiple hierarchical routers in src/routes

src/frontend/github-oauth.ts
```diff
@@ -1,187 +0,0 @@
-import crypto from "crypto";
```
mboudreau, line 1: not sure why this was considered deleted as it was moved to /src/routes/github, but changed a bit to be it's own router

src/interfaces/common.ts
```diff
@@ -8,3 +8,16 @@ export enum BooleanEnum {
...
+	// eslint-disable-next-line @typescript-eslint/no-namespace
+	namespace Express {
+		interface Request {
+			session: {
+				jiraHost?: string;
+				githubToken?: string;
+			};
+		}
+	}
+}
```
mboudreau, line 23: This used to be in src/frontend/app.ts, but since it's now being used by multiple files in the new route structure, I moved it into a more reusable area.

src/jira/connect.ts
```diff
@@ -1,111 +0,0 @@
-import { Request, Response } from "express";
```
mboudreau, line 1: Not sure why this is considered deleted, it was moved to src/routes/jira

src/middleware/cookiesession-middleware.ts
```diff
@@ -0,0 +1,12 @@
+// setup route middlewares
+import cookieSession from "cookie-session";
```
mboudreau, line 2: Moving all middlewares to this folder instead of being in routers

tsconfig.json
```diff
@@ -20,7 +20,28 @@
...
+				"src/util"
+			],
+			"config": [
+				"src/config"
+			],
+			"interfaces": [
+				"src/interfaces"
+			],
+			"services": [
+				"src/services"
```
mboudreau, line 42: Since we're redoing the folder structure, might as well add paths to make it easier to get to the imported files. 

Instead of doing `../../../../routes/api/api-router.ts` you can simply do `routes/api/api-router.ts`

joshkay10 (dismissed):

joshkay10 (approved):

PR #1088, opened 2022-02-15
ARC-946 fix janky flow

rachellerathbone:
**Current flow for Jira install** - in case you need a refresher (prod): https://www.loom.com/share/3d7a4caba70f4a32a7268b0ce372f4da

**Updated flow for Jira install** when user wants to install the app in a new org and uses the Jira host stored in a cookie: https://www.loom.com/share/f9fd0820ff91404c8521fea22b242eea

**Jira install flow but user wants to enter a different Jira site:** https://www.loom.com/share/59a2bd11a9e448be8fa2409da835f491

**GH install flow** - we have no data in the cookie: https://www.loom.com/share/e786f65b6b6b49068ed9a34dbea3c554

### DIFF
--- a/src/frontend/app.ts
+++ b/src/frontend/app.ts
@@ -199,6 +199,7 @@ export default (octokitApp: App): Express => {
 	app.get(
 		"/github/setup",
 		csrfProtection,
+		oauth.checkGithubAuth,
 		getGitHubSetup
 	);
 
--- a/src/frontend/get-github-setup.ts
+++ b/src/frontend/get-github-setup.ts
@@ -14,6 +14,7 @@ export default async (req: Request, res: Response): Promise<void> => {
 	req.log.info("Received get github setup page request");
 	const { jiraHost } = res.locals;
 
+	const installationId = req.originalUrl.split("=")[1].split("&")[0]
 	let redirectUrl = getJiraMarketplaceUrl(jiraHost);
 
 	// If we know enough about user and site, redirect to the app
@@ -21,10 +22,21 @@ export default async (req: Request, res: Response): Promise<void> => {
 		redirectUrl = getJiraAppUrl(jiraHost);
 	}
 
+	let installation;
+
+	if (jiraHost) {
+		installation = await Installation.getForHost(jiraHost);
+	}
+
+	const hasJiraHost = !!jiraHost;
+
 	res.render("github-setup.hbs", {
 		csrfToken: req.csrfToken(),
 		nonce: res.locals.nonce,
 		jiraHost,
-		redirectUrl
+		redirectUrl,
+		hasJiraHost,
+		clientKey: installation?.clientKey,
+		id: installationId
 	});
 };
--- a/src/frontend/get-github-setup.ts
+++ b/src/frontend/get-github-setup.ts
@@ -12,7 +12,7 @@ import { Installation } from "../models";
 */
 export default async (req: Request, res: Response): Promise<void> => {
 	req.log.info("Received get github setup page request");
-	const { jiraHost, isAdmin } = res.locals;
+	const { jiraHost } = res.locals;
 
 	const installationId = req.originalUrl.split("=")[1].split("&")[0]
 	let redirectUrl = getJiraMarketplaceUrl(jiraHost);
@@ -20,16 +20,9 @@ export default async (req: Request, res: Response): Promise<void> => {
 	const {	github, client } = res.locals;
 
 	const { data: { installations } } = await github.apps.listInstallationsForAuthenticatedUser();
-	const { data: { login } } = await github.users.getAuthenticated();
 	const { data: info } = await client.apps.getAuthenticated();
 
 	const installation = installations.filter((item) => item.id === Number(installationId));
-
-	const admin = await isAdmin({
-		org: installation[0]?.account?.login,
-		username: login,
-		type: installation[0]?.target_type || ""
-	});
 
 	// If we know enough about user and site, redirect to the app
 	if (jiraHost && await jiraSiteExists(jiraHost) && await Installation.getForHost(jiraHost)) {
@@ -54,7 +47,6 @@ export default async (req: Request, res: Response): Promise<void> => {
 		orgName: installation[0]?.account?.login,
 		avatar: installation[0]?.account?.avatar_url,
 		html_url: info.html_url,
-		admin,
 		id: installationId // can't use this is they switch (reset id and send to marketplace)
 	});
 };
--- a/src/frontend/get-github-setup.ts
+++ b/src/frontend/get-github-setup.ts
@@ -15,17 +15,22 @@ import { getCloudInstallationId } from "../github/client/installation-id";
 */
 export default async (req: Request, res: Response): Promise<void> => {
 	req.log.info("Received get github setup page request");
-	const { jiraHost, client } = res.locals;
-	const installationId = Number(req.query.installation_id);
+	const { jiraHost, client, github: oldGitHub } = res.locals;
+
+	const installationId = 17381973;
+	// const installationId = Number(req.query.installation_id);
 	const github = new GitHubClient(getCloudInstallationId(installationId), req.log);
-	const { data: installations } = await github.listInstallationsForAuthenticatedUser();
+	req.log.info("HERE")
+	const { data: { installations: oldInstallations } } = await oldGitHub.apps.listInstallationsForAuthenticatedUser();
+	const { data: installations } = await github.listInstallationsForAuthenticatedUser(installationId);
 
+	req.log.info("installations: ", installations, "OLD:" , oldInstallations)
 	const { data: info } = await client.apps.getAuthenticated();
-	const installation = installations.filter((item) => item.id === Number(installationId));
+	// const installation = installations.data.find((item) => item.id === Number(installationId));
 
-	if (installation.length === 0) {
-		throw new Error(`Error retrieving installation:${installationId}. App not installed on org.`);
-	}
+	// if (!installation) {
+	// 	throw new Error(`Error retrieving installation:${installationId}. App not installed on org.`);
+	// }
 
 	let redirectUrl = getJiraMarketplaceUrl(jiraHost);
 
@@ -41,8 +46,8 @@ export default async (req: Request, res: Response): Promise<void> => {
 	}
 
 	const hasJiraHost = !!jiraHost;
-	const { account } = installation[0];
-	const { login, avatar_url } = account;
+	// const { account } = installation;
+	// const { login, avatar_url } = account;
 
 	res.render("github-setup.hbs", {
 		csrfToken: req.csrfToken(),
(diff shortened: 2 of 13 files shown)


test/unit/frontend/github-setup.test.ts
```diff
@@ -22,27 +22,53 @@ describe("Github Setup", () => {
...
+		beforeEach(async () => {
+			await Installation.create({
+				jiraHost,
+				clientKey: "abc123",
+				secrets: "def234",
+				sharedSecret: "ghi345",
+			});
+		});
+
+		it("should return redirect to github oauth flow for GET request if token is missing", async () =>
```
rachellerathbone, line 34: I had to update the tests for 'GET' as we are now passing `oauth.checkGithubAuth` to the github/setup endpoint for get requests. @mboudreau would love your guidance here. Is there anything else I should be testing? Do I need to change these/

src/frontend/app.ts
```diff
@@ -199,6 +199,7 @@ export default (octokitApp: App): Express => {
 	app.get(
 		"/github/setup",
 		csrfProtection,
+		oauth.checkGithubAuth,
```
KAbakumovAtl, line 202: I wonder why it was missing and what was the effect?

src/frontend/get-github-setup.ts
```diff
@@ -14,17 +14,39 @@ export default async (req: Request, res: Response): Promise<void> => {
 	req.log.info("Received get github setup page request");
 	const { jiraHost } = res.locals;
 
+	const installationId = req.originalUrl.split("=")[1].split("&")[0]
```
KAbakumovAtl, line 17: We shouldn't parse path parameters from URL manually, because express has functionality for it.
Moreover we shouldn't rely that the installationId will be passed first.

Instead we should do:
`req.query.installationId` (where installationId is the name of the query parameter)

src/frontend/get-github-setup.ts
```diff
@@ -14,17 +14,39 @@ export default async (req: Request, res: Response): Promise<void> => {
 	req.log.info("Received get github setup page request");
```
KAbakumovAtl, line 14: I wonder if we should update the method readme? It says i handles redirects from Jira, however it doesn't seem so (according to videos)

src/frontend/get-github-setup.ts
```diff
@@ -14,17 +14,39 @@ export default async (req: Request, res: Response): Promise<void> => {
...
 
+	const installationId = req.originalUrl.split("=")[1].split("&")[0]
 	let redirectUrl = getJiraMarketplaceUrl(jiraHost);
 
+	const {	github, client } = res.locals;
+
+	const { data: { installations } } = await github.apps.listInstallationsForAuthenticatedUser();
+	const { data: info } = await client.apps.getAuthenticated();
+
+	const installation = installations.filter((item) => item.id === Number(installationId));
```
KAbakumovAtl, line 25: 1. Can we get the exact installation ID from GitHub rather than filtering them manually?

2. Should we use new GitHub Client instead of Octokit? (Maybe not now, because it might require some middleware changes)

KAbakumovAtl: - [ ] Query the exact installationId from github

src/frontend/get-github-setup.ts
KAbakumovAtl: - [ ] use query parameters

src/frontend/app.ts
rachellerathbone: It wasn't 'missing' before, it just wasn't needed. Now that we are connecting the org from this page, we need to pass this middleware.

src/frontend/get-github-setup.ts
rachellerathbone: Good catch.

src/frontend/get-github-setup.ts
rachellerathbone: 1. Not that I could find. I was hoping GH had an API that I could call with the installationId but there doesn't appear to be one. Getting all the orgs where the app was installed appears to be the only option.

2. Yes, yes we should 😄 Will update.

test/unit/frontend/github-setup.test.ts
```diff
@@ -22,27 +22,53 @@ describe("Github Setup", () => {
...
-					expect(res.status).toBe(200);
-					expect(res.headers["content-type"]).toContain("text/html");
+				.set(
+					"Cookie",
+					getSignedCookieHeader({
+						jiraHost,
+					})
+				)
+				.expect((res) => {
+					expect(res.status).toBe(302);
```
KAbakumovAtl, line 44: I am curious how does redirect happen? I see that /github/setup endpoint code always returns the handlebars template

(Does it happen somewhere in interceptors?)

src/frontend/get-github-setup.ts
KAbakumovAtl: Interesting, is it possible to get the org ID from our DB somehow and then query by Org?

src/frontend/get-github-setup.ts
KAbakumovAtl: Also, we need to make sure that it won't be possible to pass a wrong InstallationID and then connect other's org. (I guess it won't be because user needs an access to that org, right?)

src/frontend/get-github-setup.ts
```diff
@@ -14,17 +14,39 @@ export default async (req: Request, res: Response): Promise<void> => {
...
 	res.render("github-setup.hbs", {
 		csrfToken: req.csrfToken(),
 		nonce: res.locals.nonce,
 		jiraHost,
-		redirectUrl
+		redirectUrl,
+		hasJiraHost,
+		clientKey: jiraInstallation?.clientKey,
+		orgName: installation[0]?.account?.login,
+		avatar: installation[0]?.account?.avatar_url,
```
thombergs, line 48: Naming an array in singular is a bit odd. I'd also rather assign the installation to a constant instead of reading it from the error twice here. Then you can also throw an error if the array is empty for whatever reason and remove the `?` here, because is it really a valid state when the `installation` is `undefined`?

thombergs (commented):
Great videos! Makes it easy to grok the impact of your changes.

src/frontend/get-github-setup.ts
rachellerathbone: Good call out. Can't connect but it currently fails silently. Will need to chat with Jade to see how we want to display this in the UI.

src/frontend/get-github-setup.ts
rachellerathbone: Just looked. It says it handles redirects specifically for the installation flow for Jira and GH. However, the rest of the comments are outdated so have amended those.

src/frontend/get-github-setup.ts
rachellerathbone: Not at this point in the flow. The installations table doesn't have this data and, at this point, no subscription has been made (happens after connect is clicked).

rachellerathbone: https://github.com/atlassian/github-for-jira/pull/1088#discussion_r808646896 I don't think there is a way to do this.

test/unit/frontend/github-setup.test.ts
rachellerathbone: It's related to the middleware we are now passing to github/setup and the GH token check.

src/frontend/get-github-setup.ts
mboudreau: Yeah, installationId is not a first class citizen apparently. I can't find any REST or GraphQL APIs that we can use to get it.  kinda stupid really.

test/unit/frontend/github-setup.test.ts
rachellerathbone: Phew!

static/js/jira-configuration.js
```diff
@@ -16,7 +16,7 @@ function openChildWindow(url) {
 
 $(".add-organization-link").click(function(event) {
 	event.preventDefault();
-	window.AP.context.getToken(function(token) {
+	window.AP.context.getToken(function() {
```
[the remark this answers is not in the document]
KAbakumovAtl: Should we add context token authentication on that endpoint?

rachellerathbone:
https://github.com/atlassian/github-for-jira/pull/1088#pullrequestreview-886675724

@mboudreau Unfortunately, for both the Jira install flow where a user chooses to enter an alternate jiraHost and the GH install flow (where we don't have the host in the cookie) our only option is to direct the user to marketplace. From there, they either install the app or click get started (or whatever the button says if the app is already installed). This is something we talked to the marketplace team about last year when we were improving the flow. Ideally, we wanted to make a request to Jira and to check if the app was installed on a given host, if so, direct to the config page, if not, direct to marketplace. I can't remember the exact reasoning but the guy I spoke to said it was something to do with security and that all customers would need to go to marketplace whether the app was installed or not.

src/frontend/get-github-setup.ts
```diff
@@ -5,14 +5,24 @@ import { Installation } from "../models";
...
-			- If we have the users Jira host, redirect to marketplace.
-			- Otherwise, render the setup page.
+		- If we have the users Jira host, redirect to marketplace.
+		- Otherwise, render the setup page (POST).
 */
 export default async (req: Request, res: Response): Promise<void> => {
 	req.log.info("Received get github setup page request");
-	const { jiraHost } = res.locals;
+	const { jiraHost, github, client } = res.locals;
+	const { data: { installations } } = await github.apps.listInstallationsForAuthenticatedUser();
```
rachellerathbone, line 18: Looked into using our GH client but it needs to be updated to support user tokens. I've created a ticket in the backlog for us to address this and to update this afterwards.

static/js/jira-configuration.js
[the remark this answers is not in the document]
mboudreau: @KAbakumovAtl how would you suggest we pass it over to the new url?  This is the issue we had eliminated with link jacking since the JWT token was being passed in the querystring.

src/frontend/get-github-setup.ts
```diff
@@ -5,14 +5,24 @@ import { Installation } from "../models";
...
+	const { data: info } = await client.apps.getAuthenticated();
+	const installationId = req.query.installation_id;
+	const installation = installations.find((item) => item.id === Number(installationId));
```
mboudreau, line 21: /nit I'd probably add the casting to a number when retrieving value from the query instead of in the find: `const installationId = Number(req.query.installation_id);`

mboudreau (approved):

static/js/jira-configuration.js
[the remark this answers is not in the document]
KAbakumovAtl: It is perfectly fine to have JWT token in a query string, because it has expiration window. (It is quite a common practice to place them in query parameters)

KAbakumovAtl (approved):

PR #1089, opened 2022-02-15
ARC-1085: delete builds and deployments on uninstallation

thombergs:
So far, when we uninstall the app or remove an org, we have been deleting the commits, branches, and PRs from Jira, but not the builds and deployments. This adds deletion for the builds and deployments.


src/jira/client/index.ts
```diff
@@ -153,13 +153,22 @@ async function getJiraClient(
...
 						}
 					),
-				delete: (gitHubInstallationId: string | number) =>
-					instance.delete(`/rest/devinfo/0.10/bulkByProperties`,
-						{
-							params: {
-								installationId: gitHubInstallationId
-							}
-						})
+				delete: async (gitHubInstallationId: string | number) =>
```
thombergs, line 156: This `delete` function is under `devinfo`, so it's technically not the right place to add calls to the `builds` and `deployments` APIs, but I didn't want to restructure the client as it's not typed throughout the whole codebase...

src/jira/client/index.ts
```diff
@@ -153,13 +153,22 @@ async function getJiraClient(
...
+					Promise.all([
+
+						// We are sending devinfo events with the property "installationId", so we delete by this property.
+						instance.delete(
+							`/rest/devinfo/0.10/bulkByProperties?installationId=${gitHubInstallationId}`
+						),
+
+						// We are sending builds and deployments with the property "gitHubInstallationId", so we delete by this property.
+						instance.delete(
+							`/rest/builds/0.1/bulkByProperties?gitHubInstallationId=${gitHubInstallationId}`
```
thombergs, line 166: Note that the property name `gitHubInstallationId` and not `installationId` like it is when we're deleting devinfo. That's because we're sending builds and deployments with the property `gitHubInstallationId`. See lines 244 and 268.

src/jira/client/index.ts
mboudreau: yeah, this file need restructuring anyways.  Need a jira client equivalent of our github client.

src/jira/client/index.ts
```diff
@@ -153,13 +153,22 @@ async function getJiraClient(
...
 					),
-				delete: (gitHubInstallationId: string | number) =>
-					instance.delete(`/rest/devinfo/0.10/bulkByProperties`,
-						{
-							params: {
-								installationId: gitHubInstallationId
-							}
-						})
+				delete: async (gitHubInstallationId: string | number) =>
+					Promise.all([
```
mboudreau, line 157: this needs to be either returned or awaited.

src/jira/client/index.ts
```diff
@@ -153,13 +153,22 @@ async function getJiraClient(
...
+
+						// We are sending devinfo events with the property "installationId", so we delete by this property.
+						instance.delete(
+							`/rest/devinfo/0.10/bulkByProperties?installationId=${gitHubInstallationId}`
```
mboudreau, line 161: don't use string templating for URL. Use the url as a base, but all variable needs to go either through `params` or `urlParams` (in this case the former)

mboudreau (changes requested):
small changes :)

src/jira/client/index.ts
mboudreau: nevermind... side by side view cut off the arrow.

thombergs: it _is_ returned due to the `=>` notation, is it not?

mboudreau (approved):

atrigueiro (approved):

PR #1093, opened 2022-02-17
ARC-810: Emit analytics for config page

atrigueiro:
We want to start sending a `screen` event to Redash when a customer lands on the configuration page.
This metric will contain a `connectedOrgCount` attribute, which will tell us if that customer has any org installed from GitHub. 
The count of this metric will also give us an idea of how big is the drop from customers installing the app on Marketplace and never reaching the post config page.


.github/workflows/on-push.yml
```diff
@@ -15,7 +15,7 @@ jobs:
         with:
           path: ./node_modules
           key: ${{ runner.os }}-${{hashFiles('**/package-lock.json')}}-node_modules
-      - run: npm install
+      - run: npm ci --no-optional
```
atrigueiro, line 18: We need this flag because we're adding `@atlassiansox/analytics-node-client` - see `package.json` below, which is a private package to P.A.C. and GitHub actions falls back to the public npmregistry, which fails our build.

src/interfaces/common.ts
```diff
@@ -8,3 +8,17 @@ export enum BooleanEnum {
...
+	GitHubConfigScreenEventName = "githubConfigurationScreen",
+	ScreenEvent = "screen",
+	UiEvent = "ui",
+	TrackEvent = "track",
+	OperationalEvent = "operational",
+	TraitEvent = "trait",
+	ProdEnv = "prod",
+	StgEnv = "stg",
+	DevEnv = "dev"
+}
```
mboudreau, line 24: bad standard to have.  Enums should be succinct to a specific context. You should have multiple enums here.  Enums are also not to be used as constants, it's more for user/function inputs so you can help validate with compilation.

mboudreau (changes requested):
a few things to change, but great first draft :)

src/frontend/analytics-client.ts
```diff
@@ -0,0 +1,97 @@
...
+function sendAnalytics(eventType: "trait")
+function sendAnalytics(eventType: "screen", attributes:{name:string} & Record<string,unknown>)
+function sendAnalytics(eventType: "ui" | "track" | "operational", attributes: Record<string,unknown>)
+async function sendAnalytics(eventType: string, attributes?: Record<string, unknown>): Promise<any> {
+	if(!analyticsClient){
+		return;
+	}
+
+	const analyticsNodeClient = analyticsClient({
+		env: AnalyticsVariablesEnum.ProdEnv, // prod, stg or dev
```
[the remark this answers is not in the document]
atrigueiro: We only use prod Jira instances though, so when sending analytics we're actually sending to their prod env. We can filter down to which of our apps is sending when checking the `appKey` on the attributes column. But you're right, we don't need the enum bc we're always sending `prod` as the value there. And we also don't want to send for local envs, I'll add that `!isProdEnv` check back.

src/frontend/analytics-client.ts
```diff
@@ -0,0 +1,97 @@
...
+		product: AnalyticsVariablesEnum.Product, // required - do not change - https://hello.atlassian.net/browse/DE-8853
```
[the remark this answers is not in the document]
atrigueiro: I fear that, if we don't make it an enum, someone could change it to `production` instead of `prod` by mistake, which wouldn't work.

src/frontend/analytics-client.ts
```diff
@@ -0,0 +1,97 @@
...
+					}
+				}
+			}))
+			break;
+		case "trait":
+			wrapPromises(analyticsNodeClient.sendTraitEvent({
+				entityType: attributes?.entityType,
+				entityId: attributes?.entityId,
+				entityTraits: attributes?.entityTraits
+			}))
```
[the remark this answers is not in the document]
atrigueiro: It does not (source code [here]( https://bitbucket.org/atlassian/analytics-node-client/src/9aecc3e3c71ffa26138254479e03b946a82b6737/src/client.js#lines-323))

src/interfaces/common.ts
```diff
@@ -9,6 +9,27 @@ export enum BooleanEnum {
 	false = "false",
 }
 
+// All variables below were defined by DataPortal. Do not change their values as it will affect our metrics logs and dashboards.
+export enum AnalyticsConfigEnum {
+	Product = "gitHubForJira", // required - do not change - https://hello.atlassian.net/browse/DE-8853
+	ProdEnv = "prod"
+}
```
mboudreau, line 16: If the product is a constant, don't make it an enum.  That doesn't really make any sense.  It's only used when creating the client, just use it inline.

Also, `ProdEnv` is not correct here. It should just be `env` as the analytics could be be used for prod or staging or ddev.

src/util/analytics-client.ts
```diff
@@ -0,0 +1,91 @@
...
+function sendAnalytics(eventType: "ui" | "track" | "operational", attributes: Record<string,unknown>)
+function sendAnalytics(eventType: string, attributes?: Record<string, unknown>): void {
+	if(!analyticsClient && !isNodeProd()){
+		return;
+	}
+
+	const analyticsNodeClient = analyticsClient({
+		env: AnalyticsConfigEnum.ProdEnv,
+		product: AnalyticsConfigEnum.Product
+	});
```
mboudreau, line 23: You really should need to create your analytics client on every `sendAnalytics` call.  You should have a check to see if it's already created and have a variable outside the function, like 
```
let analyticsNodeClient:AnalyticsClient;

function sendAnalytics(eventType: string, attributes?: Record<string, unknown>): void {
	if(!analyticsClient && !isNodeProd()){
		return;
	}

  if(!analyticsNodeClient) {
	  analyticsNodeClient = analyticsClient({
		  env: envVars.MICROS_ENV,
		  product: "gitHubForJira"
	  });
  }
...
```

src/util/analytics-client.ts
```diff
@@ -0,0 +1,91 @@
...
+						appKey
+					}
+				}
+			}))
+			break;
+		case "ui":
+			wrapPromises(analyticsNodeClient.sendUIEvent({
+				...baseAttributes,
+				uiEvent: {
+					attributes: attributes
```
mboudreau, line 52: /nit don't need to do `attributes: atributes`, eslint will complain about this.  Just a single `attributes` will do the key and the value :)

src/util/analytics-client.ts
```diff
@@ -0,0 +1,91 @@
...
+				}
+			}))
+			break;
+		default:
+			logger.warn(`Cannot sendAnalytics: unknown eventType`)
+			break;
+	}
+}
+
+function wrapPromises(promise: Promise<unknown>) {
```
mboudreau, line 85: It's just a single promise

src/util/analytics-client.ts
```diff
@@ -0,0 +1,91 @@
...
+		case "screen":
+			wrapPromises(analyticsNodeClient.sendScreenEvent({
+				...baseAttributes,
+				name: attributes?.name,
+				screenEvent: {
+					platform: "web",
+					attributes: {
+						...omit(attributes, "name"),
+						appKey
+					}
```
mboudreau, line 44: don't need the app key here as it's already added to attributes above. just `attributes: omit(attributes, "name")` should do.

mboudreau (commented):
A few things to change.  I'm pretty sure I've mentioned these before...

src/util/analytics-client.ts
atrigueiro: That wouldn't work as `!analyticsNodeClient` would've returned the function...

mboudreau (approved):

PR #1094, opened 2022-02-17
fixing JWT QSH verification to use the correct request parameter for full path

mboudreau:
JWT verification was using `req.path` which is not the full path but the local router path.  If you want the full path, you need to concatenate `req.baseUrl + req.path` to get the full path.

### DIFF
--- a/test/unit/jira/util/jwt.test.ts
+++ b/test/unit/jira/util/jwt.test.ts
@@ -1,174 +1,155 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
-import {
-	TokenType,
-	verifyAsymmetricJwtTokenMiddleware,
-	verifySymmetricJwtTokenMiddleware
-} from "../../../../src/jira/util/jwt";
-import logger from "../../../../src/config/logger"
-import {AsymmetricAlgorithm, encodeAsymmetric, encodeSymmetric} from "atlassian-jwt";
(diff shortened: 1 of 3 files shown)


test/unit/jira/util/jwt.test.ts
```diff
@@ -1,174 +1,155 @@
...
-		query: testQueryParams,
-		method: testRequestMethod,
-		path: testRequestPath,
-		session: {
-			jiraHost: "https://test.atlassian.net"
+	let testQueryParams: Record<string, string>;
+	let res: Response;
+	let baseRequest: Request;
+
+	let next:Mock;
```
yvonnevictoria, line 18: /nit space

yvonnevictoria (approved):

joshkay10 (approved):

test/unit/jira/util/jwt.test.ts
```diff
@@ -1,174 +1,155 @@
...
+		// TODO: need to create a route with express Router and use supertest to actually
+		// test the real flow instead of mocking everything based on the assumption that
+		// we know exactly how express is going to behave
 		res = {
 			locals: {},
 			status: jest.fn().mockReturnValue(res),
 			json: jest.fn().mockReturnValue(res)
+		} as any;
+
+		// TODO: need to remove all references to 'kabakumov' in this file, just use jiraHost instead
```
KAbakumovAtl, line 73: Agree

KAbakumovAtl (approved):
🟢 Approved.

The only thing I'd like to ask is to test it in ddev/stg

PR #1095, opened 2022-02-17
ARC-1089 replace GH Client on Deployments path

KAbakumovAtl:
Replace Octokit with new GitHub Client for the deployment queue

### DIFF
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -187,6 +187,22 @@ export default class GitHubClient {
 		});
 	};
 
+	public listDeployments = async (owner: string, repo: string, environment: string, per_page: number ): Promise<AxiosResponse<Octokit.ReposListDeploymentsResponse>> => {
+		return await this.get<Octokit.ReposListDeploymentsResponse>(`/repos/{owner}/{repo}/deployments`,
+			{environment,
+				per_page},
+			{owner,
+				repo});
+	}
+
+	public listDeploymentStatuses = async (owner: string, repo: string, deployment_id: number, per_page: number) : Promise<AxiosResponse<Octokit.ReposListDeploymentStatusesResponse>> => {
+		return await this.get<Octokit.ReposListDeploymentStatusesResponse>(`/repos/{owner}/{repo}/deployments/{deployment_id}/statuses`,
+			{per_page},
+			{owner,
+				repo,
+				deployment_id});
+	}
+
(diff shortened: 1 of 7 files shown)


src/sqs/deployment.ts
```diff
@@ -16,18 +18,30 @@ export type DeploymentMessagePayload = {
...
 	const messagePayload: DeploymentMessagePayload = context.payload;
 
-	const github = await app.auth(messagePayload.installationId);
+	const {webhookId, jiraHost, installationId} = messagePayload;
+
+	context.log = context.log.child({
+		webhookId,
+		jiraHost,
+		installationId
+	})
```
KAbakumovAtl, line 31: Apologise for the adjacent change here, but that's something I've noticed during the yesterday's incident. 

The idea here is put webhookId and installationId into the context logger, so these parameters will be logged in case of the message failure. (Otherwise we can't filter errors by installation Id)

src/transforms/deployment.ts
```diff
@@ -7,25 +7,31 @@ import { LoggerWithTarget } from "probot/lib/wrap-logger";
...
+import GitHubClient from "../github/client/github-client";
+import {AxiosResponse} from "axios";
 
 // https://docs.github.com/en/rest/reference/repos#list-deployments
 async function getLastSuccessfulDeployCommitSha(
 	owner: string,
 	repoName: string,
 	github: GitHubAPI,
+	newGitHubClient: GitHubClient,
+	useNewClient: boolean,
```
KAbakumovAtl, line 19: Using boolean method parameters is an anti-pattern. 

However now I am doing it to minimise the amount of changes for the PR. Will be gone anyway with the feature flag cleanup.

src/github/client/github-client.ts
```diff
@@ -187,6 +187,22 @@ export default class GitHubClient {
 		});
 	};
 
+	public listDeployments = async (owner: string, repo: string, environment: string, per_page: number ): Promise<AxiosResponse<Octokit.ReposListDeploymentsResponse>> => {
+		return await this.get<Octokit.ReposListDeploymentsResponse>(`/repos/{owner}/{repo}/deployments`,
+			{environment,
+				per_page},
+			{owner,
+				repo});
```
KAbakumovAtl, line 195: Indentation looked much better in IDEA :(

KAbakumovAtl:
TODO Fill up the flag

go/flag-ticket/FD-30143

joshkay10 (approved):

PR #1097, opened 2022-02-17
Arc 997 - Replace OctoKit with Github client for the discovery queue

joshkay10:
In Github Client, created a new Github client function `getRepositoriesPage` which is a graphql query to get a page of repos for the logged in org.

In src/sync/discovery.ts added a feature flag to use Octokit or new github client.


src/sync/discovery.ts
```diff
@@ -3,22 +3,54 @@ import { getRepositorySummary } from "./jobs";
...
 import { sqsQueues } from "../sqs/queues";
+import GitHubClient from "../github/client/github-client";
+import { getCloudInstallationId } from "../github/client/installation-id";
+import { Repository } from "@octokit/graphql-schema";
+import { booleanFlag, BooleanFlags } from "../config/feature-flags";
 
 export const DISCOVERY_LOGGER_NAME = "sync.discovery";
+const MAX_PAGE_SIZE_REPOSITORY = 100;
+
+const getAllRepositories = async (github, hasNextPage: boolean, cursor?: string, repositories: Repository[] = []): Promise<Repository[]> => {
```
joshkay10, line 16: Originally wrote this inside a loop, but found this solution read a lot cleaner to me.....

src/sync/discovery.ts
```diff
@@ -3,22 +3,54 @@ import { getRepositorySummary } from "./jobs";
...
-import {LoggerWithTarget} from "probot/lib/wrap-logger";
+import { LoggerWithTarget } from "probot/lib/wrap-logger";
 import { sqsQueues } from "../sqs/queues";
+import GitHubClient from "../github/client/github-client";
+import { getCloudInstallationId } from "../github/client/installation-id";
+import { Repository } from "@octokit/graphql-schema";
+import { booleanFlag, BooleanFlags } from "../config/feature-flags";
 
 export const DISCOVERY_LOGGER_NAME = "sync.discovery";
+const MAX_PAGE_SIZE_REPOSITORY = 100;
```
joshkay10, line 14: Set to 100 because same as  what Octokit had, but also there is an upper limit with the graphql query of 100

src/github/client/github-queries.ts
```diff
@@ -9,6 +9,42 @@ query {
...
+}
+
+export type GetRepositoriesResponse = {
+  viewer: {
+    repositories: {
+      pageInfo,
+      edges: RepositoryNode[]
+    }
+  }
+};
```
KAbakumovAtl, line 23: We could simply use types from Octokit for responses like we do in other queries.

But this one even better 👍

src/sync/discovery.ts
KAbakumovAtl: Will cost some stack though ;) 

In case if there will be too many repos we are risking stack overflow

But our customers usually don't have more than a few thousands repos, so shoul'd be fine.

src/sync/discovery.ts
```diff
@@ -3,22 +3,54 @@ import { getRepositorySummary } from "./jobs";
...
+
+const getAllRepositories = async (github, hasNextPage: boolean, cursor?: string, repositories: Repository[] = []): Promise<Repository[]> => {
+	if (!hasNextPage) {
+		return repositories;
+	}
+
+	const result = await github.getRepositoriesPage(MAX_PAGE_SIZE_REPOSITORY, cursor);
+	const edges = result?.viewer?.repositories?.edges || [];
+	const nodes = edges.map(({ node: item }) => item);
+	const repos = [...repositories, ...nodes];
```
KAbakumovAtl, line 24: Aren't we risking OOM if we have to keep the total list of repositories fetched on each iteration?

KAbakumovAtl: That won't be garbage collected until recursion finished.

src/sync/discovery.ts
KAbakumovAtl: Actually, I was thinking more about it. 
We are risking using too much memory here. Because all entities we create within this function won't be garbage collected until we reach the end of recursion and exit from all functions. That mean all our responses will be kept in memory, list of all repos generated on each step, etc.

So the loop solution will be better.

src/sync/discovery.ts
```diff
@@ -56,4 +89,4 @@ export const discovery = (app: Application) => async (job, logger: LoggerWithTar
 	} catch (err) {
 		logger.error({ job, err }, "Discovery error");
```
KAbakumovAtl, line 90: 🤔 I've just realised if any error happens here the job won't be retried. Not sure why it was there, but I think we need to remvoe this try/catch block, so the retry will be properly handled by sqs listener

src/sync/discovery.ts
```diff
@@ -3,22 +3,54 @@ import { getRepositorySummary } from "./jobs";
...
+// To tidy up, you can replace the call of this function with the true condition block
+const getRepositories = async (app, installationId, jiraHost, logger) => {
+	if (await booleanFlag(BooleanFlags.USE_NEW_GITHUB_CLIENT_FOR_DISCOVERY, true, jiraHost)) {
+		const github = new GitHubClient(getCloudInstallationId(installationId), logger);
+		const repositories = await getAllRepositories(github, true);
+		return repositories;
+	}
+
+	const github = await app.auth(installationId);
+	enhanceOctokit(github);
```
KAbakumovAtl, line 38: Moving GH Client creating behind the feature flag. Nice!

test/safe/discovery.test.ts
```diff
@@ -43,11 +88,17 @@ describe("Discovery Queue Test", () => {
...
-			.reply(200, require("../fixtures/list-repositories.json"));
+		if (useNewGithubClient) {
+			githubNock
+				.post("/graphql")
+				.reply(200, mockGitHubClientResponse);
+		} else {
+			githubNock.get("/installation/repositories?per_page=100")
+				// eslint-disable-next-line @typescript-eslint/no-var-requires
+				.reply(200, require("../fixtures/list-repositories.json"));
+		}
```
KAbakumovAtl, line 101: /super-nit It will make sense to have 2 separate tests. 1 feature flag ON, another OFF.
Because we have too much if/else branches within our tests

test/safe/discovery.test.ts
```diff
@@ -58,6 +109,7 @@ describe("Discovery Queue Test", () => {
 		expect(states.length).toBe(2);
 	}
 
+	// eslint-disable-next-line jest/expect-expect
```
KAbakumovAtl, line 112: What does this rule do? Seems it wasn't needed before

KAbakumovAtl (commented):
So have one comment about recursion, I think we should address it before merging. 

Otherwise nits.

src/sync/discovery.ts
joshkay10: Sweet Ill swap it back.

test/safe/discovery.test.ts
joshkay10: These two tests don't have assertions directly inside the' it' block since they are hidden away inside a seperate function.
It's one of the 250 warning messages that get thrown on commit.

Alternatively we could move the assertion inside the it blocks, its more code but its much more obvious for testing.

src/sync/discovery.ts
```diff
@@ -3,57 +3,97 @@ import { getRepositorySummary } from "./jobs";
...
 import { sqsQueues } from "../sqs/queues";
+import GitHubClient from "../github/client/github-client";
+import { getCloudInstallationId } from "../github/client/installation-id";
+import { Repository } from "@octokit/graphql-schema";
+import { booleanFlag, BooleanFlags } from "../config/feature-flags";
 
 export const DISCOVERY_LOGGER_NAME = "sync.discovery";
+const MAX_PAGE_SIZE_REPOSITORY = 100;
+
+const getAllRepositories = async (github): Promise<Repository[]> => {
```
joshkay10, line 16: Alternatively, I could loop and fire off a queue event each 100 repos, I was just a bit hesitant to make too much change, since the goal is just replacing the client...... risk vs value kinda thing.

src/sync/discovery.ts
```diff
@@ -3,57 +3,97 @@ import { getRepositorySummary } from "./jobs";
...
+		return;
 	}
-};
+
+	if (repositories.length === 0) {
+		await subscription.update({
+			syncStatus: SyncStatus.COMPLETE
+		});
+		return;
+	}
```
mboudreau, line 83: Can't say I'm a fan of having this duplication of logic for sync.  Can't we reuse the code somehow?

test/safe/discovery.test.ts
```diff
@@ -58,6 +96,79 @@ describe("Discovery Queue Test", () => {
...
```
mboudreau, line 99: I'd probably put that at the top of the file instead of per test if you're going to repeat it

test/safe/discovery.test.ts
```diff
@@ -58,6 +96,79 @@ describe("Discovery Queue Test", () => {
...
+		await verify2RepositoriesInTheStateAndBackfillMessageSent();
+	});
+});
+
+
+describe("Discovery Queue Test - GitHub Client", () => {
+	const TEST_INSTALLATION_ID = 1234;
+	let sendMessageSpy: jest.SpyInstance;
+
+	beforeAll(async () => {
```
mboudreau, line 121: You shouldn't need this as the queue should already be cleared

mboudreau (commented):
A few things to go through.  We should probably talk about large orgs and pagination of repos.

src/sync/discovery.ts
KAbakumovAtl: We definitely should separate it by different queue messages (Keep cursor in DB and schedule a new SQS message for each page).

However for now we are just changing the client. We definitely need a ticket for that (If we don't have it already)

KAbakumovAtl (dismissed):

joshkay10:
Closing PR so it get a fresh look, had significant changes when moving from graphql to REST, additional syncing per API request.

src/sync/discovery.ts
```diff
@@ -3,57 +3,97 @@ import { getRepositorySummary } from "./jobs";
...
+			const result = await github.getRepositoriesPage(MAX_PAGE_SIZE_REPOSITORY, cursor);
+			const edges = result?.viewer?.repositories?.edges || [];
+			const nodes = edges.map(({ node: item }) => item);
+			repositories.push(...nodes);
+			cursor = result?.viewer?.repositories?.pageInfo?.endCursor;
+			hasNextPage = result?.viewer?.repositories?.pageInfo?.hasNextPage;
+		} catch (err) {
+			hasNextPage = false;
+			throw new Error(err);
+		}
```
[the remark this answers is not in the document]
KAbakumovAtl: That's how it work at the moment. The difference is that now it is happening inside Octokit. We definitely need to make a smarter iteration here, but it is not a scope of this PR

src/models/subscription.ts
```diff
@@ -194,6 +194,14 @@ export default class Subscription extends Sequelize.Model {
 	// This is a workaround to fix a long standing bug in sequelize for JSON data types
 	// https://github.com/sequelize/sequelize/issues/4387
 	async updateSyncState(updatedState: RepoSyncStateObject): Promise<Subscription> {
+
+		if (!updatedState?.repos) {
```
joshkay10, line 198: This was pulled out from the discovery function, brought it here so we only have to call update once from discovery.... however, looking at it now i am concerned the and update with no repos may be a legitimate case?
@mboudreau @KAbakumovAtl

src/sqs/discovery.ts
```diff
@@ -1,12 +1,18 @@
...
 
 export type DiscoveryMessagePayload = {
 	installationId: number,
 	jiraHost: string
 }
 
 export const discoveryQueueMessageHandler : MessageHandler<DiscoveryMessagePayload> = async (context) => {
-	await discovery(app)({data: context.payload}, context.log);
-}
+	const useNewGHClient = await booleanFlag(BooleanFlags.USE_NEW_GITHUB_CLIENT_FOR_DISCOVERY, false, context?.payload?.jiraHost);
```
joshkay10, line 12: Moved the feature flag up a level so that the original discovery(now discoveryOctokit) can stay as it was and the new GitHub client version can be a separate function entirely.

src/sync/discovery.ts
```diff
@@ -1,14 +1,76 @@
...
+import Subscription, { Repositories, Repository, SyncStatus } from "../models/subscription";
+import { LoggerWithTarget } from "probot/lib/wrap-logger";
 import { sqsQueues } from "../sqs/queues";
+import GitHubClient from "../github/client/github-client";
+import { getCloudInstallationId } from "../github/client/installation-id";
 
 export const DISCOVERY_LOGGER_NAME = "sync.discovery";
 
-export const discovery = (app: Application) => async (job, logger: LoggerWithTarget) => {
+const mapRepositories = (repositories: Repository[]): Repositories => {
```
joshkay10, line 12: Existing functionality, just wrapped in a function for clarity

src/sync/discovery.ts
```diff
@@ -1,14 +1,76 @@
...
+	return repositories.reduce((obj, repo) => {
+		obj[repo.id] = { repository: getRepositorySummary(repo) };
+		return obj;
+	}, {});
+};
+
+const resetNumberOfSyncedRepos = async (subscription: Subscription) => {
```
joshkay10, line 19: since updateSyncState(line 23) will be called multiple times had to split it out so we could reset the count to zero before the backfill begins without it reseting each iteration.

src/sync/discovery.ts
```diff
@@ -1,14 +1,76 @@
...
+};
+
+const updateSyncState = async (subscription: Subscription, repositories: Repository[]) => {
+	const repos = mapRepositories(repositories);
+	await subscription.updateSyncState({
+		repos
+	});
+};
+
+const checkHeaderForNextPage = (headers): boolean => {
```
joshkay10, line 30: Github returns a link header attribute if there is another page (https://docs.github.com/en/rest/guides/traversing-with-pagination) there is exceptions where no link header will be sent too, but those exceptions don't have more data.
header looks like
header: {
...
links: ["<https://api.github.com/installation/repositories?per_page=1&page=2>; rel=\"prev\"", "<https://api.github.com/installation/repositories?per_page=1&page=2>; rel=\"next\""]
...
}

src/sync/discovery.ts
```diff
@@ -1,14 +1,76 @@
...
+	if(!subscription) {
+		logger.info({ jiraHost, installationId }, "Subscription has been removed, ignoring job.");
+		return;
+	}
+
+	await syncRepositories(github, subscription, installationId, jiraHost, logger);
+	await sqsQueues.backfill.sendMessage({ installationId, jiraHost, startTime: startTime.toISOString() }, 0, logger);
+};
+
+export const discoveryOctoKit = (app: Application) => async (job, logger: LoggerWithTarget) => {
```
joshkay10, line 73: This function is exactly the same pre-pull request with a new name discoveryOctoKit, during clean up will be able to rip this out.

src/github/client/github-queries.ts
```diff
@@ -9,6 +9,42 @@ query {
...
+
+export const GetRepositoriesQuery = `query ($per_page: Int!, $cursor: String) {
```
KAbakumovAtl, line 25: Are we going to switch to GQL at some point?

src/models/subscription.ts
KAbakumovAtl: Yes sure, if there is no repos we don't need to run backfill queue.

If we pull it out from the discovery queue code we need to add a unit test for this case for this class tests

src/sqs/discovery.ts
KAbakumovAtl, line 12: /nit 

Would be great to add jiraHost / installation ID to the context.log, so when logs are logged we can see for which JiraHost and installationId.

		jiraHost: context?.payload?.jiraHost
		installationId: context?.payload?.installationId
	})

src/sqs/discovery.ts
```diff
@@ -1,12 +1,18 @@
...
+	if(useNewGHClient) {
+		await discovery({ data: context.payload }, context.log);
```
KAbakumovAtl, line 14: /nit we don't need to pass payload like "data" parameter. We can put it directly to the method parameter:
`await discovery(context.payload, context.log)`

`export const discovery = async (data: DiscoveryMessagePayload, logger: LoggerWithTarget) => {`
etc.

KAbakumovAtl (approved):
There are some /nits but looks great overall

src/sqs/discovery.ts
joshkay10: ooo yeah, agreed, nice n tidy

src/github/client/github-queries.ts
joshkay10: eventually... I guess it comes down to when we think GHs graphql is robust enough

src/sqs/discovery.ts
joshkay10: Added, was able to remove passing jirahost/installationId around the place too..... win!!

KAbakumovAtl (dismissed):

src/sync/discovery.ts
```diff
@@ -1,14 +1,93 @@
...
+	const regex = /(rel="next")/g;
+	const hasNextPage = links?.find(elem => elem.match(regex));
+	return hasNextPage;
+};
+
+/*
+* Continuosuly call the GitHub repo to fetch a page of repositories at a time and update there sync status until no more pages.
+*/
+const syncRepositories = async (github, subscription: Subscription, logger: LoggerWithTarget): Promise<void> => {
+	let page = 1;
```
joshkay10, line 52: Page = 1, page zero is invalid on the GitHub side so it defaults to page 1, this would cause our counter to be behind by 1 and receive page 1 twice.... therefore starting at page one.

src/models/subscription.ts
mboudreau: also, you're just checking to see if the repos object exist, it could be possible it's an empty object.  Maybe try something like `if(!Object.keys(updatedState?.repos || {}).length) {`

src/sync/discovery.ts
mboudreau: headers can be typed with `AxiosResponseHeaders`

src/sync/discovery.ts
```diff
@@ -1,14 +1,93 @@
...
+};
+
+/*
+* Checking the header atttribute link(type string[]) for a entry that contains "rel=next", this indicates more pages available.
+*/
+const checkHeaderForNextPage = (headers): boolean => {
+	const links = headers?.link?.split(",");
+	const regex = /(rel="next")/g;
+	const hasNextPage = links?.find(elem => elem.match(regex));
+	return hasNextPage;
```
mboudreau, line 45: if links is `undefined`, the function just returns `undefined` or if not, it either finds the element and returns that, or doens't find it which is again, `undefined`.  I'm also pretty sure based on the documentation that you can simplify this function by just doing ```
checkHeaderForNextPage = (headers): boolean => !!headers.link?.includes(`rel="next"`);
```

File: src/sync/discovery.ts (hunk 22 above)
Comment (mboudreau) [reply src/sync/discovery.ts:52]: I don't like the idea that the implementation of the API is bleeding out to this, same with the  `checkHeaderForNextPage` function.  Can we maybe add the `hasNextPage` and the next page number be returned as part of the `getRepositories` function?

Another possibility that I think might be the best way, is having a `pagination` function in the github client which knows about pages and `hasNextPage` implementation details and essentially does the same while loop and then just calls the callback when a page arrives.  Something like:

```
await github.paginate('getRepositoriesPage', async (data) => { /* do the thing */ });
```
It might be outside the scope of this PR, but would be extremely useful for the future.

File: src/sync/discovery.ts [hunk 24]
```diff
@@ -1,14 +1,93 @@
...
+		installationId
+	);
+
+	if(!subscription) {
+		logger.info({ jiraHost, installationId }, "Subscription has been removed, ignoring job.");
+		return;
+	}
+
+	await syncRepositories(github, subscription, logger);
+	await sqsQueues.backfill.sendMessage({ installationId, jiraHost, startTime: startTime.toISOString() }, 0, logger);
```
mboudreau, line 87: you don't need to do `.toISOString()`, Date will serialize to it by default.  Just adding `startTime` should do.

mboudreau (commented):
Couple small changes to make this shine.

src/sync/discovery.ts
joshkay10: Did try that but the sendmessage BackfillMessagePayload type expects a string, I have moved it to its declaration though so its a bit cleaner passing it.

src/sync/discovery.ts
KAbakumovAtl: Do we even need to supply the page number for the first page?

Also I guess the best way will be to not reinvent the wheel and see how it is done in Octokit

joshkay10: We can omit the page and it will return page one by default, in the link header it contains the full url for the next page.... i think(outside of this PR) when we to at implementing a more complete solution like `await github.paginate('getRepositoriesPage', async (data) => { /* do the thing */ })` we can try use that then/ see what Octokit does? I imagine this is going to be re-used a fair bit, so worth doing it proper.

src/models/subscription.ts
joshkay10: This will only get called once from discovery, and handles with/without repos, so discovery is not directly updating status anymore(win). It returns early if nothing to update so no unnecessary checks are run,

src/models/subscription.ts
```diff
@@ -194,6 +194,15 @@ export default class Subscription extends Sequelize.Model {
 	// This is a workaround to fix a long standing bug in sequelize for JSON data types
 	// https://github.com/sequelize/sequelize/issues/4387
 	async updateSyncState(updatedState: RepoSyncStateObject): Promise<Subscription> {
+
+		// Zero repos, means zero syncs required so set COMPLETE.
+		if(!Object.keys(updatedState?.repos || {}).length) {
+			await this.update({
+				syncStatus: SyncStatus.COMPLETE
+			});
+			return this;
```
mboudreau, line 203: You definitely don't want to `return this` as you'll need the `RepoSyncstate` update below to do it's job.  Just removing this will then work as it should.

mboudreau (approved):
finally ;)

PR #1098, opened 2022-02-20
Update express/test packages

mboudreau:
Needed to update packages for the latest folder structure PR split as an issue was found in a middleware.  Figured might as well update all middlewares while I was at it.


test/unit/sqs/index.test.ts
```diff
@@ -91,18 +88,16 @@ describe("SqsQueue tests", () => {
...
-			await queue.stop();
-
-			//delaying to make sure all asynchronous invocations inside the queue will be finished and it will stop
-			queue.start();
-
-			queue.sendMessage(testPayload);
+			queue.stop()
+				//delaying to make sure all asynchronous invocations inside the queue will be finished and it will stop
+				.then(() => queue.start())
+				.then(() => queue.sendMessage(testPayload));
```
KAbakumovAtl, line 94: Didn't we decide to prefer using "await" over "then"?

test/unit/sqs/index.test.ts
```diff
@@ -91,18 +88,16 @@ describe("SqsQueue tests", () => {
...
 			});
 
-			await queue.stop();
-
-			//delaying to make sure all asynchronous invocations inside the queue will be finished and it will stop
-			queue.start();
-
-			queue.sendMessage(testPayload);
+			queue.stop()
+				//delaying to make sure all asynchronous invocations inside the queue will be finished and it will stop
```
KAbakumovAtl, line 92: It seems that this comment doesn't make sense anymore

test/unit/sqs/index.test.ts
```diff
@@ -301,45 +296,45 @@ describe("SqsQueue tests", () => {
...
 			});
 
 			mockErrorHandler.mockImplementation((_error: Error, _context: Context<TestMessage>): ErrorHandlingResult => {
-				return {retryable: receiveCounter.receivesCounter < 3, retryDelaySec: 0, isFailure: true}
-			})
+				return { retryable: receiveCounter.receivesCounter < 3, retryDelaySec: 0, isFailure: true };
+			});
 
-			await queue.sendMessage(testPayload);
+			queue.sendMessage(testPayload);
```
KAbakumovAtl, line 337: Why "await" was removed?

KAbakumovAtl: I believe we should keep all these awaits in case if unexpected execption happen in sendMessage method though unlikeny. 

However I can see we verify test asynchronously, so this is /nit

KAbakumovAtl (approved):

test/unit/sqs/index.test.ts
mboudreau: yes, except we can't here because it has a `done` callback and you can't do both without triggering a jest error.  This is the only way to trigger a "throwaway" promise and guarantee the sequential execution of the code.

test/unit/sqs/index.test.ts
mboudreau: You can't do both a done callback and an async function as they step on each other's toes.  If something does go wrong, it'll timeout and error after 5 seconds.

test/unit/sqs/index.test.ts
KAbakumovAtl: Even if we do:
await queue.stop
await queue.start()
queue.sendMessage()

test/unit/sqs/index.test.ts
KAbakumovAtl: Ah makes sense now

test/unit/sqs/index.test.ts
mboudreau: as I said above, we can't have an async function and a done callback at the same time.  Latest jest throws an error because of the race condition. It's one or the other.

src/frontend/get-jira-configuration.ts
```diff
@@ -6,7 +6,7 @@ import { metricError } from "../config/metric-names";
 import { AppInstallation, FailedAppInstallation } from "../config/interfaces";
 import { GitHubAPI } from "probot";
 import Logger from "bunyan";
-import _ from "lodash";
+import { groupBy } from "lodash";
```
joshkay10, line 9: 🤟 Do we want to capture this preference in the conventions?
https://hello.atlassian.net/wiki/spaces/PF/pages/1424951560/Arc+Coding+Conventions

joshkay10 (dismissed):

src/frontend/get-jira-configuration.ts
mboudreau: Yeah, I think so.  Latest version of ESLint is mentioning it and it's a better standard to use.  Would be good to have it in there too.

joshkay10 (approved):

test/unit/sqs/index.test.ts
KAbakumovAtl: Ah, we can't make the function itself async

KAbakumovAtl (approved):

PR #1101, opened 2022-02-21
Allow  decimal values for the queue delays

KAbakumovAtl:
In this PR I've fixed the visibility timeout change failure in case if we were getting decimal timeout value.

We currently have all Rate Limiting error delays broken, because they pass decimal values. So they use default timeout (Which adds additional load to the rate limited orgs)

Fixing it now

### DIFF
--- a/src/sqs/index.ts
+++ b/src/sqs/index.ts
@@ -53,7 +53,8 @@ export type QueueSettings = {
 	readonly longPollingIntervalSec?: number,
 
 	/**
-	 * Timeout for processing a single message in seconds
+	 * Timeout for processing a single message in seconds.
+	 * If non-integer value passed then the timeout will be rounded to the closest integer
 	 */
 	readonly timeoutSec: number;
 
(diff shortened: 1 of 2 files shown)


src/sqs/index.ts
```diff
@@ -422,15 +423,15 @@ export class SqsQueue<MessagePayload> {
 			return;
 		}
 
-		if(!Number.isInteger(timeout) || timeout < 0) {
```
KAbakumovAtl, line 425: Having all decimal numbers silently ignored here caused some problems. 

All our rate limiting timeouts were passing decimal numbers here (Due to Date.now() / 1000 ) 

We had to either throw an exception to make it more type-safe or support decimal numbers.

I've chosen to support decimal numbers and convert to the nearest integer for the following reasons:
1. Increase type-safety. If decimal passed we still execute logic
2. On the scale of visibility timeout it doesn't matter if we delay for 40.3 seconds or 40 seconds. So rounding the number doesn't break the desired logic

test/unit/sqs/index.test.ts
```diff
@@ -175,39 +175,45 @@ describe("SqsQueue tests", () => {
 			await queue.sendMessage(testPayload);
 		});
 
-		it("Retries with the correct delay", async (done: DoneCallback) => {
 
-			const testErrorMessage = "Something bad happened";
-			const testPayload = generatePayload();
-			const receivedTime = {time: Date.now(), receivesCounter: 0, errorHandlingCounter: 0};
+		describe.each([1, 1.2])("Timeout tests", (timeout) => {
```
KAbakumovAtl, line 179: I've just added an option here, to make 2 tests from 1

mboudreau (approved):

PR #1104, opened 2022-02-23
Alert for messages building in queues

rachellerathbone:
Originally added as a detector in [fusion-monitoring ](https://bitbucket.org/atlassian/fusion-monitoring/pull-requests/444) but moved to our service descriptor to avoid delays and to be consistent with other Micros alerts.

### DIFF
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -165,7 +165,7 @@ resources:
         MetricName: ApproximateNumberOfMessagesVisible
         Namespace: AWS/SQS
         Description: "Message count for deployment queue too high! Follow the runbook: <insert_runbook_url_here>"
-        Threshold: 10
+        Threshold: 400
         Priority: High
         Dimensions:
           - Name: QueueName
@@ -190,7 +190,7 @@ resources:
         MetricName: ApproximateNumberOfMessagesVisible
         Namespace: AWS/SQS
         Description: "Message count for branch queue too high! Follow the runbook: <insert_runbook_url_here>"
-        Threshold: 20
+        Threshold: 400
         Priority: High
         Dimensions:
           - Name: QueueName
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -204,6 +204,22 @@ alarms:
     HighSeverityAlarmWhenTooManyELB5xxErrors: null
     LowSeverityAlarmWhenTooManyBackend5xxErrors: null
     LowSeverityAlarmWhenTooManyELB5xxErrors: null
+  eventsQueueSize:
+    MetricName: ApproximateNumberOfMessagesVisible
+    Namespace: AWS/SQS
+    Description: "Lots of messages in the events queue. Follow the runbook: TBD"
+    # Threshold is 10 because messages generally process fairly quickly but sometimes
+    # we'll have 10 or less that end up waiting on the queue for a short duration
+    Threshold: 10
+    Priority: High
+    Dimensions:
+    - Name: QueueName
+      Value: {"Ref": "QueueName"}
+    EvaluationPeriods: 5
+    Period: 120
+    ComparisonOperator: GreaterThanThreshold
+    Statistic: Maximum
+    Unit: Count
 
 config:
   environmentVariables:
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -87,20 +87,6 @@ resources:
       dataType:
         - UGC/Label                # name of GitHub org
         - PII/IndirectConfidential # name of GitHub org
-    alarms:
-      AlarmOnTooManyMessages:
-        MetricName: ApproximateNumberOfMessagesVisible
-        Namespace: AWS/SQS
-        Description: "Message count for backfill queue too high! Follow the runbook: <insert_runbook_url_here>"
-        Threshold: 20
-        Priority: High
-        Dimensions:
-          - Name: QueueName
-            Value: { "Ref": "QueueName" }
-        EvaluationPeriods: 1
-        Period: 300
-        ComparisonOperator: GreaterThanThreshold
-        Statistic: Maximum
 
   - type: sqs
     name: discovery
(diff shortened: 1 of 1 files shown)


github-for-jira.sd.yml
```diff
@@ -204,6 +204,22 @@ alarms:
     HighSeverityAlarmWhenTooManyELB5xxErrors: null
     LowSeverityAlarmWhenTooManyBackend5xxErrors: null
     LowSeverityAlarmWhenTooManyELB5xxErrors: null
+  eventsQueueSize:
+    MetricName: ApproximateNumberOfMessagesVisible
+    Namespace: AWS/SQS
+    Description: "Lots of messages in the events queue. Follow the runbook: TBD"
+    # Threshold is 10 because messages generally process fairly quickly but sometimes
+    # we'll have 10 or less that end up waiting on the queue for a short duration
+    Threshold: 10
```
KAbakumovAtl, line 213: We should set it to the value greater than our autoscaling threshold (See the autoscaling rules for calculation).

For example 400 for push, 
20 for Backfill queue
5 for Discovery queue (Because we rarely get messages on it, so 5 pending messages will be a huge amount)

github-for-jira.sd.yml
```diff
@@ -204,6 +204,22 @@ alarms:
     HighSeverityAlarmWhenTooManyELB5xxErrors: null
     LowSeverityAlarmWhenTooManyBackend5xxErrors: null
     LowSeverityAlarmWhenTooManyELB5xxErrors: null
+  eventsQueueSize:
```
KAbakumovAtl, line 207: This is a queue alert, we need to put it inside each queue. It wouldn't have any effect on this level (And probably fail the deployment)

Also, don't forget to validate the config using `atlas micros validate` command (You'll need to set all env variables we have in this file to make this validation work)

KAbakumovAtl, line 207: The name of the alert need to be changed `eventsQueueSize:` because we don't have events queue.

github-for-jira.sd.yml
```diff
@@ -204,6 +204,22 @@ alarms:
...
+    MetricName: ApproximateNumberOfMessagesVisible
+    Namespace: AWS/SQS
+    Description: "Lots of messages in the events queue. Follow the runbook: TBD"
```
KAbakumovAtl, line 210: Same here, it is not the events queue

 \+ We need a runbook

rachellerathbone: There's a ticket for this to be addressed separately: ARC-1094

github-for-jira.sd.yml
rachellerathbone: Tried running this but it doesn't work anymore. Not sure what the updated version is.

github-for-jira.sd.yml
```diff
@@ -550,3 +550,88 @@ environmentOverrides:
           ip_whitelist:
             - public
           routes: *blackholeSpammingIPs
+
```
KAbakumovAtl, line 553: Moved it to prod overrides

github-for-jira.sd.yml
```diff
@@ -550,3 +550,88 @@ environmentOverrides:
...
+          AlarmOnTooManyMessages:
+            MetricName: ApproximateNumberOfMessagesVisible
+            Namespace: AWS/SQS
+            Description: "Message count for backfill queue too high! Follow the runbook: https://hello.atlassian.net/wiki/spaces/PF/pages/1550488861/HOWTO+Respond+to+Too+Many+Messages+on+the+Queue+Alert"
+            Threshold: 30 #~ Autoscaling Threshold x 2. So we'll be alerted if autoscaling failed to fix the piling messages
+            Priority: High
+            Dimensions:
+              - Name: QueueName
+                Value: { "Ref": "QueueName" }
+            EvaluationPeriods: 5
```
KAbakumovAtl, line 566: Set evaluation period to 5, so we'll check 5 times with 5 minutes delay (Which mean we'll get an alert if the situation on the queue will stay for 20 minutes)

github-for-jira.sd.yml
```diff
@@ -550,3 +550,88 @@ environmentOverrides:
...
+        alarms:
+          AlarmOnTooManyMessages:
+            MetricName: ApproximateNumberOfMessagesVisible
+            Namespace: AWS/SQS
+            Description: "Message count for deployment queue too high! Follow the runbook: https://hello.atlassian.net/wiki/spaces/PF/pages/1550488861/HOWTO+Respond+to+Too+Many+Messages+on+the+Queue+Alert"
+            Threshold: 400 #~ Autoscaling Threshold x 2. So we'll be alerted if autoscaling failed to fix the piling messages
+            Priority: High
+            Dimensions:
+              - Name: QueueName
+                Value: { "Ref": "QueueName" }
```
thombergs, line 616: Where does it take the QueueName from? Is this some Micros magic?

github-for-jira.sd.yml
```diff
@@ -550,3 +550,88 @@ environmentOverrides:
...
+            Statistic: Maximum
+
+      - type: sqs
+        name: discovery
+        alarms:
+          AlarmOnTooManyMessages:
+            MetricName: ApproximateNumberOfMessagesVisible
+            Namespace: AWS/SQS
+            Description: "Message count for discovery queue too high! Follow the runbook: https://hello.atlassian.net/wiki/spaces/PF/pages/1550488861/HOWTO+Respond+to+Too+Many+Messages+on+the+Queue+Alert"
+            Threshold: 5 #Discovery queue rarely gets messages so if we see 5 messages stuck on the queue, it should be an alarm
```
thombergs, line 578: Need to keep in mind that this will change once we merge the code-as-config feature. With that feature, we add a message to this queue for each repository in an org that has just been connected, to check if there is a `.jira/config.yml` in the repo.

thombergs (approved):

github-for-jira.sd.yml
KAbakumovAtl: Right, makes sense. We should increase this value and include this queue into autoscaling rule when we expect more traffic.

We also should add pagination for the repos iteration.

PR #1109, opened 2022-03-01
JST-736037 GitHub setup issue

mboudreau:
Trying to fix: https://getsupport.atlassian.com/browse/JST-736037

Gist of it is the github setup page after app installation on github side would fail.  Probably because of eventual consistency and the API used.  I changed which API to use and made the call itself optional.  I also changed the design a bit after consulting with our designer as it wasn't consistent.


package.json
```diff
@@ -12,8 +12,8 @@
     "prestart": "run-p setup clean db:dev",
     "init": "run-p db",
     "start": "run-p start:main start:worker",
-    "start:main": "tsnd --watch=.env,db/config.json --inspect=0.0.0.0:9229 --respawn --transpile-only -- src/main.ts",
-    "start:worker": "tsnd --watch=.env,db/config.json --inspect=0.0.0.0:9230 --respawn --transpile-only -- src/worker.ts",
+    "start:main": "tsnd --watch=.env,db/config.json,**/*.hbs --inspect=0.0.0.0:9229 --respawn --transpile-only -- src/main.ts",
+    "start:worker": "tsnd --watch=.env,db/config.json,**/*.hbs --inspect=0.0.0.0:9230 --respawn --transpile-only -- src/worker.ts",
```
mboudreau, line 16: Fixing dev server to restart when changing handlebar files so that they show the latest content on refresh.

src/routes/github/github-router.ts
```diff
@@ -15,13 +17,13 @@ GithubRouter.use(GithubOAuthRouter);
 // CSRF Protection Middleware for all following routes
 GithubRouter.use(csrfMiddleware);
 
-GithubRouter.post("/setup", GithubSetupPost);
+GithubRouter.route("/setup")
+	.get(query("installation_id").isInt(), returnOnValidationError, GithubSetupGet)
+	.post(GithubSetupPost);
```
mboudreau, line 22: Don't need github oauth flow for this anymore as we're not using a user API.  We are checking to make sure `installation_id` is set in querystring though.

src/routes/github/setup/github-setup-get.ts
```diff
@@ -14,32 +14,25 @@ import { Installation } from "../../../models";
...
+		jiraSiteExists(jiraHost),
+		Installation.getForHost(jiraHost)
+	]);
 
+	const redirectUrl = siteExists && jiraInstallation ? getJiraAppUrl(jiraHost) : getJiraMarketplaceUrl(jiraHost);
 	const hasJiraHost = !!jiraHost;
-	const { account } = installation;
-	const { login, avatar_url } = account;
+	const login = githubInstallation?.data.account?.login;
+	const avatar_url = githubInstallation?.data.account?.avatar_url;
```
joshkay10, line 35: Do we want to test data**?** here too?

test/setup/setup.ts
```diff
@@ -13,7 +13,8 @@ jest.mock("lru-cache");
 
 const redis = new IORedis(getRedisInfo("test"));
 
-type AccessTokenNockFunc = (id: number, returnToken?: string, expires?: number, expectedAuthToken?: string) => void
+type GithubUserTokenNockFunc = (id: number, returnToken?: string, expires?: number, expectedAuthToken?: string) => void
```
joshkay10, line 16: 🤌 name change make it a lot more obvious.

src/routes/github/setup/github-setup-get.ts
```diff
@@ -14,32 +14,25 @@ import { Installation } from "../../../models";
...
-	if (!installation) {
-		throw new Error(`Error retrieving installation:${installationId}. You do not permissions to access this org or it does not exist.`);
-	}
-
-	let redirectUrl = getJiraMarketplaceUrl(jiraHost);
+	const githubInstallation = await client.apps.getInstallation({ installation_id: installationId })
+		.catch(() => {
+			// if we cannot get github installation, try to log as much as possible to help debug
+			req.log.warn({ appInfo: info, jiraHost, installationId }, "Cannot retrieve Github Installation from API");
+		});
```
mboudreau, line 24: Using app based API instead of user based API to get the installation directly based on ID.  Also making it optional as we only use it to show the user the current github org name/avatar that they've just connected.

mboudreau: Actually, now that I think about it, if we know the `jiraHost` at this point, we should be able to allow the user to connect to their instance right away.

src/models/installation.ts
```diff
@@ -26,6 +26,9 @@ export default class Installation extends Sequelize.Model {
 	static async getForClientKey(
 		clientKey: string
 	): Promise<Installation | null> {
```
joshkay10, line 28: super /nit   line breaks, 
static async getForClientKey(
		clientKey: string
	): Promise<Installation | null> {
	**vs**
static async getForClientKey( clientKey: string ): Promise<Installation | null> {

I find the latter considerably more readable, but personal pref

src/routes/github/setup/github-setup-get.ts
```diff
@@ -14,32 +14,25 @@ import { Installation } from "../../../models";
...
-
-	let jiraInstallation;
-
-	if (jiraHost) {
-		jiraInstallation = await Installation.getForHost(jiraHost);
-	}
+	const [siteExists, jiraInstallation] = await Promise.all([
+		jiraSiteExists(jiraHost),
+		Installation.getForHost(jiraHost)
+	]);
```
mboudreau, line 30: mmm parallelization

joshkay10 (dismissed):

test/github/client/github-client.test.ts
```diff
@@ -87,7 +87,7 @@ describe("GitHub Client", () => {
 		const pageSize = 5;
 		const page = 1;
 
-		githubAccessTokenNock(githubInstallationId, "installation token");
+		githubUserTokenNock(githubInstallationId, "installation token");
```
mboudreau, line 90: I changed the function name because we also needed to create a global github app token nock, so I changed the name to make it more explicit which token we're getting.

src/routes/github/setup/github-setup-get.ts
mboudreau: Not sure I follow

src/models/installation.ts
mboudreau: Completely agree, but currently we keep getting issues with formatting because some of it is based on the IDE and not the editorconfig file.  I need to fix it with the eslint stuff, which I think will be my next task.

joshkay10 (approved):

joshkay10 (dismissed):

joshkay10 (approved):

PR #1110, opened 2022-03-03
Arc 806 - add github client to Commits backfill.

joshkay10:
Use the GitHub client to request the commit data.

Existing request was 2 GraphQL calls, get default branch and get commits for default branch,
rewrote the GraphQL query to include defaultBranchRef instead of having a seperate request to fetch it.
Manual testing confirmed return of the same data.

Also takes away the assumption that no default branch response is master.

Did not convert to REST due to the task processor relying on the cursor from the graphQL request and felt that changing that was out of scope.


.eslintrc.json
```diff
@@ -9,19 +9,6 @@
 		"project": "tsconfig.json",
 		"sourceType": "module"
 	},
-	"settings": {
```
joshkay10, line 12: settings was duplicated.

src/config/feature-flags.ts
```diff
@@ -24,6 +24,7 @@ export enum BooleanFlags {
 	USE_NEW_GITHUB_CLIENT_FOR_BRANCHES = "use-new-github-client-for-branches",
 	USE_NEW_GITHUB_CLIENT_FOR_DISCOVERY = "use-new-github-client-for-discovery",
 	USE_NEW_GITHUB_CLIENT_FOR_DEPLOYMENTS = "use-new-github-client-for-deployments",
+	USE_NEW_GITHUB_CLIENT_FOR_BACKFILL = "use-new-github-client-for-backfill_z7ps8",
```
joshkay10, line 27: id at the end is because i used Switcheroo to generate it.

src/github/client/github-queries.ts
```diff
@@ -78,7 +78,50 @@ export const getPullRequests = `query ($owner: String!, $repo: String!, $per_pag
     }
   }`;
 
-export const getCommits = (includeChangedFiles?: boolean) => `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String, $default_ref: String!) {
+export type getCommitsResponse = {
```
joshkay10, line 81: Didn't have an appropriate Octokit type that I could find, felt like a unique case so made a custom Type.

src/github/client/github-queries.ts
```diff
@@ -78,7 +78,50 @@ export const getPullRequests = `query ($owner: String!, $repo: String!, $per_pag
...
+          edges
+        }
+      }
+    }
+  }
+};
+
+export const getCommitsQuery = (includeChangedFiles?: boolean) => `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String) {
+    repository(owner: $owner, name: $repo){
+      defaultBranchRef {
```
joshkay10, line 95: other than `defaultBranchRef` replacing `ref(qualifiedName: $default_ref)` this query is exactly the same

src/sync/commits.ts
```diff
@@ -3,10 +3,60 @@ import { GitHubAPI } from "probot";
...
+
+		return {
+			edges,
+			jiraPayload
+		};
+	}
+	return getCommitsOctoKit(logger, github, gitHubClient, jiraHost, repository, cursor, perPage) 
+};
+
+/*
```
joshkay10, line 54: Everything below here is the original implementation, it was too messy to have conditions everywhere to use octokit vs client.... clean up should be simple though. Renamed a few things for clarity.

test/fixtures/api/graphql/commit-queries.ts
```diff
@@ -28,12 +29,11 @@ const query = `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor:
     }
   }`;
 
-export const commitsNoLastCursor = {
+
+export const commitsNoLastCursor = (variables) => ({
 	query,
-	variables: {
```
joshkay10, line 33: moved these values out of here and into the test file, was easier to understand test with this data available.....

test/unit/sync/commits.test.ts
```diff
@@ -1,237 +1,245 @@
 /* eslint-disable @typescript-eslint/no-var-requires,@typescript-eslint/no-explicit-any */
-import { commitsNoLastCursor, commitsWithLastCursor, getDefaultBranch } from "../../fixtures/api/graphql/commit-queries";
```
joshkay10, line 2: Please take extra care looking through these tests. They were previously all skipped, have rebuilt them all based off the original test cases.

src/sync/commits.ts
```diff
@@ -3,10 +3,60 @@ import { GitHubAPI } from "probot";
...
+// According to the logs, GraphQL queries sometimes fail because the "changedFiles" field is not available.
+// In this case we just try again, but without asking for the changedFiles field.
+const retryFetchCommits = async (logger: LoggerWithTarget, err, gitHubClient: GitHubClient, repoOwner: string, repoName: string, cursor?: string | number, perPage?: number): Promise<getCommitsResponse> => {
+	const changedFilesErrors = err.errors?.filter(e => e.message?.includes("The changedFiles count for this commit is unavailable"));
+
+	if (changedFilesErrors?.length) {
+		logger.info("retrying without changedFiles");
+		return await gitHubClient.getCommitsPage(false, repoOwner, repoName, perPage, cursor);
+	}
+	throw new Error(err);
```
joshkay10, line 18: Not sure if this is correct, what I want to achieve is if the above error for changedFilesErrors doesn't exist then just to bubble up the error to be handled elsewhere

.eslintrc.json
mboudreau: huh, must have been a merge issue.

src/config/feature-flags.ts
mboudreau: ...that's kind of annoying

src/github/client/github-queries.ts
mboudreau: Yeah, most graphql ones are going to be unique, sadly.

src/github/client/github-queries.ts
```diff
@@ -78,7 +78,50 @@ export const getPullRequests = `query ($owner: String!, $repo: String!, $per_pag
     }
   }`;
 
-export const getCommits = (includeChangedFiles?: boolean) => `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String, $default_ref: String!) {
+export type getCommitsResponse = {
+  repository: {
+    defaultBranchRef: {
+      target: {
+        history: {
+          edges
```
mboudreau, line 86: What's `edges`? It should have it's own type.

src/sync/commits.ts
mboudreau: Personally, I'd have this retry mechanism build into the `getCommitsPage` for everyone to use.

src/github/client/github-client.ts
```diff
@@ -226,12 +224,22 @@ export default class GitHubClient {
...
-				owner: owner,
+				owner,
 				repo: repoName,
 				per_page: perPage,
 				cursor
 			});
 		return response?.data?.data;
 	}
 
+	public async getCommitsPage(includeChangedFiles: boolean, owner: string, repoName: string, perPage?: number, cursor?: string | number): Promise<getCommitsResponse> {
```
mboudreau, line 235: I don't particularly agree with having `includeChangedFiles` at the start of the argument list.  The argument list is also starting to become unmanageable.  I suggest we use a single object type for all params.  `includeChangedFiles` should be options, defaulted to false.

src/github/client/github-queries.ts
joshkay10: array of commit objects, will add a type against.

src/github/client/github-client.ts
joshkay10: ha, agreed, I had originally created a repoDetails object for exactly that but stripped it out again when I dropped the default ref value. But yes args are getting too long. I'll do some tidy up.

mboudreau (commented):
A few things I think needs to be changed. Happy to pair on this if needed.

joshkay10:
Pollinator staging green, Manual testing on stage working as expected.

mboudreau (approved):
Nice!

PR #1113, opened 2022-03-07
ARC-1176 remove octokit workflow webhook

mboudreau:
Using new github client instead.

### DIFF
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -5,25 +5,12 @@ import AppTokenHolder from "./app-token-holder";
 import InstallationTokenCache from "./installation-token-cache";
 import AuthToken from "./auth-token";
 import { GetPullRequestParams } from "./types";
-import {
-	handleFailedRequest,
-	instrumentFailedRequest,
-	instrumentRequest,
-	setRequestStartTime,
-	setRequestTimeout
-} from "./interceptors";
+import { handleFailedRequest, instrumentFailedRequest, instrumentRequest, setRequestStartTime, setRequestTimeout } from "./interceptors";
 import { metricHttpRequest } from "../../config/metric-names";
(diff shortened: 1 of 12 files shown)


src/github/client/github-client.ts
```diff
@@ -192,6 +179,17 @@ export default class GitHubClient {
 		});
 	};
 
+	public compareReferences = async (owner: string, repo: string, baseRef: string, headRef: string): Promise<AxiosResponse<Octokit.ReposCompareCommitsResponse>> => {
+		return this.get<Octokit.ReposCompareCommitsResponse>(
+			`/repos/{owner}/{repo}/compare/{basehead}`,
+			undefined,
```
joshkay10, line 185: /nit consistency, we've used {} everywhere else for an undefined params argument, I see the function defaults to {} anyway, just looked different. ¯\_(ツ)_/¯

src/transforms/workflow.ts
```diff
@@ -37,68 +38,54 @@ function mapStatus(status: string, conclusion?: string): string {
 }
 
 function mapPullRequests(
-	pull_requests: GitHubPullRequest[]
+	pull_requests: GitHubPullRequest[] = []
 ): JiraPullRequest[] {
 	return pull_requests.map((pr) => ({
 		commit: {
 			id: pr.head.sha,
-			repositoryUri: pr.head.repo.url,
```
joshkay10, line 45: dangling commas 🤢 great choice.

joshkay10 (approved):

src/github/client/github-client.ts
mboudreau: I hate the look of it anyways.  We should use the axios function signature standard anyways.  either just an options object or a url with an options object.  I'll do this change in a later PR.

PR #1114, opened 2022-03-07
Arc 1177 - Replace Octokit with GitHub Client for Branch events

joshkay10:
Added new Feature flag use-new-github-client-for-branch-event to toggle github client in place of OctoKit for the create branch process.

There were 2 locations that Octokit has been replaced for the GetRef and GetCommit API calls.

### DIFF
--- a/src/github/branch.ts
+++ b/src/github/branch.ts
@@ -9,13 +9,10 @@ import { sqsQueues } from "../sqs/queues";
 import { GitHubAPI } from "probot";
 import { LoggerWithTarget } from "probot/lib/wrap-logger";
 import getJiraClient from "../jira/client";
+import GitHubClient from "./client/github-client";
+import { getCloudInstallationId } from "./client/installation-id";
 
-export const createBranch = async (
-	context: CustomContext,
-	jiraClient,
-	_util,
-	githubInstallationId: number
-): Promise<void> => {
+export const createBranch = async (context: CustomContext, jiraClient, _util, githubInstallationId: number): Promise<void> => {
 
 	const webhookPayload: WebhookPayloadCreate = context.payload;
 
@@ -29,10 +26,15 @@ export const createBranch = async (
 		});
 	} else {
 
-		const jiraPayload = await transformBranch(context.github, webhookPayload, jiraClient.baseURL, githubInstallationId, context.log);
+		const gitHubClient = new GitHubClient(getCloudInstallationId(githubInstallationId), context.log);
+		const github = await booleanFlag(BooleanFlags.USE_NEW_GITHUB_CLIENT_FOR_BRANCH_EVENT, false, jiraHost) ? gitHubClient : context.github;
+		const jiraPayload = await transformBranch(github, webhookPayload);
 
 		if (!jiraPayload) {
-			context.log({ noop: "no_jira_payload_create_branch" }, "Halting further execution for createBranch since jiraPayload is empty");
+			context.log(
+				{ noop: "no_jira_payload_create_branch" }, 
+				"Halting further execution for createBranch since jiraPayload is empty"
+			);
 			return;
 		}
 
@@ -51,7 +53,7 @@ export const createBranch = async (
 };
 
 export const processBranch = async (
-	github: GitHubAPI,
+	github: GitHubAPI | GitHubClient,
 	webhookId: string,
 	webhookPayload: WebhookPayloadCreate,
 	webhookReceivedDate: Date,
@@ -65,7 +67,7 @@ export const processBranch = async (
 		webhookReceived: webhookReceivedDate
 	});
 
-	const jiraPayload = await transformBranch(github, webhookPayload, jiraHost, installationId, logger);
+	const jiraPayload = await transformBranch(github, webhookPayload);
 
 	if (!jiraPayload) {
 		logger.info(
@@ -75,9 +77,7 @@ export const processBranch = async (
 		return;
 	}
 
-	logger.info(
-		`Sending jira update for create branch event`
-	);
+	logger.info(`Sending jira update for create branch event`);
 
 	const jiraClient = await getJiraClient(
 		jiraHost,
--- a/src/github/branch.ts
+++ b/src/github/branch.ts
@@ -18,7 +18,6 @@ export const createBranch = async (
 ): Promise<void> => {
 
 	const webhookPayload: WebhookPayloadCreate = context.payload;
-	const logger = context.log;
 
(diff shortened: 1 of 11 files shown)


src/sync/commits.ts
```diff
@@ -20,7 +20,6 @@ const fetchCommits = async (gitHubClient: GitHubClient, repository: Repository,
 export const getCommits = async (logger: LoggerWithTarget, github: GitHubAPI, gitHubClient: GitHubClient, jiraHost: string, repository: Repository, cursor?: string | number, perPage?: number) => {
 	logger.info("Syncing commits: started");
 	if (await booleanFlag(BooleanFlags.USE_NEW_GITHUB_CLIENT_FOR_BACKFILL, false, jiraHost)) {
-		console.log("IM GETTING FROM GHCLIENT");
```
joshkay10, line 23: These 2 logs snuck in from my last update so cleaning up... my bad

test/safe/branch.test.ts
```diff
@@ -41,6 +41,137 @@ describe("Branch Webhook", () => {
 	describe("Create Branch", () => {
 		it("should queue and process a create webhook", async () => {
 
+			when(booleanFlag).calledWith(
```
joshkay10, line 44: Copied the entire test describe block and had one with Octokit(to be cleaned/removed) and a ghclient flow.
We have 3 active Feature flags on this test now..... so will be nice to clean them up soon!

src/transforms/branch.ts
```diff
@@ -1,29 +1,24 @@
...
 import { GitHubAPI } from "probot";
 import { generateCreatePullRequestUrl } from "./util/pullRequestLinkGenerator";
+import GitHubClient from "../github/client/github-client";
+import { getCloudInstallationId } from "../github/client/installation-id";
 import { booleanFlag, BooleanFlags } from "../config/feature-flags";
+import { LoggerWithTarget } from "probot/lib/wrap-logger";
+import { JiraBranchData, JiraCommit } from "src/interfaces/jira";
 
-async function getLastCommit(github: GitHubAPI, webhookPayload: WebhookPayloadCreate, issueKeys: string[]) {
+const getLastCommit = async (useNewGitHubClient: boolean, github: GitHubAPI, gitHubClient: GitHubClient, webhookPayload: WebhookPayloadCreate, issueKeys: string[]): Promise<JiraCommit> => {
```
mboudreau, line 14: Don't add in the old github client, and the new one, and the flag of which one to use as parameters of the function.

Instead you can just use a single client of type `GithubAPI | GithubClient` and then do an `instanceof` check to see which function to call.  Check my last PR as an example of this: https://github.com/atlassian/github-for-jira/pull/1113/files#diff-90aa83dd1c1806edf7ae34e6163d76150bf0e0e132556c007f8dabd0d9fbb2e7R21

It'll make the transition to removing octokit much simpler and the functions cleaner

src/transforms/branch.ts
```diff
@@ -37,31 +32,33 @@ async function getLastCommit(github: GitHubAPI, webhookPayload: WebhookPayloadCr
...
+		return;
 	}
 
-	const lastCommit = await getLastCommit(github, webhookPayload, issueKeys);
-
-	const newPrUrl = await booleanFlag(BooleanFlags.USE_NEW_GITHUB_PULL_REQUEST_URL_FORMAT, false);
+	const gitHubClient = new GitHubClient(getCloudInstallationId(installationId), logger);
+	const newPrUrl = booleanFlag(BooleanFlags.USE_NEW_GITHUB_PULL_REQUEST_URL_FORMAT, false, jiraHost);
+	const newGitHubClient = booleanFlag(BooleanFlags.USE_NEW_GITHUB_CLIENT_FOR_BRANCH_EVENT, false, jiraHost);
+	const [useNewPrUrl, useNewGitHubClient] = await Promise.all([newPrUrl, newGitHubClient]);
```
mboudreau, line 52: /nit I've never been a fan of using constants that don't stick around.

test/safe/branch.test.ts
mboudreau: Instead of copying the block, you can do `describe.each([true, false])("your description", (useNewGithubClient) => {...` since the code should work _exactly the same way_, it should pass whether it's on or off.  Here's an example: https://github.com/atlassian/github-for-jira/pull/1113/files#diff-d49caa3c6885aa6a494bf47acede68ba12da206329cb7eb0614b8fbd21d771e2R10

src/transforms/branch.ts
joshkay10: sounds reasonable, will convert

src/transforms/branch.ts
```diff
@@ -37,31 +32,34 @@ async function getLastCommit(github: GitHubAPI, webhookPayload: WebhookPayloadCr
...
 
-	const lastCommit = await getLastCommit(github, webhookPayload, issueKeys);
-
-	const newPrUrl = await booleanFlag(BooleanFlags.USE_NEW_GITHUB_PULL_REQUEST_URL_FORMAT, false);
+	const gitHubClient = new GitHubClient(getCloudInstallationId(installationId), logger);
+	const [useNewPrUrl, useNewGitHubClient] = await Promise.all([
+		booleanFlag(BooleanFlags.USE_NEW_GITHUB_PULL_REQUEST_URL_FORMAT, false, jiraHost), 
+		booleanFlag(BooleanFlags.USE_NEW_GITHUB_CLIENT_FOR_BRANCH_EVENT, false, jiraHost)
+	]);
+	const lastCommit = await getLastCommit((useNewGitHubClient ? gitHubClient : github), webhookPayload, issueKeys);
```
joshkay10, line 54: I went inline with the ternary, thought it would be a bit silly to make a throw away const after the changes above 😂
will be ripped out in clean up too....

KAbakumovAtl (dismissed):

src/github/branch.ts
```diff
@@ -64,14 +59,14 @@ export const processBranch = async (
...
 ) => {
-	const jiraPayload = await transformBranch(github, webhookPayload);
-
 	const logger = rootLogger.child({
 		webhookId: webhookId,
 		installationId,
 		webhookReceived: webhookReceivedDate
 	});
 
+	const jiraPayload = await transformBranch(github, webhookPayload, jiraHost, installationId, logger);
```
mboudreau, line 68: I noticed a few of these `transformBranch` calls are sending down information that's needed to create a github client.  I think we should be building the github client and doing the flag check at a high level and passing down the `GithubAPI | GithubClient` so there's no duplication of code across multiple functions, just a simple ternary operator to know which one to use/call.

joshkay10: tidied up and only pass through github/githubclient

mboudreau (approved):

PR #1116, opened 2022-03-08
Arc 1182 fix error handling for backfill

KAbakumovAtl:
In this PR I've done the following:
1. Refactored Backfill error handling code to make it testable
2. Added tests
3. Fixed a situation when we have 404 error and treat it like Rate Limiting (See ticket description for details)

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -313,25 +313,56 @@ async function doProcessInstallation(app, data: BackfillMessagePayload, sentry: 
 		statsd.increment(metricTaskStatus.complete, [`type: ${nextTask.task}`]);
 
 	} catch (err) {
-		await handleBackfillError(err, logger, scheduleNextTask, data, task, repositoryId, subscription, nextTask);
+
+		const ignoreCurrentRepo = async () => {
+			const edgesLeft = []; // No edges left to process since the repository doesn't exist
+			await updateJobStatus(data, edgesLeft, task, repositoryId, logger, scheduleNextTask);
+		}
+
+		const failCurrentRepoAndContinue = async () => {
+			// marking the current task as failed
+			await subscription.updateRepoSyncStateItem(nextTask.repositoryId, getStatusKey(nextTask.task as TaskType), "failed");
+
+			statsd.increment(metricTaskStatus.failed, [`type: ${nextTask.task}`]);
+
+			// queueing the job again to pick up the next task
+			scheduleNextTask(0);
+		}
+
+		await handleBackfillError(err, logger, scheduleNextTask, ignoreCurrentRepo, failCurrentRepoAndContinue);
 	}
 }
 
-const handleBackfillError = async (err,
+/**
+ * Handles and error and takes action based on the error type and parameters
+ *
+ * @param err The error
+ * @param logger
+ * @param scheduleNextTask Function which schedules next task with the specified delay
+ * @param ignoreCurrentRepo Function which ignores currently processing repo and schedules the next task
+ * @param failCurrentRepoAndContinue Function which sets the status of the current repo sync to "failed" and schedules the next task
+ */
+export const handleBackfillError = async (err,
 	logger: LoggerWithTarget,
-	scheduleNextTask: (delayMs) => void, data: BackfillMessagePayload,
-	task: "pull" | "commit" | "branch",
-	repositoryId: string,
-	subscription: SubscriptionClass,
-	nextTask: Task) : Promise<void> => {
+	scheduleNextTask: (delayMs: number) => void,
+	ignoreCurrentRepo: () => Promise<void>,
+	failCurrentRepoAndContinue: () => Promise<void>) : Promise<void> => {
 
-	const rateLimit = (err instanceof RateLimitingError || err instanceof OldRateLimitingError) ? err.rateLimitReset : Number(err?.headers?.["x-ratelimit-reset"]);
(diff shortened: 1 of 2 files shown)


src/sync/installation.ts
```diff
@@ -313,25 +313,56 @@ async function doProcessInstallation(app, data: BackfillMessagePayload, sentry:
 		statsd.increment(metricTaskStatus.complete, [`type: ${nextTask.task}`]);
 
 	} catch (err) {
-		await handleBackfillError(err, logger, scheduleNextTask, data, task, repositoryId, subscription, nextTask);
+
```
KAbakumovAtl, line 316: I've extrated some code from `handleBackfillError` into functions `ignoreCurrentRepo` and `failCurrentRepoAndContinue`. 

This is to make the code testable without changing much

src/sync/installation.ts
```diff
@@ -313,25 +313,56 @@ async function doProcessInstallation(app, data: BackfillMessagePayload, sentry:
...
-	scheduleNextTask: (delayMs) => void, data: BackfillMessagePayload,
-	task: "pull" | "commit" | "branch",
-	repositoryId: string,
-	subscription: SubscriptionClass,
-	nextTask: Task) : Promise<void> => {
+	scheduleNextTask: (delayMs: number) => void,
+	ignoreCurrentRepo: () => Promise<void>,
+	failCurrentRepoAndContinue: () => Promise<void>) : Promise<void> => {
+
+	const isRateLimitError = (err instanceof RateLimitingError || err instanceof OldRateLimitingError) || Number(err?.headers?.["x-ratelimit-remaining"]) == 0;
```
KAbakumovAtl, line 351: That's the bug fix. Check explicitly if it is rate limiting based on the error type and `x-ratelimit-remaining` header

src/sync/installation.ts
```diff
@@ -356,22 +387,14 @@ const handleBackfillError = async (err,
...
 	// Continue sync when a 404/NOT_FOUND is returned
 	if (isNotFoundError(err, logger)) {
-		const edgesLeft = []; // No edges left to process since the repository doesn't exist
-		await updateJobStatus(data, edgesLeft, task, repositoryId, logger, scheduleNextTask);
+		await ignoreCurrentRepo();
 		return;
 	}
 
-	// TODO: add the jiraHost to the logger with logger.child()
-	const host = subscription.jiraHost || "none";
```
KAbakumovAtl, line 365: There is no need to extract host from subscription and adding it to the log anymore. Jira Hostname is already added to the logger in the backfill message handler.

src/sync/installation.ts
mboudreau: I'm not sure I understand the point of doing these changes.  Maybe it's making it easier to test, but it's also making it extremely hard to read and breaks multiple software practices.

Can we please keep the functionality as it was?

KAbakumovAtl: The point is to make it testable. How would you suggest to test it otherwise?

test/unit/sync/installation.test.ts
```diff
@@ -1,14 +1,23 @@
-import { isRetryableWithSmallerRequest, maybeScheduleNextTask, processInstallation } from "../../../src/sync/installation";
+import {
+	handleBackfillError,
+	isRetryableWithSmallerRequest, maybeScheduleNextTask,
+	processInstallation,
+} from "../../../src/sync/installation";
 
-import { DeduplicatorResult } from "../../../src/sync/deduplicator";
+import * as installation from "../../../src/sync/installation";
```
KAbakumovAtl, line 7: Seems that star import works well alongside with named imports. Not sure if it is the best practice, however seems that the rest of the code will be more clear if I don't have to specify "installation." prefix before each imported entity

src/sync/installation.ts
KAbakumovAtl: As we discussed. I was able to achieve the same result by using jest.spyOn on installation module functions

src/sync/installation.ts
```diff
@@ -313,25 +314,41 @@ async function doProcessInstallation(app, data: BackfillMessagePayload, sentry:
...
-		scheduleNextTask(delay);
+		if (delay) {
+			// if not NaN or 0
+			logger.info({delay}, `Delaying job for ${delay}ms`);
+			scheduleNextTask(delay);
+		} else {
+			//Retry immediately if rate limiting reset already
+			logger.info("Rate limit was reset already. Scheduling next task")
+			scheduleNextTask(0);
+		}
```
mboudreau, line 351: /nit `scheduleNextTask(delay)` and `scheduleNextTask(0)` is the same thing so long as `rateLimit` defaults to 0, or you do  `scheduleNextTask(delay || 0)`

mboudreau (approved):

src/sync/installation.ts
KAbakumovAtl: Yes, true

KAbakumovAtl: I tried to refactor, but I'd prefer to keep it like this, it seems more readable to me to make it explicit.

```
		if (delay) {
			// if not NaN or 0
			logger.info({delay}, `Delaying job for ${delay}ms`);
		} else {
			//Retry immediately if rate limiting reset already
			logger.info("Rate limit was reset already. Scheduling next task")
		}
		scheduleNextTask(delay);
		return;
```

PR #1117, opened 2022-03-09
ARC-1172 updating issue update to use github client

mboudreau:


src/github/issue.ts
```diff
@@ -36,3 +45,6 @@ export default async (context: CustomContext, _: JiraClient, util): Promise<void
 		githubResponse?.status
 	);
 };
+
+const updateIssue = async (githubClient:GitHubAPI | GitHubClient, issue: Octokit.IssuesUpdateParams) =>
```
joshkay10, line 49: I do the same call in the PR event, is it worth me calling this one or having a service for both events(pr and issue) to access?

joshkay10 (approved):
Starting to feel like some more consistency with these events!!

src/github/issue.ts
mboudreau: always share code where we can.  I'd say move it to it's own file somewhere with a TODO once the flags have been removed.

PR #1118, opened 2022-03-09
ARC-1174 pr events

joshkay10:
Replacing  Octokit with Github client for fetch a list of review data and updating github issues.

### DIFF
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -159,6 +159,17 @@ export default class GitHubClient {
 	}
 
 	/**
+	 * Get all reviews for a specific pull request.
+	 */
+	public async getPullRequestReviews(owner: string, repo: string, pullNumber: string | number): Promise<AxiosResponse<Octokit.PullsListReviewsResponse>> {
+		return await this.get<Octokit.PullsListReviewsResponse>(`/repos/{owner}/{repo}/pulls/{pullNumber}/reviews`, {}, {
+			owner,
+			repo,
+			pullNumber
+		});
+	}
+
+	/**
 	 * Get publicly available information for user with given username.
 	 */
 	// TODO: add a unit test
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -242,10 +242,6 @@ export default class GitHubClient {
 	 * Update an issue
 	 */
 	public async updateIssue(data = {}, owner: string, repo: string, issueNumber: string | number): Promise<AxiosResponse<Octokit.IssuesUpdateResponse>> {
-		console.log("WHOAMI");
-		console.log(`/repos/${owner}/${repo}/issues/${issueNumber}`);
-		console.log(data);
-
 		return await this.patch<Octokit.IssuesUpdateResponse>(`/repos/{owner}/{repo}/issues/{issueNumber}`, data, {}, {
 			owner,
 			repo,
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -210,6 +210,17 @@ export default class GitHubClient {
 	};
 
 	/**
+	 * Returns a single reference from Git. The {ref} in the URL must be formatted as heads/<branch name>
+	 */
+	public getRef = async (owner: string, repo: string, ref: string): Promise<AxiosResponse<Octokit.GitGetRefResponse>> => {
+		return await this.get<Octokit.GitGetRefResponse>(`/repos/{owner}/{repo}/git/ref/{ref}`, {}, {
+			owner,
+			repo,
+			ref
+		});
+	};
+
+	/**
 	 * Get a page of repositories.
 	 */
 	public getRepositoriesPage = async (page = 1): Promise<PaginatedAxiosResponse<Octokit.AppsListReposResponse>> => {
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -247,20 +247,6 @@ export default class GitHubClient {
 			});
 	}
 
-	// public async updateIssue({ owner, repo, issue_number, body }: Octokit.IssuesUpdateParams): Promise<AxiosResponse<Octokit.IssuesUpdateResponse>> {
-	// 	return await this.axios.patch<Octokit.IssuesUpdateResponse>(
-	// 		`/repos/{owner}/{repo}/issues/{issue_number}`, {
-	// 			body
-	// 		}, {
-	// 			...await this.installationAuthenticationHeaders(),
-	// 			urlParams: {
-	// 				owner,
-	// 				repo,
-	// 				issue_number
-	// 			}
-	// 		});
-	// }
-
 	public async getNumberOfReposForInstallation(): Promise<number> {
 		const response = await this.graphql<{ viewer: { repositories: { totalCount: number } } }>(ViewerRepositoryCountQuery);
 
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -294,4 +294,17 @@ export default class GitHubClient {
 		return response?.data?.data;
 	}
 
+	public async updateIssue({ owner, repo, issue_number, body }: Octokit.IssuesUpdateParams): Promise<AxiosResponse<Octokit.IssuesUpdateResponse>> {
+		return await this.axios.patch<Octokit.IssuesUpdateResponse>(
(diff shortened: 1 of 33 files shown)


src/github/client/github-client.ts
```diff
@@ -158,6 +166,17 @@ export default class GitHubClient {
 		});
 	}
 
+	/**
+	 * Get all reviews for a specific pull request.
+	 */
+	public async getPullRequestReviews(owner: string, repo: string, pullNumber: string | number): Promise<AxiosResponse<Octokit.PullsListReviewsResponse>> {
```
joshkay10, line 172: I will align this method arguments with https://github.com/atlassian/github-for-jira/pull/1117/files

src/github/pull-request.ts
```diff
@@ -111,3 +95,29 @@ export default async (
 		jiraResponse?.status
 	);
 };
+
+const updateGithubIssues = async (github: GitHubClient | GitHubAPI, context: CustomContext, util, pullRequest) => {
```
joshkay10, line 99: @mboudreau  - this seems to do similar as https://github.com/atlassian/github-for-jira/pull/1117/files, I am considering re-using your issues method or moving them both to a generic location?

src/github/pull-request.ts
```diff
@@ -111,3 +95,29 @@ export default async (
...
+	const linkifiedBody = await util.unfurl(pullRequest.body);
+	if (!linkifiedBody) {
+		return;
+	}
+
+	context.log("Updating pull request");
+	const editedPullRequest = context.issue({
+		body: linkifiedBody,
+		id: pullRequest.id
```
joshkay10, line 108: ID doesn't seem to be required, but it is in the tests so didn't pull it out incase there was some other reason its required that I didn't spot.

src/github/pull-request.ts
mboudreau: yep, sounds like a great idea.

src/github/pull-request.ts
mboudreau: yeah, pretty sure it's not.  Also, I'd remove the `context.issue` data object generator and build it yourself (you can look at my issue PR for an example).

src/github/pull-request.ts
```diff
@@ -111,3 +95,29 @@ export default async (
...
+	});
+
+	const { body, id, owner, repo, number } = editedPullRequest;
```
mboudreau, line 111: yeah, you're already creating the object here really.  Cut out the middle man ;)

src/transforms/pull-request.ts
```diff
@@ -49,28 +50,27 @@ function mapReviews(reviews: Octokit.PullsListReviewsResponse = []) {
 }
 
 // TODO: define arguments and return
-export const transformPullRequest = async (github: GitHubAPI, pullRequest: Octokit.PullsGetResponse, reviews?: Octokit.PullsListReviewsResponse, log?: LoggerWithTarget) => {
+export const transformPullRequest = async (github: GitHubAPI | GitHubClient, pullRequest: Octokit.PullsGetResponse, reviews?: Octokit.PullsListReviewsResponse, log?: LoggerWithTarget) => {
 	const { title: prTitle, head, body } = pullRequest;
 
 	// This is the same thing we do in sync, concatenating these values
-	const textToSearch = await booleanFlag(BooleanFlags.ASSOCIATE_PR_TO_ISSUES_IN_BODY, true) ? `${prTitle}\n${head.ref}\n${body}}` : `${prTitle}\n${pullRequest.head.ref}`
+	const textToSearch = await booleanFlag(BooleanFlags.ASSOCIATE_PR_TO_ISSUES_IN_BODY, true) ? `${prTitle}\n${head.ref}\n${body}}` : `${prTitle}\n${pullRequest.head.ref}`;
```
mboudreau, line 57: This could be simplified.  Always try to not duplicate string interpolation if you can to reduce possible errors in the future.
```
const prBody = await booleanFlag(BooleanFlags.ASSOCIATE_PR_TO_ISSUES_IN_BODY, true) ? body : "";
const textToSearch = `${prTitle}\n${head.ref}\n${prBody}}`
```
You could also inline it all, but seems harder to read/messier.

src/github/client/github-client.ts
KAbakumovAtl: I wonder should we make these methods return PullsListReviewsResponse rather than AxiosResponse? (Cuz we handle axios response error in the middleware (if I am not wrong))

Also, half of the methods in this file return data classes like `getCommitsPage` return `getCommitsResponse`

I believe we should incapsulate Axios details inside the class and return the domain model type

src/github/pull-request.ts
```diff
@@ -1,61 +1,60 @@
 import { transformPullRequest } from "../transforms/pull-request";
 import issueKeyParser from "jira-issue-key-parser";
-
 import { emitWebhookProcessedMetrics } from "../util/webhooks";
 import { CustomContext } from "./middleware";
-import _ from "lodash";
-
-export default async (
```
KAbakumovAtl, line 8: Nice 👍

src/github/pull-request.ts
```diff
@@ -1,61 +1,60 @@
...
+	const baseUrl = jiraClient.baseUrl || "none";
+	const githubClient =
+		await booleanFlag(BooleanFlags.USE_NEW_GITHUB_CLIENT_FOR_PULL_REQUEST_WEBHOOK, false, jiraClient.baseURL) ?
+			new GitHubClient(getCloudInstallationId(githubInstallationId), context.log)
+			: context.github;
+
+	context.log = context.log.child({
+		number: pullRequestNumber,
+		body: pullRequestBody,
+		id: pullRequestId
```
KAbakumovAtl, line 32: the "id" parameter is already used for sqs message id.
We need to choose other names.

Also, we shoudln't log PR body, because it is UGC (Only log in case of errors)
- [ ] Adjust logger parameters

src/github/pull-request.ts
```diff
@@ -111,3 +95,29 @@ export default async (
...
+		await github.updateIssue({ body, id }, owner, repo, number) :
+		await github.issues.update(editedPullRequest);
+};
+
+const getReviews = async (githubCient: GitHubAPI | GitHubClient, owner: string, repo: string, pull_number: number): Promise<Octokit.PullsListReviewsResponse> => {
+	const response = githubCient instanceof GitHubClient ?
+		await githubCient.getPullRequestReviews(owner, repo, pull_number) :
+		await githubCient.pulls.listReviews({ owner, repo, pull_number });
+	return response.data;
+};
```
KAbakumovAtl, line 123: Nice, that looks clean

src/github/pull-request.ts
KAbakumovAtl: We are more interested at adding jiraHostname and InstallationID to logger params

joshkay10: Have dropped the body....

and used more descriptive names for the log attributes, squeezed in install and jirahost too
```
{
 jiraHostName: jiraClient.baseURL,
 installationId: githubInstallationId,
 pullRequestNumber,
 pullRequestId
}
```

src/github/client/github-client.ts
joshkay10: The Axios fella is wraps it up nicely and gives us the data in a data attribute and a few more attributes like headers, status etc.

The `getCommitsPage` and `getCommitsResponse` are GraphQL calls, but you are probably right that we should try have some consistency!! Calling the GH client and having one return the data and one with an object containing the data might cause some confusion🤔

src/github/client/github-client.ts
```diff
@@ -219,6 +238,29 @@ export default class GitHubClient {
...
+	// 			body
+	// 		}, {
+	// 			...await this.installationAuthenticationHeaders(),
+	// 			urlParams: {
+	// 				owner,
+	// 				repo,
+	// 				issue_number
+	// 			}
+	// 		});
+	// }
```
KAbakumovAtl, line 262: /nit Should we remove commented code?

KAbakumovAtl (dismissed):

KAbakumovAtl (approved):

src/github/client/github-client.ts
mboudreau: yeah, noticed the inconsistency as well.  I think we need to have a bit of a chat around the Github Client and how we want it to work because I think there are better ways of doing it that is much more consistent.

mboudreau (approved):

PR #1126, opened 2022-03-16
ARC-1128 - Created new  GitHub user client class

joshkay10:
Created new Github User Client class, uses the apptoken from Github Middleware to authenticate requests.

Renamed GithubClient to GitHubAppClient


mboudreau (approved):
Very nice :)

PR #1133, opened 2022-03-18
JST-736037 trying to fix issue with getBranch sync graphql

mboudreau:
Trying to fix sync for getBranch call for large enterprise customer.  Adding better logging to see what's going on.

### DIFF
--- a/src/sync/branches.ts
+++ b/src/sync/branches.ts
@@ -1,37 +1,19 @@
 import { transformBranches } from "./transforms/branch";
 import { GitHubAPI } from "probot";
 import { Repository } from "../models/subscription";
-import { Repository as OctokitRepository } from "@octokit/graphql-schema";
 import { GitHubAppClient } from "../github/client/github-app-client";
 import { booleanFlag, BooleanFlags } from "../config/feature-flags";
 import { LoggerWithTarget } from "probot/lib/wrap-logger";
-import { GetBranchesQuery } from "../github/client/github-queries";
 
 // TODO: better typings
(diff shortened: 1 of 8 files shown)


src/sync/branches.ts
```diff
@@ -1,37 +1,19 @@
...
-import { GetBranchesQuery } from "../github/client/github-queries";
 
 // TODO: better typings
-export default async (logger: LoggerWithTarget, github: GitHubAPI, newGithub: GitHubAppClient, jiraHost: string, repository:Repository, cursor?:string | number, perPage?:number) => {
+export default async (logger: LoggerWithTarget, _github: GitHubAPI, newGithub: GitHubAppClient, jiraHost: string, repository:Repository, cursor?:string | number, perPage?:number) => {
 	// TODO: fix typings for graphql
 	logger.info("Syncing branches: started");
-
-	const useNewGHClient = await booleanFlag(BooleanFlags.USE_NEW_GITHUB_CLIENT_FOR_BRANCHES, false, jiraHost);
+	perPage = perPage || 20;
```
joshkay10, line 12: /nit can we set  perPage = 20 in the function def instead o reassigning parameter.
/nit name the export since you're in the neighbourhood.

joshkay10 (approved):

src/sync/branches.ts
mboudreau: I can, but would have to change the signature of the function as you can't but an initialized parameter in front of an optional one.  Decided against changing the signature as it would mean a larger change.

PR #1134, opened 2022-03-18
ARC-672 readd test file move

mboudreau:
Issue was only happening in production because of the shortened paths.  This has been resolved.


tsconfig.json
```diff
@@ -40,5 +40,5 @@
 	"include": [
```
joshkay10, line 40: getting an error from this file ': Unexpected token } in JSON at position 1069' during tests

joshkay10 (approved):
🤞

PR #1139, opened 2022-03-25
ARC-361 Added POCO config and enabled Servbice Proxy

KAbakumovAtl:

* Added POCO configuration
* Enabled Service Proxy in a transition mode.

About POCO Configuration.
* For the endpoints which are used in the app business logic the authentication will be handled in the app. (As it is now). In POCO configuration we'll allow public access to these endpoints
* For admin API endpoints we'll allow only github-for-jira service team to call them using SLAuth CLI.

Current configuration change don't enforce any of these POCO rules, it simply adds access logs which show if endpoints access correctly.

I am also updating Deployment Scripts to publish POCO config. There will be a PR in the deployment repo.

After this the next steps are the following:
- we'll test in staging and prod that we can access Admin Endpoints using SLAuth CLI (by still including GitHub token). We'll see in logs if we doing it correctly
- we'll add SLAuth authentication for Admin endpoints in the app and remove "transitionMode" from the POCO cofnig.
- we'll remove GitHub App token authentication so we won't need to pass it to Admin SLAuth CLI request


etc/poco/bundle/main.json
```diff
@@ -0,0 +1,52 @@
...
+        "/api/**"
+      ],
+      "methods": [
+        "GET","POST","PUT","DELETE"
+      ],
+      "principals": {
+				"staff":{"groups": ["micros-sv--github-for-jira-dl-admins"]}
+			}
+    }
+  ]
```
KAbakumovAtl, line 51: TODO:
- [ ] Add per-env access to Pollinator Tests
- [ ] Add ASAP Authentication for Pollinator Test clients

atrigueiro (dismissed):

etc/poco/README.md
```diff
@@ -0,0 +1,58 @@
...
+  -b etc/poco/bundle/extras-stg.json \
+  -t etc/poco/bundle/main-test.json \
+  -t etc/poco/bundle/extras-stg-test.json
+```
+
+# Deployment to production
+
+Make sure all changes in the poco policy are backward compatible i.e. they can safely be applied to the current and new version of service.
+
+Deployment to production happens automatically in the build.
```
atrigueiro, line 58: We should add a section at the end on how to access admin endpoints locally using the `atlas slauth curl` command and saying it's only available if you're part of the admins group.

atrigueiro: - [ ] Add instructions for Admin access via CLI

atrigueiro (approved):

etc/poco/README.md
KAbakumovAtl: Sure, we'll add it. But not now, with the following PRs, when we enable SLAuth

PR #1140, opened 2022-03-27
ARC-672 ALL THE CODE UPDATES!

mboudreau:
Removed all `export default` and replaced them with named exports, changed filenames to kebab case everywhere for consistency, updated paths to use shorthand version where possible, renamed all `index.ts` file to be specific and changed models to be consistent throughout.


joshkay10 (approved):
Nice!!
Can we start looking at enforcing these with eslint errors!?

mboudreau:
@joshkay10 that's next with the eslint ticket in backlog :)

PR #1141, opened 2022-03-28
Disabled transition mode for POCO/SLAuth and disabled GitHub token au…

KAbakumovAtl:
This PR disables transition mode for Service Proxy.

Since it will be deployed, all requests to GitHub App will be authenticated and authorized by Service Proxy.

We won't need GitHub Token authentication anymore for admin endpoints. So I removed the code which was handling it. 
From the APP point of view, these endpoints will be OPEN for anyone. However, in fact these requests will be authenticated by Service Proxy and will require proper Slauth headers or ASAP tokens (for Pollinator tests). (I am going to write a runbook on how to access them)

### DIFF
--- a/src/routes/api/api-router.ts
+++ b/src/routes/api/api-router.ts
@@ -1,9 +1,8 @@
-import { NextFunction, Request, Response, Router } from "express";
+import { Request, Response, Router } from "express";
 import { param } from "express-validator";
 import rateLimit from "express-rate-limit";
 import RedisStore from "rate-limit-redis";
 import IORedis from "ioredis";
-import { GithubAPI } from "config/github-api";
 import { Subscription } from "models/subscription";
 import { returnOnValidationError, serializeSubscription } from "./api-utils";
 import { getRedisInfo } from "config/redis-info";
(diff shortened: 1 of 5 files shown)


KAbakumovAtl:
We've just tested in DDEV and Staging, we are going to schedule a Blitz with the team for Today before merging to prod

src/routes/api/api-router.ts
```diff
@@ -43,78 +28,6 @@ ApiRouter.use(rateLimit({
...
-			req.log.info({ err });
-
-			if (err.status === 401) {
-				res.status(401).send(err.HttpError);
-				return;
-			}
-			res.sendStatus(500);
-		}
-	}
-);
```
mboudreau, line 116: By removing this, you're making the API completely open.  I think the `/api` route is being blocked by a higher level routing at the micros/infrastructure level, but it still irks me that we don't do a single check at this level (like a header check or something) because if the infrastructure changes by mistake, how will we know that our admin api is open?

mboudreau: also, what happens if we change the API route in the future to `/admin` or something?

KAbakumovAtl: >> also, what happens if we change the API route in the future to /admin or something?
If we do it we won't be able to call it unless we change POCO config accordingly. All open endpoints need to be explicitly whitelisted


We can address adding extra protection layer later

KAbakumovAtl: SLAuth adds headers to the incoming request, which tell what authentication mechanism've been used.

We can add the middleware which verifies that these headers present there. However, these headers don't contain any token or signature we can verify to ensure that these headers were added by Service Proxy. 

However adding a check to make sure that these headers present and Authentication Mechanism was used will be able to protect us from accidentially making them open in POCO config. 

I'll create a ticket

KAbakumovAtl: Done, ticket created. Will do it shortly after we roll out the changes

atrigueiro (approved):

src/routes/api/api-router.ts
atrigueiro: [ARC-1282] for future reference

PR #1143, opened 2022-03-29
Arc 1282 add slauth headers verification

KAbakumovAtl:
Follow up on the previous PR.

Added a verification that SLAuth headers are set

Currently testing in staging

### DIFF
--- a/src/routes/api/api-router.ts
+++ b/src/routes/api/api-router.ts
@@ -29,11 +29,6 @@ ApiRouter.use(
 		const issuer = req.get("X-Slauth-Issuer");
 		const principal = req.get("X-Slauth-Principal");
 
-		if(!mechanism || mechanism === "open") {
-			res.status(401).json({error: "Open access not allowed"});
-			return;
-		}
-
 		req.log = req.log.child({slauth: {
 			mechanism,
 			issuer,
@@ -42,6 +37,12 @@ ApiRouter.use(
 			aaid: req.get("X-Slauth-User-Aaid"),
 			username: req.get("X-Slauth-User-Username")
 		}});
+
+		if(!mechanism || mechanism === "open") {
+			req.log.warn("Attempt to access Admin API without authentication")
+			res.status(401).json({error: "Open access not allowed"});
+			return;
+		}
 
 		req.log.info("API Request successfully authenticated");
 
--- a/src/routes/api/api-router.ts
+++ b/src/routes/api/api-router.ts
@@ -1,4 +1,4 @@
-import { Request, Response, Router } from "express";
+import {NextFunction, Request, Response, Router} from "express";
 import { param } from "express-validator";
 import rateLimit from "express-rate-limit";
 import RedisStore from "rate-limit-redis";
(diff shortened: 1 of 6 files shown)


KAbakumovAtl:
It works, tested in staging

src/routes/api/api-router.ts
```diff
@@ -20,6 +20,36 @@ ApiRouter.use(urlencoded({ extended: false }));
...
+		}
+
+		req.log = req.log.child({slauth: {
+			mechanism,
+			issuer,
+			principal,
+			userGroup: req.get("X-Slauth-User-Groups"),
+			aaid: req.get("X-Slauth-User-Aaid"),
+			username: req.get("X-Slauth-User-Username")
+		}});
```
mboudreau, line 44: You'd probably want to set this before the conditional check

src/routes/api/api-router.ts
```diff
@@ -20,6 +20,36 @@ ApiRouter.use(urlencoded({ extended: false }));
...
+
+ApiRouter.use(
+	async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+		const mechanism = req.get("X-Slauth-Mechanism");
+		const issuer = req.get("X-Slauth-Issuer");
+		const principal = req.get("X-Slauth-Principal");
+
+		if(!mechanism || mechanism === "open") {
+			res.status(401).json({error: "Open access not allowed"});
+			return;
```
mboudreau, line 34: Should we maybe log this as a warn when someone's trying to authenticate without correct credentials?

KAbakumovAtl: Makes sense, will do

mboudreau: Also, I'm not sure I'm a fan of this conditional.  When you're doing a security check, you check for what you expect to continue, not check for what _not_ to continue.

Shouldn't we also check the other headers? is there something atlassian specific in the issues, principal, etc?  Shouldn't we also check the usergroup to make sure that whoever they are in atlassian has access to this API?

mboudreau: @KAbakumovAtl :arrow_up:

src/routes/api/api-router.ts
```diff
@@ -20,6 +20,37 @@ ApiRouter.use(urlencoded({ extended: false }));
...
+			userGroup: req.get("X-Slauth-User-Groups"),
+			aaid: req.get("X-Slauth-User-Aaid"),
+			username: req.get("X-Slauth-User-Username")
+		}});
+
+		if(!mechanism || mechanism === "open") {
+			req.log.warn("Attempt to access Admin API without authentication")
+			res.status(401).json({error: "Open access not allowed"});
+			return;
+		}
```
mboudreau, line 45: Still need to fix this conditional.  You can't base your security on what not to expect.

src/routes/api/api-router.ts
KAbakumovAtl: No, it is the responsibility of POCO

src/routes/api/api-router.ts
KAbakumovAtl: Don't understand why. 
We are checking that headers present, we are not repeating POCO/SLAuth logic here. That's how it works in DSS and any other services which use SLAuth.

Our goal of checking this header to make sure that the SP is enabled (and that the endpoint is not public in the config). That's it.

Please refer to SLAuth docs go/slauth to get better understanding on how it works.

joshkay10 (approved):

PR #1144, opened 2022-03-31
ARC-1017 show IP blocked errors to user

mboudreau:
When a user tries to connect an organization, we are now showing a much more helpful message if that org has IP allowlisting enabled and misconfigured.  We are now pointing them to documentation on how to fix this correctly.


joshkay10 (approved):
/nit include the logger values as the object argument for splunks help
e.g logger.warn(**{value},** `message with ${value}`)

I see conflicts in my PRs future. :(

src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -187,6 +193,10 @@ export const GithubConfigurationGet = async (req: Request, res: Response, next:
 			log
 		);
 
+		// Sort to that orgs ready to be connected are at the top
+		const isReady = (i: MergedInstallation) => i.isAdmin && !i.isIPBlocked && i.syncStatus !== "FINISHED" && i.syncStatus !== "IN PROGRESS" && i.syncStatus !== "PENDING";
```
joshkay10: // todo - would be nice to type the syncstatus's too

joshkay10 (dismissed):

src/routes/github/configuration/github-configuration-get.ts
mboudreau: yeah, I already have that on the interface that uses it.

joshkay10 (approved):

PR #1146, opened 2022-04-04
ARC-1286 gh client jira config get

joshkay10:
use github client for get installations call on jira-configuration-get

Updated tests to use gh client and octokit.

### DIFF
--- a/src/routes/jira/configuration/jira-configuration-get.ts
+++ b/src/routes/jira/configuration/jira-configuration-get.ts
@@ -12,7 +12,6 @@ import { getCloudInstallationId } from "~/src/github/client/installation-id";
 import { GitHubAppClient } from "~/src/github/client/github-app-client";
 import { booleanFlag, BooleanFlags } from "~/src/config/feature-flags";
 
-
 const mapSyncStatus = (syncStatus: SyncStatus = SyncStatus.PENDING): string => {
 	switch (syncStatus) {
 		case "ACTIVE":
--- a/src/routes/jira/configuration/jira-configuration-get.ts
+++ b/src/routes/jira/configuration/jira-configuration-get.ts
@@ -64,13 +64,13 @@ const getInstallation = async (client: GitHubAPI, subscription: Subscription, lo
 		};
 
 	} catch (err) {
-		log?.error(
+		log.error(
 			{ installationId: gitHubInstallationId, error: err, uninstalled: err.code === 404 },
 			"Failed connection"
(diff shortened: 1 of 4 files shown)


src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -38,28 +42,35 @@ export const getInstallations = async (client: GitHubAPI, subscriptions: Subscri
 	};
 };
 
-const getInstallation = async (client: GitHubAPI, subscription: Subscription, log?: Logger): Promise<AppInstallation> => {
-	const id = subscription.gitHubInstallationId;
+const getInstallation = async (client: GitHubAPI, subscription: Subscription, log: Logger): Promise<AppInstallation> => {
```
mboudreau, line 45: You should be parsing which github client to use at a higher level and passing it to this function like we've done with the other PRs

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -38,28 +42,35 @@ export const getInstallations = async (client: GitHubAPI, subscriptions: Subscri
...
 	} catch (err) {
 		log?.error(
-			{ installationId: id, error: err, uninstalled: err.code === 404 },
+			{ installationId: gitHubInstallationId, error: err, uninstalled: err.code === 404 },
 			"Failed connection"
 		);
 		statsd.increment(metricError.failedConnection);
 
-		return Promise.reject({ error: err, id, deleted: err.code === 404 });
+		return Promise.reject({ error: err, gitHubInstallationId, deleted: err.code === 404 });
```
mboudreau, line 73: Is this right? I don't see the change to the `FailedAppInstallation` interface to use `gitHubInstallationId` instead of `id`

joshkay10: Turns out. You are correct. Will update to id: gitHubInstallationId

src/routes/jira/configuration/jira-configuration-get.ts
joshkay10: Did try originally but  we don't have the installation Id until we start mapping per subscription, so was allowing for different installation ids.

Could probably adjust the githubapp class to work but seems more effort than value at this point?

mboudreau: ah right.

mboudreau (approved):

PR #1149, opened 2022-04-05
ARC-328 import and improve Jira issue parser

mboudreau:
Importing the code and tests from jira issue parser into the codebase.  Improving the regex and tests to better reflect our use of it.  We cannot feature flag this easily without some massive refactoring because of the async/await hierarchy would need to change drastically.  I think the thorough testing and staging should be more than enough to derisk.


src/util/jira-utils.ts
```diff
@@ -57,3 +58,36 @@ interface Author {
...
+ * Parses strings for Jira issue keys for commit messages,
+ * branches, and pull requests.
+ *
+ * Accepted patterns:
+ *      - JRA-123 (all uppercase)
+ *      - jRA-123 (some uppercase, some lowercase - any order)
+ *      - jra-123 (all lowercase - any order)
+ *      - J2-123 (first part of key alphanumeric)
+ *      - [JRA-123] (keys inside square brackets)
+ *      - (JRA-123) (keys inside square brackets)
```
joshkay10, line 71: *(keys inside parenthesis)

src/util/jira-utils.ts
```diff
@@ -57,3 +58,36 @@ interface Author {
 	};
 }
 
+/**
+ * Parses strings for Jira issue keys for commit messages,
+ * branches, and pull requests.
+ *
+ * Accepted patterns:
```
joshkay10, line 65: This would be really useful on the Readme too🤔

joshkay10 (approved):

src/util/jira-utils.ts
mboudreau: Eh, seems a bit much to put it in the readme.  We'll just be duplicating the documentation for this function and it seems to be a very specific implementation that most won't care about until they change the function itself.

joshkay10 (approved):

PR #1150, opened 2022-04-05
WIP ARC-1090: delete builds and deployments on repository delete

thombergs:
When a user deletes a GitHub repo, the GitHub app currently deletes the commits, branches, and pullrequests from Jira, but not the builds and deployments.

This PR adds deletion of builds and deployments as well.

I chose not to feature flag this change as I tested it successfully in staging and deleting a repository is not a function that is used too often.

Issue view before deleting the repo in GitHub: 
image

Issue view after deleting the repo in GitHub:
image

No builds and deployments anymore :)


src/github/repository.ts
```diff
@@ -3,15 +3,30 @@ import { emitWebhookProcessedMetrics } from "utils/webhook-utils";
...
+	]);
+
 	const { webhookReceived, name, log } = context;
 
 	webhookReceived && emitWebhookProcessedMetrics(
 		webhookReceived,
 		name,
 		log,
-		jiraResponse?.status
+		202
```
thombergs, line 30: We can just return a `202` here, because the Jira client will throw an error when it gets a non-successful response, and then this code isn't being executed anyway.

src/jira/client/jira-client.ts
```diff
@@ -261,7 +261,8 @@ export const getJiraClient = async (
 				const payload = {
 					builds: data.builds,
 					properties: {
-						gitHubInstallationId
+						gitHubInstallationId,
+						repositoryId
```
thombergs, line 265: Adding the `repositoryId` as a property when sending a build, so we can delete by that repositoryId later.

src/jira/client/jira-client.ts
```diff
@@ -286,7 +294,8 @@ export const getJiraClient = async (
 				const payload = {
 					deployments: data.deployments,
 					properties: {
-						gitHubInstallationId
+						gitHubInstallationId,
+						repositoryId
```
thombergs, line 298: Adding the `repositoryId` as a property when sending a deployment, so we can delete by that repositoryId later.

src/github/repository.ts
```diff
@@ -3,15 +3,30 @@ import { emitWebhookProcessedMetrics } from "utils/webhook-utils";
...
+
+	const deleteDeploymentsPromise = jiraClient.deployment.delete(
+		context.payload.repository?.id
+	);
+
+	const deleteBuildsPromise = jiraClient.workflow.delete(
+		context.payload.repository?.id
+	);
+
+	await Promise.all([
```
bgvozdev, line 18: Do we have a test for this?

mboudreau (approved):

bgvozdev (changes requested):
Marking as "request changes" as there are unresolved comments and also no FF: cannot be released during thaw.

src/github/repository.ts
gxueatlassian: How about transaction/retry mechanism? Are bellow the intended mechanism?
1. This msg is taken from sqs, so assuming this call error, the msg will `stay` in sqs for retry (up to ? times)
2. The deletion is idempotent, so if some call failed, some calls success, it is okay, next retry is safe to call whole thing again (like the success calls, say `deleDevinfoPromise` won't error if called again with a deleted `id`)

rachellerathbone:
WIP that has been open for almost a year... killing.

PR #1156, opened 2022-04-07
Arc-1285 ghclient GitHub setup get

joshkay10:
Updated github detup get to use new github client instead of Octokit

### DIFF
--- a/src/routes/github/setup/github-setup-get.ts
+++ b/src/routes/github/setup/github-setup-get.ts
@@ -18,20 +18,18 @@ import { GitHubAPI } from "probot";
 		- Otherwise, render the setup page (POST).
 */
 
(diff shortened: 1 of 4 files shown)


src/routes/github/setup/github-setup-get.ts
```diff
@@ -12,17 +17,40 @@ import { Installation } from "models/installation";
 		- If we have the users Jira host, redirect to marketplace.
 		- Otherwise, render the setup page (POST).
 */
+
+const getInstallationData = async (githubAppClient: GitHubAppClient | GitHubAPI, githubInstallationId: number, logger: Logger) => {
```
joshkay10, line 21: Split the installation get to a seperate function for calrity and was able to handle the error cases more appropriately.

mboudreau (approved):

PR #1162, opened 2022-04-14
ARC-1072 include deployments and  build in backfill

joshkay10:
Updated the backfill process to include Deployments and Builds.

Reused most of the existing transform code from the webhooks.

**Actions read permissions is a new permission requirement**

Successful example
Screen Shot 2022-05-17 at 2 23 17 PM

If the Org has not accepted the new permissions we will display a warning triangle with a popup on hover with the following copy.

Screen Shot 2022-05-17 at 2 08 56 PM


src/config/interfaces.ts
```diff
@@ -153,82 +153,23 @@ interface WorkflowRunPullRequests {
 }
 
 interface WorkflowRun {
-	id: number;
```
joshkay10, line 156: Removing all unused types

src/models/reposyncstate.ts
```diff
@@ -204,6 +212,10 @@ export class RepoSyncState extends Model {
 			this.commitCursor = repo.lastCommitCursor;
 			this.pullStatus = repo.pullStatus;
 			this.pullCursor = repo.lastPullCursor?.toString();
+			this.buildStatus = repo.buildStatus;
+			this.buildCursor = repo.lastBuildCursor?.toString();
```
joshkay10, line 216: uses a page number for the cursor for builds, so tostring was required

src/sync/build.ts
```diff
@@ -0,0 +1,79 @@
...
+import { Octokit } from "@octokit/rest";
+import { LoggerWithTarget } from "probot/lib/wrap-logger";
+import { Repository } from "models/subscription";
+import { GitHubInstallationClient } from "../github/client/github-installation-client";
+import { transformWorkflow } from "../transforms/transform-workflow";
+import { WorkflowPayload } from "../config/interfaces";
+
+type BuildWithCursor = { cursor: number } & Octokit.ActionsListRepoWorkflowRunsResponse;
+
+const getTransformedBuilds = async (workflowRun, gitHubInstallationClient, logger) => {
```
joshkay10, line 11: Split this task out to simplify things a bit

src/sync/installation.ts
```diff
@@ -264,9 +267,18 @@ async function doProcessInstallation(app, data: BackfillMessagePayload, sentry:
 		const { edges, jiraPayload } = await execute();
 		if (jiraPayload) {
 			try {
-				await jiraClient.devinfo.repository.update(jiraPayload, {
-					preventTransitions: true
-				});
+				switch (task) {
```
bgvozdev, line 270: Feature flag?

src/transforms/transform-deployment.ts
```diff
@@ -101,7 +100,7 @@ async function getCommitMessagesSinceLastSuccessfulDeployment(
 // https://developer.atlassian.com/cloud/jira/software/rest/api-group-builds/#api-deployments-0-1-bulk-post
 // Deployment state - Jira: Can be one of unknown, pending, in_progress, cancelled, failed, rolled_back, successful
 function mapState(state: string): string {
-	switch (state) {
+	switch (state.toLowerCase()) {
```
bgvozdev, line 103: Is it happening? Is there a guarantee that `state` is always defined/not null?

bgvozdev (changes requested):
Some unresolved comments (including author's one ;) ) + no feature flag, cannot be merged during thaw

src/transforms/transform-deployment.ts
joshkay10: unsure about always defined, we could `state?.toLowerCase()` to protect that case.

src/sync/installation.ts
joshkay10: Added feature flag, will add comments at location. ended up being awkward to put in nicely due to the async requirements

src/sync/installation.ts
```diff
@@ -418,6 +435,11 @@ export const processInstallation =
 		return async (data: BackfillMessagePayload, sentry: Hub, logger: LoggerWithTarget): Promise<void> => {
 			const { installationId, jiraHost } = data;
 
+			// This is prefered at taskType definition, but feature flag requires async so was moved to starting of process
+			if (!await booleanFlag(BooleanFlags.BACKFILL_FOR_BUILDS_AND_DEPLOYMENTS, true, jiraHost)) {
+				taskTypes = taskTypes.filter((task) => task !== "build" && task !== "deployment")
+			}
+
```
joshkay10, line 442: Feature flag added at this point, ideally wanted it up top when taskTypes is defined but need async for flag,
this was the least intrusive way.

mboudreau:
@joshkay10 don't forget to change the `countSyncedReposFromSubscription` in RepoSyncState with your new statuses :)

joshkay10:
> @joshkay10 don't forget to change the `countSyncedReposFromSubscription` in RepoSyncState with your new statuses :)

src/models/reposyncstate.ts
```diff
@@ -51,7 +51,9 @@ export class RepoSyncState extends Model {
 			where: {
 				pullStatus: "complete",
 				branchStatus: "complete",
-				commitStatus: "complete"
+				commitStatus: "complete",
+				buildStatus: "complete",
+				deploymentStatus: "complete"
```
joshkay10, line 56: @mboudreau 
> @joshkay10 don't forget to change the `countSyncedReposFromSubscription` in RepoSyncState with your new statuses :)

This should do the trick?

mboudreau: yup yup :D

mboudreau:
Please change PR title to reflect jira issue

src/sync/installation.ts
```diff
@@ -252,6 +256,14 @@ async function doProcessInstallation(app, data: BackfillMessagePayload, sentry:
...
 				return await processor(logger, github, gitHubInstallationClient, jiraHost, repository, cursor, perPage);
 			} catch (err) {
+				// In the event that the customer has not accepted the required permissions.
+				// We will continue to process the data per usual while omitting the tasks the app does not have access too.
+				if (err.status === 403 && task === "build" && err.message?.includes("Resource not accessible by integration")) {
+					await subscription?.update({ syncWarning: "Invalid permissions for build task" });
+					logger.error({ err }, `Invalid permissions for build task`);
+					// Return undefined objects so the sync can complete while skipping this task
+					return 	{ edges: undefined, jiraPayload: undefined };
+				}
```
mboudreau, line 266: This is probably not the best way to do this.  Furthermore, the same thing can happen with any other tasks and should probably be handled.

src/sync/installation.ts
```diff
@@ -275,9 +287,18 @@ async function doProcessInstallation(app, data: BackfillMessagePayload, sentry:
...
+					case "build":
+						await jiraClient.workflow.submit(jiraPayload);
+						break;
+					case "deployment":
+						await jiraClient.deployment.submit(jiraPayload);
+						break;
+					default:
+						await jiraClient.devinfo.repository.update(jiraPayload, {
+							preventTransitions: true
+						});
```
mboudreau, line 300: I'm not sure I understand this point of the switch statement.

src/transforms/transform-deployment.ts
```diff
@@ -101,7 +100,7 @@ async function getCommitMessagesSinceLastSuccessfulDeployment(
 // https://developer.atlassian.com/cloud/jira/software/rest/api-group-builds/#api-deployments-0-1-bulk-post
 // Deployment state - Jira: Can be one of unknown, pending, in_progress, cancelled, failed, rolled_back, successful
 function mapState(state: string): string {
-	switch (state) {
+	switch (state?.toLowerCase()) {
```
mboudreau, line 103: if the state an be undefined, then you should specify it in the params with an optional annotation.

views/jira-configuration.hbs
```diff
@@ -195,11 +211,10 @@
                             &#x21bb;
                           </button>
                           <input type="hidden" id="_csrf" name="_csrf" value="{{../csrfToken}}">
-                          <div class="jiraConfiguration__retryMsg">Retry</div>
+                          <div class="jiraConfiguration__retryMsg">Retry2</div>
```
mboudreau, line 214: Retry2?

src/sync/deployment.ts
```diff
@@ -0,0 +1,73 @@
+import { GitHubAPI } from "probot";
+import { Repository } from "models/subscription";
+import { WebhookPayloadDeploymentStatus } from "@octokit/webhooks";
+import { GitHubInstallationClient } from "../github/client/github-installation-client";
+import { LoggerWithTarget } from "probot/lib/wrap-logger";
+import { transformDeployment } from "../transforms/transform-deployment";
+
+const fetchDeployments = async (gitHubInstallationClient: GitHubInstallationClient, repository: Repository, cursor?: string | number, perPage?: number) => {
+	const deploymentData = await gitHubInstallationClient.getDeploymentsPage(repository.owner.login, repository.name, perPage, cursor);
+	const edges = deploymentData.repository.deployments.edges;
```
mboudreau, line 10: I'd probably set the `|| []` on edges here :)

src/sync/deployment.ts
```diff
@@ -0,0 +1,73 @@
...
+	const deployments = edges?.map(({ node: item }) => item) || [];
+
+	return {
+		edges,
+		deployments
+	};
+};
+
+const getTransformedDeployments = async (deployments, _github: GitHubAPI, gitHubInstallationClient: GitHubInstallationClient, jiraHost: string, logger: LoggerWithTarget) => {
+	const transformTasks = await deployments.reduce(async (acc, current) => {
```
mboudreau, line 20: I don't recommend mixing awaits with reduce.  Reduce is complex enough as is without having to think about promises

mboudreau: instead, you probably want to do a `await Promise.all(deployments.map(...))`

src/sync/installation.ts
joshkay10: For all current tasks except build/deployment we use the `/rest/devinfo/0.10/bulk` endpoint, however deployment and build need to hit a different path so we are just handling where to send it based on task... could convert it to a function to return the appropriate function instead....?

views/jira-configuration.hbs
joshkay10: oh my!!! good catch!

src/sync/deployment.ts
joshkay10: yes, agree was replacing a map.filter, but the reduce/promises do make it unnecessarily more complicated.

src/github/client/github-client-interceptors.ts
```diff
@@ -111,6 +111,14 @@ export const handleFailedRequest = (logger: Logger) =>
...
 				return Promise.reject(new BlockedIpError(error, status));
 			}
+
+			if (status === 403 && response.data?.message?.includes("Resource not accessible by integration")) {
+				logger.warn({
+					err: error,
+					remote: response.data?.message
+				}, "unauthorized");
+				return Promise.reject(new InvalidPermissionsError(error, status));
+			}
```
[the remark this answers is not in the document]
joshkay10: Why this error happens or why this block of code exists?

This error is the result of the app not having sufficient permissions. Specifically in this use case, they don't have workflow read. Most likely they have not accepted updated permission set.

This block of code exists so we can 'bubble' up the error message at `installation.ts` or `build.ts` level intead of a generic `Error executing Axios Req.....`


Not a massive fan of relying on a hardcoded string to match a message is maintained outside of our app....

src/sync/installation.ts
joshkay10: Yeah, not a big fan of this,
In terms of placement of the logic, originally tried to isolate it within the build function itself, but then thought a more generic approach was beneficial, but added on that task === "build" last minute :(

Keen to discuss alternatives though!

src/github/client/github-client-interceptors.ts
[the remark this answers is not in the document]
mboudreau: agreed, but github isn't really giving us much choice.  I do think we need to centralize all possible GH error handling though, but that will be another PR

src/sync/installation.ts
mboudreau: I think leaving it there for now is okay as long as there's a TODO as we need a better way to manage GH errors globally.  I don't think we need to specify `task === 'build'` because it should be showing for any task that might not have access

src/sync/installation.ts
mboudreau: oh shoot, I didn't see it was hitting workflow vs deployment in my late night haze, nvm, carry on :)

src/sync/deployment.ts
joshkay10: Simplified!

src/sync/installation.ts
joshkay10: added a todo.
Dropped the `task === "build"`
updated the `invalid permissions for build task` to a template string `invalid permissions for${task} task`


Am currently updating the messaging on the UI to be generic also.

joshkay10: Any failure on the permissions side will be treated the same, be it builds, pulls, commits etc

mboudreau (approved):

mboudreau (approved):

src/sync/installation.ts
```diff
@@ -252,6 +256,16 @@ async function doProcessInstallation(app, data: BackfillMessagePayload, sentry:
 			try {
 				return await processor(logger, github, gitHubInstallationClient, jiraHost, repository, cursor, perPage);
 			} catch (err) {
+				// TODO - need a better way to manage GitHub errors globally
+				// In the event that the customer has not accepted the required permissions.
+				// We will continue to process the data per usual while omitting the tasks the app does not have access too.
+				// The GraphQL errors do not return a status so we check 403 or undefined
+				if ((err.status === 403 || err.status === undefined) && err.message?.includes("Resource not accessible by integration")) {
+					await subscription?.update({ syncWarning: `Invalid permissions for ${task} task` });
```
atrigueiro, line 264: I'm not too sure what's wrong with this piece of code - not even sure if the problem is actually here - but I can't see the syncWarning message for deployments when I haven't accepted permissions in staging.
Screen Shot 2022-05-24 at 3 23 14 pm
Screen Shot 2022-05-24 at 3 08 19 pm
Screen Shot 2022-05-24 at 3 10 18 pm

atrigueiro (dismissed):
SyncWarning message not appearing in staging :(

PR #1165, opened 2022-04-20
[ARC-328] fixing regex temporarily and adding tests

mboudreau:
Reverting back to previous regex to fix devinfo api issues with "misformated" issue keys


joshkay10 (dismissed):

src/util/jira-utils.ts
```diff
@@ -58,6 +59,10 @@ interface Author {
 	};
 }
 
+let regexFixFeature = false;
+onFlagChange(BooleanFlags.REGEX_FIX, async () => {
```
joshkay10, line 63: A fun way around the async flag issue :)
Assuming the onflagchange event is driven from changing it on launch darkly?

What happens if the flag is true on deployment, looks like it can only set regexFixFeature to true on a flag change event. But if the flag start true it will never fire?

mboudreau: as per documentation, this isn't an "on flag change on server" but a "on flag change on client" which includes when the client initializes and get the values from the server.  Either way, this won't be long running.  Essentially just leave it there for a few days after deployment then remove it ASAP.

bgvozdev: I believe the values are pulled from LD each 30 seconds (which is the default config  I guess https://github.com/launchdarkly/node-server-sdk/blob/ddfcd805c400d5445f5ad94b350cac7c241ef4db/configuration.js#L17), so the value should be updated.

bgvozdev (approved):

joshkay10 (approved):

PR #1168, opened 2022-04-21
Better logging

mboudreau:
Adding better debug logs, setting staging to log level debug, removing pollinator runonce check on staging as it won't actually test the queues correctly.


atrigueiro:
I'm not too sure if I agree with removing the pollinator check in staging, as it has saved us from previous incidents in prod before.
My main 2 questions are:
* Why's only an issue in staging and not in prod?
* Can we not replicate the prod env in staging so we can have the check for both envs?

mboudreau:
> * Why's only an issue in staging and not in prod?
It's an issue on both.  We just haven't broken it out of sheer luck.  If an issue is introduced on the sync side, it probably won't ever be caught until we actually hit production because all runonce checks don't disable other SQS listeners and it doesn't have it's own queue, which means that statistically, the currently deployed infra has a much higher chance of receiving the message than the new one.

Either way, pollinator tests are inadequate and is not an appropriate solution for doing e2e testing correctly for things like SQS.

> * Can we not replicate the prod env in staging so we can have the check for both envs?

The only way to "test" if the prod stack is going to work is to fully deploy to staging, confirm the old stack is removed, then do the checks (not using runonce).  However, something we should investigate is if we can do our own e2e tests within a local docker container - if we can accomplish this, we can guarantee that the currently deployed stack wouldn't affect the tests and would catch 90+% of issues we might encounter.  The rest would mostly have to do with micros implementation and making sure things are hooked up correctly, which hopefully staging can achieve to gain 100% confidence when deploying to prod.

joshkay10 (approved):

PR #1170, opened 2022-04-25
ARC-1261 limit message length

joshkay10:
Data depot is getting 400's from github integration.

Data sent to Jira is fine, but the migration project has noticed errors due to different rules.

One of those rules is a maximum character limit on commit messages, max being 1024 in length.

I have limited the commit message to 1024 characters.

Issue keys are pulled out before the substring.

Tested in stage, errors are no longer being generated for
'submitDevInfo failed parse/validate: [ErrorMessage(message='message' cannot exceed 1024 characters, key=devInformation.repository.commit.message.tooLong, errorTraceId=null)]'


src/sync/transforms/branch.ts
```diff
@@ -69,7 +70,7 @@ const mapCommit = (commit) => {
 		hash: commit.oid,
 		id: commit.oid,
 		issueKeys: issueKeys || [],
-		message: commit.message,
+		message: commit?.message?.substring(0, MAX_COMMIT_MESSAGE_LENGTH),
```
mboudreau, line 73: `commit` will always be non-null.  the message value has to be non-null or else jira will reject it.  So maybe add a default in case?  `commit.message?.substring(0, MAX_COMMIT_MESSAGE_LENGTH) || ""`

src/transforms/transform-branch.ts
```diff
@@ -5,6 +5,7 @@ import { WebhookPayloadCreate } from "@octokit/webhooks";
 import { generateCreatePullRequestUrl } from "./util/pull-request-link-generator";
 import { GitHubInstallationClient } from "../github/client/github-installation-client";
 import { JiraBranchData, JiraCommit } from "src/interfaces/jira";
+const MAX_COMMIT_MESSAGE_LENGTH = 1024;
```
mboudreau, line 8: I don't like the idea of having multiple places for this.  I mean, I don't like the idea that we're duplicating the logic between sync and webhook either...

Maybe this constant can sit in the env.ts file?

src/transforms/transform-branch.ts
```diff
@@ -18,7 +19,7 @@ const getLastCommit = async (github: GitHubInstallationClient, webhookPayload: W
 		hash: sha,
 		id: sha,
 		issueKeys,
-		message: commit.message,
+		message: commit?.message?.substring(0, MAX_COMMIT_MESSAGE_LENGTH),
```
mboudreau, line 22: same thing as above

src/transforms/transform-commit.ts
```diff
@@ -16,7 +17,7 @@ export const mapCommit = (commit): JiraCommit | undefined => {
 		hash: commit.oid,
 		id: commit.oid,
 		issueKeys,
-		message: commit.message,
+		message: commit?.message?.substring(0, MAX_COMMIT_MESSAGE_LENGTH),
```
mboudreau, line 20: hm, I'm starting to thing we need to create a function that does this as a util so we don't have to replicate logic.

joshkay10: Sweet, I will move it all to one location.

src/util/jira-utils.ts
```diff
@@ -43,6 +43,11 @@ export const getJiraAuthor = (...authors: (Author | undefined)[]): JiraAuthor =>
 	};
 };
 
+export const limitCommitMessage = (message: string ): string => {
+	const MAX_COMMIT_MESSAGE_LENGTH = 1024;
+	return message.substring(0, MAX_COMMIT_MESSAGE_LENGTH);
```
joshkay10, line 48: I believe i can omit the `|| ""` in the return now since type string is required?

thombergs (approved):

src/util/jira-utils.ts
mboudreau: the type is there, but doesn't mean the data will always be not undefined in runtime.  I'd probably set the function definition as `(message = "")` so we have a default if nothing is specified :)

src/util/jira-utils.ts
```diff
@@ -43,6 +43,11 @@ export const getJiraAuthor = (...authors: (Author | undefined)[]): JiraAuthor =>
 	};
 };
 
+export const limitCommitMessage = (message: string): string => {
+	const MAX_COMMIT_MESSAGE_LENGTH = 1024;
```
mboudreau, line 47: I'd probably add length as an input to the function with a default of 1024: `(message = "", length = 1024):string`

mboudreau (approved):

PR #1171, opened 2022-04-26
ARC-1312 fix discovery for enterprise

mboudreau:
Making the repository discovery part of the backfill process since it is resilient to rate limits and other issues.  Need to use graphql to get a proper cursor while iterating through repos as the rest API only provides pagination and might miss repos if new ones are being added at the same time.


src/sync/discovery.ts
```diff
@@ -1,62 +1,86 @@
...
+	logger: LoggerWithTarget,
+	_github: GitHubAPI,
+	newGithub: GitHubInstallationClient,
+	jiraHost: string,
+	_repository: Repository,
+	cursor?: string | number,
+	perPage?: number
+): Promise<TaskPayload> => {
+	logger.debug("Repository Discovery: started");
+	const installationId = newGithub.githubInstallationId.installationId;
```
joshkay10, line 42: Will there be a difference for GHE installationId or will this work for both?

Just thinking it might be handy to wrap this function instead of exposing the property.....however might pay to wait until GHE and not preempt without all the GHE details ¯\_(ツ)_/¯

joshkay10 (approved):

src/sync/discovery.ts
mboudreau: frankly, I'm not sure and didn't want to work on a solution before we know it's a problem.  I think this is fine for now.  We still have a lot of things to improve along the way.

src/config/feature-flags.ts
```diff
@@ -13,7 +13,6 @@ const launchdarklyClient = LaunchDarkly.init(envVars.LAUNCHDARKLY_KEY || "", {
 
 export enum BooleanFlags {
 	MAINTENANCE_MODE = "maintenance-mode",
-	SIMPLER_PROCESSOR = "simpler-processor",
```
thombergs, line 16: Please try to avoid changes that don't address the PR topic. I know that it's annoying to create a separate PR for these changes, but it gives so much more confidence when reviewing the PR.

thombergs (approved):

PR #1173, opened 2022-04-26
FD-27717 remove feature flag

thombergs:
Removes the feature flag that controls usage of the new GitHub client in the processing of pull request events.

The feature flag has been enabled in prod for quite a while:

image


src/sync/pull-request.ts
```diff
@@ -41,41 +40,31 @@ type PullRequestWithCursor = { cursor: number } & Octokit.PullsListResponseItem;
 
 export const getPullRequestTask = async (
 	logger: LoggerWithTarget,
-	github: GitHubAPI,
+	_oldGithub: GitHubAPI,
 	newGithub: GitHubInstallationClient,
-	jiraHost: string,
+	_jiraHost: string,
```
thombergs, line 45: Had to rename these vars, as they aren't used anymore (couldn't remove them from the signature completely, because they are still used in other implementations)

joshkay10 (dismissed):

atrigueiro (approved):

PR #1182, opened 2022-05-08
Arc 1302 yarn replace

joshkay10:
Replaced NPM with yarn.

~~don't forget to delete NODE_MODULES first~~


bgvozdev (approved):

mboudreau:
I've just tried it locally.  No need to delete `node_modules` folder from what I could see.

I also added yarn as part of the `engines` and `volta` configuration of package.json.  Just needs the yarnrc file and I think we're good to go :)

package.json
```diff
@@ -134,6 +134,7 @@
 		"yaml-lint": "^1.2.4"
 	},
 	"volta": {
-		"node": "14.17.0"
```
joshkay10, line 137: NICE!!

atrigueiro:
Looks good to me. We also need to make changes in our deployment pipelines, should probably try and deploy this to ddev/staging first before merging, otherwise we'll break our prod deployment as well.

src/config/.yarnrc
```diff
@@ -0,0 +1,2 @@
+registry "https://registry.npmjs.org/"
+"@atlassian:registry" "https://registry.npmjs.org/"
```
mboudreau, line 2: this needs to be at the top level of the project.

mboudreau (changes requested):

src/config/.yarnrc
joshkay10: moved, good catch!

bgvozdev (dismissed):

mboudreau (dismissed):

.github/workflows/on-push.yml
```diff
@@ -8,32 +8,36 @@ jobs:
...
-        uses: actions/cache@v1
+      - uses: actions/cache@v2
         with:
-          path: ./node_modules
-          key: ${{ runner.os }}-${{hashFiles('**/package-lock.json')}}-node_modules
-      - run: npm install
-      - run: npm run db:test
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+      - run: volta pin yarn@latest
```
mboudreau, line 38: I don't think you need to pin it.  You were missing yarn in volta config :)

.github/workflows/on-push.yml
```diff
@@ -8,32 +8,36 @@ jobs:
...
+      - uses: actions/cache@v2
         with:
-          path: ./node_modules
-          key: ${{ runner.os }}-${{hashFiles('**/package-lock.json')}}-node_modules
-      - run: npm install
-      - run: npm run lint
-      - run: npm run build
+          path: '**/node_modules'
+          key: ${{ runner.os }}-modules-${{ hashFiles('**/yarn.lock') }}
+      - run: volta pin yarn@latest
```
mboudreau, line 19: I don't think you need to pin it.  You were missing yarn in volta config :)

.github/workflows/on-push.yml
```diff
@@ -8,32 +8,36 @@ jobs:
 
     steps:
       - uses: actions/checkout@v1
+      - uses: actions/setup-node@v3
+        with:
+          node-version: 14
```
mboudreau, line 13: pretty sure volta takes care of that too

.github/workflows/on-push.yml
joshkay10: 1 way to find out!

mboudreau (approved):

atrigueiro (approved):

PR #1184, opened 2022-05-09
ARC-964 Added script file to create GithubApps table

Harminder84:


Harminder84:
issue with github user

PR #1185, opened 2022-05-10
ARC-964 create githubapps table

Harminder84:


db/migrations/20220509002323-create-github-app.js
```diff
@@ -0,0 +1,47 @@
...
+				type: Sequelize.ENUM("cloud", "ghe")
+
+			},
+			githubClientId: {
+				allowNull: false,
+				type: Sequelize.STRING
+			},
+			secrets: {
+				allowNull: false,
+				type: Sequelize.BLOB
```
Harminder84, line 31: githubClientSecret, webhookSecret and privateKey will be stored in secrets vault.

mboudreau:
Don't we want to include the model as part of this?  Also, how are we adding the current staging/prod apps to this DB?

db/migrations/20220509002323-create-github-app.js
thombergs: What do you mean with "secrets vault"? Does that mean they will be stored in this `secrets` blob, but encrypted?

mboudreau: Blob is non-encrypted.  DB is encrypted at rest.  We need to be careful how we name things since we have Github Client Secret and Github Webhook Secret.  We also need to go over how we'll be listening to webhooks for each github enterprise instance as we need listening code for each one or have a way to decipher every event by iterating through all  github instances to see which webhook secret will work with the event, which I'm not sure if actually scalable.

We need to brainstorm some solutions for this.

Harminder84: We need to store githubClientSecret, webhookSecret and privateKey as encrypted values. The idea here is to put all these values in a secure vault. The vault will be stored as blob in secrets column. The same pattern is being used in installations table with sequelize-encrypted library.

db/migrations/20220509002323-create-github-app.js
```diff
@@ -0,0 +1,47 @@
...
+				type: Sequelize.INTEGER
+			},
+			githubBaseUrl: {
+				allowNull: false,
+				type: Sequelize.STRING
+			},
+			githubApiBaseUrl: {
+				allowNull: false,
+				type: Sequelize.STRING
+			},
```
mboudreau, line 19: I don't think we need this as we can figure out the API url based on if it's a ghe instance or not

db/migrations/20220509002323-create-github-app.js
```diff
@@ -0,0 +1,47 @@
...
+			githubServerType: {
+				allowNull: false,
+				type: Sequelize.ENUM("cloud", "ghe")
+
+			},
```
mboudreau, line 24: if we only have 2 choices, we could just make this a boolean

db/migrations/20220509002323-create-github-app.js
mboudreau: ah right, didn't know that was the idea without he model being set

src/models/git-hub-server-app.ts
```diff
@@ -0,0 +1,38 @@
...
+	id: {
+		type: DataTypes.INTEGER,
+		primaryKey: true,
+		allowNull: false,
+		autoIncrement: true
+	},
+	uuid: {
+		type: DataTypes.STRING,
+		unique: true
+	},
```
mboudreau, line 16: I believe that sequelize has a UUID datatype and can create a default value on model creation.  Also, you need to set the `allowNull: false` here.

```
uuid: {
    type: DataTypes.UUID,
    defaultValue: Sequelize.UUIDV4,
   unique: true,
   allowNull: false
}
```

mboudreau (approved):

PR #1191, opened 2022-05-16
ARC-1294 Fix Retry Logic

mboudreau:
Adding feature flag to basically ignore the retry logic and always retry because sometimes jira just doesn't accept the devinfo payload and needs to be retried.  This jira error is being investigated further but in the meantime, we need something that works for our customers.  It's started happening since mid march and shows the error `HTTP 403 - The JWT token used does not correspond to an app that defines the jiraDevelopmentTool module, or the app does not define the 'WRITE' scope`.  It seems that the connect app authentication layer is blocking the request.


joshkay10 (dismissed):

atrigueiro (approved):

src/sync/installation.ts
```diff
@@ -157,7 +157,10 @@ const getEnhancedGitHub = async (app: Application, installationId) =>
  * with a smaller request (i.e. with fewer pages).
  * @param err the error thrown by Octokit.
  */
-export const isRetryableWithSmallerRequest = (err): boolean => {
+export const isRetryableWithSmallerRequest = async (err): Promise<boolean> => {
+	if(await booleanFlag(BooleanFlags.RETRY_ALL_ERRORS, false)) {
+		return true;
```
bgvozdev, line 162: Shouldn't we check if this was a 403?

PR #1193, opened 2022-05-17
ARC-1239: Return 404 instead of 503 if Jira instance is suspended.

atrigueiro:
This PR adds a check after we receive a 503 error from Jira - we now validate if the Jira instance is suspended by checking the `status` endpoint of that instance. If so, we change the error code to 404 on our end and log it as a warning.

Tested in staging - triggered a backfill for my suspended Jira instance using the admin API.

Screen Shot 2022-05-17 at 6 26 07 pm

This will, hopefully, improve our SLOs, since 503 errors from Jira represent the majority of our SLO failure events.


joshkay10 (approved):

mboudreau (approved):
nice

PR #1199, opened 2022-05-20
ARC-1239: Get the right status from AxiosError

atrigueiro:
🤦‍♀️

### DIFF
--- a/src/jira/client/axios.ts
+++ b/src/jira/client/axios.ts
@@ -149,10 +149,12 @@ const instrumentRequest = (response) => {
  * @param logger
  */
 const instrumentFailedRequest = (instance: AxiosInstance, logger: Logger) => {
-	return async (error:AxiosError) => {
+	return async (error: AxiosError) => {
 		instrumentRequest(error?.response);
-		logger.info(`ERROR OBJ IS ${error}`)
-		if (error.response?.status === 503) {
+		logger.info(`ERROR OBJ IS ${error}`);
+		logger.info(`ERROR RESPONSE IS ${error.response}`);
+		logger.info(`ERROR RESPONSE CODE IS ${error.response?.status}`);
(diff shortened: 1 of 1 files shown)


joshkay10 (approved):
🤟

src/jira/client/axios.ts
```diff
@@ -149,13 +149,13 @@ const instrumentRequest = (response) => {
...
-	return async (error) => {
+	return async (error: AxiosError) => {
 		instrumentRequest(error?.response);
-		if (error?.statusCode === 503) {
+		if (error.response?.status === 503) {
 			const statusResponse: AxiosResponse = await instance.get("/status");
 			if (statusResponse?.status === 503) {
 				logger.info(`503 from Jira: Jira instance ${instance} is suspended or does not exist. Returning 404 to our application.`);
-				error.statusCode = 404;
+				error.response.status = 404;
```
bgvozdev, line 158: Looks like this codepath is not cover by any tests. Do you think it worth adding one?

mboudreau: yeah, I think there's some value in adding a test for this codepath as it would have caught the issue before going to production.

src/jira/client/axios.ts
```diff
@@ -149,13 +149,13 @@ const instrumentRequest = (response) => {
  * @param logger
  */
 const instrumentFailedRequest = (instance: AxiosInstance, logger: Logger) => {
-	return async (error) => {
+	return async (error: AxiosError) => {
 		instrumentRequest(error?.response);
-		if (error?.statusCode === 503) {
+		if (error.response?.status === 503) {
 			const statusResponse: AxiosResponse = await instance.get("/status");
```
mboudreau, line 155: I'm pretty sure that  doing an await on that request will throw an error if it returns a 503.  You need to either catch it or do a try/catch around it

PR #1200, opened 2022-05-20
Revert "ARC-1239: Get the right status from AxiosError"

atrigueiro:
Reverts atlassian/github-for-jira#1199

Will need to wrap it around a try/catch to actually catch the 503.


atrigueiro:
Raising a follow up PR soon.

PR #1201, opened 2022-05-23
ARC-1239: Adding try/catch and a test

atrigueiro:


src/jira/client/axios.test.ts
```diff
@@ -72,6 +72,24 @@ describe("Jira axios instance", () => {
...
+		let error;
+		try {
+			await getAxiosInstance(jiraHost, "secret").post("/foo/bar", requestPayload);
+		} catch (e) {
+			error = e;
+		}
+
+		expect(error?.status).toEqual(404);
+		expect(error?.message).toEqual("Error executing Axios Request HTTP 404 - Bad REST path, or Jira instance not found or temporarily suspended.");
+	});
```
mboudreau, line 93: You should also check that if an error happens with the original request, but the instance returns 200, that it still throws the original error :)

mboudreau (approved):
:clap:

bgvozdev (approved):

yvonnevictoria (dismissed):

yvonnevictoria (approved):

PR #1207, opened 2022-05-24
SAI-11777 Refactor to split retrieval of commits from processing

jason-hooker:
**What's in this PR?**

- Split retrieval of commit data from processing commit messages.
- Refactored two calling locations
- Refactored tests

**What's next?**

- Add code to extract list of commit hashes from commit data
- Add feature flag to control sending commits in deployment data object
- Add list of associated commit hashes to deployment data object


src/transforms/transform-deployment.ts
```diff
@@ -88,11 +88,13 @@ async function getCommitMessagesSinceLastSuccessfulDeployment(
...
 	};
 
-	return await getAllCommitMessagesBetweenReferences(
+	const commitSummaries = await getAllCommitsBetweenReferences(
 		compareCommitsPayload,
 		useNewClient ? newGitHubClient : github,
 		logger
 	);
+
+	return await extractMessagesFromCommitSummaries(commitSummaries);
```
mboudreau, line 97: I'm not a fan of this pattern.  The function just wants the messages and nothing else, why does it need to call an api, then extract the data it needs after the fact?  It's adding a lot more emphasis on the developer to know the system deeply to use this pattern correctly.  

Instead, I'd suggest you just make 2 function, one that calls to get the API commit data only, and another one that does the same thing, but then extract only the messages.  The latter would of course call the former.  You greatly reduce the code entropy for any callers by doing this.

jason-hooker: Funnily enough, I did actually implement it that way for a while, then changed to this. Will refactor and resubmit.

jason-hooker: I'll fix this for the usage in `transform-workflow.ts` because that will not change again for this feature, however, for this usage, I'll leave it as is, because in the next PR, I'll also be extracting commit hashes from the same commit summary array.

mboudreau (approved):

PR #1214, opened 2022-05-25
SAI-11777 Modified to send related commits (controlled by FF) with deployment entities

jason-hooker:
**What's in this PR?**

- Added code to send related commits with deployments (as a JiraCommitAssociation)
- Added feature flag to control whether to send related commits with deployments
- Added tests


src/interfaces/jira.ts
```diff
@@ -158,6 +159,16 @@ export interface JiraAssociation {
 	values: string[];
 }
 
+export interface JiraCommitKey {
+	commitHash: string;
+	repositoryId: string;
+}
+
+export interface JiraCommitAssociation {
```
mboudreau, line 167: Instead of creating 2 types for the same association, why not just include both in the same?
```
interface JiraAssociation {
   associationType: string;
   values: string[] | JiraCommitKey[];
}
```

src/interfaces/jira.ts
```diff
@@ -138,6 +138,7 @@ export interface JiraDeployment {
 		displayName: string;
 		type: string;
 	};
+	associations?: Array<JiraAssociation | JiraCommitAssociation>;
```
mboudreau, line 141: We prefer to use the `[]` annotation for arrays.  Also, I think this is more readable if we do `JiraAssociation[]` if you follow my comment below.

src/transforms/transform-deployment.ts
```diff
@@ -57,7 +61,7 @@ async function getCommitMessagesSinceLastSuccessfulDeployment(
 	newGitHubClient: GitHubInstallationClient,
 	useNewClient: boolean,
 	logger: LoggerWithTarget
-): Promise<string | void | undefined> {
+): Promise<Array<CommitSummary> | void | undefined> {
```
mboudreau, line 64: `CommitSummary[]`

src/transforms/transform-deployment.ts
```diff
@@ -155,6 +157,26 @@ export function mapEnvironment(environment: string): string {
 	return jiraEnv;
 }
 
+// Maps commit summaries to a commit association containing the commit keys (commit hash and repository id)
+function mapCommitSummariesToCommitAssociation(
```
mboudreau, line 161: we prefer to use arrow functions with `const` instead of `function`

src/transforms/transform-deployment.ts
```diff
@@ -155,6 +157,26 @@ export function mapEnvironment(environment: string): string {
...
+	commitSummaries: Array<CommitSummary>,
```
mboudreau, line 162: `CommitSummary[]`

src/transforms/transform-deployment.ts
```diff
@@ -155,6 +157,26 @@ export function mapEnvironment(environment: string): string {
...
+	repositoryId: string
+): JiraCommitAssociation {
+
+	const commitKeys = commitSummaries
```
mboudreau, line 166: if you're doing `commitSummaries?.` if means that commitSummaries can be undefined.  As such, the param needs to specify that with `commitSummaries?: Array<CommitSummary>`.  Also, if that's the case, it means that `commitKeys` can be undefined as well and the devinfo API will definitely fail as it wouldn't accept an association with missing values.

IMO, we've seen enough situations that it is possible that data might be missing, so we need to handle this edge case.  I think the function needs to return `undefined` if `commitSummaries` is undefined

src/transforms/transform-deployment.ts
```diff
@@ -182,7 +205,13 @@ export const transformDeployment = async (githubClient: GitHubAPI, newGitHubClie
 			logger
 		);
 
+		const allCommitsMessages = commitSummaries ? await extractMessagesFromCommitSummaries(commitSummaries) : "";
```
mboudreau, line 208: I'd rather the function handle undefined `commitSummaries` than have it outside of it.  I'm not a fan of the 2 different types being returned.  one is a `CommitSummary[]` and the other is a string.  I think the `undefined` route should return either `undefined` or `[]`.  I tend to lean towards the latter as there's less possibilities of runtime errors being thrown.

src/transforms/transform-deployment.ts
```diff
@@ -219,7 +248,8 @@ export const transformDeployment = async (githubClient: GitHubAPI, newGitHubClie
 				id: deployment_status.environment,
 				displayName: deployment_status.environment,
 				type: environment
-			}
+			},
+			associations: associations
```
mboudreau, line 252: in typescript, you don't have to duplicate the name of key and value if they're the same, you can just do `associations` and it does it for you :)

src/transforms/transform-deployment.ts
```diff
@@ -155,6 +157,34 @@ export function mapEnvironment(environment: string): string {
...
+					commitHash: commitSummary.sha,
+					repositoryId: repositoryId
+				};
+			});
+
+		associationArray = [
+			{
+				associationType: "commit",
+				values: commitKeys
+			}
```
mboudreau, line 181: Documentation for this says:

> associationType REQUIRED
> string
> Defines the asssociation type.
> Valid values: issueKeys, issueIdOrKeys

Or is `commit` a new type that the documentation hasn't been updated with?

src/transforms/transform-deployment.ts
```diff
@@ -155,6 +157,34 @@ export function mapEnvironment(environment: string): string {
...
 
+// Maps commit summaries to an array of a single association containing the commit keys (commit hash and repository id)
+const mapCommitSummariesToAssociationArray = (
+	repositoryId: string,
+	commitSummaries?: CommitSummary[]
+): JiraAssociation[] | undefined => {
+
+	let associationArray;
+
+	if (commitSummaries && commitSummaries.length) {
```
mboudreau, line 168: better way of doing this would be to add the condition check to return undefined at the top instead of wrapping most of the function in an conditional:

`if(commitSummaries?.length) { return undefined; }` at the top will do, and the rest can just continue as in.

There's also the question, should the function return undefined or just an empty array?  Would the devinfo api error on an empty array?

src/transforms/transform-deployment.ts
jason-hooker: `commit` is a new type that we're adding. Documentation is due to be updated later in the project.

mboudreau: Cool, in that case, can you please update the type for `associationType` to include the possible values and not just a generic string?  Thanks.

src/transforms/transform-deployment.ts
jason-hooker: I did a quick test, and it appears that receiving an empty array causes issues, so it's safer to not send it at all - ie. set to undefined, rather than []

mboudreau (approved):

src/transforms/transform-deployment.ts
mboudreau: yeah, figured as much

jason-hooker:
Ready for review and merge (I just merged conflicts from main which means I lost the existing approval).

joshkay10 (approved):

mboudreau (approved):

mboudreau:
I'll deploy this to staging after I've tested my PR.

PR #1216, opened 2022-05-25
ARC-918 auth and connect flow

rachellerathbone:
Updated GH clients to dynamically set urls and the accept header so that they work with both cloud and GHE server. PR looks pretty big but that is mostly because we create GH user, app, and installation clients in a lot of places and they all needed to be modified to use the factory functions in `get-github-client-config` so that I could wrap changes at the top level in the `GHE_SERVER` feature flag.

Main changes to look at are:

- get-github-client-config: this is a new file and has a number of functions to set instances of each client based on whether the ff is on or not
- GitHubClient superclass: new class to handle setting the rest and graphql endpoints for GHE sever and github.com. Can be used in future to clean up the clients
- how the baseUrl is now being set in each client
- change to the graphql method in the installation client
- removal of GHE_HOST from env.ts: if we're pointing to a GHE server instance, this url will be pulled from the GitHubserverApps table so this env is no longer needed


src/config/feature-flags.ts
```diff
@@ -30,7 +30,8 @@ export enum BooleanFlags {
 	USE_NEW_GITHUB_CLIENT_FOR_GET_REPO = "use-new-github-client-for-get-repo",
 	USE_NEW_GITHUB_CLIENT_FOR_INSTALLATION_API = "use-new-github-client-for-installation-api",
 	USE_NEW_GITHUB_CLIENT_FOR_REDIRECT = "use-new-github-client-for-redirect",
-	RETRY_ALL_ERRORS = "retry-all-errors"
+	RETRY_ALL_ERRORS = "retry-all-errors",
+	GHE_SERVER = "ghe_server",
```
rachellerathbone, line 34: This is our 'global' flag that we will use to wrap all GHE changes.

src/github/client/github-client-server.test.ts
```diff
@@ -0,0 +1,108 @@
...
+	function verifyMetricsSent(path: string, status) {
+		expect(statsdHistogramSpy).toBeCalledWith("app.server.http.request.github", expect.anything(), expect.objectContaining({
+			client: "axios",
+			method: "GET",
+			path,
+			status
+		}));
+	}
+
+	it("works with a non-cloud installation", async () => {
```
rachellerathbone, line 63: Moved this test out into its own file to clairfy that this one test is focused on GHE server. I was going to update the test setup and add more tests in this PR but this PR is already pretty big. Everything is wrapped in a FF so we can address updating the test setup and adding more tests separately (ARC-1374)

src/github/client/github-installation-client.ts
```diff
@@ -207,10 +214,17 @@ export class GitHubInstallationClient {
 
 	public async getNumberOfReposForInstallation(): Promise<number> {
 		const response = await this.graphql<{ viewer: { repositories: { totalCount: number } } }>(ViewerRepositoryCountQuery);
-
 		return response?.data?.data?.viewer?.repositories?.totalCount;
 	}
 
+	// TODO - remove this when we have replaced Probot and can make graphql requests for both cloud and server
+	public async getNumberOfReposForInstallationRest(): Promise<number> {
```
rachellerathbone, line 221: Graphql (imported by Probot) fail for GHE at this point. We can clean this up once we have our own webhook receiver and have configured graphql to work with cloud and server.

src/util/get-github-client-config.ts
```diff
@@ -0,0 +1,89 @@
...
+export async function getGitHubHostname(jiraHost: string) {
+	const gitHubClientConfig = await getGitHubClientConfig(jiraHost);
+
+	return await booleanFlag(BooleanFlags.GHE_SERVER, false, jiraHost) && gitHubClientConfig
+		? gitHubClientConfig.hostname
+		: GITHUB_CLOUD_HOSTNAME;
+}
+
+const getGitHubClientConfig = async (jiraHost: string): Promise<GitHubClientConfig> => {
+	// TODO: remove duplicate jiraHosts and find issue that causes duplicates
```
rachellerathbone, line 38: We have ARC-772 to remove duplicate entries from the db and ARC-769 to add unique constraints to jiraHost and clientKey in the Installations table. Both are assigned to the `DISTURBED` epic. We should probably get whoever is disturbed in upcoming weeks to tackle these 2 issues.

.env.test
```diff
@@ -19,6 +19,7 @@ SQS_DEPLOYMENT_QUEUE_REGION=us-west-1
 SQS_BRANCH_QUEUE_URL=http://127.0.0.1:4566/000000000000/test-branch
 SQS_BRANCH_QUEUE_REGION=us-west-1
 BULL_QUEUE_PREFIX=bulltest
+DEBUG=nock.*
```
rachellerathbone, line 22: Added this because degubbing nock (and trying to figure out how to debug nock) is laaaaaaaaame. It does create noise in the terminal but... who cares? It means we actually can figure out why nock tests fail. Without it, we have zero insight into why a mock request fails.

src/github/client/github-installation-client.ts
mboudreau: imported by probot?  This is our own implementation though... this should work with GHE.  If it doesn't, we have a bigger issue that the APIs are not consistent between github.com and ghe

src/github/client/github-installation-client.ts
```diff
@@ -293,7 +307,7 @@ export class GitHubInstallationClient {
 		const appToken = this.appTokenHolder.getAppToken(this.githubInstallationId);
 		return {
 			headers: {
-				Accept: "application/vnd.github.v3+json",
```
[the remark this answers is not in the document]
mboudreau: I also don't think that they should be duplicated between the apptoken and the installationtoken, this header should be global for all requests...

src/github/client/github-installation-client.ts
```diff
@@ -33,17 +37,19 @@ export class GitHubInstallationClient {
 	private readonly appTokenHolder: AppTokenHolder;
 	private readonly installationTokenCache: InstallationTokenCache;
 	public readonly githubInstallationId: InstallationId;
+	private readonly gitHubEnterprise: GitHubClientConfig | undefined;
 	private readonly logger: Logger;
 
 	constructor(
 		githubInstallationId: InstallationId,
 		logger: Logger,
+		gitHubEnterprise?: GitHubClientConfig | undefined,
```
mboudreau, line 46: you don't have to do ` | undefined` if your param is already optional, that's implied

src/github/client/github-installation-client.ts
mboudreau: configuring what?  Graphql?  look at the graphql function in the class

src/util/get-github-client-config.ts
mboudreau: Probably need to remove it form disturbed and instead put it as part of the GHE work instead

src/github/client/github-installation-client.ts
mboudreau: that's because your `githubEnterprise` in the class is set to `GithubClientConfig` only and needs to be `GithubClientConfig | undefined`.  

And since we're not using the accept headers anymore, we could just remove this and use the `baseURL` we used to have that combines both the domain and the path (for the GHE case) which should work for both GHE and github.com :)

.env.test
joshkay10: agreed, is a nightmare without this!

joshkay10 (dismissed):
Exciting, Getting much  more friendlier ghclient versions each time!!

mboudreau (approved):
:pinched_fingers:

mboudreau (approved):

PR #1217, opened 2022-05-25
ARC-1366 hashing util

joshkay10:
**Whats in this Pull Request**
A generic hashing function that can be used globally with the same hash secret from envvars
The intention is to use this for restricted UGC/PII
A symmetrical key will be stored in lastpass and micros stash (per environment)

**Whats Next**
- Slauth endpoint so that devs can generate the same hash value ( useful for splunking)
- Change the current sha1 hashing on the feature flags of jirahost to new function


rachellerathbone (dismissed):

src/util/encryption.ts
```diff
@@ -0,0 +1,13 @@
+import { createHmac } from "crypto";
+import { envVars } from "../config/env";
+
+const getSharedSecret = (): string => {
+	return envVars.HASH_SECRET;
+}
```
mboudreau, line 6: do you really need a function if it's just a constant?

.env.test
```diff
@@ -19,6 +19,7 @@ SQS_DEPLOYMENT_QUEUE_REGION=us-west-1
 SQS_BRANCH_QUEUE_URL=http://127.0.0.1:4566/000000000000/test-branch
 SQS_BRANCH_QUEUE_REGION=us-west-1
 BULL_QUEUE_PREFIX=bulltest
+HASH_SECRET=test
```
mboudreau, line 22: I feel like we should rename this to be more explicit, something like `GLOBAL_HASH_KEY`

src/config/env.ts
```diff
@@ -99,6 +100,7 @@ export interface EnvVars {
 	GITHUB_HOSTNAME: string;
 	DEPLOYMENT_DATE: string;
 	BULL_QUEUE_PREFIX?: string;
+	HASH_SECRET: string;
```
mboudreau, line 103: This should also be part of the required env vars function above to prevent hashing with 'undefined'

src/util/encryption.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const createHashWithSharedSecret = (data): string => {
```
mboudreau, line 8: you could make this a generic "createHash" function with a secondary parameter of `hashKey = envVars.HASH_SECRET`

src/util/encryption.ts
joshkay10: I accept.

src/util/encryption.ts
joshkay10: I had it as something similar to begin with, but felt since the hash is global wanted to make it explicit to avoid whoopsies

src/config/env.ts
```diff
@@ -20,7 +20,8 @@ const requiredEnvVars = [
 	"SQS_DEPLOYMENT_QUEUE_REGION",
 	"SQS_BRANCH_QUEUE_URL",
 	"SQS_BRANCH_QUEUE_REGION",
-	"MICROS_AWS_REGION"
+	"MICROS_AWS_REGION",
+	"GLOBAL_HASH_SECRET"
```
mboudreau, line 24: it's not a secret, it's a key.

src/util/encryption.ts
mboudreau: I mean, having it in the params is making it explicit...  I'm not sure I understand the concern

src/config/env.ts
mboudreau: don't forget this one

bgvozdev (approved):

src/util/encryption.ts
```diff
@@ -0,0 +1,8 @@
+import { BinaryLike, createHmac } from "crypto";
+import { envVars } from "../config/env";
+
+export const createHashWithSharedSecret = (data: BinaryLike, secret = envVars.GLOBAL_HASH_SECRET): string => {
```
bgvozdev, line 4: Wondering why it is "shared"? :)

bgvozdev (approved):

src/util/encryption.ts
joshkay10: you mean just have it as createHash instead of createHashWithSharedSecret or createHashWithGlobalSecret?

bgvozdev: Well, since you are passing the secret as a parameter, I think it should be just "createHash". TBH I liked more when it was createHashWithGlobalSecret and without the parameter 🤷 . Anyway, pretty sure this is a "nit" 😃

bgvozdev (approved):

src/util/encryption.ts
joshkay10: sigh.. your right, ready to re-re-re-reapprove lolol
its either createHashWithSharedSecret without the default value or
createHash with the default value. combination is just messy.

1 moment.

PR #1218, opened 2022-05-27
ARC-1367 slauth hash endpoint

joshkay10:
**Stuck in draft until ARC-1366 PR has been merged**

**Whats in this Pull Request**
Added a new POST api/hash endpoint that you can pass a value too and receive the hashed value back.

**request** `{ "data": "hashmeeee"}`
**response** `{ "originalValue": "hashmeeee","hashedValue":"4cee752352ee515bd7d79e28a4648b3e63729ebe3e0265c6c37e0617db836a33"}`

**Why**
Soon, due to security concerns we need to start hashing UGC/PII and jirahost will be one of those values, so we need a way to still filter against the hashed value in splunk when we can't use plaintext.

**Extra attention areas**
_Naming_ :( went back and forth. If its no obvious to you then do raise it, am not too commited to any of it.

Location of the new `api-hash-post.ts` file, originally had it inside a new  folder like jira and installation but settled on just the file, couldn't think of a good name for the folder either lol :(

**Whats Next**
Now we have a way to generate the hashes, we can start implementing them in the code. 
First thing to do is change the feature flags hashing of JiraHost to the new function

**Supporting documentation**
A runbook will be created before merging to main.


src/routes/api/api-hash-post.ts
```diff
@@ -0,0 +1,22 @@
+import { Request, Response } from "express";
+import { createHashWithSharedSecret } from "~/src/util/encryption";
+
+export const ApiHashPost = async (req: Request, res: Response): Promise<void> => {
+
+	const { data } = req.body;
+
+	if (!data) {
+		res.status(400)
+			.json({
```
joshkay10, line 10: We have a combination of
`.json({message: "problem oh no"}` and `.send("problem oh no")` for our error responses, not sure if theres a agrred upon preference but i like it being decorated with a message attribute....¯\_(ツ)_/¯

src/routes/api/api-hash-post.ts
```diff
@@ -0,0 +1,22 @@
...
+				message: "Please provide a value to be hashed."
+			});
+		return;
+	}
+
+	const hashedValue = createHashWithSharedSecret(data);
+
+	res.json({
```
joshkay10, line 18: Could just return the hashed value and nothing else.... mainly being used from a terminal so was nice to get a confirmation of what i was hashing.... maybe a security issue though pairing them up like that?

src/routes/api/api-router.ts
```diff
@@ -94,6 +95,9 @@ ApiRouter.post(
 	}
 );
 
+// Hash incoming values the same as the internal hashing, used to help filter logs.
+ApiRouter.post("/hash", ApiHashPost);
```
joshkay10, line 99: "/hash" was as imaginative as I could get, other options welcomes

src/routes/api/api-hash-post.ts
mboudreau: I don't see the point of a JSON error unless it's part of a convention (like graphql with an `errors` array). Just `send` with the error is fine IMO.

src/routes/api/api-hash-post.ts
mboudreau: I'm not sure I understand the point of this API?  We call this first API to be able to get data from other APIs?

We don't need to be this drastic.  Our API can still take jirahost to return relevant data.

bgvozdev (approved):

gxueatlassian:
Sorry I don't understand of story of this API. Am I right saying that:
1. Someone A in our team get paged, A logins to splunk to check the logs
2. Apps already hashed the UGC/PII data in log, so A couldn't filter the splunk logs
3. So instead A call this API manually, grab the hashed return value, and use it to filter in splunk?
------
If that's the case, this API is purely used by us? If so, what sort of security protection should we put it? Like can we stop public from calling it?

bgvozdev:
> Sorry I don't understand of story of this API. Am I right saying that:
> 
> 1. Someone A in our team get paged, A logins to splunk to check the logs
> 2. Apps already hashed the UGC/PII data in log, so A couldn't filter the splunk logs
> 3. So instead A call this API manually, grab the hashed return value, and use it to filter in splunk?

Correct.


> If that's the case, this API is purely used by us? 

Correct

> If so, what sort of security protection should we put it? Like can we stop public from calling it?

I think all our admin endpoints are available only to us via slauth calls (go/slauth) and not available to public calls.

gxueatlassian:
> > Sorry I don't understand of story of this API. Am I right saying that:
> > 
> > 1. Someone A in our team get paged, A logins to splunk to check the logs
> > 2. Apps already hashed the UGC/PII data in log, so A couldn't filter the splunk logs
> > 3. So instead A call this API manually, grab the hashed return value, and use it to filter in splunk?
> 
> Correct.
> 
> > If that's the case, this API is purely used by us?
> 
> Correct
> 
> > If so, what sort of security protection should we put it? Like can we stop public from calling it?
> 
> I think all our admin endpoints are available only to us via slauth calls (go/slauth) and not available to public calls.

👍 Okay, didn't aware of that. All good.

gxueatlassian (approved):

PR #1220, opened 2022-05-29
ARC-1072 move the task filtering to getnext task to stop pollution the global …

joshkay10:
Moved the filtering of task types to the getnextask function so that it doesn't pollute global values.


src/sync/installation.ts
```diff
@@ -62,7 +62,11 @@ export const sortedRepos = (repos: Repositories): [string, RepositoryData][] =>
 			new Date(a[1].repository?.updated_at || 0).getTime()
 	);
 
-const getNextTask = async (subscription: Subscription): Promise<Task | undefined> => {
+const getNextTask = async (subscription: Subscription, jiraHost: string): Promise<Task | undefined> => {
+
+	const includeBuildAndDeployments = await booleanFlag(BooleanFlags.BACKFILL_FOR_BUILDS_AND_DEPLOYMENTS, false, jiraHost);
+	const tasks = taskTypes.filter(task => includeBuildAndDeployments ? taskTypes : task !== "build" && task !== "deployment");
```
mboudreau, line 68: filter need to return true or false.  I think the function you're looking for is `task => includeBuildAndDeployments || task !== "build" && task !== "deployment"`

src/sync/installation.ts
```diff
@@ -62,7 +62,11 @@ export const sortedRepos = (repos: Repositories): [string, RepositoryData][] =>
 			new Date(a[1].repository?.updated_at || 0).getTime()
 	);
 
-const getNextTask = async (subscription: Subscription): Promise<Task | undefined> => {
+const getNextTask = async (subscription: Subscription, jiraHost: string): Promise<Task | undefined> => {
```
mboudreau, line 65: instead of passing in the `jiraHost`, it might make more sense to pass in the boolean of `includeBuildAndDeployments`

joshkay10: it makes more sense to pass the boolean, it just got awkward with line 
162-`} else if (!(await getNextTask(subscription, jiraHost))) {`
	
Could have FF evaluating in 2 places(above each getNext) but seemed convoluted to evaluate a flag that might not get called....

If I wanted to evaluate it once, it got pretty messy fast passing it all over the show.....

src/sync/installation.ts
joshkay10: 😮‍💨  - fixing

src/sync/installation.ts
mboudreau: yeah, fair.

mboudreau (approved):

PR #1221, opened 2022-05-30
ARC-1300 new wehook receiver

Harminder84:
In order to create a new webhook receiver, below changes have been made :-

- A new endpoint is exposed for webhook receiver 
- Webhook path for Github cloud - `webhook/event/cloud` and for Github Enterprise it is `webhook/event/:uuid` where uuid is unique identifier for every Github Enterprise customer.

When the incoming webhook request is received, it validates the payload against the stored webhook secret. The webhook secret is stored in `GithubServerApps` table for Github Enterprise whereas in case of Github cloud, it gets the secret from environment variable. Once the payload is verified, it forwards the request to the registered webhook handlers. The handlers are dummy right now and these will be replaced with actual handlers in coming PRs.


package.json
```diff
@@ -49,6 +50,7 @@
 		"@atlassian/aui": "^9.4.0",
 		"@octokit/auth-app": "^2.6.0",
 		"@octokit/graphql-schema": "^10.73.0",
+		"@octokit/webhooks-methods": "^2.0.0",
```
rachellerathbone, line 53: What was this being used for?

src/webhook/webhook-receiver.ts
```diff
@@ -0,0 +1,80 @@
...
+	const signatureSHA256 = request.headers["x-hub-signature-256"] as string;
+	const id = request.headers["x-github-delivery"] as string;
+	const uuid = request.params.uuid;
+	let webhookSecret: string = process.env.WEBHOOK_SECRET!;
+	try {
+		if (uuid != "cloud") {
+			const gitHubServerApp = await GitHubServerApp.getGitHubServerAppForUuid(uuid)
+			if (!gitHubServerApp) {
+				response.statusCode = 404;
+				response.end("GitHub app does not found\n");
```
rachellerathbone, line 42: Edit: GitHub app not found.

src/server.ts
```diff
@@ -0,0 +1,23 @@
+import "config/env";
+import express, { Router } from "express";
+import { getLogger } from "./config/logger";
+import { setupGithubWebhooks } from "./webhook/setupGithubWebhooks";
+import { webhookReceiver } from "./webhook/webhook-receiver";
+import { Webhooks } from "./webhook/webhooks";
+const app = express();
+const port = 8082;
```
rachellerathbone, line 8: I'm not sure we can have the app run on port 8082. Because it runs on micros, I think it needs to be 8080.

Also, we don't necessarily need to do this as part of this PR because we're mainly focusing on the webhook aspect, but we should at least think about other places we're calling a Probot method in the app. For example, both main and worker call probot.start which is a method on Probot's server class that creates a httpServer. 

We can focus on adding methods that aren't webhook specific later but it might be good to keep them in mind now. I'm thinking refactoring this into a Server class might be a better approach. Instead of adding an additional script, I'm guessing we would want a similar implementation where we can call something like `webhookReceiver.start()` from main and worker and have everything 'just work'. 

For the simplicity of getting webhooks working without Probot, I'm happy for us to leave this as is for now but definitely think we need to modify this to completely replace Probot. Might be worthwhile setting up some time with some of the team to discuss how we want to tackle this.

package.json
Harminder84: This is being used for validating the payload with webhook secret

rachellerathbone: Gotcha. Is there another way we can do this? We're currently in the process of ripping Octokit out so not sure if introducing more Octokit packages is the way to go.

src/server.ts
Harminder84: Definitely, this is going to be change a lot in future. Port 8080 is currently blocked by Probot

package.json
Harminder84: This is a lightweight library with only two methods i.e. sign and verify the payload. If you still want to remove it, we can in incorporate the code in our app.

mboudreau: I would imagine that the code underneath it is just using https://nodejs.org/api/crypto.html

We probably don't need it...

src/server.ts
mboudreau: there's no need to run a new server - just hook into the current one.

package.json
```diff
@@ -14,6 +14,7 @@
 		"start": "run-p start:main start:worker",
 		"start:main": "tsnd -r tsconfig-paths/register --watch=.env,db/config.json,**/*.hbs --inspect=0.0.0.0:9229 --respawn --transpile-only -- src/main.ts",
 		"start:worker": "tsnd -r tsconfig-paths/register --watch=.env,db/config.json,**/*.hbs --inspect=0.0.0.0:9230 --respawn --transpile-only -- src/worker.ts",
+		"start:server": "tsnd -r tsconfig-paths/register --watch=.env,db/config.json,**/*.hbs --inspect=0.0.0.0:9231 --respawn --transpile-only -- src/server.ts",
```
mboudreau, line 17: Don't need another server, please remove.

src/webhook/webhook-receiver.ts
```diff
@@ -0,0 +1,80 @@
...
+			if (!gitHubServerApp) {
+				response.statusCode = 404;
+				response.end("GitHub app not found\n");
+				clearTimeout(timeout);
+				return;
+			}
+			webhookSecret = gitHubServerApp?.webhookSecret;
+		}
+
+		const payload = await getPayload(request);
```
mboudreau, line 49: why is `getPayload` necessary?  express should be handling this automatically for us.  If you need to do something else, you should probably use a middleware instead.

src/webhook/webhook-receiver.ts
```diff
@@ -0,0 +1,80 @@
...
+		response.end(`Required headers missing: ${missingHeaders}`);
+		return;
+	}
+	// GitHub will abort the request if it does not receive a response within 10s
+	let didTimeout = false;
+	const timeout = setTimeout(() => {
+		didTimeout = true;
+		response.statusCode = 202;
+		response.end("still processing\n");
+	}, 9000).unref();
```
mboudreau, line 30: Why are we setting and clearing timeouts everywhere?

package.json
joshkay10: There is a Pull Request that will hopefully go in tomorrow tha signs things, but would be a good spot for a verify as well. Already importing the crypto class mentioned.
https://github.com/atlassian/github-for-jira/pull/1217

src/webhook/get-payload.ts
```diff
@@ -0,0 +1,26 @@
+
+import { Request } from "express";
+
+export function getPayload(request: Request): Promise<object> {
```
joshkay10, line 4: preferred function styles as `export const getPayload.....`

src/webhook/webhook-receiver.ts
Harminder84: GitHub will abort the request if it does not receive a response within 10s. So, setting timeout to response within that timeframe if our handler took more time to response and clearing timeout if event handler responds to the request.

src/webhook/get-payload.ts
joshkay10: maybe we add func-style to eslint?

src/webhook/webhook-receiver.ts
```diff
@@ -0,0 +1,80 @@
...
+
+	const eventName = request.headers["x-github-event"] as string;
+	const signatureSHA256 = request.headers["x-hub-signature-256"] as string;
+	const id = request.headers["x-github-delivery"] as string;
+	const uuid = request.params.uuid;
+	let webhookSecret: string = process.env.WEBHOOK_SECRET!;
+	try {
+		if (uuid != "cloud") {
+			const gitHubServerApp = await GitHubServerApp.getGitHubServerAppForUuid(uuid)
```
joshkay10, line 39: / nit ;

src/webhook/setupGithubWebhooks.ts
```diff
@@ -0,0 +1,27 @@
...
+import { Webhooks } from "./webhooks";
+
+export const setupGithubWebhooks = (webhooks: Webhooks) => {
+	webhooks.on(
+		[
+			"issue_comment.created",
+			"issue_comment.edited"
+		],
+		(event: WebhookContext) => {
+			console.log("issue_comment created and edited ", event.id)
```
joshkay10, line 16: Feels like a Logger instance instead of console.log would be useful.

package.json
Harminder84: Yep @mboudreau , it is using crypto package underneath it.  

Thanks @joshkay10 for the update. I'll remove this package once your PR is in.

src/models/git-hub-server-app.ts
```diff
@@ -64,6 +64,20 @@ export class GitHubServerApp extends Model {
...
 		return gitHubServerApp;
 	}
+
+	/**
+	 * Get GitHubServerApp
+	 *
+	 * @param {{uuid: string}} uuid
+	 * @returns {GitHubServerApp}
+	 */
+	static async getGitHubServerAppForUuid(uuid: string): Promise<GitHubServerApp | null> {
```
mboudreau, line 74: I'd probably name this `findForUuid` since you don't need the object context in the function name (already in the model class) and we are also more explicit as to which operation the function is doing (find)

src/webhook/get-payload.ts
mboudreau: probably a good idea @joshkay10.  Tag, you're it :P

src/webhook/setupGithubWebhooks.ts
Harminder84: This is temporary one. Will be removed in upcoming PRs.

src/webhook/webhook-receiver.ts
mboudreau: github doesn't need to wait for our processing, it's just waiting for a confirmation that we received the event.  We can response right away with a '204' while we're still processing the payload in the background.

src/webhook/setupGithubWebhooks.ts
mboudreau: might as well do it now then?

src/server.ts
```diff
@@ -0,0 +1,23 @@
...
+const app = express();
+const port = 8082; 
+
+const logger =  getLogger("server");
+
+const webhooks = new Webhooks();
+setupGithubWebhooks(webhooks);
+
+const RootRouter = Router();
+RootRouter.post("/webhook/event/:uuid", webhookReceiver.bind(null, webhooks));
```
mboudreau, line 16: Why are you using `bind` here?

src/webhook/webhooks.ts
```diff
@@ -0,0 +1,17 @@
...
+	public receive: (eventPayload: WebhookContext) => Promise<void[]>;
+
+	constructor() {
+		const state: State = {
+			hooks: {}
+		};
+		this.on = receiverOn.bind(null, state);
+		this.receive = receiverHandle.bind(null, state);
+	}
+}
```
mboudreau, line 17: Not sure I understand the point of this class

src/webhook/webhook-receiver.ts
```diff
@@ -0,0 +1,80 @@
...
+	"x-github-delivery",
+];
+
+export const webhookReceiver = async (webhooks: Webhooks, request: Request, response: Response) => {
+	const missingHeaders = (WEBHOOK_HEADERS.filter((header) => !(header in request.headers))).join(", ");
+	if (missingHeaders) {
+		response.writeHead(400, {
+			"content-type": "application/json",
+		});
+		response.end(`Required headers missing: ${missingHeaders}`);
```
mboudreau, line 21: You should be using the express functions to send information: `return response.send(400, `Required headers missing: ${missingHeaders}`);

src/webhook/webhook-receiver.ts
```diff
@@ -0,0 +1,80 @@
...
+import { WebhookContext } from "./types";
+
+const WEBHOOK_HEADERS = [
+	"x-github-event",
+	"x-hub-signature-256",
+	"x-github-delivery",
+];
+
+export const webhookReceiver = async (webhooks: Webhooks, request: Request, response: Response) => {
+	const missingHeaders = (WEBHOOK_HEADERS.filter((header) => !(header in request.headers))).join(", ");
```
mboudreau, line 16: there's a better way to do this kind of validation at the route level using `express-validator`

src/webhook/webhook-receiver.ts
mboudreau: I'm also kinda curious as to what would be taking 10s to process...

src/webhook/types.ts
```diff
@@ -0,0 +1,42 @@
...
+
+type Hooks = {
+	[key: string]: Function[];
+};
+
+export interface State {
+	hooks: Hooks;
+}
+
+export class WebhookContext {
```
mboudreau, line 13: Not much point using a class for a simple object that holds data.  Could just use an interface for this.

src/webhook/webhooks.ts
Harminder84: This class holds the information about registered events and their handler.  For e.g . to register for push events
webhooks.on("push", (event: WebhookContext) => {... }); 

The receive method is invoked when any incoming webhook event is received. This method decides based on event and action, which event handlers to be executed.

src/server.ts
Harminder84: The idea here is to webhookReceiver also get webhooks  as parameter along with request and response. The webhooks holds information about subscribed events and their handlers.

src/webhook/types.ts
Harminder84: We're going to replace Probot context's class with WebhookContext. We've used various context's class method such as repo in our exisiting code, so WebhookContext is going to have similar methods.

src/webhook/webhook-receiver.ts
Harminder84: It is to handle edge use cases where it takes more than 10 secs and I found this in @octokit/webhooks source code which is being used by Probot. If we're not wait for processing to complete before sending response to github, then I'll remove this code as you mentioned earlier.

src/webhook/get-payload.ts
joshkay10: **Dodges** .... Unsuccessfully - https://github.com/atlassian/github-for-jira/pull/1226

src/server.ts
bgvozdev: I'm wondering why are we not using the existing router that all other endpoints are using?

src/webhook/setupGithubWebhooks.ts
bgvozdev: I agree, it is better to use a context logger because it adds some useful info, e.g. request id that we can use later for tracing in splunk

src/webhook/webhook-receiver.ts
bgvozdev: What could be a use-case when we handle a request more than 10 seconds? Shouldn't it be handled asynchronously if such is possible at all?

src/webhook/webhooks.ts
bgvozdev: Maybe some other name (e.g. WebhooksRouter) would make it more clear? (Although, Router is kinda express thingy... maybe Orchestrator or Relay...)

src/webhook/webhook-receiver.ts
Harminder84: Used middleware now

src/webhook/webhook-receiver.ts
Harminder84: Updated flow to send response back to incoming webhook request without waiting for processing to complete. Removed timeout statements as well.

src/webhook/get-payload.ts
Harminder84: Removed file as now using middleware to parse request body.

src/webhook/setupGithubWebhooks.ts
Harminder84: Now I'm using logger instance.

src/webhook/webhook-receiver.ts
Harminder84: Now using `express-validator` to check for headers

src/webhook/webhook-receiver.ts
Harminder84: Now using `express-validator` to check for headers

src/models/git-hub-server-app.ts
Harminder84: Updated method name to `findForUuid`

package.json
joshkay10: If you merge main you will have new utils/encryption.ts file that it can sneak into.

src/routes/webhook/webhook-context.ts
```diff
@@ -0,0 +1,30 @@
...
+	}
+
+	repo() {
+		const repo = this.payload?.repository;
+		if (!repo) {
+			throw new Error(
+				"context.repo() is not supported for this webhook event."
+			);
+		}
+		return Object.assign(
```
joshkay10, line 23: Did you mean to use Object.assign without the ..sources argument?
maybe wanted .freeze or drop Object.assign completely?

src/routes/webhook/webhook-receiver-post.ts
```diff
@@ -0,0 +1,74 @@
...
+	}
+	const logger = getLogger("webhook.receiver");
+	const eventName = request.headers["x-github-event"] as string;
+	const signatureSHA256 = request.headers["x-hub-signature-256"] as string;
+	const id = request.headers["x-github-delivery"] as string;
+	const uuid = request.params.uuid;
+	let webhookSecret: string = process.env.WEBHOOK_SECRET!;
+	const payload = request.body;
+	try {
+		if (uuid != "cloud") {
```
thombergs, line 22: Doesn't need to be now, but in the long run, I would like to avoid if/else blocks like this one and instead handle all webhooks the same, be they from cloud or server. In this case, we could create a GitHubApp entry in the database for the cloud app.

src/routes/webhook/webhook-receiver-post.ts
```diff
@@ -0,0 +1,74 @@
...
+			name: eventName,
+			payload: payload,
+			log: logger
+		});
+
+		const action = "action" in payload ? payload.action : null;
+		if (action) {
+			invokeEventHandler(`${eventName}.${action}`, webhookContext);
+		}
+		invokeEventHandler(eventName, webhookContext);
```
thombergs, line 47: Isn't there an `else` missing here? `invokeEventHandler` is called twice.

thombergs (changes requested):

src/routes/webhook/webhook-receiver-post.ts
Harminder84: In one of the discussions, it is decided to keep Github cloud related configuration as it is i.e. they remain in environment variables. That is why database table named as `GitHubServerApps`

src/routes/webhook/webhook-receiver-post.ts
```diff
@@ -0,0 +1,74 @@
...
+
+		response.sendStatus(204);
+
+	} catch (error) {
+		logger.error(error);
```
gxueatlassian, line 52: Should it `response.sendStatus(5xx)`?
And if that's the cases, will Github retry this webhook again?

src/routes/webhook/webhook-receiver-post.ts
```diff
@@ -0,0 +1,74 @@
...
+	}
+};
+
+function invokeEventHandler(event: string, context: WebhookContext) {
```
gxueatlassian, line 56: Is this function not finished yet? I mean, should it push any msg to the sqs?

Harminder84: It is not finished yet, will hook with actual handlers in upcoming PRs.

src/routes/router.ts
```diff
@@ -54,6 +56,10 @@ RootRouter.use(MaintenanceRouter);
 RootRouter.use("/github", GithubRouter);
 RootRouter.use("/jira", JiraRouter);
 
+RootRouter.post("/webhook/event/:uuid",
```
mboudreau, line 59: I'd put this under the `GithubRouter` since it is only github webhooks we're handing and not a global webhook handler.

The final url should be something like `github/webhooks/some-id`, but you also need to specify that the uuid is optional since we want to handle the github cloud webhooks as well, so the route should be something like `/webhooks/:uuid?` to make it optional.  If uuid isn't set, it's github.com, if it is set, it's ghe.

src/routes/webhook/webhook-context.ts
```diff
@@ -0,0 +1,28 @@
+import Logger from "bunyan";
+
+export class WebhookContext {
+	id: string;
+	name: string;
+	payload: any;
```
mboudreau, line 6: I want a way to make the payload typable.  Maybe add a generic to the context that's defaulted to any if you don't specify it and that type is set as the payload type?

src/routes/webhook/webhook-context.ts
```diff
@@ -0,0 +1,28 @@
...
+	log: Logger;
+
+	constructor({ id, name, payload, log }) {
+		this.id = id;
+		this.name = name;
+		this.payload = payload;
+		this.log = log;
+	}
+
+	repo() {
```
mboudreau, line 16: We shouldn't need this function as it should be used from the payload type.

src/routes/router.ts
```diff
@@ -54,6 +56,10 @@ RootRouter.use(MaintenanceRouter);
...
+	header(["x-github-event", "x-hub-signature-256", "x-github-delivery"]).not().isEmpty().withMessage("Missing Header"),
```
mboudreau, line 60: I think the convent we use is would be something like `header(["x-github-event", "x-hub-signature-256", "x-github-delivery"]).exists(), returnOnValidationError`

src/routes/webhook/webhook-receiver-post.ts
mboudreau: the uuid would never be set at `cloud` though.  It's either undefined (coud) or defined (ghe)

src/routes/webhook/webhook-receiver-post.ts
```diff
@@ -0,0 +1,77 @@
...
+	let webhookSecret: string = process.env.WEBHOOK_SECRET!;
+	const payload = request.body;
+	try {
+		if (uuid != "cloud") {
+			const gitHubServerApp = await GitHubServerApp.findForUuid(uuid);
+			if (!gitHubServerApp) {
+				response.status(400).send("GitHub app not found");
+				return;
+			}
+			webhookSecret = gitHubServerApp?.webhookSecret;
```
mboudreau, line 28: you don't need optional chaining here since you just validated that it's defined already in the previous if statement

src/routes/webhook/webhook-receiver-post.ts
```diff
@@ -0,0 +1,77 @@
...
+	const signatureSHA256 = request.headers["x-hub-signature-256"] as string;
+	const id = request.headers["x-github-delivery"] as string;
+	const uuid = request.params.uuid;
+	let webhookSecret: string = process.env.WEBHOOK_SECRET!;
+	const payload = request.body;
+	try {
+		if (uuid != "cloud") {
+			const gitHubServerApp = await GitHubServerApp.findForUuid(uuid);
+			if (!gitHubServerApp) {
+				response.status(400).send("GitHub app not found");
```
mboudreau, line 25: this can be simplified with `return response.send(400, "GitHub app not found")`

Harminder84: `response.send` takes only body as parameter, there is no way you can pass Http status code to `response.send`

src/routes/webhook/webhook-receiver-post.ts
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		const webhookContext = new WebhookContext({
+			id: id,
+			name: eventName,
+			payload: payload,
+			log: logger
+		});
+
+		const action = "action" in payload ? payload.action : null;
```
mboudreau, line 43: I don't understand the point of this.  It's the same as doing `const action = payload.action`.  We use `undefined` instead of `null` wherever possible

src/routes/webhook/webhook-receiver-post.ts
mboudreau: I think this line should be above the if condition as I believe the "parent" should be called first because the sub-action listener

src/routes/webhook/webhook-receiver-post.ts
mboudreau: There are no retries with GH webhooks sadly unless you want to do them manually from the GH console :/

src/routes/webhook/webhook-receiver-post.ts
```diff
@@ -0,0 +1,77 @@
...
+
+		response.sendStatus(204);
+
+	} catch (error) {
+		response.sendStatus(500);
+		logger.error(error);
+	}
+};
+
+const invokeEventHandler = (event: string, context: WebhookContext) => {
```
mboudreau, line 57: I'd probably call this `webhookRouter` or something and would pass in `context:WebhookContext, event:string, action?:string` so that the logic to do both top level event and event.action is in this router instead of the request handler.

src/routes/webhook/webhook-receiver-post.ts
```diff
@@ -0,0 +1,77 @@
...
+			context.log.info("push event Received!");
+			break;
+		case "pull_request":
+			context.log.info("pull req event Received!");
+			break;
+		case "pull_request.opened":
+			context.log.info("pull req opened event Received!");
+			break;
+		default:
+			throw new Error(`Event '${event}' not supported `);
```
mboudreau, line 69: This should _definitely_ not throw an error.  There are dozens of events that we don't listen to or handle.

src/routes/webhook/webhook-receiver-post.ts
```diff
@@ -0,0 +1,77 @@
...
+			}
+			webhookSecret = gitHubServerApp?.webhookSecret;
+		}
+		const verification = createHash(JSON.stringify(payload), webhookSecret);
+		if (verification != signatureSHA256) {
+			response.status(400).send("signature does not match event payload and secret");
+			return;
+		}
+
+		const webhookContext = new WebhookContext({
```
mboudreau, line 36: WebhookContext is still missing important github app information to be able to continue like the app url, clientId, webhooksecret, etc (pretty much everything in the GithubServerApp table).  You also need to a way to add the github.com app information in there if it's not a ghe app

mboudreau: I'd also add the action to the webhook context

src/routes/webhook/webhook-receiver-post.ts
mboudreau: actually, we can just pass in `WebhookContext` only if we add the action in there as well and let the router decipher who to call.

src/routes/webhook/webhook-receiver-post.ts
Harminder84: I think more specific should be called first. This is how other frameworks such as Probot works.

src/routes/webhook/webhook-receiver-post.ts
Harminder84: As discussed with @rachellerathbone , webhhook url for github cloud contains `cloud` as uuid.

src/routes/webhook/webhook-receiver-post.ts
Harminder84: Should we log the error in that case?

src/routes/webhook/webhook-receiver-post.ts
Harminder84: I'll add missing information in `WebhookContext` in upcoming PRs

Harminder84: action is now part of `WebhookContext`

src/routes/router.ts
Harminder84: Moved under `GithubRouter'.  The webhook url for GHE looks like `github/webhooks/some-id` where as for github cloud it is `github/webhooks/cloud`

src/routes/webhook/webhook-receiver-post.ts
mboudreau: We could, but I wouldn't consider it an error at all, just that we're not handling the webhook

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -0,0 +1,78 @@
...
+import { getLogger } from "~/src/config/logger";
+import { GitHubServerApp } from "~/src/models/git-hub-server-app";
+import { WebhookContext } from "./webhook-context";
+
+export const WebhookReceiverPost = async (request: Request, response: Response): Promise<void> => {
+	const errors = validationResult(request);
+	if (!errors.isEmpty()) {
+		response.status(400).json({ errors: errors.array() });
+		return;
+	}
```
mboudreau, line 14: we don't need this anymore as it's handled by the `returnOnValidationError` middleware

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -0,0 +1,78 @@
...
+	const logger = getLogger("webhook.receiver");
+	const eventName = request.headers["x-github-event"] as string;
+	const signatureSHA256 = request.headers["x-hub-signature-256"] as string;
+	const id = request.headers["x-github-delivery"] as string;
```
mboudreau, line 18: aren't all headers already strings?

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -0,0 +1,78 @@
...
+	const uuid = request.params.uuid;
+	let webhookSecret: string = process.env.WEBHOOK_SECRET!;
+	const payload = request.body;
+	try {
+		if (uuid != "cloud") {
```
mboudreau, line 23: as part of the model talk, I think we might be placing the github.com app in the gh app table so we don't need this conditional anymore

src/routes/github/webhook/webhook-receiver-post.ts
Harminder84: It is of type `string  |  string[]  |  undefined`

src/routes/github/webhook/webhook-receiver-post.ts
Harminder84: I missed to remove it. Good catch

mboudreau (approved):
nice

joshkay10 (approved):

PR #1223, opened 2022-05-30
ARC-933 limiting the history of commits taken from all branches during sync

mboudreau:
Everything's behind a feature flag so we can dial up or down the commits date limit we want from both the main and any other branch.  Since it's feature flags, we could always enable certain customers if they ask for more commit or change the defaults across the board very easily.


src/sync/commits.ts
```diff
@@ -26,14 +27,23 @@ export const getCommitTask = async (
 	logger: LoggerWithTarget,
 	_github: GitHubAPI,
 	gitHubClient: GitHubInstallationClient,
-	_jiraHost: string,
+	jiraHost: string,
 	repository: Repository,
 	cursor?: string | number,
 	perPage?: number): Promise<CommitData> => {
-	logger.info("Syncing commits: started");
+	const itemNumber = Number(cursor?.toString().split(" ").pop() || 0);
```
joshkay10, line 34: whats going on here?
`const itemNumber = Number(cursor?.toString().split(" ").pop() || 0);`
I would expect the cursor to look like **Y3VycOAKYOHs4Apg4e** since its from graphQL?

rachellerathbone:
What are we defaulting to 10? I only see 50 as the default.

bgvozdev:
Is it an experiment or the way we are going? Do we have some footprint somewhere about this decision?

src/sync/commits.ts
mboudreau: The commit cursor is unique here. It's in the format of `6ee0a9e7b43033ca54f3e313c32017e05182c691 3107`, the number at the end is actually the amount of commits down from the HEAD.  Not the best way to do this, I'll agree, since GH could change their cursor format tomorrow and break everything, but this is temporary until we change how we're processing syncs.

mboudreau:
@rachellerathbone We are getting the number from the FF https://github.com/atlassian/github-for-jira/pull/1223/files#diff-a21f86e19d3e5af07bafb085d36ca3128d36dbbe75b1f0ecb53e6fcfdc456de5R20

@bgvozdev I wanted to see how it works first, but it's a FF and we can change it to whatever we want.  We can deploy it right now, the FF will be off and it'll work as normal, but we can then decide how many we want to backfill which we can experiment with and change over time.  I've personally noticed a dramatic increase in speed for backfilling because of this.  Something else we can do is have a 2 part backfill: 1st part is getting everything except all commit (just get say 100 commits from main branch), then the second part is get all the historical commits.  We can do this by adding another task and changing the logic slightly.

This first approach was just something simple we can add right away and improve backfilling quickly, and we can decide now (or later) what will be the new values for commits.

bgvozdev:
> @rachellerathbone We are getting the number from the FF https://github.com/atlassian/github-for-jira/pull/1223/files#diff-a21f86e19d3e5af07bafb085d36ca3128d36dbbe75b1f0ecb53e6fcfdc456de5R20
> 
> @bgvozdev I wanted to see how it works first, but it's a FF and we can change it to whatever we want. We can deploy it right now, the FF will be off and it'll work as normal, but we can then decide how many we want to backfill which we can experiment with and change over time. I've personally noticed a dramatic increase in speed for backfilling because of this. Something else we can do is have a 2 part backfill: 1st part is getting everything except all commit (just get say 100 commits from main branch), then the second part is get all the historical commits. We can do this by adding another task and changing the logic slightly.
> 
> This first approach was just something simple we can add right away and improve backfilling quickly, and we can decide now (or later) what will be the new values for commits.

Probably we could try it on Equifax? (Cough @thombergs cough)

bgvozdev:
A general commit: I see 7 files changed here, including major behaviour change, and not a single test file changed...

src/sync/commits.ts
bgvozdev: Shouldn't this be covered/described in a test? (And similar things?)

bgvozdev (changes requested):

mboudreau:
@thombergs mentioned in a meeting that instead of having a hard "limit" of commits, maybe it should be time based instead.  example: get the commits for branches in the last week except for the main branch where we get the last month's worth.  I quite like this approach as it's more fluid for our customers.  Large customers will have many more commits in a month than smaller ones.

I'll try to implement this new approach as part of this PR.

src/sync/commits.ts
```diff
@@ -30,23 +26,23 @@ export const getCommitTask = async (
...
-		};
-	}
+	perPage?: number): Promise<TaskPayload<CommitQueryNode, JiraCommitData>> => {
+
 	logger.debug("Syncing commits: started");
 	const { edges, commits } = await fetchCommits(gitHubClient, repository, cursor, perPage);
 	const jiraPayload = await transformCommit({ commits, repository });
 	logger.debug("Syncing commits: finished");
 
+	const timeCutoff = Date.now() - await numberFlag(NumberFlags.SYNC_MAIN_COMMIT_TIME_LIMIT, Date.now(), jiraHost);
```
gxueatlassian, line 36: I'd like to suggest move the `timeCutoff` to close where they are used, which is inside the `if(edges?.length)`

src/github/client/github-installation-client.ts
```diff
@@ -213,12 +213,12 @@ export class GitHubInstallationClient extends GitHubClient {
 		return response?.data?.data?.viewer?.repositories?.totalCount;
 	}
 
-	public async getBranchesPage(owner: string, repoName: string, perPage = 1, commitAmount = 50, cursor?: string): Promise<getBranchesResponse> {
+	public async getBranchesPage(owner: string, repoName: string, perPage = 1, timeCutoff = Date.now(), cursor?: string): Promise<getBranchesResponse> {
```
gxueatlassian, line 216: Maybe change to `branchCommitTimeLimit`?
According to the code processing main branch, `timeCutoff` is a target Date, while here, `timeCutoff` represent miliseconds of gap between two Date.

gxueatlassian: And maybe remove the default value? Since the calling of this function all provides default value.

src/sync/branches.ts
```diff
@@ -17,8 +17,8 @@ export const getBranchTask = async (
 	// TODO: fix typings for graphql
 	logger.info("Syncing branches: started");
 	perPage = perPage || 20;
-	const commitAmount = await numberFlag(NumberFlags.SYNC_BRANCH_COMMIT_AMOUNT, 50, jiraHost);
-	const result = await newGithub.getBranchesPage(repository.owner.login, repository.name, perPage, commitAmount, cursor as string);
+	const timeCutoff = await numberFlag(NumberFlags.SYNC_BRANCH_COMMIT_TIME_LIMIT, Date.now(), jiraHost);
```
gxueatlassian, line 20: same as above, maybe call it `branchCommitTimeLimit`?

src/sync/commits.ts
gxueatlassian: This is minor I assume: Will the task be executed in multiple loop/async-subtasks/etc ( I don't know how to describe, lol) <-- but I mean, each time the task is executed, the `await numberFlag` is called. Not sure how does this work, but will it coz some http traffic and performance issue?

mboudreau: launchdarkly doesn't call the API every time we request the FF.  From my understanding it caches the results.  The initial wait time is a bit longer to call the API, but every subsequent call is then using the cached value which is essentially instantaneous.

gxueatlassian (dismissed):
Tentatively approve it.
I think we could add some tests as well.

gxueatlassian (dismissed):

gxueatlassian (dismissed):

src/sync/commits.ts
mboudreau: tests added

src/github/client/github-installation-client.ts
```diff
@@ -213,28 +213,24 @@ export class GitHubInstallationClient extends GitHubClient {
...
-			this.logger.warn("retrying branch graphql query without changedFiles");
-			return this.graphql<getBranchesResponse>(getBranchesQueryWithoutChangedFiles,
-				{
-					owner,
-					repo: repoName,
-					per_page: perPage,
-					cursor
-				});
-		});
+	public async getBranchesPage(owner: string, repoName: string, perPage = 1, timeCutoff?: number, cursor?: string): Promise<getBranchesResponse> {
```
bgvozdev, line 216: nit: I think `timestampCutoff` might be more explicit

src/github/client/github-installation-client.ts
```diff
@@ -213,28 +213,24 @@ export class GitHubInstallationClient extends GitHubClient {
...
+		const variables = {
+			owner,
+			repo: repoName,
+			per_page: perPage,
+			commitSince: timeCutoff ? new Date(Date.now() - timeCutoff).toISOString() : undefined,
```
bgvozdev, line 221: Shouldn't this be feature-flagged?

src/github/client/github-installation-client.ts
```diff
@@ -213,28 +213,24 @@ export class GitHubInstallationClient extends GitHubClient {
...
+			cursor
+		};
+		const response = await this.graphql<getBranchesResponse>(getBranchesQueryWithChangedFiles, variables)
+			.catch((err) => {
+				// Is it a changedFiles error?
```
bgvozdev, line 226: nit: comment duplicating the code

bgvozdev: nevermind, I see this was like that before

src/github/client/github-installation-client.ts
```diff
@@ -252,27 +248,22 @@ export class GitHubInstallationClient extends GitHubClient {
...
-					cursor
-				});
-		});
+	public async getCommitsPage(owner: string, repoName: string, perPage?: number, timeCutoff?: number, cursor?: string | number): Promise<getCommitsResponse> {
+		const variables = {
+			owner,
+			repo: repoName,
+			per_page: perPage,
+			cursor,
+			commitSince: timeCutoff ? new Date(Date.now() - timeCutoff).toISOString() : undefined
```
bgvozdev, line 257: feature flag?

src/github/client/github-installation-client.ts
bgvozdev: ah, that's not the timestamp. 🤦

bgvozdev: I'd suggest then to include the units into the name, e.g. `timeCutoffMsecs`

src/github/client/github-queries.ts
```diff
@@ -112,12 +112,12 @@ export type getCommitsResponse = {
...
 };
 
-export const getCommitsQueryWithChangedFiles = `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String) {
+export const getCommitsQueryWithChangedFiles = `query ($owner: String!, $repo: String!, $per_page: Int!, $commitSince: GitTimestamp, $cursor: String) {
   repository(owner: $owner, name: $repo){
     defaultBranchRef {
       target {
         ... on Commit {
-          history(first: $per_page, after: $cursor) {
+          history(first: $per_page, after: $cursor, since: $commitSince) {
```
bgvozdev, line 120: feature flag (here and everywhere in this file)?

src/sync/commits.ts
```diff
@@ -5,14 +5,11 @@ import { GitHubInstallationClient } from "../github/client/github-installation-c
...
+import { TaskPayload } from "~/src/sync/installation";
 
-type CommitData = {
-	edges: CommitQueryNode[],
-	jiraPayload: JiraCommitData | undefined
-}
-
-const fetchCommits = async (gitHubClient: GitHubInstallationClient, repository: Repository, cursor?: string | number, perPage?: number) => {
-	const commitsData = await gitHubClient.getCommitsPage(repository.owner.login, repository.name, perPage, cursor);
+const fetchCommits = async (gitHubClient: GitHubInstallationClient, repository: Repository, timeCutoff?: number, cursor?: string | number, perPage?: number) => {
```
bgvozdev, line 11: `timeCutoffMsecs`?

src/sync/commits.ts
```diff
@@ -26,14 +23,16 @@ export const getCommitTask = async (
...
-	logger.info("Syncing commits: started");
-	const { edges, commits } = await fetchCommits(gitHubClient, repository, cursor, perPage);
+	perPage?: number): Promise<TaskPayload<CommitQueryNode, JiraCommitData>> => {
+
+	logger.debug("Syncing commits: started");
+	const timeCutoff = await numberFlag(NumberFlags.SYNC_MAIN_COMMIT_TIME_LIMIT, NaN, jiraHost);
+	const { edges, commits } = await fetchCommits(gitHubClient, repository, timeCutoff, cursor, perPage);
 	const jiraPayload = await transformCommit({ commits, repository });
-	logger.info("Syncing commits: finished");
+	logger.debug("Syncing commits: finished");
```
bgvozdev, line 35: Just thinking if info is actually the right level here... would be easy to understand if the sync is stuck during commits sync, wdyt? (And same above where you changed it for "started")

src/sync/installation.ts
```diff
@@ -46,14 +46,15 @@ interface TaskProcessors {
...
+export interface TaskPayload<E = any, P = any> {
+	edges?: E[];
+	jiraPayload?: P;
+	isDone?: boolean;
 }
 
 type TaskType = "repository" | "pull" | "commit" | "branch" | "build" | "deployment";
 
-const taskTypes: TaskType[] = ["pull", "commit", "branch", "build", "deployment"];
+const taskTypes: TaskType[] = ["pull", "branch", "commit", "build", "deployment"];
```
bgvozdev, line 57: Does the order changes the priority?

src/sync/installation.ts
```diff
@@ -128,13 +129,15 @@ export const updateJobStatus = async (
 		logger.info("Organization has been deleted. Other active syncs will continue.");
 		return;
 	}
+	const { edges } = taskPayload;
+	const isComplete = !edges?.length || taskPayload.isDone;
```
bgvozdev, line 133: I see this changes the logic (`isDone` wasn't there before) and there's neither a test nor a feature-flag...

bgvozdev: Actually it is not used at all, because edges is undefined in this case...

bgvozdev: So I guess that's OK :)

src/github/client/github-installation-client.ts
bgvozdev: Or maybe not to feature-flag but no to include the key to the object if `timeCutoffMsecs` is not provided.

src/github/client/github-installation-client.ts
bgvozdev: Or maybe not to feature-flag but no to include the key to the object if timeCutoffMsecs is not provided.

src/github/client/github-queries.ts
bgvozdev: Soo... I see it is passing as undefined in case `timeCutoffMsecs` is not provided. Wondering how this query would work in this case?

I think we should include this variable only if it is provided, or is graphql protocol will understand `undefined`?

src/sync/branches.ts
```diff
@@ -3,20 +3,22 @@ import { GitHubAPI } from "probot";
...
-	_jiraHost: string,
+	jiraHost: string,
 	repository: Repository,
 	cursor?: string | number,
 	perPage?: number) => {
 	// TODO: fix typings for graphql
 	logger.info("Syncing branches: started");
 	perPage = perPage || 20;
-	const result = await newGithub.getBranchesPage(repository.owner.login, repository.name, perPage, cursor as string);
+	const timeCutoff = await numberFlag(NumberFlags.SYNC_BRANCH_COMMIT_TIME_LIMIT, NaN, jiraHost);
```
bgvozdev, line 20: Curious why the default value is NaN, not `undefined` ?

src/sync/commits.ts
```diff
@@ -26,14 +23,16 @@ export const getCommitTask = async (
...
+	jiraHost: string,
 	repository: Repository,
 	cursor?: string | number,
-	perPage?: number): Promise<CommitData> => {
-	logger.info("Syncing commits: started");
-	const { edges, commits } = await fetchCommits(gitHubClient, repository, cursor, perPage);
+	perPage?: number): Promise<TaskPayload<CommitQueryNode, JiraCommitData>> => {
+
+	logger.debug("Syncing commits: started");
+	const timeCutoff = await numberFlag(NumberFlags.SYNC_MAIN_COMMIT_TIME_LIMIT, NaN, jiraHost);
```
bgvozdev, line 32: Shouldn't the default value be `undefined`, not NaN? (Since you are passing it over later as an optional parameter)

src/github/client/github-queries.ts
bgvozdev: I guess if you do this, no FF is needed (because we are essentially switching it in the caller and when off the old code will work).

src/sync/installation.ts
mboudreau: yes

src/sync/branches.ts
mboudreau: Mentioned below.

src/github/client/github-installation-client.ts
mboudreau: The FF is the one that populate the variable.  No FF, variable is undefined and never set

src/github/client/github-installation-client.ts
mboudreau: I used `timeCutoffMsecs` as you mentioned below

src/github/client/github-queries.ts
bgvozdev: Well, we change it. Even though it is undefined, we send the modified query + a new key in the variables, relying/hoping GitHub will work as we expect.  FF should allow us to rollback to the old behaviour fast. With this change this is impossible.

mboudreau: ...we can disable the FF and get the previous behavior though.  This has been confirmed through unit tests and pollinator.  Just because any code has changed doesn't mean _everything_ needs to be behind a FF.  If that was the case, half the codebase would be duplicated.  It doesn't make logical sense to do this and will add more code/complexity to try to FF everything behind it

bgvozdev: Discussed offline, summary:
1. this FF will stay for a while, therefore duplicating code is bad (not a temp code)
2. we will be enabling FF not for everyone but just for canaries at start, therefore this change will be tested on them anyway

bgvozdev (approved):

src/github/client/github-queries.ts
bgvozdev: Actually no, I'm wrong, the query will be changed for everyone. Ok, let's see how it go anyway 🍿

PR #1224, opened 2022-05-30
ARC-1386: hide avatar if not GitHub cloud app

rachellerathbone:
**UPDATE:** all tasks mentioned below have been done and merged so this bad boy is finally ready to go.

**NOTE:**
* Changed this to a draft PR. To be considered ready, I need to add tests. However, before this can be done, I need to update the GitHubServerApps model and before that can be done, I need to update our other models.

--------------------------------------------------------------------------------------------------------------------------

We were trying to decide what to do regarding assets from GHE server accounts and whether or not we should update our content security policy (CSP). You can read about the decision in our [decision doc](https://hello.atlassian.net/wiki/spaces/PF/pages/1687919516/Decision+We+need+to+update+our+Content+Security+Policy+but+what+do+we+do+with+GHE+instances+running+on+http) but basically the summary is: for github.com we're going to render the avatar, otherwise, hide it 😄 

This change needed to happen on 3 pages: GH config, connect an org page, and the setup page.

As mentioned, for cloud we'll still see the avatar:

Screen Shot 2022-05-30 at 4 13 49 pm

For GHE server customers, the avatar will be hidden:
Screen Shot 2022-05-30 at 4 47 25 pm

Screen Shot 2022-05-30 at 5 28 09 pm

Screen Shot 2022-05-30 at 4 27 07 pm


src/util/jira-utils.ts
```diff
@@ -105,3 +107,13 @@ export const jiraIssueKeyParser = (str: string): string[] => {
 };
 
 export const hasJiraIssueKey = (str:string): boolean => !isEmpty(jiraIssueKeyParser(str));
+
+export const isGitHubCloudApp = async (jiraHost: string): Promise<Promise<boolean> | boolean>=> {
+	if (await booleanFlag(BooleanFlags.GHE_SERVER, false, jiraHost)) {
+		const installation = await Installation.getForHost(jiraHost);
+		const gitHubAppId = installation?.githubAppId;
+		return gitHubAppId ? !await GitHubServerApp.getForGitHubServerAppId(gitHubAppId) : true;
+	}
```
joshkay10, line 116: Got in a tangle reading this one.
Is this right?
if GHE FF is off then cloud?
if no gitHubAppId then cloud?
if githubcloudId but doesn't find anything then cloud?

my opinion piece is I would break it up a bit just for readability.

rachellerathbone: Yeah basically, we want to return true if cloud and false if GHE, and we detect a GHE if there is an entry in the GitHubServerApps table.

Any suggestions on how you want to break it out?

joshkay10: untested.. but something like this?  ¯\_(ツ)_/¯
Can rip out the feature flag easy when the time comes too.
```
export const isGitHubCloudApp = async (jiraHost: string): Promise<boolean>=> {
	if (await booleanFlag(BooleanFlags.GHE_SERVER, false, jiraHost)) {
		return false;
	}

	const installation = await Installation.getForHost(jiraHost);
	const gitHubAppId = installation?.githubAppId;
////githubcomment - don't need the **gitHubAppId &&** since getForGitHubServerAppId will return null, but easier to understand and can stop unnecessary calls......
/// i'm also assuming that this.findOne in getForGitHubServerAppId returns null or undefined if it doesn't find anything
if(gitHubAppId && await GitHubServerApp.getForGitHubServerAppId(gitHubAppId)) {
		return false;
	}
	return true;
}
```

rachellerathbone: Screen Shot 2022-05-31 at 9 34 20 am

IntelliJ is complaining that we can simplify the last conditional. It suggests changing it to `return !(gitHubAppId && await GitHubServerApp.getForGitHubServerAppId(gitHubAppId));` Thoughts? Listen to IntelliJ or ignore it?

bgvozdev: I'd suggest to have some temp var like `isGitHubEnterpriseServer=gitHubAppId && await GitHubServerApp.getForGitHubServerAppId(gitHubAppId)`, and then `return !isGitHubEnterpriseServer`...

rachellerathbone: Updated. Lmk what you think. I'd added some comments for clarity (which is not ideal but we'll be able to clean this function up when we remove the FF). I had to modify the suggestions a little. Couldn't have:
```
if (await booleanFlag(BooleanFlags.GHE_SERVER, false, jiraHost)) { return false; }
```
because it would return false in instances where the feature flag is turned on but there is no gitHubAppId or entry in the GitHubServerApps table (github cloud app).

bgvozdev (approved):

src/util/jira-utils.ts
```diff
@@ -104,4 +106,24 @@ export const jiraIssueKeyParser = (str: string): string[] => {
 	return uniq(Array.from(str.matchAll(regex), m => m[2].toUpperCase()));
 };
 
-export const hasJiraIssueKey = (str:string): boolean => !isEmpty(jiraIssueKeyParser(str));
+export const hasJiraIssueKey = (str: string): boolean => !isEmpty(jiraIssueKeyParser(str));
+
+export const isGitHubCloudApp = async (jiraHost: string): Promise<boolean>=> {
```
bgvozdev, line 111: Should we add a unit-test maybe? and for other changes, too...

bgvozdev (changes requested):

src/util/jira-utils.ts
joshkay10: a lot easier to understand I reckon.
Yea once that FF goes you could probably do what intelliJ said and it will be clean as!

mboudreau: do you even need the FF in there?  githubAppId will be null if it's cloud.  You could literally just do `return !(await Installation.getForHost(jiraHost)).githubAppId`

views/jira-configuration.hbs
```diff
@@ -127,13 +127,15 @@
...
                     {{! Organization }}
                     <td class="jiraConfiguration__table__cell">
-                      <img
-                        class="jiraConfiguration__table__cell__avatar"
-                        src="{{account.avatar_url}}"
-                        height="24"
-                        width="24"
-                        data-installation-id="{{id}}"
-                      />
+                      {{#if ../isGitHubCloudApp}}
```
bgvozdev, line 130: Curious what does `../` mean?

bgvozdev (approved):

PR #1234, opened 2022-06-02
ARC-1355 add jwt cookie

gxueatlassian:
Adding a new mechanism to protect user from connecting to other jira host. 

https://developer.atlassian.com/cloud/confluence/cacheable-app-iframes-for-connect-apps/#retrieving-context-using-ap-context-gettoken-- - essentially we are doing what's recommended in the docs but passing the value in cookies, not in a payload/query parameter


src/jira/util/jwt.ts
```diff
@@ -28,18 +29,22 @@ export enum TokenType {
 	context = "context"
 }
 
-export function extractJwtFromRequest(req: Request): string | undefined {
-	const tokenInQuery = req.query?.[JWT_PARAM];
+let secureJiraHostInCookies;
```
mboudreau, line 32: This var will bleed between different requests

src/jira/util/jwt.ts
```diff
@@ -28,18 +29,22 @@ export enum TokenType {
...
+
+
+function extractJwtFromRequest(req: Request): string | undefined {
 
-	// JWT appears in both parameter and body will result query hash being invalid.
+	booleanFlag(BooleanFlags.SECURE_JIRAHOST_IN_COOKIES, false)
```
mboudreau, line 37: please use `await` instead

bgvozdev: We don't want to change the whole chain to be async @mboudreau cause it is pretty long.

src/jira/util/jwt.ts
bgvozdev: That's what we want :) Once the FF is on, the value will eventually be updated (2nd call after the flip). Given the distributed nature of feature-flags, this is OK imho

gxueatlassian: Yup, understand that, and the full story is:
1. We want to have some feature flags to guard this code. But the middleware is NOT async at the moment, therefore can't use the feature flag (unless we make it async)
2. I understand that making the middleware async should not have problem. But I also think `turning middleware async` vs `have a global var` here, the latter one is less risky.
3. I understand the race condition of multi request, but I think they will all lead to expected result (which is "the flag will be used eventually" if turned on. 

Unless you think there's still impact on this global var?

views/session.hbs
```diff
@@ -46,8 +46,13 @@
...
   $(function() {
     if (window.jiraHost) {
-      // Set jiraHost value to cookie
+      {{!--/*
+            We cannot pass jwt as a query string parameter because of security considerations!
+            (E.g. they can be stored in browser's session or hijacked.)
+            More secure is to pass it using the cookies.
+      */--}}
       document.cookie = "jiraHost=" + window.jiraHost + "; path=/; max-age=" + 60 * 5 + "; samesite=none; secure";
+      document.cookie = "jwt=" + window.jwt + "; path=/; max-age=" + 60 * 5 + "; samesite=none; secure";
```
mboudreau, line 55: We could probably do `encodeURIComponent` of the cookie values in case there's a semi-colon or something else in there.

src/middleware/jira-jwt-middleware.ts
```diff
@@ -2,7 +2,7 @@ import { Installation } from "models/installation";
 import { NextFunction, Request, Response } from "express";
 import { sendError, TokenType, verifySymmetricJwtTokenMiddleware } from "../jira/util/jwt";
 
-const verifyJiraJwtMiddleware = (tokenType: TokenType) => async (
+export const verifyJiraJwtMiddleware = (tokenType: TokenType) => async (
```
mboudreau, line 5: This shouldn't be exported

bgvozdev: Well, we kinda need that...

mboudreau:
I disagree.  We're in this predicament for exactly the same reason: we had a security issue, so someone coded the verifyJwtMiddleware super quickly in a bad manner, but then never readdressed it and now we're still using that bad code to implement more bad code.  In the time we are debating around this issue, creating a new verifyJwtToken function would have already been done with a lot less risk involved as it wouldn't affect any other system and the new code wouldn't rely on the old verify middleware.  I personally think my version is less risky because there's less code in total, less files changed and less dependency on bad code.

But I won't stop you I guess.

mboudreau (approved):

PR #1236, opened 2022-06-03
Hash secret example

joshkay10:
added the global hash secret to example env to help set up env file correctly!


joshkay10:
nope nope nope - has other stuff I dont want

PR #1237, opened 2022-06-06
ARC-1387 Select GitHub version screen

rachellerathbone:
This PR includes UI for GHE: this new screen will allow users to select either cloud or server.

**States of this new page:**
Initial view

Screen Shot 2022-06-06 at 11 03 11 am

Hover state

There is a pointer cursor in the hover state (you can't see it as it disappears when taking a screenshot).
Screen Shot 2022-06-06 at 11 04 46 am

Version selected

Screen Shot 2022-06-06 at 11 04 58 am

Screen Shot 2022-06-06 at 11 05 07 am

**What's not included in this PR:**

- navigation between screens. This PR focuses solely on the page and its various states. Navigation between screens will be addressed in ARC-1406 after all new screens have been built.
- GHE_SERVER feature flag: not needed until we introduce navigation for the GHE flow (this page is currently not rendered at any point)


static/js/jira-select-github-version.js
```diff
@@ -0,0 +1,17 @@
+$('.jiraSelectGitHubVersion__options__card').click(function (event) {
+	event.preventDefault();
+	const selectedCard = event.target.className;
```
gxueatlassian, line 3: maybe we could just do `$('.jiraSelectGitHubVersion__options__card').removeClass('selected')`
and
`$(event.target).addClass('selected')`

------

`selectedCard.substring(selectedCard.indexOf(' ') + 1)` seems not very robust to me as it depends on the order or class names.

krazziekay (changes requested):

static/assets/arrow-left.svg
```diff
@@ -0,0 +1,4 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
mboudreau, line 1: Shouldn't we be using something from the icon library? like `arrow-left` or `arrow-left-circle`?

static/assets/question.svg
```diff
@@ -0,0 +1,5 @@
+<svg width="24" height="24" viewBox="0 0 24 24" fill="none" xmlns="http://www.w3.org/2000/svg">
```
mboudreau, line 1: Same here, there's `question-circle` and `question-filled`

static/css/jira-select-github-version.css
```diff
@@ -0,0 +1,82 @@
...
+}
+
+.jiraSelectGitHubVersion__options__prompt {
+    color: #5E6C84;
+    font-size: 1.2rem;
+    margin-bottom: 1em;
+}
+
+.jiraSelectGitHubVersion__options__cardsContainer {
+    border-radiusL 10px;
```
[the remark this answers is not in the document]
mboudreau: definitely

static/assets/arrow-left.svg
rachellerathbone: I opted to use the svg from Jade's figma designs because the arrow left from AUI is completely different.

Top arrow is the AUI component and the second is the one from the figma designs (and what the Atlaskit arrow looks like):

Screen Shot 2022-06-06 at 2 33 05 pm

static/js/jira-select-github-version.js
rachellerathbone: Good point.

static/js/jira-select-github-version.js
```diff
@@ -0,0 +1,17 @@
...
+	const selectedVersion = selectedCard.substring(selectedCard.indexOf(' ') + 1);
+	const className = `.jiraSelectGitHubVersion__options__card.${selectedVersion}`;
+
+	if (selectedVersion === "cloud") {
+		$(".jiraSelectGitHubVersion__options__card.server").removeClass("selected");
+	}
+
+	if (selectedVersion === "server") {
```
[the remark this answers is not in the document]
rachellerathbone: Sorry, "but the condition as displayed in the screen is, the user can not select both options." is a little confusing. Are you saying the user should be able to should be able to select both options?

static/assets/arrow-left.svg
mboudreau: Are we use Jade is using the correct one? or maybe we need to update AUI?

static/assets/question.svg
rachellerathbone: Same issue as above. AUI component vs figma deisgns (and there doesn't appear to be a way to remove the outer cirlce).

Screen Shot 2022-06-06 at 3 04 52 pm

static/js/jira-select-github-version.js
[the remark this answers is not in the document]
krazziekay: From the UI, it can be observed that the user can select only one option at a time. So, if we do multiple `if`, that means all those `if` will be checked. But if we do a `if-elseif`, then if the first condition is true, it won't do any other checks. It will go to the next `elseif` check only if the first `if` fails.

**Multiple if**
```
if (conditionA) -> do A
if (conditionB) -> do B
```
Here, if `conditionA` is true, then we don't need to check for `conditionB`. But since another `if` is being used, it will do that check, which will surely fail. It seems redundant.

So, instead we can do a `if-elseif`.
**if-elseif**
```
if (conditionA) -> do A
else if (conditionB) -> do B
```
So now if `conditionA` is true, then it won't do the other check. The only time if it would do the other check is if `conditionA` fails.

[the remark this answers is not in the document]
rachellerathbone: lol I know how conditionals work. I meant the wording of your comment was confusing.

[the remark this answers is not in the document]
krazziekay: ah apologies, I was just mentioning that both conditions can not be true, so its better to use `if-elseif`....

static/js/jira-select-github-version.js
gxueatlassian: How about `$(event.target).closest('.jiraSelectGitHubVersion__options__card').addClass('selected')`?
I think above may allow us to remove the whole block bellow:
```javascript
        const selectedVersion = selectedCard.substring(selectedCard.indexOf(' ') + 1);
	const className = `.jiraSelectGitHubVersion__options__card.${selectedVersion}`;

	if (selectedVersion === "cloud") {
		$(".jiraSelectGitHubVersion__options__card.server").removeClass("selected");
	}

	if (selectedVersion === "server") {
		$(".jiraSelectGitHubVersion__options__card.cloud").removeClass("selected");
	}

	$(`${className}`).addClass("selected");
```

static/js/jira-select-github-version.js
```diff
@@ -0,0 +1,15 @@
...
+	event.preventDefault();
+	const selectedCard = event.target.className;
+	const selectedVersion = selectedCard.includes("cloud") ? "cloud" : "server";
+	const className = `.jiraSelectGitHubVersion__options__card.${selectedVersion}`;
+
+	if (selectedVersion === "cloud" && $(".jiraSelectGitHubVersion__options__card.server").hasClass("selected")) {
+		$(".jiraSelectGitHubVersion__options__card.server").removeClass("selected");
+	} else if (selectedVersion === "server" && $(".jiraSelectGitHubVersion__options__card.cloud").hasClass("selected")) {
+		$(".jiraSelectGitHubVersion__options__card.cloud").removeClass("selected");
+	}
```
mboudreau, line 11: Wouldn't it be easier to just remove 'selected' from all cards, then add it to the appropriate one?

```
$(".jiraSelectGitHubVersion__options__card").removeClass("selected");
$(`${className}`).addClass("selected");
```
No conditionals needed.

static/assets/arrow-left.svg
rachellerathbone: I had to add the same arrow to the nav in Jenkins and it's the same is what's in the figma designs for GHE.

How would we go about updating AUI?

static/js/jira-select-github-version.js
rachellerathbone: That looks like it would select the correct element but we would still need to remove it from the other card if the user than selects the other option. How would this help us remove the conditional block?

gxueatlassian: I meant we could first remove the `'.selected'` from both of the `'.jiraSelectGitHubVersion__options__card'`, then add `'.selected'` only to the `$(target).closest('.jiraSelectGitHubVersion__options__card')`
---
But having say that, I am happy to leave the code as it is. I don't see a big problem here.

gxueatlassian (dismissed):
I don't see a big problem. Approved.

static/js/jira-select-github-version.js
rachellerathbone: Nice. Much simpler 😄

static/js/jira-select-github-version.js
rachellerathbone: Did you see the most recent change?

rachellerathbone: 🤦 realised I forgot to push them.

static/js/jira-select-github-version.js
```diff
@@ -0,0 +1,10 @@
+$('.jiraSelectGitHubVersion__options__card').click(function (event) {
+	event.preventDefault();
+	const selectedCard = event.target.className;
+	const selectedVersion = selectedCard.includes("cloud") ? "cloud" : "server";
```
mboudreau, line 4: `$(event.target).hasClass("cloud")` would be better IMO

mboudreau: actually, another way of doing it would be to use `data-` attributes on the HTML element.  So in your html, you'd do something like `<div data-type="cloud">...` and for your code you just need to do `const selectedVersion = $(event.target).data("type");`

mboudreau (approved):

joshkay10 (approved):
🤟

krazziekay (approved):

static/js/jira-select-github-version.js
gxueatlassian: Oh, haha, all good. Now much better :) 
ps: And I forgot the approve it again, sorry.

PR #1239, opened 2022-06-07
ARC-1390 manual vs auto app creation

rachellerathbone:
This PR includes UI for GHE: this new screen will allow users to select either automatic or manual app creation.

**States of this new page:**
Initial view

Screen Shot 2022-06-07 at 10 25 25 am

Hover state

There is a pointer cursor in the hover state (you can't see it as it disappears when taking a screenshot).

Screen Shot 2022-06-07 at 10 25 32 am

Version selected

Screen Shot 2022-06-07 at 10 25 39 am

Screen Shot 2022-06-07 at 10 25 45 am

**What's not included in this PR:**

- navigation between screens. This PR focuses solely on the page and its various states. Navigation between screens will be addressed in issue 1406 after all new screens have been built.
- GHE_SERVER feature flag: not needed until we introduce navigation for the GHE flow (this page is currently not rendered at any point)


static/js/jira-select-card-option.js
```diff
@@ -0,0 +1,8 @@
+$(".optionsCard").click(function (event) {
+	event.preventDefault();
+	const selectedVersion = $(event.target).data("type");
```
gxueatlassian, line 3: Okay this is really minor. I am happy with the code as it it. It is just suggestions:

It seems we don't need the `.data("type")` here. We can just do the following:

```javascript
$(".optionsCard").click(function (event) {
	event.preventDefault();
	$(".optionsCard").removeClass("selected");
  $(event.target).closet('.optionsCard').addClass('selected');
	$(".optionBtn").prop("disabled", false).attr("aria-disabled", "false").addClass("aui-button-primary");
});
```

So @mboudreau opinions? You mentioned the "data('type')" yesterday, I want to ask your thoughts. 

---

for [this](https://stackoverflow.com/a/10086501), we could replace the  
`$(event.target).closet('.optionsCard').addClass('selected');`
to
`$(event.currentTarget).addClass('selected');`

gxueatlassian (dismissed):

mboudreau (approved):

PR #1242, opened 2022-06-07
Added a hitbox for the sync warning hover icon

joshkay10:
**In this PR**

Added a hitbox to the sync warning icon to keep popup open when transitioning from icon to popup.

Couple of options here in how to achieve it... went with least intrusive.

**Proof**
The thing to look for in the screenshot is the location of the cursor(just above the warning icon) before the change the popup would vanish if the cursor went to the deadzone between icon and popup, much annoying.

It now has a generous hitbox around it so that its more consistent with that atlas feel.

Screen Shot 2022-06-08 at 10 23 06 AM

### DIFF
--- a/static/css/jira-configuration.css
+++ b/static/css/jira-configuration.css
@@ -159,8 +159,7 @@
 /* Hit box */
 .jiraConfiguration__syncWarning:before {
   content: '';
-  width: 3rem;
-  height: 3rem;
+  padding: 1.25em;
   position: absolute;
 }
 
--- a/static/css/jira-configuration.css
+++ b/static/css/jira-configuration.css
@@ -156,6 +156,14 @@
   padding: 0.2rem;
 }
 
+/* Hit box */
+.jiraConfiguration__syncWarning:before {
+  content: '';
+  width: 3rem;
+  height: 3rem;
+  position: absolute;
+}
+
 .jiraConfiguration__syncWarning:hover {
   background: #0914E208;
 }


bgvozdev (dismissed):

rachellerathbone:
Can we see a screenshot?

joshkay10:
> Can we see a screenshot?

Seems reasonable, will get one now.

static/css/jira-configuration.css
```diff
@@ -156,6 +156,14 @@
   padding: 0.2rem;
 }
 
+/* Hit box */
+.jiraConfiguration__syncWarning:before {
+  content: '';
+  width: 3rem;
+  height: 3rem;
```
mboudreau, line 163: Not a fan of having set height and width as it can affect ux if the warning gets bigger or it might go over another components hitbox nearby.

rachellerathbone: Just noticed we're using a lot of mixed measurements in this css file. I think we're going to update this to align with what we decided on for the standard of the Jenkins app but right now we're using:
- px for borders
- em for margins and padding (so the margins and padding are relative to the parent element)
- rem for font (so it's relative to the root unit of measurement of the page)

gxueatlassian (approved):

static/css/jira-configuration.css
joshkay10: @mboudreau  - Changed it to a delightful 1.25em of **padding** instead of fixed size.
@rachellerathbone - replaced filthy **rem** with **em**.

bgvozdev (approved):

gxueatlassian (approved):
Good trick

static/css/jira-configuration.css
mboudreau: image

mboudreau (approved):

PR #1243, opened 2022-06-07
Feature Flag remove - USE_NEW_GITHUB_CLIENT_FOR_GET_INSTALLATION

joshkay10:
**Whats in this PR**
Removing feature flag USE_NEW_GITHUB_CLIENT_FOR_GET_INSTALLATION

Removed the octokit object entirely from getInstallations function so had to tidy up calls to getInstallations and tests.


src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -40,18 +38,14 @@ export const getInstallations = async (client: GitHubAPI, subscriptions: Subscri
 	};
 };
 
-const getInstallation = async (client: GitHubAPI, subscription: Subscription, log: Logger): Promise<AppInstallation> => {
+const getInstallation = async (subscription: Subscription, log: Logger): Promise<AppInstallation> => {
 
 	const { jiraHost } = subscription;
-	const useNewGitHubClient = await booleanFlag(BooleanFlags.USE_NEW_GITHUB_CLIENT_FOR_GET_INSTALLATION, false, jiraHost) ;
```
bgvozdev, line 46: Shouldn't the flag also be removed from feature-flags file?

joshkay10: oo thank you, good catch!

bgvozdev (approved):

gxueatlassian (approved):

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -63,12 +57,12 @@ const getInstallation = async (client: GitHubAPI, subscription: Subscription, lo
...
 	} catch (err) {
 		log.error(
-			{ installationId: gitHubInstallationId, error: err, uninstalled: err.code === 404 },
+			{ installationId: gitHubInstallationId, error: err, uninstalled: err.status === 404 },
 			"Failed connection"
 		);
 		statsd.increment(metricError.failedConnection);
 
-		return Promise.reject({ error: err, id: gitHubInstallationId, deleted: err.code === 404 });
+		return Promise.reject({ error: err, id: gitHubInstallationId, deleted: err.status === 404 });
```
mboudreau, line 65: is it from code to status correct?  Fairly certain this is an error of type `AxiosError` and the correct check is `err.response.status`

joshkay10: investigating

joshkay10: Good news everyone.
Throws `AxiosError` with `err.response.status` but then gets intercepted and becomes a `GithubClientError` before it reaches the catch block.... at this point it becomes err.status.

We will get failed tests if this changes :)

mboudreau: ah right.  In that case, maybe change the error type? :)

joshkay10: It's a try/catch error so typescript wont allow me to type it as anything other than any or undefined.
Unless I misunderstood?

mboudreau: ah right, stupid try/catch

PR #1245, opened 2022-06-08
updating data model for GHE app security

mboudreau:
After talking to @rachellerathbone about the GHE flow, I had a thought around how we would secure GHE apps to a specific Jira instance since we have no association except through the subscription, but if we can "reuse" the app to add an org, we need to lock it down as a user could input someone else's GHE url and see their data.

As such, we need to force/limit GHE apps per Jira instance.  Even if a company has multiple Jira instances, they cannot reuse the same app and needs to create a new one minimum per instance for the integration to work as there's no other way to "validate" the access to that GHE app without implementing our own permission model on top of it.

I've updated the model documentation and diagram to reflect these changes.  Please let me know what you think.


docs/draw.io/data-model.drawio.xml
```diff
@@ -1 +1 @@
...
...l5ziPojSeI+z/pCCAXIEBFhCZWqnGlJ3JAcgZVLl7dUuUeV80ZeKAIYtQnwWAWma6AzERXalNIST5r+NRMfnePYeslerqqk7flnNF1VKXvi2B5le7Zzlr6FGf1V9ODg27zANBQJ0UAgLHDLRYomI2zlewU96moPXZLWTqzPdO4FOv38JNwUbhxdCdQgdDUjB/9zF4hygJqowWoAfwgbwgVaAZiOOH3l1y1pVVYatRm+wmdOSTiJTxBJBCmRJJELXCPv2C/e6eMWS39vM83JQRfo53o4Mt2ssrTtPlLHawH3Vq1JIambuy4pdXI0NeViipkZfMb4rh871bUaBLhWDlylRpKZ0HXUQRbcacnI5wZSrwhoRLXly5hxGabkKHCgqBnXK1oFy2agflNJVLXi4qKVdcCLJMs05GaYY7U+xYlMY0JXhOYh1FOH5/82Z+6mfkJ3tNHH6zeldXtSGd4fOz9n+rinEWbtWbcjbQxZImYom0p6VQd121grr1WYtfIZiY8rZtKZj4dQlwOlsvmSLsP9HOlhPiFZzNEXqUpvBLHJxOaMp04FWhSUlQ7WCxSbXv2m3os2s2taH/Yp3TzVFZeHR5795QCM/B9u5N1T51R6K3RSJRbo1E9fvjHYneCImyDKwtEllyKt4sif7oSPRaEpmavJfWLIlUb9p0JDpqEpnVV5NETtLW84yW6t2DLpy9LRIpEutmSdSX0Gp25TCZV0seTyj6/Mph/mhQr/hgUP6cUM2jQbVIPrviKLb7io/03F//+IBuifdxeDv7MNv8NX335YN4senQK5PVJ/osEdoaWpm0utn9mxOhTE1EOtR6Tt3N7t8ciaqRLNvcai2SjdqIZHt4OPVVEXC/kWzrfuiRPJxqtz0B78Ti1RNwS/HiQaMRR7zl2E3A3zCJBs2t4ihT/FZeXqkLONmxmoCzl0CybabzbBwRW8TNxxF4T35urMvNxxAZf0c3w0//WLOW0Fv75Gvhc2G6Sks5dKzwMuQqXj0slHZDwG3rZEG6P91LWOI5WsyWcY24FRDZRypsjs7tyoxKliZLIU3WHqRJSW7jVyH361Z4XnpTbH3fZEeB2/sSj61XZlrmwZZ4lFxqe6/iNPNtxZcNSOTdeYVH/UUDh0mVlByq/9KYLt0+Vg71qxwyD8UhWsy/+TCVrfzrI83r/wE=</diagram></mxfile>
```
gxueatlassian, line 1: If you uncheck this, it will make xml un-compressed, which should be better for git version control.

image

gxueatlassian: it is in the preference settings of the the drawio app

gxueatlassian: Can't comments in the jpg image, but just curious is the "1__*" relation between "Installation" and "GithubServerApp" a concret FK constraint or just a virtual relation?

gxueatlassian (approved):

docs/draw.io/data-model.drawio.xml
Harminder84, line 1: @mboudreau, I think there will be one to many relationship b/w `GithubServerApp` and `Subscription` tables. This is use case where  one Jira host and they choose to create 1 GH app that will be used to connect multiple orgs.

gxueatlassian: > @mboudreau, I think there will be one to many relationship b/w `GithubServerApp` and `Subscription` tables. This is use case where one Jira host and they choose to create 1 GH app that will be used to connect multiple orgs.

But won't that end up multiple `Subscription`?

Harminder84: You will have one entry in `GithubServerApp` and multiple entries in `Subscription` for each connected org. So, it is "1 to *"

gxueatlassian: I think @mboudreau mentioned `As such, we need to force/limit GHE apps per Jira instance` in the PR description. Not sure the whole story, but I suspect that is related to your question?

Harminder84: That is forced by having `installationId` in `GithubServerApp` table

docs/data-model.md
```diff
@@ -13,7 +13,9 @@ Installations are only specific to the Jira app, 1 per Jira instance.  It holds
 
 ### GitHubServerApp
 
-Each GitHub App created in a GitHub Enterprise Server instance is added to this table to keep track of them.  We do not hold the GitHub Cloud app in this table since we need it as a default, but still use the same data structure in the code.  Each App will have access to many Organizations under it and we much show all results to the users when needed, which means multiple API calls .
+Each GitHub App created in a GitHub Enterprise Server instance is added to this table to keep track of them.  A Github app _can only be associated with one Jira instance for security reasons_.  This means that each new Jira instance needs to create a new GHE app.
```
thombergs, line 16: "for security reasons" is correct, of course, but it's so often used as a blanket argument that I would want a sentence here that explains why exactly (because each Jira instance can be owned by a different customer and we don't want to allow accidental communication from a GitHub org of customer A with a Jira instance of customer B)

docs/draw.io/data-model.drawio.xml
mboudreau: Yep, good suggestion.  We should probably add a foreign key to the GithubServerApp table to the Installation.  We need to clean up our DB schema frankly, including removing duplicate installations and a better foreign key between installations and subscriptions.

mboudreau: actually, it's a `* to 1` relationship, you're right.  Many subscriptions can point to a single GHapp model

docs/data-model.md
mboudreau: yeah, good point.

docs/draw.io/data-model.drawio.xml
```diff
@@ -1 +1,141 @@
...
\ No newline at end of file
+<mxfile host="app.diagrams.net" modified="2022-06-09T00:49:27.456Z" agent="5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/101.0.0.0 Safari/537.36" etag="3GRBDWr6KHYVjdFBL5xY" compressed="false" version="19.0.3" type="device">
```
gxueatlassian, line 1: Hey @mboudreau I think you forgot to re-generate the jpg.

mboudreau: whoops, good catch

atrigueiro (approved):

Harminder84 (approved):

PR #1250, opened 2022-06-09
ARC-1419 Make connected org list scrollable

Harminder84:
I've made org list scrollable to view large number of connected orgs.
Screen Shot 2022-06-09 at 3 38 22 pm

### DIFF
--- a/views/jira-configuration.hbs
+++ b/views/jira-configuration.hbs
@@ -107,161 +107,163 @@
           </div>
 
           {{#if successfulConnections}}
-            <table class="jiraConfiguration__table aui">
-              <thead class="jiraConfiguration__table__head">
-                <tr class="jiraConfiguration__table__head_row">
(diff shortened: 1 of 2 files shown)


krazziekay (dismissed):
Looks good...
(Non blocker: The indent spaces in the `hbs` are a bit too much, better to fix that)

views/jira-configuration.hbs
```diff
@@ -107,161 +107,163 @@
           </div>
```
gxueatlassian, line 107: Are all the changes in this file just format/style change?
Can we exclude those in this PR? So that we can keep the PR absolutely focus on its intention.

gxueatlassian (approved):

rachellerathbone (approved):

Harminder84:
> 

Fixed indenation issue

bgvozdev (approved):

PR #1251, opened 2022-06-09
ARC-1348: testing a new PRIVATE_KEY

bgvozdev:
The new GitHub app's PRIVATE_KEY is generated and is in vault, however before the switch we want to make sure it works. That's why we are doing this temp. change. Once we confirm nothing was broken, we can assume that the key is fine and swap it with the current one.


src/github/client/app-token-holder.ts
```diff
@@ -5,13 +5,23 @@ import { AuthToken, ONE_MINUTE, TEN_MINUTES } from "./auth-token";
...
 export type KeyLocator = (installationId: InstallationId) => string;
 
 /**
  * By default, we just look for a key in the `PRIVATE_KEY` env var.
  */
+let shouldUseNewSecret = false;
 export const cloudKeyLocator: KeyLocator = () => {
+	booleanFlag(BooleanFlags.USE_NEW_GITHUB_PRIVATE_KEY, false)?.then(newValue => shouldUseNewSecret = newValue);
+	if (shouldUseNewSecret) {
+		return Buffer.from(envVars.PRIVATE_KEY_VAULT, 'base64').toString();
```
bgvozdev, line 23: We are mimicking probot here: https://github.com/probot/get-private-key/blob/094fe45cf040c345802bc81a12dac576b92a4589/src/index.ts#L32

rachellerathbone (approved):

gxueatlassian (dismissed):

gxueatlassian (approved):

src/github/client/app-token-holder.ts
gxueatlassian: Tbh I don't think we should say "we are mimicking probot", as this is required because we put base64 encoded key in vault, not because we are or not using probot.

gxueatlassian (approved):

PR #1252, opened 2022-06-09
ARC-916 - GHE UI: Connect GitHub Enterprise Server page

krazziekay:
**Rationale**
- This PR includes the addition of a new view where users can add the URL for their GHE server.

**Changes**
- Added a new view `hbs`, with its respective `js` and `css` file

- The view has an empty field and a disabled button by default.
Screen Shot 2022-06-09 at 3 46 48 pm

- The button is enabled once the input field has some value.
Screen Shot 2022-06-09 at 3 47 14 pm

- For any invalid URLs typed, the error message is displayed once you click on the button.
Screen Shot 2022-06-09 at 4 09 14 pm
Screen Shot 2022-06-09 at 4 09 54 pm

- For a valid URL typed, clicking on the button will simply change the text on the button to a spinner. (Nothing will happen for now)
Screen Shot 2022-06-09 at 3 47 33 pm

**Whats not done**
- URLs for the link `Learn More`
- Tooltips for the _QuestionMark_ icon.
- The action for the `Next` button.
- The links for the left arrow and the Help link.
- The flow when user types in a **GitHub Cloud** url, i.e. `github.com`
    - https://softwareteams.atlassian.net/browse/ARC-1418


static/js/jira-connect-ghe.js
```diff
@@ -0,0 +1,62 @@
+const ALLOWED_PROTOCOLS = ["http:", "https:"];
```
gxueatlassian, line 1: shall we only allow https?

static/js/jira-connect-ghe.js
```diff
@@ -0,0 +1,62 @@
...
+const ALLOWED_DOMAIN_EXTENSIONS = [".com", ".net"];
```
gxueatlassian, line 2: there could be `.org`. Not sure we really need to restrict our top level domain?

static/js/jira-connect-ghe.js
```diff
@@ -0,0 +1,62 @@
...
+			return false;
+		}
+
+		return true;
+	} catch (e) {
+		return false;
+	}
+};
+
+$("#gheServerURL").on("keyup", event => {
```
gxueatlassian, line 32: might need some `debounce` here for better UX, jquery has a plugin, or we could write our own `debounce` (should be easy, just some `setTimeout/clearTimeout`)

--

Or actually it is okay to NOT `debounce`, as this piece of code is small?

static/js/jira-connect-ghe.js
```diff
@@ -0,0 +1,62 @@
...
+/**
+ * Method that checks the validity of the passed URL
+ *
+ * @returns {boolean}
+ */
+const checkValidUrl = inputURL => {
+	try {
+		const { protocol, hostname} = new URL(inputURL);
+
+		if (!ALLOWED_PROTOCOLS.some(allowedProtocol => protocol.startsWith(allowedProtocol))) {
```
gxueatlassian, line 14: Probably should use `===` instead of `startsWith`?

static/js/jira-connect-ghe.js
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Method that checks the validity of the passed URL
+ *
+ * @returns {boolean}
+ */
+const checkValidUrl = inputURL => {
+	try {
+		const { protocol, hostname } = new URL(inputURL);
+
+		if (!ALLOWED_PROTOCOLS.some(allowedProtocol => protocol === allowedProtocol)) {
```
joshkay10, line 14: /nit `some` is good if you need the function to compare, in this case I prefer to use a `includes`
e.g. `ALLOWED_PROTOCOLS.includes(protocol)`

static/js/jira-connect-ghe.js
rachellerathbone: We need to allow both as we have no control which protocol our GHE customers use. For instance, our own internal instance uses http so it's safe to assume we're not the only GHE account holders with the same setup.

static/css/jira-connect-ghe.css
```diff
@@ -0,0 +1,55 @@
...
+  margin-top: 2px;
+}
+
+label {
+  font-size: 12px;
+  font-weight: 600;
+  color: #6B778C;
+}
+
+.ghe-server-input {
```
rachellerathbone, line 35: Why has this class name broken from the jiraConnectGH__ pattern? All class names in this file should be prefixed with the same thing.

static/js/jira-connect-ghe.js
```diff
@@ -0,0 +1,63 @@
+const ALLOWED_PROTOCOLS = ["http:", "https:"];
+const GITHUB_CLOUD = ["github.com", "www.github.com"];
+
+/**
+ * Method that checks the validity of the passed URL
+ *
+ * @returns {boolean}
+ */
+const checkValidUrl = inputURL => {
+	try {
```
rachellerathbone, line 10: Don't think we need to wrap this in a try/catch block. We really just want to return a boolean: if valid return true, if invalid return false. An invalid result will tell the UI to render the error.

static/js/jira-connect-ghe.js
```diff
@@ -0,0 +1,63 @@
+const ALLOWED_PROTOCOLS = ["http:", "https:"];
+const GITHUB_CLOUD = ["github.com", "www.github.com"];
+
+/**
+ * Method that checks the validity of the passed URL
+ *
+ * @returns {boolean}
+ */
+const checkValidUrl = inputURL => {
```
rachellerathbone, line 9: We need to add typing here. What are we passing in and what are we returning?

static/js/jira-connect-ghe.js
```diff
@@ -0,0 +1,63 @@
+const ALLOWED_PROTOCOLS = ["http:", "https:"];
+const GITHUB_CLOUD = ["github.com", "www.github.com"];
```
rachellerathbone, line 2: Is this is the check for users who are meaning to connect to cloud but accidentally go down the server path? Shouldn't we do a hard check against the github.com url? https://github.com. My thoughts are it would need to be an exact match for us to assume this is what we are trying to do, otherwise, we make a request to what they enter and render the error when it fails. Plus, the first item 'github.com' could cause problems for users who have internal instance running on a domain that is something like 'http://my-internal-github.com'

static/js/jira-connect-ghe.js
```diff
@@ -0,0 +1,63 @@
...
+};
+
+$("#gheServerURL").on("keyup", event => {
+	const hasUrl = event.target.value.length > 0;
+	$("#gheServerBtn").attr({
+		"aria-disabled": !hasUrl,
+		"disabled": !hasUrl
+	});
+	$("#gheServerURLError").hide();
+	$("#gheServerURL").removeClass("has-error ");
```
rachellerathbone, line 38: Nit: why the trailing space?

views/jira-connect-ghe.hbs
```diff
@@ -0,0 +1,102 @@
...
+            src="/public/assets/jira-and-github.png"
+            alt="Jira and GitHub logos"
+          />
+        </div>
+        <h2 class="jiraConnectGHE__content__title">GitHub Enterprise Server</h2>
+
+        <div class="jiraConnectGHE__prompt">
+          Connecting GitHub Enterprise Server to Jira requires a hole in your firewall. More copy/details here.
+          <!--TODO: Add a link here-->
+          <a href="">Learn more.</a>
```
rachellerathbone, line 48: We'll be linking off to a page outside our app here so we need to add target=_blank.

views/jira-connect-ghe.hbs
```diff
@@ -0,0 +1,102 @@
...
+            Enter the URL for the server of the GitHub Enterprise organization that you want to connect. This is unique to your server.
+          </div>
+          <div>
+            <label for="gheServerURL">Server URL</label>
+            <input type="text" class="ghe-server-input" id="gheServerURL" placeholder="https://example.com" />
+            <div class="jiraConnectGHE__error" id="gheServerURLError">
+              <i class="aui-icon aui-iconfont-error"></i>
+              The entered URL is not valid.
+              <!--TODO: Add a link here-->
+              <a href="">Learn more.</a>
```
rachellerathbone, line 67: Same here. Need target attribute.

rachellerathbone (changes requested):

static/js/jira-connect-ghe.js
krazziekay: This has been removed

static/js/jira-connect-ghe.js
krazziekay: I think debounce here would be an overkill, as there is no API call nor any major performance affecting things happening. The main function happens once you click the button.

static/js/jira-connect-ghe.js
krazziekay: Updated

static/js/jira-connect-ghe.js
krazziekay: Updated

static/js/jira-connect-ghe.js
krazziekay: Removed the extra space

views/jira-connect-ghe.hbs
krazziekay: Added

views/jira-connect-ghe.hbs
krazziekay: Added

static/css/jira-connect-ghe.css
```diff
@@ -0,0 +1,55 @@
...
+}
+
+.jiraConnectGHE__error {
+  display: none;
+  color: #DE350B;
+  font-size: 12px;
+  margin-top: 2px;
+}
+
+label {
```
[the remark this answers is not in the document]
krazziekay: Class name added `jiraConnectGHE__label`

static/js/jira-connect-ghe.js
krazziekay: Defined the types in the annotations above...

static/css/jira-connect-ghe.css
```diff
@@ -0,0 +1,55 @@
+.jiraConnectGHE__content {
+  margin: 4em auto 0;
+  max-width: 560px;
+}
+.jiraConnectGHE__content__title {
+  text-align: center;
+  margin-bottom: 32px;
```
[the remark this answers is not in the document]
krazziekay: The `px` for margins, paddings and font size were updated to `em` and `rem`

static/js/jira-connect-ghe.js
krazziekay: Yes this simply checks if the hostname is one of the `GITHUB_CLOUD`. I am doing a `===` check, so it is doing an exact match with the user's input URL.  
I am doing a `hostname` check, so shouldn't be a problem for domain names like `http://my-internal-github.com`.

static/js/jira-connect-ghe.js
rachellerathbone: But why do we need to throw an error here? This should be a simple check to determine if the url is valid or not.

krazziekay: To separate out the host and protocols and thus checking if passed string is a correct URL or not, I am using `new URL()`. This command throws a **JS error** in console for any invalid strings. So those invalid errors are being captured by `catch` which returns a false value.

static/js/jira-connect-ghe.js
```diff
@@ -0,0 +1,64 @@
+const ALLOWED_PROTOCOLS = ["http:", "https:"];
+const GITHUB_CLOUD = ["github.com", "www.github.com"];
+
+/**
+ * Method that checks the validity of the passed URL
+ *
+ * @param {string} inputURL
+ * @returns {boolean}
+ */
+const checkValidUrl = inputURL => {
```
joshkay10, line 10: Worthwhile calling out that its checking a valid GHE url so `const checkValidGHEUrl` or something of the sort would be nice

static/js/jira-connect-ghe.js
```diff
@@ -0,0 +1,64 @@
...
+	try {
+		const { protocol, hostname } = new URL(inputURL);
+
+		if (!ALLOWED_PROTOCOLS.includes(protocol)) {
+			return false;
+		}
+		if (hostname.split('.').length < 2) {
+			return false;
+		}
+		if (GITHUB_CLOUD.some(ghCloud => ghCloud === hostname)) {
```
joshkay10, line 20: i'd use includes again

static/js/jira-connect-ghe.js
```diff
@@ -0,0 +1,64 @@
...
+ * @returns {boolean}
+ */
+const checkValidUrl = inputURL => {
+	try {
+		const { protocol, hostname } = new URL(inputURL);
+
+		if (!ALLOWED_PROTOCOLS.includes(protocol)) {
+			return false;
+		}
+		if (hostname.split('.').length < 2) {
```
joshkay10, line 17: .... this is to check .com .org .net etc??
Why do we need to check this?

what you originally had was clearer with intent, but if this catches everything then this could be fine, 
is localhost or server names valid hosts?

maybe just a comment above the if statement what its trying to achieve?

static/js/jira-connect-ghe.js
krazziekay: Renamed to `checkValidGHEUrl`

static/js/jira-connect-ghe.js
krazziekay: Replaced `some` by `includes`

static/js/jira-connect-ghe.js
krazziekay: As mentioned by @rachellerathbone , the reason for this change is to ensure that there is an extension at the end like `.com` or `.net`, but we don't want to do any specific checks like before

joshkay10 (dismissed):

rachellerathbone:
I just pulled your branch to have a look at the UI locally and the font looks really small. Remember, when working in the iframe you can't use the exact measurements that you see in Figma because we're working with a smaller real estate thanks to the left panel and navigation. Makes it hard to figure out the exact measurements but you should be able to look at the design and your rendering and not really see any differences. The font in the UI is noticeably smaller.

static/js/jira-connect-ghe.js
```diff
@@ -0,0 +1,65 @@
...
+			return false;
+		}
+		// This checks whether the hostname whether there is an extension like `.com`, `.net` etc.
+		if (hostname.split('.').length < 2) {
+			return false;
+		}
+		if (GITHUB_CLOUD.includes(hostname)) {
+			// TODO: Need to alert the users that this URL is GitHub cloud and not Enterprise
+			// Waiting for design: https://softwareteams.atlassian.net/browse/ARC-1418
+			return false;
```
rachellerathbone, line 24: I don't think we should return false here otherwise it will render the invalid URL message. I'd say, until we have the design update with how we should render the 'Were you meaning to connect to cloud?' message, you just drop a console.log in here. Something like `console.log('Cloud URL entered.');` just so it's clear this is not an error state.

views/jira-connect-ghe.hbs
```diff
@@ -0,0 +1,103 @@
...
+            Connecting GitHub Enterprise Server to Jira requires opening a hole in your firewall,
+            as Atlassian’s IP addresses must be able to call your GitHub Enterprise Server.
+          <!--TODO: Add a link here-->
+          <a href="" target="_blank">Learn more.</a>
+        </div>
+
+        <div>
+          <h3 class="jiraConnectGHE__options__subTitle">
+            GitHub Enterprise Server URL&nbsp;
+            <!--TODO: Add a tooltip for this icon-->
```
rachellerathbone, line 55: Going off Jade's comment in slack, let's remove this todo and the question mark.

static/js/jira-connect-ghe.js
krazziekay: Agreed. 
This method `checkValidGHEUrl` was build to check if the URL is valid or not and this GitHub cloud check is unrelated to that. 
So, I'm separating this into a new method which only checks if the passed valid URL is a GitHub cloud or enterprise.

views/jira-connect-ghe.hbs
krazziekay: Removed

rachellerathbone (approved):

krazziekay:
This PR was accidentally merged before testing in staging. So has been reverted.
The new version of this PR was created here: https://github.com/atlassian/github-for-jira/pull/1267.

PR #1253, opened 2022-06-09
ARC-1372 send unsafe logs to restricted environment

joshkay10:
**Whats in this PR**

We can create a new logger that will sends logs to the 'unsafe' environment. This means that for some data e.g. jirahost, orgname we don't need to hash it.

**How is it consumed**

With this option we would have to wrap up the logging each time with the `LOG_UNSAFE_DATA` feature flag
Similar to how we do currently with `VERBOSE_LOGGING`

```
const unsafeLogger = getUnsafeLogger("nameoflogger");

if (await booleanFlag(BooleanFlags.LOG_UNSAFE_DATA, false, jiraHost)) {
  unsafeLogger.warn({ jiraHost }, "unsafe logger - SHOW ME UGC BEHIND THE FF"); 
}
```

Things will probably become a little bit clearer once we start hashing and removing some of the UGC/PII.
Once this works starts we may find this solution is trash or its fine.... this left the most flexibility IMO.....

```
                                  .-.
     (___________________________()6 `-,
     (   ______________________   /''"`
     //\\                      //\\
    "" ""                     "" ""
```


src/config/logger.ts
```diff
@@ -91,6 +91,11 @@ export const getLogger = (name: string): LoggerWithTarget => {
 	return logger.child({ name });
```
joshkay10, line 91: Adding the feature flag here and returning safe or unsafe logger was considered.
Reasons against... 
- would convert this function to async :( its consumed everywhere so would be a headache(onflagchange event might be an option though...)
- When we log the confidential data we need to hash for safe and not hash for unsafe so its not quite as simple as a gate to the environment there is unique logic likely per log event.

src/config/logger.ts
```diff
@@ -91,6 +91,11 @@ export const getLogger = (name: string): LoggerWithTarget => {
...
 };
 
+// This will log data to a restricted environment [env]-unsafe
+export const getUnsafeLogger = (name: string): LoggerWithTarget => {
+	return logger.child({ name, env_suffix: "unsafe" });
```
joshkay10, line 96: this little fella `env_suffix: "unsafe"` is the difference, Splunk will index these logs off into there own env.

So on splunk you can search safe logging `env=stg-west` or unsafe logging `env="sg-west-unsafe"`

rachellerathbone (approved):

gxueatlassian (approved):

src/config/logger.ts
bgvozdev: What's the process of acquiring access to the unsafe logs?

joshkay10: Still confirming that things will work.. but the intention is to have a ACL that will be controlled by Christian and is assigned to the unsafe environment on the microscopes observability page.... at this point it would be manual, but can look into other options so that access expires and we have better visibility of who used it.

PR #1255, opened 2022-06-10
ARC-1348 start using the private key from vault

bgvozdev:
After this one is merged, the new private key will be used. After that, we can remove the old key. We've tested the key itself for some requests and it worked just fine!

I'm currently testing the swap in staging, won't merge until confirm it works as expected.


src/github/client/app-token-holder.ts
```diff
@@ -5,23 +5,16 @@ import { AuthToken, ONE_MINUTE, TEN_MINUTES } from "./auth-token";
...
 export type KeyLocator = (installationId: InstallationId) => string;
 
 /**
  * By default, we just look for a key in the `PRIVATE_KEY` env var.
  */
-let shouldUseNewSecret = false;
 export const cloudKeyLocator: KeyLocator = () => {
-	booleanFlag(BooleanFlags.USE_NEW_GITHUB_PRIVATE_KEY, false)?.then(newValue => shouldUseNewSecret = newValue);
-	if (shouldUseNewSecret) {
-		return Buffer.from(envVars.PRIVATE_KEY_VAULT, 'base64').toString();
```
gxueatlassian, line 23: Forgot to remove `envVars.PRIVATE_KEY_VAULT` within the `envVars.ts` file?

bgvozdev: Good point! Thanks :)

joshkay10 (approved):

bgvozdev:
Tested in ddev.

joshkay10 (approved):

PR #1256, opened 2022-06-10
ARC-1240: Truncate issue keys when sending PRs to Jira.

atrigueiro:
We have a bunch of pull request webhooks failing with 400 errors.
Turns out we weren't truncating issue keys when sending them to Jira.


src/jira/client/jira-client.test.ts
```diff
@@ -54,6 +54,71 @@ describe("Test getting a jira client", () => {
 		expect(subscription.syncWarning).toEqual("Exceeded issue key reference limit. Some issues may not be linked.");
 	});
 
+	it("Should truncate issueKeys for branches if over the limit", async () => {
```
atrigueiro, line 57: Adding tests for branches since we didn't have one

src/jira/client/jira-client.ts
```diff
@@ -221,15 +221,17 @@ export const getJiraClient = async (
 						}
 					}),
 				update: async (data, options?: { preventTransitions: boolean }) => {
-					dedupIssueKeys(data, logger);
```
atrigueiro, line 224: We're not logging anything in this function, so just removing it.

src/jira/client/jira-client.ts
```diff
@@ -393,12 +395,12 @@ const dedupCommits = (commits: JiraCommit[] = []): JiraCommit[] =>
 /**
  * Deduplicates issueKeys field for branches and commits
  */
-const dedupIssueKeys = (repositoryObj, logger?) => {
```
atrigueiro, line 396: We're not logging anything in this function, so just removing it.

joshkay10 (approved):

mboudreau (approved):
:hyperblob:

src/jira/client/jira-client.ts
```diff
@@ -393,12 +395,12 @@ const dedupCommits = (commits: JiraCommit[] = []): JiraCommit[] =>
...
+const dedupIssueKeys = (repositoryObj) => {
+	updateRepositoryIssueKeys(repositoryObj, uniq);
 };
 
 /**
- * Truncates branches and commits to first 100 issue keys for branch or commit
+ * Truncates branches, commits and PRs to their first 100 issue keys
  */
 const truncateIssueKeys = (repositoryObj) => {
 	updateRepositoryIssueKeys(repositoryObj, truncate);
```
gxueatlassian, line 406: How does this line currently work in main? It is passing in a `truncate` function, but inside `updateRepositoryIssueKeys` calls it `logger` (which you removed in this PR) and never use it.

----

So is it true that the truncate never work?

gxueatlassian (approved):

src/jira/client/jira-client.ts
atrigueiro: The `updateRepositoryIssueKeys` should always receive a `mutatingFunc` in its second parameter. You're completely right, in main we were just receiving a logger which is not really what we wanted here. I caught that when writing my test and fixed it.

PR #1258, opened 2022-06-13
ARC-1369 review logs child

joshkay10:
**Whats in this PR**
All calls to the loggers `.child` function has been inspected for UGC/PII,
All UGC/PII has been either hashed or removed.

I will be keeping a record of data that got hashed(and didn't in some cases) on the data classification table [here](https://hello.atlassian.net/wiki/spaces/PF/pages/1649328545/Logging+UGC+PII).
This is to help me with consistency during this process, but also add reasoning to decision made(it can be very murky).

**Whats Next**
Doing the same thing for
~~`.child`~~
`.log`
`.warn`
`.info`
`.error`
`.debug`

Deep dive into Splunk and pull out all fields that could be UGC/PII and confirm they are hashed or go hash them!!


src/github/pull-request.ts
```diff
@@ -25,12 +26,11 @@ export const pullRequestWebhookHandler = async (context: CustomContext, jiraClie
...
 
 	context.log = context.log.child({
-		jiraHostName: jiraClient.baseURL,
+		jiraHostName: createHashWithSharedSecret(jiraClient.baseURL),
 		installationId: githubInstallationId,
-		orgName: owner,
+		orgName: createHashWithSharedSecret(owner),
 		pullRequestNumber,
-		pullRequestId,
-		payload: context.payload
```
joshkay10, line 33: Completely removing the payload here, seems a bit silly to hash and we can't guarantee the contents is not UGC/PII so leaning towards safer.

src/jira/extract-installation-from-jira-callback.ts
```diff
@@ -25,7 +26,7 @@ export const extractInstallationFromJiraCallback = async (req: Request, res: Res
 	const { jiraHost, clientKey } = installation;
 
 	req.addLogFields({
-		jiraHost,
+		jiraHost: createHashWithSharedSecret(jiraHost),
 		jiraClientKey: `${clientKey.substr(0, 5)}***}`
```
joshkay10, line 30: We are already doing some obfuscation here with the JiraClientKey so left it as is.

src/github/pull-request.ts
```diff
@@ -25,12 +26,11 @@ export const pullRequestWebhookHandler = async (context: CustomContext, jiraClie
 	const gitHubInstallationClient = await createInstallationClient(githubInstallationId, jiraClient.baseURL, context.log);
 
 	context.log = context.log.child({
-		jiraHostName: jiraClient.baseURL,
+		jiraHost: createHashWithSharedSecret(jiraClient.baseURL),
```
joshkay10, line 29: changed to just jiraHost because jiraHostName is a lie.
Plus it is the only time we refer to it as jiraHostName.

rachellerathbone (approved):

src/github/pull-request.ts
mboudreau: wouldn't it be simpler to just add a top level serializer in `logger.ts` to check for all keys that should be encrypted and do it there instead of duplicating this logic everywhere?

src/github/pull-request.ts
```diff
@@ -25,12 +26,11 @@ export const pullRequestWebhookHandler = async (context: CustomContext, jiraClie
...
 		installationId: githubInstallationId,
-		orgName: owner,
+		orgName: createHashWithSharedSecret(owner),
 		pullRequestNumber,
```
atrigueiro, line 32: Are `pullRequestNumber` and `pullRequestId` the same thing?

atrigueiro: Oh nvm just saw you didn't actually removed the id

src/github/pull-request.ts
gxueatlassian: I agree with @mboudreau ! But did tried couple way and could not figure out how to do this. Try to reference to [this](https://probot.github.io/docs/logging/) and [this](https://getpino.io/#/docs/redaction). 

---

And btw, just figure out that, the probot is using pino log, but in the code, we are typing it as buyan. Although most of the interface are same.

gxueatlassian (approved):

src/github/pull-request.ts
joshkay10: Yeah, makes sense, I'll spend some time digging around today.

@gxueatlassian Cheers for the Pino find!!
We are slowly ripping out Probot, so might be a case of top level serializer for our bunyan logge  and do it this way for Pino(depending on effort) for now.

joshkay10: Made a serializer, see logger.ts,
**Fun fact**, the values added in the .child function don't go through the serializers :(
so its there for all the log function e.g. `logger.info({**serialize me**}, "some message");`

There is a story after the unsafe data logging to standardize all our logs, so I will leave it until that story to replace .child with create loggers (if appropriate) or move where values are being passed into the logger.

src/config/logger.ts
```diff
@@ -67,7 +84,8 @@ const logger = wrapLogger(createLogger(
 		serializers: {
 			err: errorSerializer,
 			res: responseSerializer,
-			req: requestSerializer
+			req: requestSerializer,
+			...unsafeDataSerilaizers
```
gxueatlassian, line 88: Hi @joshkay10 , it seems in the code we use a lot of `context.log` and `req.log`. Can you plz show me how are they configured to use this instance of `const logger = wrapLogger(...)`?

mboudreau, line 88: fix your typo ;)

mboudreau:
Hey mate, if we're going to be redoing the logger, maybe we want to look into using a different/better library?  We were just using bunyan because probot was using it.  Kinda annoying that 'child' doesn't seem to take the settings from the parent...

src/config/logger.ts
```diff
@@ -45,6 +46,13 @@ const errorSerializer = (err) => (!err || !err.stack) ? err : {
 	stack: getFullErrorStack(err)
 };
 
+const hashSerializer = (data) => {
+	if (!data) {
+		return data;
+	}
+	return createHashWithSharedSecret(data);
```
mboudreau, line 53: Just for my understanding, we're encrypting all the values, not removing them, and we can still decrypt it if needed right?

We have a way of doing this easily?  Is this using cryptor?

joshkay10:
Closing due to different approach being taken. Raising new PR.

PR #1259, opened 2022-06-14
Arc-1408 update models

rachellerathbone:
Moved `gitHubAppId` from Installations to Subscriptions to align with https://github.com/atlassian/github-for-jira/blob/main/docs/data-model.md

A lot of files have been modified but most changes are trivial. We're now using `gitHubInstallationId` to query the Subscriptions table so I had to update the params in the 3 GH client factory functions to:
a. Add gitHubInstallationId (if not already included in the func definition)
b. Move `jiraHost` so that it is the last param: this is currently only used to set the FF and will later be removed so makes more sense to pass it as the final argument wherever we're calling these functions.

What's to come after this PR:

1. Updates to the GitHubServerApp table (update method to save an entry to the table, addition of new field)
2. Completion of the PR to hide the avatar if GHE (need to be able to create an entry in the GitHubServerApps table to finish writing tests)
3. Updating code so that we can render multiple GHE and cloud connections because right now this happens when we try to render both

Screen Shot 2022-06-14 at 10 22 58 am


db/migrations/20220609044825-move-github-app-id-to-subscriptions.js
```diff
@@ -0,0 +1,21 @@
+'use strict';
+
+module.exports = {
+	up: async (queryInterface, Sequelize) => {
```
rachellerathbone, line 4: Moved gitHubAppId from Installations to Subscriptions to align with our data model update.

src/github/issue-comment.ts
```diff
@@ -18,7 +18,7 @@ export const issueCommentWebhookHandler = async (
 	} = context.payload;
 	let linkifiedBody;
 
-	const gitHubInstallationClient = await createInstallationClient(githubInstallationId, jiraClient.baseURL, context.log);
+	const gitHubInstallationClient = await createInstallationClient(githubInstallationId, context.log, jiraClient.baseURL);
```
rachellerathbone, line 21: We're now using the `gitHubInstallationId` to check against entries in the Subscriptions table. The `jiraHost` param in the 3 client factory functions is only used to set the FF. Seeing this will later be removed, I moved all `jiraHost` args so that they are the last to be passed.

db/migrations/20220609044825-move-github-app-id-to-subscriptions.js
```diff
@@ -0,0 +1,21 @@
...
+		await queryInterface.removeColumn("Installations", "githubAppId");
```
gxueatlassian, line 5: I assume this column is for GHE at the moment and therefore NO production data in it, right? Otherwise we can preserve this column until prod data is migrated.

gxueatlassian (approved):
I don't see any problems related to bugs/performance. So approved this PR. 

Leave some minor comments though, mainly for coding style. Not required to be addressed.

src/util/get-github-client-config.ts
```diff
@@ -66,8 +63,8 @@ export async function createAppClient(logger: Logger, jiraHost: string): Promise
  * Factory function to create a GitHub client that authenticates as the installation of our GitHub app to get
  * information specific to an organization.
  */
-export async function createInstallationClient(githubInstallationId: number, jiraHost: string, logger: Logger): Promise<GitHubInstallationClient> {
-	const gitHubClientConfig = await getGitHubClientConfig(jiraHost);
+export async function createInstallationClient(githubInstallationId: number, logger: Logger, jiraHost: string): Promise<GitHubInstallationClient> {
```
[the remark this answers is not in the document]
gxueatlassian: Okay, just see this line. `Move jiraHost so that it is the last param: this is currently only used to set the FF and will later be removed so makes more sense to pass it as the final argument wherever we're calling these functions.`

If it is related to that comment, can we still figure out a way? Not sure, but happy to discuss more.

db/migrations/20220609044825-move-github-app-id-to-subscriptions.js
```diff
@@ -0,0 +1,21 @@
...
+
+		await queryInterface.addColumn("Subscriptions", "gitHubAppId", {
+			type: Sequelize.INTEGER,
+			allowNull: true
+		});
+	},
+
+	down: async (queryInterface, Sequelize) => {
+		await queryInterface.addColumn("Installations", "gitHubAppId", {
```
joshkay10, line 14: Should this be back to `githubAppId` instead of  `gitHubAppId`?

rachellerathbone: No. Changed it to `gitHubAppId` because we're trying to standardise the way we use the word GitHub in the app. The 'H' should be captialised and should have been that way when we first introduced this field.

src/models/subscription.ts
```diff
@@ -77,12 +78,16 @@ export class Subscription extends Model {
...
 	}
 
-	static getAllForInstallation(
-		installationId: number
-	): Promise<Subscription[]> {
+	static getAllForGitHubInstallationId(
+		gitHubInstallationId: number
+	): Promise<Subscription[] | null> | null {
+		if (!gitHubInstallationId) {
+			return null;
```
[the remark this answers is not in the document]
rachellerathbone: Hmm good point. It was throwing in error in the get github config but definitely better to add additional checks there than modify this seeing it's being used elsewhere in the codebase.

src/routes/api/installation/api-installation-get.ts
```diff
@@ -7,12 +7,13 @@ import { createAppClient } from "~/src/util/get-github-client-config";
 export const ApiInstallationGet = async (req: Request, res: Response): Promise<void> => {
 	const { installationId } = req.params;
 	const { client, jiraHost } = res.locals;
-	const gitHubAppClient = await createAppClient(req.log, jiraHost);
+	const { gitHubInstallationId } = req.body;
+	const gitHubAppClient = await createAppClient(gitHubInstallationId, req.log, jiraHost);
 
 	try {
-		const subscriptions = await Subscription.getAllForInstallation(Number(installationId));
+		const subscriptions = await Subscription.getAllForGitHubInstallationId(Number(installationId));
```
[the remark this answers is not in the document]
gxueatlassian: Ignore above, I think the best way will be somehow validates all the params first, and here still just doing the `Number(installationId)` for sake of simply code.

src/util/get-github-client-config.ts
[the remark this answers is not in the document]
rachellerathbone: These functions were created as part of the GHE track so we can use a FF to determine what to call the GH clients with the additional params or not. The FF is currently off so existing code isn't affected by this. The args being passed in only take effect once we start rolling out the FF. It is possible to expect that there may be further changes to these factory functions params as we continue to work on this project but as long as the FF is off, we don't need to be concerned about making trivial changes such as this.

db/migrations/20220609044825-move-github-app-id-to-subscriptions.js
rachellerathbone: Yep new column added specifically for ghe so no prod data.

src/util/get-github-client-config.ts
```diff
@@ -17,46 +17,43 @@ export interface GitHubClientConfig {
...
 	return await booleanFlag(BooleanFlags.GHE_SERVER, false, jiraHost) && gitHubClientConfig
 		? `${gitHubClientConfig.baseUrl}`
 		: GITHUB_CLOUD_API_BASEURL;
 }
 
-const getGitHubClientConfig = async (jiraHost: string): Promise<GitHubClientConfig> => {
-	// TODO: remove duplicate jiraHosts and find issue that causes duplicates
-	const installation = await Installation.getForHost(jiraHost);
-	const gitHubAppId = installation?.githubAppId;
+const getGitHubClientConfig = async (gitHubInstallationId: number): Promise<GitHubClientConfig> => {
```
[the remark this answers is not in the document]
rachellerathbone: No, it will never be empty. A `gitHubInstallationId` is created when an entry is made in the Subscriptions table and is not a new field. The `gitHubInstallationId` is unique to the org where the app has been installed.

src/routes/api/installation/api-installation-get.ts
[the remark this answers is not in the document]
rachellerathbone: I'd prefer not to make that change as part of this PR as that is existing code. The only reason there is a change to this file is because I had to update the subscription method name.

src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -165,7 +164,7 @@ export const GithubConfigurationGet = async (req: Request, res: Response, next:
 			return;
 		}
 
-		const gitHubAppClient = await createAppClient(log, jiraHost);
+		const gitHubAppClient = await createAppClient(req.body.gitHubInstallationId, log, jiraHost);
```
gxueatlassian, line 167: 1. The function is called `GithubConfigurationGet`, so I assume this route is a `HTTP GET` request. And in case of those request, I don’t think using `req.body` is a good way. Because `GET` request usually is NOT supposed to have a body? Possibly solution is to attach the id to url, like `/github/configiguration?ghid=xxxx`?

2. Second, will it break the existing github cloud? As I don’t see any gitHubInstallationId in the existing github-configuration page (see screenshot)

image

gxueatlassian (approved):

db/migrations/20220609044825-move-github-app-id-to-subscriptions.js
joshkay10: I agree with the case change in general, I was more meaning the specific `down` function for `Installations` since it probably has to go back to the dumb spelling it was, if it does the down-grade? It might not care ¯\_(ツ)_/¯

PR #1262, opened 2022-06-14
ARC-1393 -GHE UI: Redirecting to GHE Server Instance 

krazziekay:
**Rationale**
- This PR includes a new screen for the loading state of GHE.

**Changes**
- Added a new view `ghe-loading.hbs` with its css file.
- Added a new `jira-logo.svg`.
  - Used `aui-spinner` and used **Jira** and **GitHub** svg logos for the image

**Preview**
demo

### DIFF
--- a/static/css/ghe-loading.css
+++ b/static/css/ghe-loading.css
@@ -1,0 +1,44 @@
+.gheLoading___body {
+    background: #fff;
+}
+
+.gheLoading___container {
+    align-items: center;
+    display: flex;
+    flex-direction: column;
+    height: 100%;
+    justify-content: center;
+    margin: auto;
+    max-width: 470px;
+}
+
+.gheLoading___header {
+    font-size: 1.5rem;
+}
+
+.gheLoading___imgDivider {
+    font-size: 2.8rem;
+}
+
+.gheLoading___imgContainer {
+    margin: 4.688em 0 4.063em;
+    display: flex;
+    width: 100%;
+    justify-content: space-evenly;
+    align-items: center;
+}
+
+.gheLoading___jiraImg {
+    height: 80px;
+}
+
+.gheLoading___GitHubImg {
+    height: 60px;
+}
+
+.gheLoading___message {
+    text-align: center;
+    max-width: 364px;
+    color: #344563;
+    font-size: 0.875rem;
+}
--- a/static/assets/jira-logo.svg
+++ b/static/assets/jira-logo.svg
@@ -1,0 +1,11 @@
+
+<svg height="80" viewBox="2.59 0 214.091 224" width="80" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <linearGradient id="a" gradientTransform="matrix(1 0 0 -1 0 264)" gradientUnits="userSpaceOnUse" x1="102.4" x2="56.15" y1="218.63" y2="172.39">
+        <stop offset=".18" stop-color="#0052cc"/>
+        <stop offset="1" stop-color="#2684ff"/>
+    </linearGradient>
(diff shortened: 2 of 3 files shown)


static/css/ghe-loading.css
```diff
@@ -0,0 +1,45 @@
...
+    font-size: 1.5rem;
+}
+
+.gheLoading___imgDivider {
+    font-size: 2.8rem;
+    color: #6B778C;
+}
+
+.gheLoading___imgContainer {
+    margin: 4.688em 0 4.063em;
```
rachellerathbone, line 25: Do we need to be this precise?

static/css/ghe-loading.css
```diff
@@ -0,0 +1,45 @@
...
+    height: 80px;
+}
+
+.gheLoading___GitHubImg {
+    height: 60px;
+}
+
+.gheLoading___message {
+    text-align: center;
+    max-width: 364px;
```
rachellerathbone, line 42: Why do we need a max-width here? Normally you'd just set a max-width on the outer container so that all child elements don't extend beyond that point.

krazziekay: This content has a small width than the parent, so used `max-width` here so that the content doesn't exceed beyond this. Can use `width` too, but don't really see much difference in this case?

static/css/ghe-loading.css
krazziekay: Not necessarily. I simply converted the pixels to em and added those values.
For avoiding any confusion, rounding off these values.

static/css/ghe-loading.css
mboudreau: not sure we need any width for a paragram element. The child would never exceed the parent width unless you had set text overflow to none;

static/css/ghe-loading.css
mboudreau: we could probably just use pixels in that case then?

mboudreau:
We should probably reuse this template as a partial and use it for the current github cloud session screen

static/css/ghe-loading.css
krazziekay: The width was added to match the design to make it two lines.

static/css/ghe-loading.css
krazziekay: Following the following convention for GitHub for Jira: 
- px for borders and max-width of a section
- em for padding and margins (so it is relative to the parent container)
- rem for font size (so it's relative to the root unit and stays consistent regardless of other measurements used elsewhere)

joshkay10 (dismissed):

mboudreau (approved):

static/css/jira-configuration.css
```diff
@@ -199,6 +199,15 @@
   top: -7.5em;
 }
 
+/*
+The tooltip for the first two rows are hidden due to the `overflow: auto` of the parent div,
+so changing the position of the tooltips for the first two rows to bottom
+ */
+.jiraConfiguration__table__row:nth-of-type(1) .jiraConfiguration__syncWarningMessage,
```
rachellerathbone, line 206: What does this change have to do with the ghe loading screen? PRs shoudl stay focused on a specific change. This should be in a separate PR with a description of why this change is needed + screenshot.

krazziekay: This is related to a small UI issue that was discovered today. There were no tickets and seeing how small it was, decided to do the fix here. 
Discussed here: https://atlassian.slack.com/archives/C01PERFAUHZ/p1655252266197379
Should we do this separately in a new ticket then?

rachellerathbone: We can let it slide this time but we generally keep PRs focused on a specific change, even if the additional change is tiny.

On the css here though: we can simplify this by using the nth-child selector and setting a range.

krazziekay: Used `nth-child` with range

rachellerathbone (approved):

PR #1268, opened 2022-06-15
ARC-1429: Return 404 if Jira has been renamed.

atrigueiro:
This is closely related to https://github.com/atlassian/github-for-jira/pull/1193

Because we don't listen to callbacks from Jira when an instance is renamed we end up making requests to the old and new Jira hosts.

We obviously want to treat this better in the future and disable suspended/renamed installations, but for now, this is a better way for us to track this.

I'll also update the [Vacuuming page](https://hello.atlassian.net/wiki/spaces/PF/pages/1726777910/Vacuuming+installations+and+listening+to+lifecycle+events) to include this scenario.


krazziekay (dismissed):

src/jira/client/axios.ts
```diff
@@ -157,6 +157,16 @@ const instrumentFailedRequest = (baseURL: string, logger: Logger) => {
...
+			try {
+				await axios.get("/status", { baseURL });
+			} catch (e) {
+				if (e.response.status === 302) {
+					logger.info(`405 from Jira: Jira instance '${baseURL}' has been renamed. Returning 404 to our application.`);
+					error.response.status = 404;
+				}
+			}
+		}
 		return Promise.reject(error);
```
Harminder84, line 170: Similar handling when we're getting HTTP error code `503`, can both be merged?

mboudreau: yeah, I agree with this, we should keep that logic in the same place

atrigueiro: Fair I moved the first check to be in the same place, but I kept the check from the `/status` endpoint separated so we have better visibility between suspended vs renamed.

mboudreau (dismissed):

src/jira/client/axios.ts
```diff
@@ -147,14 +147,18 @@ const instrumentRequest = (response) => {
...
-		if (error.response?.status === 503) {
+		if (error.response?.status === 503 || error.response?.status === 405) {
 			try {
 				await axios.get("/status", { baseURL });
 			} catch (e) {
 				if (e.response.status === 503) {
 					logger.info(`503 from Jira: Jira instance '${baseURL}' has been deactivated, is suspended or does not exist. Returning 404 to our application.`);
 					error.response.status = 404;
 				}
+				if (e.response.status === 302) {
```
Harminder84, line 158: nit: use `else/if` instead of `if`

atrigueiro: If I use `else if` I will have to define another `else` at the end, but there's nothing else I want to do there. Do I just do else { return; } ?

Harminder84: `else if` doesn't force to have `else` at the end, so skip `else` if not required

atrigueiro: Ooo nice I didn't know that, I though `else` was required. Thanks Harminder, just updated :)

src/jira/client/axios.ts
```diff
@@ -147,13 +147,16 @@ const instrumentRequest = (response) => {
 const instrumentFailedRequest = (baseURL: string, logger: Logger) => {
 	return async (error: AxiosError) => {
 		instrumentRequest(error?.response);
-		if (error.response?.status === 503) {
+		if (error.response?.status === 503 || error.response?.status === 405) {
```
bgvozdev, line 150: Ideally this should be feature-flagged...

src/jira/client/axios.ts
```diff
@@ -147,13 +147,16 @@ const instrumentRequest = (response) => {
...
 			try {
 				await axios.get("/status", { baseURL });
 			} catch (e) {
 				if (e.response.status === 503) {
 					logger.info(`503 from Jira: Jira instance '${baseURL}' has been deactivated, is suspended or does not exist. Returning 404 to our application.`);
 					error.response.status = 404;
+				} else if (e.response.status === 302) {
```
bgvozdev, line 157: ... and this one too...

bgvozdev (approved):

src/jira/client/axios.ts
bgvozdev: but probably not a big deal cause this is error handling

atrigueiro: Yeah I think it's ok because I'm just adding more logging to an error that is already being thrown. If these conditions aren't met, we will reject it anyways on line 163.

Harminder84 (approved):

src/jira/client/axios.ts
mboudreau: Why?

mboudreau (approved):

src/jira/client/axios.ts
bgvozdev: Because we are changing the behaviour?

atrigueiro: I'm considering it fine because of the comment below.

PR #1269, opened 2022-06-15
ARC-1348: use new client secret

bgvozdev:
Tested in ddev.


Harminder84 (approved):

atrigueiro:
Is this change something we test in staging as well?

src/config/env.ts
```diff
@@ -78,6 +78,7 @@ export interface EnvVars {
 	WEBHOOK_SECRET: string;
 	GITHUB_CLIENT_ID: string;
 	GITHUB_CLIENT_SECRET: string;
+	GITHUB_CLIENT_SECRET_VAULT: string;
```
mboudreau, line 81: Should we really use a new variable here or just keep the old one?  Would make it simpler in code since we wouldn't have to update it everywhere or have to have conditionals

bgvozdev: The new variable is just for the feature flag. I'll rename it to `GITHUB_CLIENT_SECRET` once the FF is removed.

bgvozdev:
> Is this change something we test in staging as well?

yep :)

joshkay10 (approved):

PR #1270, opened 2022-06-15
ARC-925 push webhook handler for new webhook receiver

Harminder84:
I’ve updated all the handlers to use `WebhookContext` instead of `ProbotContext`. Created a wrapper function `Convert-to-webhook-context` that transforms the `ProbotContext` to `WebhookContext` so that existing flows work seamlessly with Probot webhook receiver. Also hooked in `push` handler with the new webhook receiver.


bgvozdev (approved):

gxueatlassian (changes requested):
Sorry @Harminder84 , still looking at it. But think this PR is important, so temporary blocking it for now until I fully review it.

src/github/push.ts
```diff
@@ -3,8 +3,10 @@ import { Context } from "probot/lib/context";
 import { getCurrentTime } from "utils/webhook-utils";
 import { hasJiraIssueKey } from "utils/jira-utils";
 import { GitHubPushData } from "../interfaces/github";
+import { WebhookContext } from "../routes/github/webhook/webhook-context";
 
-export const pushWebhookHandler = async (context: Context, jiraClient): Promise<void> => {
+export const pushWebhookHandler = async (probotContext: Context, jiraClient, _util, _githubInstallationId: number, webhookContext?: WebhookContext): Promise<void> => {
+	const context = webhookContext ? webhookContext : probotContext;
```
mboudreau, line 9: Shouldn't we be removing the dependency on probot for this new handler?

Harminder84: Same handler is being used for both old (Probot) and new webhook receiver. When push handler is called with new webhook receiver, `probotContext ` empty whereas `WebhookContext` populated.

rachellerathbone, line 9: We should probably set context based on the `new-webhook-receiver` FF you created.

src/middleware/github-webhook-middleware.test.ts
```diff
@@ -53,6 +53,10 @@ describe("Probot event middleware", () => {
 			const spy = jest.fn();
 			await expect(GithubWebhookMiddleware(spy)(context)).toResolve();
 			expect(spy).toHaveBeenCalledTimes(3);
+
+			const anotherSpy = jest.fn();
+			await expect(GithubWebhookMiddleware(anotherSpy)(undefined, context)).toResolve();
+			expect(anotherSpy).toHaveBeenCalledTimes(3);
 		});
 
 		it("when one call fails", async () => {
```
rachellerathbone, line 62: Can we change this to 'should <what we expect to happen> when one call fails'? Just makes it easier to understand what we're expecting from the test.

src/middleware/github-webhook-middleware.test.ts
```diff
@@ -63,6 +67,10 @@ describe("Probot event middleware", () => {
 			const spy = jest.fn().mockRejectedValueOnce(new Error("Failed!"));
 			await expect(GithubWebhookMiddleware(spy)(context)).toResolve();
 			expect(spy).toHaveBeenCalledTimes(3);
+
+			const anotherSpy = jest.fn().mockRejectedValueOnce(new Error("Failed!"));
+			await expect(GithubWebhookMiddleware(anotherSpy)(undefined, context)).toResolve();
+			expect(anotherSpy).toHaveBeenCalledTimes(3);
 		});
 
 		it("when all calls fail", async () => {
```
rachellerathbone, line 76: Same here.

src/middleware/github-webhook-middleware.test.ts
```diff
@@ -104,5 +126,10 @@ describe("Probot event middleware", () => {
 		context.log.info("test");
 		expect(loggedStuff).toContain("foo");
 		expect(loggedStuff).toContain("123");
+
+		await GithubWebhookMiddleware(jest.fn())(undefined, context);
+		context.log.info("test");
```
rachellerathbone, line 131: Is this meant to be here?

src/middleware/github-webhook-middleware.ts
```diff
@@ -11,14 +11,16 @@ import { booleanFlag, BooleanFlags } from "config/feature-flags";
...
 
 const warnOnErrorCodes = ["401", "403", "404"];
 
 // Returns an async function that reports errors errors to Sentry.
 // This works similar to Sentry.withScope but works in an async context.
 // A new Sentry hub is assigned to context.sentry and can be used later to add context to the error message.
-const withSentry = function(callback) {
-	return async (context) => {
+const withSentry = function (callback) {
+	return async (customContext?: CustomContext, webhookContext?: WebhookContext) => {
```
rachellerathbone, line 22: Why is customContext optional? In the original code context is required. My understanding is:

- we should have a FF. If feature flag is true, use webhookContext
- if feature flag is off, default to context currently used. This should not be an optional param.

src/middleware/github-webhook-middleware.ts
```diff
@@ -11,14 +11,16 @@ import { booleanFlag, BooleanFlags } from "config/feature-flags";
...
+		const context = webhookContext ? webhookContext : customContext!;
```
rachellerathbone, line 23: This should be set via your FF.

src/middleware/github-webhook-middleware.ts
```diff
@@ -84,7 +87,10 @@ export const GithubWebhookMiddleware = (
 			event: webhookEvent,
 			action: context.payload?.action,
 			id: context.id,
-			repo: context.payload?.repository ? context.repo() : undefined,
+			repo: context.payload?.repository ? {
+				owner: context.payload.repository.owner.login,
```
rachellerathbone, line 91: This was context.repo() before. Won't this break if we're using the current context?

src/github/push.ts
Harminder84: I'm thinking of using FF `new-webhook-receiver` at the start of new webhook receiver flow.  It doesn't make sense to use FF flag here because `webhookContext` will only be populated with new flow. Please create another ticket for FF.

mboudreau: Sure, but the whole point of the move is to remove probot.  We should try to standardize it to use the same `WebhookContext` for both instead of have the probot context pollute the codebase and making it harder to code because of conditionals.

mboudreau: If this is not possible, we should probably just create a new push handler since we have a new entrypoint for webhooks anyways.  Once we move to the new webhook url, we can delete everything probot.

src/middleware/github-webhook-middleware.test.ts
Harminder84: Yes, testing child log data preserves or not.

src/middleware/github-webhook-middleware.ts
Harminder84: I need to call this method with arguments `(undefined, webhookContext)` from new webhook receiver. If it make it mandatory, I won't be able to call with `undefined`. Regarding FF, I explained above.

src/middleware/github-webhook-middleware.ts
Harminder84: Same here for FF

src/github/push.ts
Harminder84: @mboudreau, I'm fine with creating a new push handler. We just need to make sure that if we fix any bug in exisiting Probot webhook handler, that needs to be done in new webhook codebase until we move away from Probot.

src/middleware/github-webhook-middleware.ts
mboudreau: context.repo() was just a utility function to get the repo information out from the payload in a certain format :)

gxueatlassian (approved):
👍 looks good.

mboudreau (approved):
:clap: much better

PR #1271, opened 2022-06-15
Arc-1401 update GitHubServerApps table

rachellerathbone:
Made some changes to the GitHubServerApps table:
- changed the casing of github to gitHub in: gitHubBaseUrl, gitHubClientId and gitHubClientSecret. We've talked as a team before about keeping this consistent with the casing used by GitHub. **UPDATE:** @mboudreau had a good point: even though we should technically be safe because we aren't using these values anywhere, better to err on the side of caution. Updated the migration to add new columns with an upper case H and then changed the nonnull value to false to the existing ones. Once this is safely in prod I will open a new pr to delete the original columns with lower casing.
- added new column `installationId`: this is the installation id from the Installations table. This value will allow us to verify against the jira host.
- added new column `gitHubAppName`: when a user enters their ghe sever url, we're going to query the db to see if we already have existing GH apps for that server instance. If we find some, we need to surface the name of the apps in the UI.


src/models/git-hub-server-app.ts
```diff
@@ -5,21 +5,25 @@ const encrypted = EncryptedField(Sequelize, process.env.STORAGE_SECRET);
 
 interface GitHubServerAppPayload {
 	uuid: string;
```
mboudreau, line 7: could you please update the file name? it's bothering me, everywhere else `github` is one word

db/migrations/20220615035112-update-github-server-apps.js
```diff
@@ -0,0 +1,28 @@
...
+			allowNull: false
+		});
+
+		await queryInterface.addColumn(tableName, "installationId", {
+			type: Sequelize.INTEGER,
+			allowNull: false
+		});
+
+		await queryInterface.renameColumn(tableName, 'githubBaseUrl', 'gitHubBaseUrl');
+		await queryInterface.renameColumn(tableName, 'githubClientId', 'gitHubClientId');
```
mboudreau, line 18: Adding columns is fine, but renaming columns might affect current production.  I don't think anyone is using the model though, so you might be safe?  Try it in Staging first.  If it doesn't work, the better way is to create the new column, copy the data over from the old to the new one, then after deploying that change, do another DB migration to remove the old column.

mboudreau (approved):

PR #1273, opened 2022-06-15
PCS-87130 Adding ability for reposyncstate to remove duplicates in DB

mboudreau:
As I was looking at a large customer's sync state after a temporary service interruption, I notice that there were more repositories in the DB than they had available.  I soon noticed that there was duplicate entries for `repoId` under this customer's subscription, which shouldn't be possible.

I will be updating the DB schema soon to not allow this to ever happen at the DB level, but for now, we need to update the code  to remove any duplicates created by mistake.


src/models/reposyncstate.ts
```diff
@@ -139,13 +139,28 @@ export class RepoSyncState extends Model {
 			}
 		});
```
Harminder84, line 141: This is a one time process to remove duplicate records from the table,  can't we add it as a DB migration script?

rachellerathbone (approved):

rachellerathbone:
Down with the duplicates! 🎉

src/models/reposyncstate.ts
joshkay10: I think until we have the fix in it will be an ongoing issue, so migration script would just clear the current duplicates but then we will continue to get more in between times??

src/models/reposyncstate.ts
```diff
@@ -139,13 +139,28 @@ export class RepoSyncState extends Model {
...
+					// Remove duplicates of repoIds
+					values.slice(1).forEach(state => acc.push(state.id));
+				}
+				return acc;
+			}, []);
+
 		return RepoSyncState.sequelize?.transaction(async (transaction) => {
-			// Delete all repos that's not in state anymore
+			// Delete all repos that's not in state anymore or are duplicates
 			await RepoSyncState.destroy({
```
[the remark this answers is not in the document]
mboudreau: relic of the old JSON repo sync state column in subscription.  By "state" I mean the repoIds that aren't in the JSON passed over, which means they've been removed and we need to update the table as such

src/models/reposyncstate.ts
mboudreau: We can, but this is also the code that added the duplicates in the first place because of a partial backfill.  I figured removing it at the source first, then doing a DB script (that'll be another PR) to remove the duplicates in the DB is the best way to go.  Then I need to update the DB schema so that this can never happen ever again :)

gxueatlassian (approved):

PR #1276, opened 2022-06-16
ARC-1240:  Limit createPullRequestUrl to 2000

atrigueiro:
Our `generateCreatePullRequestUrl` generates a create pull request link that's shown on the screen below: (you can see the formatted URL on the very bottom of the page)

Screen Shot 2022-06-16 at 11 21 38 am

However, the `createPullRequestUrl` has a max length of 2000 that we were not checking for, and therefore, we've been receiving a lot of 400 Bad Request errors for pull_requests events.

Screen Shot 2022-06-16 at 3 43 21 pm

I have added a test to make sure we're not sending the createPullRequestUrl parameter if length > 2000. 

I've also changed the logic a bit to:
1.  Only add the branch as the title if there's an issue key there - which is the most important one; if not
2. only add the first issue key from the issueKeys array - no need to add 100+ issue keys on the title when creating a PR, the first one should be the most important one.


src/interfaces/jira.ts
```diff
@@ -50,7 +50,7 @@ export interface JiraBuildData {
 }
 
 export interface JiraBranch {
-	createPullRequestUrl: string;
+	createPullRequestUrl?: string;
```
atrigueiro, line 53: This field is not mandatory https://developer.atlassian.com/cloud/jira/software/rest/api-group-development-information/#api-devinfo-0-10-bulk-post

src/transforms/transform-pull-request.test.ts
```diff
@@ -134,4 +134,88 @@ describe("pull_request transform", () => {
...
+						"https://github.com/integrations/test/tree/use-the-force",
+					status: "OPEN",
+					timestamp: updated_at,
+					title: title,
+					url: "https://github.com/integrations/test/pull/51",
+					updateSequenceId: 12345678
+				}
+			],
+			branches: [
+				{
```
atrigueiro, line 193: So for other scenarios where generateCreatePullRequestUrl < 2000 (example of test above), we would have a `createPullRequestUrl` parameter here. Since in this test url > 2000, that field is not going in the payload anymore.

src/transforms/util/pull-request-link-generator.ts
```diff
@@ -1,11 +1,17 @@
...
+ * Max length: 2000
  */
 export const generateCreatePullRequestUrl = (baseUrl: string, name: string, issueKeys: string[] = []) => {
 	const keys = issueKeys?.length ? `${issueKeys.join(" ")} - ` : "";
 	const title = encodeURIComponent(keys + name);
 	const branchName = encodeURIComponent(name);
 	const url = `${baseUrl}/compare/${branchName}?title=${title}&quick_pull=1`;
+
+	if (url.length > 2000) {
+		return undefined;
```
atrigueiro, line 13: Returning undefined here will remove the `createPullRequestUrl` from the pullrequest payload

~
```diff
@@ -0,0 +1,24 @@
+Revert "Revert "ARC-810: Emit analytics for config page (#1093)" (#1112)"
```
rachellerathbone, line 1: What's this? :blobconfused:

atrigueiro: lol no idea

src/transforms/util/pull-request-link-generator.ts
joshkay10: Instead of returning undefined could we truncate the title?
(this is assuming the title that gets super duper long)

~
atrigueiro: File deleted

src/transforms/util/pull-request-link-generator.ts
atrigueiro: If we truncate it won't be a valid URL though, so I guess it's better to not send any URL instead of sending a broken one

atrigueiro:
[] Test in staging

src/transforms/util/pull-request-link-generator.ts
mboudreau: Yeah, @gxueatlassian has a good point, if the currently url with the title is too long, we should be able to use the url without the title querystring and should be way under 2000 character as a fallback, but it means the user need to input the title manually, which I think is still a better experience than not having that create pr link :)

gxueatlassian (approved):

PR #1277, opened 2022-06-16
Arc-1408 model updates

rachellerathbone:
The return of `that` PR! :laughcry:

So what's changed since the last one? 

- There are no migrations in this one as we merged those in a separate PR while fixing the above mentioned issue.
- new function in `get-github-client-config` called `getGitHubClientConfigFromAppId`. The user and app clients now take in the `gitHubAppId` from res.locals and call this new function to set the baseUrl and hostname.
- New middleware `githubServerAppMiddleware`: handles checking if there is an id provided in the url and, if there is, validating that a) a GH app with that id exists in GitHubServerApps and b) the installation id matches the id in the Installations table.

**NOTE:** same as the last time, there are lots of file changes in this PR but most are trivial as we have had to update each every client creation.

What's not in the PR:
- passing the GitHub app id from jira/configuration and adding it to the path. This work will be done in ARC 1438 when we update the layout of the GitHub config page


src/github/issue-comment.ts
```diff
@@ -18,7 +18,7 @@ export const issueCommentWebhookHandler = async (
 	} = context.payload;
 	let linkifiedBody;
 
-	const gitHubInstallationClient = await createInstallationClient(githubInstallationId, jiraClient.baseURL, context.log);
+	const gitHubInstallationClient = await createInstallationClient(githubInstallationId, context.log, jiraClient.baseURL);
```
mboudreau, line 21: Why the signature change?

src/middleware/github-server-app-middleware.ts
```diff
@@ -0,0 +1,31 @@
...
+		const gitHubServerApp = await GitHubServerApp.getForGitHubServerAppId(Number(id));
+
+		if (!gitHubServerApp) {
+			req.log.error({ id, jiraHost }, "No GitHub app found for provided id.");
+			throw new Error("No GitHub app found for provided id.");
+		}
+
+		const installation = await Installation.getForHost(jiraHost);
+
+		if (installation?.id !== gitHubServerApp?.installationId) {
```
mboudreau, line 20: I'd probably get the installation from the `gitHubServerApp.installationId` then compare the jiraHost instead :)

mboudreau: You also don't need the `gitHubServerApp?` since you did the conditional check above

src/models/subscription.ts
```diff
@@ -239,12 +253,17 @@ Subscription.init({
...
+	gitHubAppId: {
+		type: DataTypes.INTEGER,
+		allowNull: true
+	}
 }, { sequelize });
 
 export interface SubscriptionPayload {
 	installationId: number;
 	host: string;
+	gitHubAppId?: number | null;
```
mboudreau, line 266: don't need to add `| null` if it's already optional

src/models/subscription.ts
```diff
@@ -68,6 +68,7 @@ export class Subscription extends Model {
 	numberOfSyncedRepos?: number;
 	repositoryCursor?: string;
 	repositoryStatus?: TaskStatus;
+	gitHubAppId: number;
```
mboudreau, line 71: This should be optional

src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -51,10 +51,11 @@ const mergeByLogin = (installationsWithAdmin: InstallationWithAdmin[], connected
 const installationConnectedStatus = async (
 	jiraHost: string,
 	installationsWithAdmin: InstallationWithAdmin[],
-	reqLog: Logger
+	reqLog: Logger,
+	gitHubAppId: number
```
mboudreau, line 55: this should be optional

src/middleware/github-server-app-middleware.ts
```diff
@@ -0,0 +1,31 @@
+import { NextFunction, Request, Response } from "express";
+import { GitHubServerApp } from "models/github-server-app";
+import { Installation } from "models/installation";
+
+export const githubServerAppMiddleware = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
```
mboudreau, line 5: The standard is to use PascalCase for middlewares

src/github/issue-comment.ts
rachellerathbone: This was something I mentioned in the previous PR: I was originally using jiraHost to check for a github app but that's changed now since we've updated the models. We're literally just passing the jiraHost in so we can check if the FF is on or not. Once the FF is removed, all the jiraHost args will be deleted. Figured it made more sense to move the temporary param in the function definition to the end.

src/middleware/github-server-app-middleware.ts
```diff
@@ -0,0 +1,31 @@
...
+			req.log.error({ id, jiraHost }, "Installation ids do not match.");
+			throw new Error("Installation ids do not match.");
+		}
+
+		req.log.info("Found GitHub server app for installation");
+		res.locals.gitHubAppId = id;
+		next();
+	} else {
+		next();
+	}
```
mboudreau, line 30: remove the else here and just have `next()` after the if conditional:

```
	  req.log.info("Found GitHub server app for installation");
		res.locals.gitHubAppId = id;
	}
	next();
```

mboudreau (approved):

PR #1278, opened 2022-06-17
ARC 1434 - Create a view for selecting server

krazziekay:
**Rationale**
- A new view for displaying the list of already existing GHE servers was added.

**Changes**
- Created new partials to display the header and the list of servers/apps.
- Created a new view which uses the newly created partials to display the list of GHE servers

**Assumptions**
- The action for the `Connect` buttons in the list are all assumed to be similar to the connect button in `github-configuration`. So, the necessary parameters are passed to replicate the same action.

**How to Test**
- We don't have any routes or ways to get to the screen yet. So, a dummy route was created for showing this screen.
  - To create a dummy test, add this in `jira-route.ts`
  `````
   JiraRouter.get("/selectServer",  async (_: Request, res: Response): Promise<void> => {
	res.render("jira-select-server.hbs", {
		id: 1,
		host: 'My Jira Instance',
		csrfToken: 2,
		nonce: res.locals.nonce,
		failedConnections: [],
		servers: [{ url: "http://github.company1.com" }, { url: "http://github.company2.com" }, { url: "http://github.company3.com" }]
	});
});
``` 
   _This has been removed_
- Go to `https://INSTANCE_TUNNELED_URL/jira/selectServer)`.

**Screenshot**:
![Screen Shot 2022-06-17 at 1 10 29 pm](https://user-images.githubusercontent.com/13076255/174217309-279828c8-45fb-4a1c-8e0e-9b5e23f6baf9.png)

### CODE REVIEW

File: static/js/jira-select-server.js [hunk 1]
```diff
@@ -0,0 +1,17 @@
...
+  const csrfToken = document.getElementById('_csrf').value;
+  const clientKey = document.getElementById('clientKey').value;
+
+  $.post('/github/configuration', {
+    installationId,
+    _csrf: csrfToken,
+    clientKey
+  }, function (data) {
+    if (data.err) {
+      return console.log(data.err)
```
gxueatlassian, line 13: Probably need input from jade to see what's the minimum acceptable prompt for user. UX related.

gxueatlassian (dismissed):

static/js/jira-select-server.js
```diff
@@ -0,0 +1,17 @@
...
+    }
+    window.close()
```
joshkay10, line 15: /nit ;'s

joshkay10 (approved):

static/js/jira-select-server.js
krazziekay: This part will surely be revisited later, so leaving it as it is with a TODO

static/js/jira-select-server.js
krazziekay: `;` added

views/partials/select-server.hbs
```diff
@@ -0,0 +1,36 @@
+<div class="selectServer__title">
```
rachellerathbone, line 1: This should probably be broken out into 2 separate components: select-server-header and select-server-table. We should componentize by smaller sections, not the entire page. 

174217309-279828c8-45fb-4a1c-8e0e-9b5e23f6baf9

static/js/jira-select-server.js
```diff
@@ -0,0 +1,18 @@
+$('.install-link').click(function (event) {
+  event.preventDefault();
+  const installationId = $(event.target).data('installation-id');
```
rachellerathbone, line 3: I would just delete all of this for now. Not much sense having code in here that's going to change. Leave the event.preventDefault() and just delete the rest. When we start wiring up routing we can update this to query GitHubServerApps for the provided installationId and then either navigate to the screen with listed GH apps or to the auto/manual screen.

rachellerathbone (changes requested):

views/partials/select-server.hbs
```diff
@@ -0,0 +1,36 @@
...
+    </a>
+  </div>
+
+  <div class="selectServer__list">
+    {{#each servers as |server|}}
+      <div class="selectServer__server">
+        <!--TODO: Update the URL by actual data-->
+        <div class="selectServer__serverURL">{{server.url}}</div>
+        <button
+          class="aui-button aui-button-label install-link"
```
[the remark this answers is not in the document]
krazziekay: Name updated to `select-server`

views/partials/select-server.hbs
krazziekay: Updated the partial `select-server` into 2 partials, `select-server-header` and `select-server-table`

static/js/jira-select-server.js
krazziekay: The logic has been removed and a TODO has been added

views/jira-select-server.hbs
```diff
@@ -0,0 +1,74 @@
...
+         {{> navigation previousPageName="GitHub Enterprise Server app creation" }}
+      </header>
+
+      <section class="jiraSelectServer__content">
+        {{> select-server-header
+            headerImgURL="/public/assets/jira-enterprise-server-connection.png"
+            header="Select a GitHub Server"
+            subHeader="We found the following GitHub servers"
+        }}
+        <!-- servers refer to the array of cloud/GHE apps/servers-->
```
rachellerathbone, line 47: Not cloud. You won't see this screen unless you go down the server path.

views/partials/select-server-header.hbs
```diff
@@ -0,0 +1,7 @@
+<div class="selectServer__title">
```
mboudreau, line 1: are these going to be re-used in the future?

mboudreau:
@krazziekay Can the PNG be switched for an SVG?  It's a better format to use as it's clearer with high density screens and can be resized very easily.

Also, can you just add a route for this screen even if it's just temporary?  Devs shouldn't need to completely replace another route just to see your view.

views/partials/select-server-header.hbs
krazziekay: Yes, there is another screen similar to this one, i.e. the list of GH apps. It can be re-used there for [ticket](https://softwareteams.atlassian.net/browse/ARC-1437): 
There could be more in the future too.

krazziekay:
> @krazziekay Can the PNG be switched for an SVG? It's a better format to use as it's clearer with high density screens and can be resized very easily.
> 
> Also, can you just add a route for this screen even if it's just temporary? Devs shouldn't need to completely replace another route just to see your view.

The PNG image has been switched by a SVG and a dummy route has been created(information added in the PR description).

rachellerathbone (approved):

rachellerathbone (approved):

PR #1279, opened 2022-06-17
PCS-87130 remove duplicates in reposyncstate table

mboudreau:
This script is to remove duplicates in the RepoSyncState table where `subscriptionId` and `repoId` are the same.  It should be impossible to have more than one `repoId` of the same number with the same `subscriptionId`.  I consider this being low risk since this is temporary data only and if data loss does occur, we can simply re-backfill everything.

I don't want to do that for one customer as it's taken days just to backfill to this very moment but they're currently stuck because of the duplicates.  I've tested this thoroughly on my local and from running the same script (minus the deletion bit) on read-replica, I know for a fact that `21684` duplicates will be deleted.  Only 2 customers will be affected with `subscriptionId` of 110062 and 126603.  One of which is the one I'm trying to unblock currently, which will have 21100 duplicates removed.  After the duplicates are gone, I can then continue the backfill by kicking off a partial for both these customers to make sure it completes.

### DIFF
--- a/db/migrations/20220617130400-remove-duplicates-reposyncstate.js
+++ b/db/migrations/20220617130400-remove-duplicates-reposyncstate.js
@@ -1,0 +1,25 @@
+"use strict";
+
+module.exports = {
+	up: async (queryInterface) => {
+		// Deletes duplicates in RepoSyncState based on the partition between
+		// subscriptionId and repoId (no repoId can be the same under the same subscriptionId)
+		// The inner select gets all duplicates and returns the id (our primary key) in
+		// ascending order (first one created is first) and the current row number in a temp table.
+		// The outer select then only selects the id from the temp table for those of row number
+		// above 1 (leaving the first row out) essentially only selecting the extra duplicates.
+		// Then we just delete everything with those ids in RepoSyncState.
+		await queryInterface.sequelize.query(`
+			DELETE FROM "RepoSyncStates" WHERE id IN (
+				SELECT id FROM (
+					SELECT id, ROW_NUMBER() OVER(
+					    PARTITION BY "subscriptionId", "repoId" ORDER BY id
+					) AS row_num FROM "RepoSyncStates"
+			) t WHERE t.row_num > 1);
+		`);
+	},
+
+	down: async (queryInterface) => {
+		// There's no coming back from this....
+	}
+};


db/migrations/20220617130400-remove-duplicates-reposyncstate.js
```diff
@@ -0,0 +1,25 @@
+"use strict";
+
+module.exports = {
```
bgvozdev, line 3: I think it is better to do such things via a temp admin endpoint...

bgvozdev (approved):

db/migrations/20220617130400-remove-duplicates-reposyncstate.js
mboudreau: I disagree, there's a lot of security considerations of doing SQL changes at the admin level, plus it doesn't really change the end result

bgvozdev: This is not a migration, but an administrative task. Your change will change the version of the schema, which might cause conflicts if someone else is doing database work, too. (Not saying it is shouldn't be changed.)

Not sure I understand what kind of security problems you think of: an admin endpoint is slauth-protected and can be invoked only by our team.

Given that our migration is currently part of CI, this makes it even worse.

To conclude, I'm disagree with your disagreement, but not blocking you ;)

bgvozdev: Another point for having this as an admin endpoint: it can be unit-tested, while this migration can't and I only can hope you tested it thoroughly before you will merge/deploy it :)

db/migrations/20220617130400-remove-duplicates-reposyncstate.js
```diff
@@ -0,0 +1,25 @@
...
+				SELECT id FROM (
+					SELECT id, ROW_NUMBER() OVER(
+					    PARTITION BY "subscriptionId", "repoId" ORDER BY id
+					) AS row_num FROM "RepoSyncStates"
+			) t WHERE t.row_num > 1);
+		`);
+	},
+
+	down: async (queryInterface) => {
+		// There's no coming back from this....
```
joshkay10, line 23: 😂

joshkay10 (approved):

gxueatlassian (changes requested):
I marking this as blocking until further agreement from broader team.

---------

Ideally, we should comes to an agreement on what task will required a kickoff / rolling plan documented (and potential backup person during rolling out).

^^^ until we have that agreement, I suggest this task need such plans above. Let's bring this up next week.

PR #1280, opened 2022-06-17
ARC-1435 - Update the validation error for GitHub Cloud URL when creating GHE server

krazziekay:
**Rationale**
- When users type in a GitHub cloud URL when trying to connect to a GitHub Enterprise server, it is now treated as a URL validation error, displaying an error message with a link that redirects to GitHub Cloud site.

**Changes**
- The separate method `checkGHEServerUrl` was removed and the logic was added to `checkValidGHEUrl`.
- A new method was created which simply displays the error messages, based on the error object passed to it.
- Minor code cleanup 
  - A new method `hideErrorMessage` was created.

**Video**
https://www.loom.com/share/0605e2c39ac5443a8b76089cbb5bf267

**Whats not done**
- No validation in terms of duplication nor security has been implemented. It will be done in backend.

### DIFF
--- a/static/js/jira-connect-ghe.js
+++ b/static/js/jira-connect-ghe.js
@@ -1,5 +1,16 @@
 const ALLOWED_PROTOCOLS = ["http:", "https:"];
 const GITHUB_CLOUD = ["github.com", "www.github.com"];
+const defaultError = {
+	message: 'The entered URL is not valid.',
+	linkMessage: 'Learn more',
+	// TODO: add URL for this
+	linkUrl: '#'
+}
+const cloudURLError = {
+	message: 'The entered URL is a GitHub Cloud site.',
+	linkMessage: 'Connect a GitHub Cloud site',
+	linkUrl: '/session/github/configuration'
+}
 
 /**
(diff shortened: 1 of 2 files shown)


gxueatlassian (approved):
I okay to approve it. 
But please be aware that:
1. If the `validation` is only for UX purpose, then it is fine.
2. If the `validation` is also for security/correctness purpose, then we need to do similar stuff to the backend route as well.

krazziekay:
> krazziekay

For this UI task, only frontend validation has been implemented. For security, duplicate or any other proper check, a backend validation will be needed, which I think will be done in a separate ticket in the future.

krazziekay:
> I okay to approve it. But please be aware that:
> 
> 1. If the `validation` is only for UX purpose, then it is fine.
> 2. If the `validation` is also for security/correctness purpose, then we need to do similar stuff to the backend route as well.



For this UI task, only frontend validation has been implemented. For security, duplicate or any other proper check, a backend validation will be needed, which I think will be done in a separate ticket in the future.

PR #1281, opened 2022-06-20
PCS-87130 Removing repo json from  reposyncstate

mboudreau:
**What's in this PR?**
The removal of all RepoJson objects in the codebase

**Why**
RepoJson was the old way of storing RepoSyncState in a `repoSyncState` JSON column under the subscription.  This was incredibly slow, annoying to update and caused issues for larger customers.  We had to do the change over from the column to it's own table without downtime, so we slowly moved the data over to the `RepoSyncState` table while still using the same data structure as the old `repoSyncState` column, until we didn't need the column anymore and 100% relied on the new table.  However, we never removed the old data structure which made things much more complicated than needed to do and after an issue was discovered from a larger customer while updating `RepoSyncState`, we now need to remove it completely to finally end the migration process and rely 100% on proper database architecture.

**Added feature flags**
None.  Impossible to do without duplicating all code using RepoSyncState

**Affected issues**  
PCS-87130

**How has this been tested?**  
Unit tests, pollinator and manual

**Whats Next?**
Deploy then backfill customer which this has been causing issues for.


src/config/statsd.ts
```diff
@@ -21,12 +21,12 @@ export const statsd = new StatsD({
 	port: 8125,
 	globalTags,
 	errorHandler: (err) => {
-		if (!isNodeDev()) {
+		if (isNodeProd()) {
```
joshkay10, line 24: Good find!

db/config.json
```diff
@@ -3,7 +3,8 @@
     "dialect": "postgres",
     "use_env_variable": "DATABASE_URL",
 		"pool": {
-			"max": 1,
+			"acquire": 60000,
+			"max": 15,
```
joshkay10, line 7: For my benefit lolol, why 60000 and 15?

joshkay10 (approved):
Juicy change! Find this new way simpler to understand too so thats nice.
A second review might not be the worst thing if you can get one lololololol

src/backfill/prioritizer.ts
```diff
@@ -46,10 +47,18 @@ export class Prioritizer implements StepPrioritizer<JobId, JobState> {
...
+				repo.pullStatus = "failed";
+			} else {
+				const cursorParts = repo.pullCursor.match(this.cursorRegex);
+				if (!cursorParts || cursorParts.length != 3) {
+					repo.pullStatus = "failed";
+				} else {
+					const cursorId = cursorParts[1];
+					const index = cursorParts[2];
+					repo.pullCursor = `${cursorId} ${+index + this.pullrequestSkipCount}`;
+				}
```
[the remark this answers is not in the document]
mboudreau: I agree, it's taken from below and is a horrible standard, but as far as I can tell, this file isn't being used but I had to change it because the old pullCursor was for numbers and not strings.

I think the idea is to remove this file eventually or just fix our  backfilling mechanism

db/config.json
mboudreau: 15 is what we use in production, but also noticed that for some operations on my local, stuff was being bottlenecked and taking forever to backfill stuff one a single connection.  I added the acquire as I sometimes get a weird timeout bug when starting out, not sure why, but this seems to fix it after reading some stuff online.  I think we need to update to latest sequelize to fix some of these issues.

joshkay10 (approved):

src/backfill/looper/looper.ts
```diff
@@ -1,143 +0,0 @@
-import Logger from "bunyan";
```
thombergs, line 1: Sad to see this code go as it took care of quite some issues with ratelimiting, retrying, etc. But I never pushed for adopting this backfilling style. cc @joshkay10 is there anything we still need to fix in the backfilling mechanism to make it more stable? We can still use this for inspiration

src/backfill/prioritizer.ts
```diff
@@ -1,14 +1,15 @@
-import { RepositoryData } from "models/subscription";
```
thombergs, line 1: If `looper.ts` is gone, `prioritizer.ts` can go, too. Unless we want to finally refactor the backfilling mechanism

src/models/subscription.test.ts
```diff
@@ -1,289 +1,55 @@
...
-				expect(results).toEqual([expect.objectContaining({
-					jiraClientKey: "myClientKey"
-				}), expect.objectContaining({
-					jiraClientKey: "myClientKey_ghe_2"
-				})]);
-			});
-		});
+
+	describe.skip("syncStatusCounts", () => {
+		// TODO: add tests
```
thombergs, line 53: Do we have a task somewhere that keeps track of adding these tests?

src/models/subscription.ts
```diff
@@ -213,23 +184,7 @@ export class Subscription extends Model {
...
-		// TODO: this is temporary until we redo sync
-		if (key === "repositoryStatus" || key === "repositoryCursor") {
-			await this.update({ [key]: value });
-		} else {
-			await RepoSyncState.updateRepoForSubscription(this, Number(repositoryId), key, value);
-		}
-		return this;
-	}
-
+	// TODO: remove this, not necessary
```
thombergs, line 187: Maybe explain why it's not necessary? Whoever looks at this TODO will ask exactly that.

thombergs (approved):
some small comments.

Before merging, I would like to have it run in staging / dev once and check that the Pollinator tests are still running through and manually check that the backfilling is still working, because a lot of the code touched the backfilling.

src/models/subscription.test.ts
mboudreau: would be nice if we had a plugin that could take TODOs and add them to jira ;)

src/backfill/prioritizer.ts
mboudreau: we're not using prioritizer at all it seems, I'm glad to remove it since we need to redo the backfill properly.

src/models/subscription.test.ts
mboudreau: but yeah, I was going to add some tasks to jira

src/backfill/looper/looper.ts
joshkay10: We don't have the preemptive ratelimiting smarts yet.... I still want to spike this solution and bring over a couple of the recent enhancements, it does feel like a shame to drop it without an honest go. The best backfill solution will be an ever going debate, but it's worth evaluating the effort vs reward.

mboudreau:
@thombergs I've already deployed it to staging previously, pollinator passed and I've done manual testing of all aspects with success.  Redeploying to staging now for others to test :)

joshkay10 (dismissed):

src/models/subscription.test.ts
```diff
@@ -1,289 +1,57 @@
...
 
 	beforeEach(async () => {
-		sub = await Subscription.create({
-			gitHubInstallationId: GITHUHB_INSTALLATION_ID ,
+		/*sub = await Subscription.create({
+			gitHubInstallationId: 123,
 			jiraHost,
 			jiraClientKey: "myClientKey"
-		});
+		});*/
```
thombergs, line 11: Can the commented code here be removed completely?

src/sync/discovery.ts
```diff
@@ -1,31 +1,22 @@
...
-const updateSyncState = async (subscription: Subscription, repositories: Repository[], totalNumberOfRepos?: number): Promise<void> => {
+/*const updateSyncState = async (subscription: Subscription, repositories: Repository[], totalNumberOfRepos?: number): Promise<void> => {
 	await Promise.all([
 		subscription.updateSyncState({
 			repos: mapRepositories(repositories)
 		}),
 		totalNumberOfRepos && subscription.update({ totalNumberOfRepos })
 	]);
-};
+};*/
```
thombergs, line 19: Can this code be deleted?

thombergs (approved):

gxueatlassian (approved):

PR #1282, opened 2022-06-20
Arc-1408 model updates part 1

rachellerathbone:
Breaking https://github.com/atlassian/github-for-jira/pull/1277 into some smaller PRs. This one starts with the addition of the new middleware.

### DIFF
--- a/src/middleware/github-server-app-middleware.test.ts
+++ b/src/middleware/github-server-app-middleware.test.ts
@@ -1,0 +1,120 @@
+import Mock = jest.Mock;
+import { GithubServerAppMiddleware } from "middleware/github-server-app-middleware";
+import { getLogger } from "config/logger";
+import { mocked } from "ts-jest/utils";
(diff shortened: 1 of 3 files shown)


src/middleware/github-server-app-middleware.test.ts
```diff
@@ -0,0 +1,120 @@
...
+		await expect(GithubServerAppMiddleware(req, res, next))
+			.rejects
+			.toThrow('Jira hosts do not match');
+	});
+
+	it("should call next() when GH app is found and installation id matches", async () => {
+		req = {
+			log: getLogger("request"),
+			params: {
+				id: 3
```
rachellerathbone, line 91: It doesn't matter what ID gets passed here. We include an ID as a query param so that we fall into the conditional block of the middleware. On line 113 we mock querying the database and finding an entry with the ID that was passed and return the provided payload.

gxueatlassian (approved):

PR #1289, opened 2022-06-22
[ARC-1486] Adding e2e tests

mboudreau:
**What's in this PR?**
E2E Test framework working

**Why**
Because pollinator ain't cutting it

**Added feature flags**
None

**Affected issues**  
ARC-1486

**How has this been tested?**  
I mean, I added tests

**Whats Next?**
After this is working in main, need to add more tests to replace pollinator


.github/workflows/on-push.yml
```diff
@@ -23,35 +20,107 @@ jobs:
...
       - run: yarn run build:release
 
-  test:
-    runs-on: ubuntu-latest
+  unit-test:
+    runs-on: ubuntu-22.04
 
     steps:
       - uses: actions/checkout@v1
-      - uses: actions/setup-node@v3
```
gxueatlassian, line 31: 👍

.github/workflows/on-push.yml
```diff
@@ -23,35 +20,107 @@ jobs:
...
+      - name: create .env file
+        run: |
+          echo "${{ secrets.E2E_GITHUB_PRIVATE_KEY }}" > jira-test.pem
+          echo "APP_URL=http://localhost" > .env
+          echo "WEBHOOK_PROXY_URL=http://localhost" >> .env
+          echo "INSTANCE_NAME=fusion-test" >> .env
+          echo "NODE_ENV=test" >> .env
+          echo "APP_NAME=jira" >> .env
+          echo "APP_ID=${{ secrets.E2E_GITHUB_APP_ID }}" >> .env
+          echo "WEBHOOK_SECRET=${{ secrets.E2E_GITHUB_WEBHOOK_SECRET }}" >> .env
```
gxueatlassian, line 43: The step is called `unit-test`, but the secret is called `E2E_xxxx`, seems a bit strange to me. Am I missing something?

src/config/env.ts
```diff
@@ -63,11 +43,26 @@ export const envVars: EnvVars = new Proxy<object>({}, {
...
 
-// Check to see if all required environment variables are set
-const missingVars = requiredEnvVars.filter(key => envVars[key] === undefined);
-if (missingVars.length) {
-	throw new Error(`Missing required Environment Variables: ${missingVars.join(", ")}`);
-}
+envCheck(
+	"APP_ID",
+	"APP_URL",
+	"INSTANCE_NAME",
```
gxueatlassian, line 49: `INSTANCE_NAME` is new in the check compare to existing code. Just to confirm this is intended correct?

gxueatlassian (approved):
I am happy with the code structure and the tests so far.
Happy to get involved in more discussion though if anyone feels like to discuss this PR.

.github/workflows/on-push.yml
mboudreau: nah, just copy pasted it from the e2e tests.  Not sure this is even needed as unit tests never go out to the internet.  I'll review this

src/config/env.ts
mboudreau: Yes, it's needed or else stuff fails, was just something that was overlooked in the past.

gxueatlassian:
Hi @mboudreau just have a quick one, once merged, will each PR's build conflicting each other (since they are installing/uninstalling same app? )

test/e2e/utils/github.ts
```diff
@@ -0,0 +1,49 @@
+import { Page } from "@playwright/test";
+import { GithubTestDataRoles, testData } from "test/e2e/constants";
+import { e2eEnvVars } from "test/e2e/env-e2e";
+
+const data = testData.github;
+export const githubLogin = async (page: Page, roleName: keyof GithubTestDataRoles, saveState = false) => {
```
joshkay10, line 6: This looks like it is never used? it's commented out in setup.ts, is that just commented while you solve it or didn't work?

test/e2e/utils/jira.ts
```diff
@@ -0,0 +1,75 @@
...
+	const role = data.roles[roleName];
+	if (!role.username || !role.password) {
+		throw "Jira username or password missing";
+	}
+	await page.goto(data.urls.login);
+	await page.waitForLoadState();
+	// This is a hack because atlassian auth is all frontend, it shows the login page and form
+	// but then redirects to the correct page if the token is available.  However, playwright
+	// will continue after the page loads and try to fill in the form before being redirected,
+	// causing a flow issue.  This is non-standard for any authentication as it should
```
joshkay10, line 17: 🤣

test/setup/env-test.ts
```diff
@@ -1,5 +1,11 @@
-import { EnvVars } from "config/env";
+import {  EnvVars } from "config/env";
```
joshkay10, line 1: /nit/nit Double space 🌈 🌈

joshkay10 (approved):
Gee wizz, a lot of stuff going on there. Would definitely be keen for a walkthrough too

test/e2e/utils/github.ts
mboudreau: it works, but it gets blocked and asks for 2FA.  The next PR will have that in, this is just the start :)

mboudreau:
@gxueatlassian Yes, the e2e tests can collide.  This is just the base and the next bit is to start building a new app for every run so they don't collide.  right now though, the tests are just installing the app, so there's no much there in terms of functionality tested, I just wanted to get this in first so the scope of this PR doesn't blow out

PR #1292, opened 2022-06-24
Dev loop CI/CD

krazziekay:
**Rationale**
- This PR includes the automation of deploying to different environments after merging to main.

**Changes**
- A new lambda function was created which checks if the commit of the main branch is same to the commit of deployed version or not.
- If not, then it sends a request to run the pipelines in [Bitbucket](https://bitbucket.org/atlassian/github-for-jira-deployment)

**Whats remaining**
- The URL for running the pipelines still needs to be updated.
- The environment variables in Microscope needs to be confirmed and finalized.

### DIFF
--- a/deploy/lambda/auto-deployment.js
+++ b/deploy/lambda/auto-deployment.js
@@ -53,7 +53,7 @@ exports.handler = async function (event, context) {
 function getRequest(url) {
     const options = {
         headers: {
-            "User-Agent": "GitHub-for-Jira-App"
+            "User-Agent": "Auto-Deployment-Lambda-GitHub-for-Jira-App"
         }
     };
     return new Promise((resolve, reject) => {
--- a/deploy/lambda/auto-deployment.js
+++ b/deploy/lambda/auto-deployment.js
@@ -1,22 +1,23 @@
 const https = require('https');
 
 exports.handler = async function (event, context) {
+    const autoDeployUsername = process.env.AUTO_DEPLOY_USERNAME;
+    const autoDeployToken = process.env.AUTO_DEPLOY_TOKEN;
+
     if (!process.env.AUTO_DEPLOY_ENABLED) {
         console.log('Auto Deploying disabled');
         return;
     }
+    if (!autoDeployUsername) {
+        console.error('No auto deploy username was provided!');
+        return;
+    }
+    if (!autoDeployToken) {
+        console.error('No auto deploy token was provided!');
+        return;
+    }
 
     try {
-        const autoDeployUsername = process.env.AUTO_DEPLOY_USERNAME;
-        const autoDeployToken = process.env.AUTO_DEPLOY_TOKEN;
-
-        if (!autoDeployUsername) {
-            console.error('No auto deploy username was provided!');
-        }
-        if (!autoDeployToken) {
-            console.error('No auto deploy token was provided!');
-        }
-
         const versionURL = `${process.env.PROD_URL}/version`;
         const mainBranchURL = 'https://api.github.com/repos/atlassian/github-for-jira/branches/main';
 
--- a/deploy/lambda/auto-deployment.js
+++ b/deploy/lambda/auto-deployment.js
@@ -11,10 +11,10 @@ exports.handler = async function (event, context) {
         const autoDeployToken = process.env.AUTO_DEPLOY_TOKEN;
 
         if (!autoDeployUsername) {
-            throw new Error('No auto deploy username was provided!');
+            console.error('No auto deploy username was provided!');
         }
         if (!autoDeployToken) {
-            throw new Error('No auto deploy token was provided!');
+            console.error('No auto deploy token was provided!');
         }
 
         const versionURL = `${process.env.PROD_URL}/version`;
@@ -76,7 +76,7 @@ function getRequest(url) {
         });
 
         req.on('error', (err) => {
-            reject(new Error(err.message));
+            reject(new Error(err));
         });
     });
 }
@@ -108,10 +108,9 @@ function postRequest(url, body, headers = {}) {
         });
 
         req.on('error', err => {
-            reject(new Error(err.message));
+            reject(new Error(err));
         });
 
-        // 👇️ write the body to the Request object
         req.write(JSON.stringify(body));
         req.end();
     });
--- a/deploy/lambda/auto-deployment.js
+++ b/deploy/lambda/auto-deployment.js
@@ -1,0 +1,117 @@
+const https = require('https');
+
+exports.handler = async function (event, context) {
+    if (!process.env.AUTO_DEPLOY_ENABLED) {
+        console.log('Auto Deploying disabled');
+        return;
+    }
+
+    try {
+        const autoDeployUsername = process.env.AUTO_DEPLOY_USERNAME;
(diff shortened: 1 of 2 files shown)


deploy/lambda/auto-deployment.js
```diff
@@ -0,0 +1,118 @@
...
+        const mainBranchURL = 'https://api.github.com/repos/atlassian/github-for-jira/branches/main';
+
+        const deployedCommitSHA = (await getRequest(versionURL)).commit;
+        const mainCommitSHA = (await getRequest(mainBranchURL)).commit.sha;
+
+        if (deployedCommitSHA !== mainCommitSHA) {
+            console.log('Changes found, starting deployment...');
+
+            // TODO: Need to update this URL with the actual one and store in Microscope
+            const pipelinesURL = 'https://api.bitbucket.org/2.0/repositories/harminder5/bitbucket-pipeline/pipelines/';
```
atrigueiro: We still need to update this, right?

deploy/lambda/auto-deployment.js
```diff
@@ -0,0 +1,118 @@
...
+            const body = {
+                "target": {
+                    "type": "pipeline_ref_target",
+                    "ref_type": "branch",
+                    "ref_name": "master",
+                    // TODO: Change the pattern to the actual one from GH4J
+                    "selector": {"type": "custom", "pattern": "deployment-to-prod"}
```
atrigueiro: Isn't this one called `deploy-to-prod` instead of `deployment-to-prod`?

deploy/lambda/auto-deployment.js
krazziekay: This has been updated already...

deploy/lambda/auto-deployment.js
```diff
@@ -0,0 +1,118 @@
...
+        }
+    } catch (e) {
+        console.error('Error: ', e);
+    }
+};
+
+function getRequest(url) {
+    const options = {
+        headers: {
+            "User-Agent": "GitHub-for-Jira-App"
```
atrigueiro, line 56: Why do we need this for?

deploy/lambda/auto-deployment.js
krazziekay: I think you're seeing the old version ,this has been updated too...

deploy/lambda/auto-deployment.js
krazziekay: Without this, the get request throws an error asking for User-Agent...

github-for-jira.sd.yml
```diff
@@ -147,6 +147,22 @@ resources:
...
+  - type: lambda
+    name: auto-deployment-github-app
+    attributes:
+      runtime: nodejs14.x
+      prefetchMicrosEnvVars: true
+      artifact: "github-for-jira/auto-deployment-github-app-${BITBUCKET_BUILD_NUMBER}.zip"
+      handler: auto-deployment.handler
+      private: true
+      concurrentExecutions: 2
+      memory: 128
```
atrigueiro: This is the default value, so we can just remove this line https://hello.atlassian.net/wiki/spaces/MICROS/pages/505840672/Lambda+Resource+Type#memory

deploy/lambda/auto-deployment.js
atrigueiro: I would rename this to explicitly say it's the lambda inside the app making the request. Maybe something like `Auto-Deployment-Lambda-GitHub-for-Jira-App` maybe? Just because we'll be calling the app ourselves within this function.

deploy/lambda/auto-deployment.js
atrigueiro: Ops, yeah you're right, forgot to refresh

deploy/lambda/auto-deployment.js
krazziekay: The User-Agent has been updated...

github-for-jira.sd.yml
krazziekay: `memory` has been removed

deploy/lambda/auto-deployment.js
```diff
@@ -0,0 +1,115 @@
...
+    if (!process.env.AUTO_DEPLOY_ENABLED) {
+        console.log('Auto Deploying disabled');
+        return;
+    }
+
+    try {
+        const autoDeployUsername = process.env.AUTO_DEPLOY_USERNAME;
+        const autoDeployToken = process.env.AUTO_DEPLOY_TOKEN;
+
+        if (!autoDeployUsername) {
```
joshkay10, line 13: Do we want to return early if we are missing these values `autoDeployUsername` or `autoDeployToken`

deploy/lambda/auto-deployment.js
```diff
@@ -0,0 +1,115 @@
...
+            console.error('No auto deploy token was provided!');
+        }
+
+        const versionURL = `${process.env.PROD_URL}/version`;
+        const mainBranchURL = 'https://api.github.com/repos/atlassian/github-for-jira/branches/main';
+
+        const deployedCommitSHA = (await getRequest(versionURL)).commit;
+        const mainCommitSHA = (await getRequest(mainBranchURL)).commit.sha;
+
+        if (deployedCommitSHA !== mainCommitSHA) {
```
joshkay10, line 26: /nit - personal preference for me, I would change the `if` to `deployedCommitSHA === mainCommitSHA`then return early and drop the else.

deploy/lambda/auto-deployment.js
krazziekay: Moved above the try catch

deploy/lambda/auto-deployment.js
krazziekay: Doesn't really make much difference, so leaving it as it is...

atrigueiro (approved):

deploy/lambda/auto-deployment.js
```diff
@@ -0,0 +1,116 @@
...
+    if (!autoDeployToken) {
+        console.error('No auto deploy token was provided!');
+        return;
+    }
+
+    try {
+        const versionURL = `${process.env.PROD_URL}/version`;
+        const mainBranchURL = 'https://api.github.com/repos/atlassian/github-for-jira/branches/main';
+
+        const deployedCommitSHA = (await getRequest(versionURL)).commit;
```
gxueatlassian, line 24: This is checking the deployed production commit sha, so if code is deployed to stage, but we haven't manually deploy to prod yet, this code will keep deploying to stage env. Could be a bit of wastes of compute resources.

atrigueiro: How is it going to keep deploying to staging? This lambda only runs in prod.

krazziekay: No, it won't. This lambda will be zipped and uploaded as an artefact to **PROD** environment only(which is done in [bitbucket repo](https://bitbucket.org/atlassian/github-for-jira-deployment/pull-requests/40)). This lambda will run only for the **PROD**  environment and will only run pipelines for the `main` branch of GH4J. 
Deploying to staging or dev will have no effect on this lambda.

gxueatlassian: All good thanks guys.

PR #1294, opened 2022-06-27
ARC-1441 start using cryptor

gxueatlassian:
**Rationale:**
By Atlassian policy, all encryption will need to use Atlassian Cryptor when running on Atlassian internal platform.

**Ticket PRs tracking**
1. **[This PR]** Setup foundation for using Cryptor client
2. [Up-coming] Use Cryptor for GitHubServerApp Table secrets columns.

**Changes:**
1. Setup service descriptor to have cryptor sidecar enabled
2. Setup mock-cryptor docker image for local development
3. Add Cryptor client to make calls to sidecar
4. Add temporary api to test Cryptor in production


src/util/cryptor-http-client.ts
```diff
@@ -0,0 +1,114 @@
...
+		}, logger);
+
+		return plainText;
+	}
+
+	async healthcheck() {
+		await axios.get("/healthcheck", this.axiosCommonConfig);
+	}
+
+	async _post(operation: string, path: string, payload: EncryptionPayload | DecryptionPayload, rootLogger: Logger) {
```
gxueatlassian, line 91: @gxueatlassian rename `rootLogger`

docker-compose.yml
```diff
@@ -32,11 +32,19 @@ services:
     volumes:
       - ./.localstack:/docker-entrypoint-initaws.d
 
+  cryptor-sidecar:
```
mboudreau, line 35: Just `cryptor` would do.

.env.test
```diff
@@ -45,3 +45,9 @@ ATLASSIAN_URL=https://test-atlassian-instance.atlassian.net
 # Use `trace` to get verbose logging or `info` to show less
 LOG_LEVEL=info
 WEBHOOK_PROXY_URL=http://localhost:8080/github/events
+
+# Cryptor
+CRYPTOR_SIDECAR_BASE_URL=http://127.0.0.1:26272
```
mboudreau, line 50: Will this work?  Shouldn't it be `http://cryptor-sidecar:26272`?

.env.test
```diff
@@ -45,3 +45,9 @@ ATLASSIAN_URL=https://test-atlassian-instance.atlassian.net
...
+CRYPTOR_SIDECAR_CLIENT_IDENTIFICATION_CHALLENGE="GitHub for Jira client"
```
mboudreau, line 51: What's this?

src/util/cryptor-http-client.ts
```diff
@@ -0,0 +1,107 @@
...
+
+	private readonly keyAlias: string;
+	private readonly axiosCommonConfig: AxiosRequestConfig;
+	private readonly axiosPostConfig: AxiosRequestConfig;
+
+	constructor(opts: CryptorHttpClientOptions) {
+		this.keyAlias = opts.keyAlias;
+		this.axiosCommonConfig = {
+			baseURL: opts.baseUrl || envVars.CRYPTOR_SIDECAR_BASE_URL,
+			timeout: opts.timoutMSec || Number(envVars.CRYPTOR_SIDECAR_TIMEOUT_MSEC)
```
mboudreau, line 54: Do we really need a configurable timeout here?  Is the timeout even going to ever be triggered?

src/util/cryptor-http-client.ts
```diff
@@ -0,0 +1,107 @@
...
+export class CryptorHttpClient {
+
+	private readonly keyAlias: string;
+	private readonly axiosCommonConfig: AxiosRequestConfig;
+	private readonly axiosPostConfig: AxiosRequestConfig;
+
+	constructor(opts: CryptorHttpClientOptions) {
+		this.keyAlias = opts.keyAlias;
+		this.axiosCommonConfig = {
+			baseURL: opts.baseUrl || envVars.CRYPTOR_SIDECAR_BASE_URL,
```
mboudreau, line 53: Do we need options at all is this is all configurable via env vars?  Adding options is inviting bad code like hardcoding the value.

src/util/cryptor-http-client.ts
```diff
@@ -0,0 +1,107 @@
...
+			timeout: opts.timoutMSec || Number(envVars.CRYPTOR_SIDECAR_TIMEOUT_MSEC)
+		};
+		this.axiosPostConfig = {
+			...this.axiosCommonConfig,
+			headers: {
+				"X-Cryptor-Client": opts.cryptorChanllenge || envVars.CRYPTOR_SIDECAR_CLIENT_IDENTIFICATION_CHALLENGE,
```
mboudreau, line 59: So, this is just a random string that's used as an auth token?

src/util/cryptor-http-client.ts
```diff
@@ -0,0 +1,114 @@
...
+				"Content-Type": "application/json; charset=utf-8"
+			}
+		};
+	}
+
+	async encrypt(logger: Logger, plainText: string, encryptionContext: EncryptionContext = {}): Promise<string> {
```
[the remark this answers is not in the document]
mboudreau: I agree, and make it optional with a default

.env.test
gxueatlassian: I think it is related to this [bit](https://developer.atlassian.com/platform/cryptor/integration/integrating-sidecar/#enabling-ssrf-protection)

src/util/cryptor-http-client.ts
```diff
@@ -0,0 +1,107 @@
...
+		}, logger);
+
+		return plainText;
+	}
+
+	async healthcheck() {
+		await axios.get("/healthcheck", this.axiosCommonConfig);
+	}
+
+	async _post(operation: string, path: string, payload: EncryptionPayload | DecryptionPayload, logger: Logger) {
```
mboudreau, line 84: We don't do underscore with typescript as we can mark the function `private` instead

src/util/cryptor-http-client.ts
```diff
@@ -0,0 +1,107 @@
...
+		const childLogger = logger.child({ keyAlias: this.keyAlias, operation });
+
+		try {
+			const started = new Date().getTime();
+
+			const result = (await axios.post(path, payload, this.axiosPostConfig)).data;
+
+			const finished = new Date().getTime();
```
mboudreau, line 92: use `Date.now()` instead

src/util/cryptor-http-client.ts
mboudreau: Do we really need `operation`? can't we just use the path?

mboudreau: Actually, do we even need the post function?  Can't we just call axios post directly and simplify the process?

etc/cryptor-mock/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM	node:alpine
+
+WORKDIR /src
+
+RUN npm i express body-parser
+COPY	cryptor-mock.js /src
+
+EXPOSE	26272
+CMD	node /src/cryptor-mock.js
```
mboudreau, line 9: Do we know for sure there's not a cryptor mock already in existence?  If there isn't, should there be one on docker hub to make it easier for everyone to dev with it in the future?  

I'm not sure which team we'd need to talk to about this.

src/routes/api/api-router.ts
```diff
@@ -107,5 +108,51 @@ ApiRouter.delete(
...
+
+	const startedTime = new Date().getTime();
+	const encrypted = await cryptor.encrypt(logger, data, { a: "b" });
+	logger.info({
+		elapsed: new Date().getTime() - startedTime
+	}, `Data encrypted: ${encrypted}`);
+
+	try {
+		const decrypted = await cryptor.decrypt(logger, encrypted, { a: "b" });
+		logger.info({
```
mboudreau, line 142: Do we really need this info? What's the purpose?

src/routes/api/api-router.ts
```diff
@@ -107,5 +108,51 @@ ApiRouter.delete(
...
+			elapsed: new Date().getTime() - startedTime
+		}, `Data decrypted (round-trip): ${decrypted}`);
+
+		await cryptor.decrypt(logger, encrypted, { a: "b2" });
+		logger.info({
+			elapsed: new Date().getTime() - startedTime
+		}, `Data decrypted (round-trip): ${decrypted}`);
+	} catch (e) {
+		logger.error({ err: e });
+		logger.error({ error: e.toJSON() });
```
mboudreau, line 152: Why are we logging 2 errors?

src/routes/api/api-router.ts
```diff
@@ -107,5 +108,51 @@ ApiRouter.delete(
...
+ApiRouter.use("/cryptor", async (req: Request, resp: Response) => {
+
+	resp.status(204).send("ack");
+
+	let data = "";
+	for (let i = 0; i < 10; i++) {
+		data = data + "-" + Math.floor((Math.random() * 10));
+	}
+	const logger = req.log.child("testing-cryptor");
+	const cryptor = new CryptorHttpClient({ keyAlias: "micros/github-for-jira/github-server-app-secrets" });
```
mboudreau, line 132: Shouldn't this key be global for our entire app?

src/util/cryptor-http-client.ts
```diff
@@ -0,0 +1,107 @@
...
+ *   const encrypted = await client.encrypt(req.logger, "super-secret-secret");
+ *
+ */
+export class CryptorHttpClient {
+
+	private readonly keyAlias: string;
+	private readonly axiosCommonConfig: AxiosRequestConfig;
+	private readonly axiosPostConfig: AxiosRequestConfig;
+
+	constructor(opts: CryptorHttpClientOptions) {
```
mboudreau, line 50: I'm not convinced we need a constructor at all as we're not using multiple cryptor endpoint, we just need 2 static functions: `encrypt` and `decrypt`

github-for-jira.sd.yml
```diff
@@ -211,6 +215,12 @@ config:
     DEPLOYMENT_DATE: ${DEPLOYMENT_DATE}
     PRIVATE_KEY: vault://secret/data/builds/micros-sv--github-for-jira-dl-admins/github-app-private-key-stg
     GITHUB_CLIENT_SECRET: vault://secret/data/builds/micros-sv--github-for-jira-dl-admins/github-app-client-secret-stg
+
+    GLOBAL_AGENT_NO_PROXY: cryptor-sidecar # needed because of proxy.ts
+    CRYPTOR_SIDECAR_BASE_URL: http://cryptor-sidecar:26272
+    CRYPTOR_SIDECAR_ENCRYPTION_KEY_ALIASES: micros/github-for-jira/github-server-app-secrets
+    CRYPTOR_SIDECAR_CLIENT_IDENTIFICATION_CHALLENGE: 'GitHub for Jira client' # https://developer.atlassian.com/platform/cryptor/integration/integrating-sidecar/#enabling-ssrf-protection
```
mboudreau, line 222: Shouldn't this be something a bit more secure?

src/util/cryptor-http-client.ts
gxueatlassian: 1. remove statsd (cryptor sidecar already has metrics)
2. remove post, make into each of the method.

src/util/cryptor-http-client.ts
```diff
@@ -0,0 +1,107 @@
...
+			const finished = new Date().getTime();
+
+			statsd.histogram(cryptorMetrics.clientHttpCallDuration, finished - started, { operation });
+			// TODO: add statsd counter (maybe check with #help-cryptor if we already have this metric in sfx)
+			// statsd.increment(cryptorMetrics.clientHttpSuccessCount, { operation });
+
+			return result;
+		} catch (e) {
+			// Do not add { err: e } param to avoid logging payload
+			childLogger.warn("Cryptor request failed: " + e?.message?.replace(payload["plainText"], "<censored>"));
```
gxueatlassian, line 101: 1. Leave it as try/catch.
2. Remove childLogger.

.env.example
```diff
@@ -41,3 +41,7 @@ NGROK_AUTHTOKEN=<your-ngrok-authtoken>
 # Use `trace` to get verbose logging or `info` to show less
 LOG_LEVEL=debug
 WEBHOOK_PROXY_URL=
+
+# Cryptor
+CRYPTOR_BASE_URL=http://cryptor-sidecar:26272
```
mboudreau, line 46: just `CRYPTOR_URL` to follow the convention

.env.example
```diff
@@ -41,3 +41,7 @@ NGROK_AUTHTOKEN=<your-ngrok-authtoken>
...
+CRYPTOR_CLIENT_IDENTIFICATION_CHALLENGE=random-string
```
mboudreau, line 47: `CRYPTOR_CLIENT_ID` is probably better here

docker-compose.yml
```diff
@@ -32,11 +32,19 @@ services:
     volumes:
       - ./.localstack:/docker-entrypoint-initaws.d
 
+  cryptor:
+    build: etc/cryptor-mock
+    container_name: cryptor
+    environment:
+      CRYPTOR_CLIENT_IDENTIFICATION_CHALLENGE: "12345" #has to match the one in env.test
```
mboudreau, line 39: instead of injecting it here (and duplicating the value) you could get the cryptor mock to use to .env file to get the value

etc/cryptor-mock/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM	node:alpine
+
+WORKDIR /src
+
+RUN npm i express body-parser
+COPY	cryptor-mock.js /src
```
mboudreau, line 6: you know, I was thinking we could use the same dockerfile as the main app with a different yarn run command to spin up the cryptor mock.  No need for an extra Dockerfile.

mboudreau (approved):

PR #1295, opened 2022-06-27
ARC-1406 navigation between config and select gh version screens

rachellerathbone:
**Rationale:**
We have built most of the UI screens and bits and pieces for the backend so it's time to start wiring it all together. I've started with one point of navigation (from the config page to the GH version selection screen) to keep the number of file changes down. Each subsequent PR will add a new step in the flow.

**Changes:**
See comments: https://github.com/atlassian/github-for-jira/pull/1295/files

**Demo:**
The new path is hidden behind the FF for everyone at this point so no one will wind up on this new flow (expected behaviour).

With FF set to false we still navigate from the config screen to the connect an org screen:

https://user-images.githubusercontent.com/37155488/175880701-f4516df6-96a1-48fd-b061-7e0ccf3660de.mov

With FF set to true (done manually on my local for demo purposes only) we now navigate from the config view to our new GH select version page. We can also navigate back to the config page.

https://user-images.githubusercontent.com/37155488/175880855-be2b051b-3647-44fc-9540-e251fda3b33d.mov

**To test locally:**
In src/routes/jira/configuration/jira-configuration-get.ts change the value from false to true for the FF check for handleNavigationClassName. When the next step of the flow is complete I'll turn the FF on for our segment so we can dogfood before we start rolling out to customers.

**What's up next:**
In ARC 1453 I'll be adding navigation from the GH version select screen 

Screen Shot 2022-06-27 at 8 37 37 pm

to the GHE server URL screen (if the user has selected server)

Screen Shot 2022-06-27 at 8 38 07 pm

or to continue along the current flow if cloud is selected. Once this is done, I'll have enough to turn on the FF for our segment. We'll still be able to go down the current path, with the addition of the extra screen, and start testing the GHE path as more is added.


src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -112,6 +113,10 @@ export const JiraConfigurationGet = async (
 				isGlobalInstall: installation.repository_selection === "all"
 			}));
 
+		const handleNavigationClassName = await booleanFlag(BooleanFlags.GHE_SERVER, false, jiraHost)
```
rachellerathbone, line 116: I've wrapped this change in our GHE global FF so current flow in prod will remain unaffected. If the FF is off (as it is currently is for everyone) we still handle navigating via the classname add-organization-link.

src/routes/jira/jira-atlassian-connect-get.ts
```diff
@@ -1,18 +1,101 @@
...
-
 const key = `com.github.integration${instance ? `.${instance}` : ""}`;
-const conditions = [
+
+const adminCondition = [
 	{
 		condition: "user_is_admin"
 	}
 ];
+const modules = {
```
rachellerathbone, line 17: This was abstracted from `JiraAtlassianConnectGet` so that we can dynamically set the `moduleUrls` below on line 98.

src/routes/jira/jira-atlassian-connect-get.ts
```diff
@@ -1,18 +1,101 @@
...
+	},
+	postInstallPage: {
+		key: "github-post-install-page",
+		name: {
+			value: "GitHub Configuration"
+		},
+		url: postInstallUrl,
+		conditions: adminCondition
+	},
+	generalPages: [
```
rachellerathbone, line 66: This is the only new section added. Each new page we need to navigate to within Jira will be added as an object here.

static/js/jira-configuration.js
```diff
@@ -23,6 +23,17 @@ $(".add-organization-link").click(function(event) {
 	});
 });
 
+$(".select-github-version-link").click(function(event) {
+	event.preventDefault();
+
+	AP.navigator.go(
+		'addonmodule',
+			{
+				moduleKey: "github-select-version-page"
```
rachellerathbone, line 32: This will eventually need to be dynamic as we will be able to hit this fro 2 different points in the flow: 1. from the config page when state is empty or 2. from the connect an org page when state is not empty. However, I'm not close to that point so hardcoded works for this PR.

views/jira-configuration.hbs
```diff
@@ -321,8 +321,9 @@
                 Shortcut
               </span>
             </a>
+            <input type="hidden" id="_csrf" name="_csrf" value="{{../csrfToken}}">
             <button
-              class="ak-button ak-button__appearance-primary add-organization-link"
+              class="ak-button ak-button__appearance-primary {{handleNavigationClassName}}"
```
rachellerathbone, line 326: Only modified the `Connect GH organization` button for empty state. The other button will be updated when we update the GH config screen to render multiple servers + cloud (button becomes a dropdown).

gxueatlassian (approved):
This PR is small enough, quite like it.

src/routes/jira/select/jira-select-version-get.ts
```diff
@@ -0,0 +1,19 @@
...
+		req.log.info("Received Jira select GitHub version page request");
+
+		res.render("jira-select-github-version.hbs", {
+			previousPagePath: "github-post-install-page"
+		});
+
+		req.log.info("Jira select rendered successfully.");
+	} catch (error) {
+		return next(new Error(`Failed to render Jira select GitHub version page: ${error}`));
+	}
```
mboudreau, line 18: Why would this render fail?

PR #1300, opened 2022-06-28
Arc 1441 use cryptor in GitHub app

gxueatlassian:
**Rationale**

Need to use cryptor to save and retrieve secrets from DB.

**Changes**

1. Disable the write to the secrets fields, prevent accidental writing plain text into db
2. Provide both explicit read and write access via cryptor, and mark as async.
3. Keep the read access from field directly, but it only return encrypted data.

**Assumption**

1. Do not favor using the `getter` and `setter` per this [doc](https://sequelize.org/v5/manual/models-definition.html#getters--amp--setters). Because it is both a) does not support async and b) implicit, hiding the async/network call anda 3) may incur bad performance if doing weird js stuff like (...spread, object.assign etc). 

**Upcoming**

1. Think about prod db migration of removing the `secret` column in the db schema.


src/models/github-server-app.ts
```diff
@@ -130,18 +169,27 @@ GitHubServerApp.init({
 		allowNull: false
 	},
 	secrets: encrypted.vault("secrets"),
```
gxueatlassian, line 171: Leave the removal of this field alone for now. Will need to consider prod db (both schema and data) migration properly in another PR later.

db/migrations/20220630002212-add-github-client-secrets-to-github-server-apps.js
```diff
@@ -0,0 +1,37 @@
...
+		await queryInterface.removeColumn(tableName, "gitHubClientSecret");
+		await queryInterface.removeColumn(tableName, "webhookSecret");
+		await queryInterface.removeColumn(tableName, "privateKey");
+		/*
+		 * Please note, rollback of following won't work, or will have bugs in the app IF THE APP still use sequalize.encrypt, as data type is different (and allowNull will messup)
+		 * Assuming we are using Cryptor for this table in the future, so above concern need not to be worried.
+		 * If we switch back to sequalize.encrypt, we will need to use another name for the secret vault column, can't reuse this one (secret).
+		 * Having this line merely because to keep the db migration scripts backward/forward compatible, able to roll forward and backwards.
+		 */
+		await queryInterface.addColumn(tableName, "secrets", {
```
gxueatlassian, line 32: Thoughts please.

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -24,7 +26,8 @@ export const WebhookReceiverPost = async (request: Request, response: Response):
 			return;
 		}
 		webhookSecret = gitHubServerApp.webhookSecret;
-		const verification = createHash(JSON.stringify(payload), webhookSecret);
+		if (!webhookSecret) throw new Error("Cannot find webhookSecret for gitHubServerApp uuid: " + uuid);
```
gxueatlassian, line 29: Actually, is empty webhook secret one of the use cases?
If so, we need another PR to bypass the `createhash` and `verification` 
If on, all good, leave it.

gxueatlassian: Well, as DB allow null is false, so we don't have to worry webhookSecret is empty.

src/models/github-server-app.ts
```diff
@@ -28,6 +31,15 @@ export class GitHubServerApp extends Model {
 	updatedAt: Date;
 	createdAt: Date;
 
+
+	getCryptorKeyAlias() {
+		return CryptorHttpClient.GITHUB_SERVER_APP_SECRET;
+	}
+
+	async getEncryptContext(): Promise<Record<string, string | number>> {
+		return {};
```
gxueatlassian, line 40: future flexibility for encryption context.

src/models/github-server-app.ts
```diff
@@ -174,3 +186,22 @@ GitHubServerApp.init({
 		allowNull: false
 	}
 }, { sequelize });
+
+const encryptSecretFields = async (app: GitHubServerApp, fields?: string[]) => {
```
gxueatlassian, line 190: Per @mboudreau 's comments, I extract this. But I am having trouble moving this up into the base class.

src/models/encryption-model.ts
```diff
@@ -0,0 +1,20 @@
...
+import Logger from 'bunyan';
+
+export abstract class EncryptionModel<TSecretFields> extends Model{
+
+	abstract getEncryptContext(field: TSecretFields): Promise<Record<string, string|number>>;
+	abstract getCryptorKeyAlias(): CryptorSecretKey;
+
+	async decrypt(field: TSecretFields , logger: Logger) {
+		const ctx = await this.getEncryptContext(field);
+		return await CryptorHttpClient.decrypt(this.getDataValue[field as any], ctx, logger);
```
gxueatlassian, line 12: I can't get something like `<TSecretField extends keyof SubClass>`, have to fallback to `as any` here when using.

thombergs, line 12: Not from this PR, but in general: The `CryptorClient` should actually be renamed to `EncryptionClient` because the fact that it's using Cryptor under the hood is an implementation detail.

src/models/github-server-app.ts
thombergs: yeah, this is a bit asymmetric with this encryption logic in `GitHubServerApp`, but the rest in `EncryptionModel`.

How about moving this function to `EncryptionModel` and replacing the `app` parameter with a parameter of type `EncryptionModel`. We'd have to add an abstract function to `EncryptionModel` then that returns all the secret fields.

src/models/encryption-model.ts
```diff
@@ -0,0 +1,20 @@
+import { Model } from "sequelize";
+import { CryptorHttpClient, CryptorSecretKey  } from "../util/cryptor-http-client";
+import Logger from 'bunyan';
+
+export abstract class EncryptionModel<TSecretFields> extends Model{
```
thombergs, line 5: I'd rather call this `EncryptedModel`, because it's not a model of an encryption, but rather a model of which parts are encrypted.

src/models/encrypted-model.ts
```diff
@@ -0,0 +1,30 @@
+import { Model } from "sequelize";
+import { EncryptionClient, CryptorSecretKey } from "../util/encryption-client";
+import Logger from "bunyan";
+
+export abstract class EncryptedModel<TModel extends EncryptedModel<TModel, TSecretField>, TSecretField extends keyof TModel> extends Model {
```
gxueatlassian, line 5: /magic !

src/models/encrypted-model.ts
```diff
@@ -0,0 +1,30 @@
...
+	abstract getEncryptContext(field: TSecretField): Promise<Record<string, string | number>>;
+	abstract getCryptorKeyAlias(): CryptorSecretKey;
+	abstract getAllSecretFields(): TSecretField[];
+
+	async decrypt(field: TSecretField, logger?: Logger) {
+		const ctx = await this.getEncryptContext(field);
+		return await EncryptionClient.decrypt(this.getDataValue(field as any), ctx, logger);
+	}
+
+	protected async encrypt(field: TSecretField, logger?: Logger) {
```
gxueatlassian, line 16: Didn't open it as:
1. We don't need it outside,
2. Avoid double-encryption accidentally by consumer. (Like `app.secrete = await EncryptClient.encrypt('plain-text')` )

src/models/github-server-app.ts
```diff
@@ -161,19 +177,18 @@ GitHubServerApp.init({
 		type: DataTypes.STRING,
 		allowNull: false
 	},
-	secrets: encrypted.vault("secrets"),
-	gitHubClientSecret: encrypted.field("gitHubClientSecret", {
-		type: DataTypes.STRING,
+	gitHubClientSecret: {
+		type: DataTypes.TEXT,
```
gxueatlassian, line 181: Might need some discussion here, but I read some article about `TEXT` is not worse than `STRING` anyway as underhood impl same

src/models/encrypted-model.ts
```diff
@@ -0,0 +1,32 @@
...
+export abstract class EncryptedModel<C extends Model, K extends keyof Pick<C, StringValues<C>>> extends Model {
+
+	abstract getEncryptContext(field: K): Promise<Record<string, string | number>>;
+
+	abstract getEncryptionSecretKey(): EncryptionSecretKeyEnum;
+
+	abstract getSecretFields(): readonly K[];
+
+	async decrypt(value: string, context?: EncryptionContext): Promise<string> {
+		return await EncryptionClient.decrypt(value, context);
```
gxueatlassian, line 17: hi @mboudreau , shouldn't this be `async decrypt(value: K` instead of `async decrypt(value: string`

src/models/github-server-app.ts
```diff
@@ -30,6 +30,23 @@ export class GitHubServerApp extends Model {
 	updatedAt: Date;
 	createdAt: Date;
 
+	getEncryptionSecretKey() {
+		return EncryptionSecretKeyEnum.GITHUB_SERVER_APP;
+	}
+
+	async getEncryptContext() {
+		//For example: we can call database to fetch sharedSecret for use as EncryptionContext
```
gxueatlassian, line 38: I am hesitate to put it here right now, as this is under discussion.

db/migrations/20220630002212-add-github-client-secrets-to-github-server-apps.js
```diff
@@ -0,0 +1,37 @@
...
+		});
+		await queryInterface.addColumn(tableName, "webhookSecret", {
+			type: Sequelize.TEXT,
+			allowNull: false
+		});
+		await queryInterface.addColumn(tableName, "privateKey", {
+			type: Sequelize.TEXT,
+			allowNull: false
+		});
+		await queryInterface.removeColumn(tableName, "secrets");
```
mboudreau, line 19: should we be explicit in the naming and call it `encryptedPrivateKey`, `encryptedWebhookSecret` and `encryptedGithubClientSecret`

gxueatlassian: But, in the code we will say `app.encryptedWebhookSecret = "some-plain-text"`, wouldn't that be strange?

mboudreau: we can change the variable name in the code as part of the model init :)

The code would stay the same, just the DB column would have a different name

gxueatlassian: you mean just change the db column name? Ah, I am cool with that.

gxueatlassian: implemented

src/models/github-server-app.ts
mboudreau: Well, we should probably decide this before merging it before someone uses the model and we can potentially lose data if we don't migrate appropriately.

src/models/github-server-app.ts
mboudreau: if it's the same implementation, I'd suggest we stick with `STRING` as it makes more sense from a code perspective

mboudreau:
Looking good, just need to decide on encryption context @thombergs

src/models/github-server-app.ts
gxueatlassian: I thought about that, yup we can do this way. But I suspect in reality we will sooner or later encounter a situation that we need to change the encryptionContext.

I suggest if that happen, we can change that `async getEncryptContext()` to return `EncryptionContext[]` instead of `EncryptionContext`. And in the base class, it decrypt by the order of the context in the array.

mboudreau: hm, not a bad solution, we would need to re-encrypt after decryption to work with the "main" context or figure out a way to do this via sequelize migrations as a once off...

thombergs (dismissed):

src/models/github-server-app.ts
gxueatlassian: Oh I mean same underline impl, but the constraints are still differents (https://sequelize.org/v5/manual/data-types.html)
STRING: varchar(255) and TEXT no constraint but the db storage engine. (1gb?)

thombergs (approved):

thombergs (approved):

src/models/github-server-app.ts
mboudreau: ah right

db/migrations/20220630052212-add-github-client-secrets-to-github-server-apps.js
```diff
@@ -0,0 +1,37 @@
...
+		});
+		await queryInterface.addColumn(tableName, "encryptedWebhookSecret", {
+			type: Sequelize.TEXT,
+			allowNull: false
+		});
+		await queryInterface.addColumn(tableName, "encryptedPrivateKey", {
+			type: Sequelize.TEXT,
+			allowNull: false
+		});
+		await queryInterface.removeColumn(tableName, "secrets");
```
mboudreau, line 19: this might cause an issue on prod on deployment.  remove this for now, we can delete it once all code refers only to the new columns

PR #1306, opened 2022-06-28
ARC-1369 log serializer

joshkay10:
**Whats in the PR**

- The bulk of the files are converting Probots `LogggerWithTarget` to Bunyans `Logger`
- There is also new serializers to hash some(not yet all) sensitive data (UGC/PII)

**Whats next**
- There are still UGC/PII that aren't being hashed or removed, so will find them and update them appropriately.


gxueatlassian (approved):

src/config/logger.ts
```diff
@@ -22,9 +22,10 @@ const LOG_STREAM = filteringHttpLogsStream(FILTERING_FRONTEND_HTTP_LOGS_MIDDLEWA
 	bformat({ outputMode, levelInString: true })
 );
 
+// Todo existing config serializer is throwing occasional error, investigate.
 const responseSerializer = (res: AxiosResponse) => ({
 	...stdSerializers.res(res),
-	config: JSON.parse(util.inspect(res?.config)), // removes circular dependency in json
+	config: JSON.parse(inspect(res?.config)), // removes circular dependency in json
```
mboudreau, line 28: wait, I'm fairly sure this is wrong.  I have some ideas on how to improve the logger here, might be simpler if we just go on a call I think.

src/config/logger.ts
```diff
@@ -45,6 +46,22 @@ const errorSerializer = (err) => (!err || !err.stack) ? err : {
 	stack: getFullErrorStack(err)
 };
 
+const hashSerializer = (data: any): string => {
+	if (data === undefined || data == null) {
+		return data;
```
mboudreau, line 51: this should return `undefined` and not null so it gets removed from the payload when stringifying it.

src/config/logger.ts
```diff
@@ -45,6 +46,22 @@ const errorSerializer = (err) => (!err || !err.stack) ? err : {
 	stack: getFullErrorStack(err)
 };
 
+const hashSerializer = (data: any): string => {
```
mboudreau, line 49: return type should be `string | undefined`

mboudreau (dismissed):

gxueatlassian (dismissed):

src/config/logger.ts
```diff
@@ -45,6 +46,22 @@ const errorSerializer = (err) => (!err || !err.stack) ? err : {
...
+	if (data === undefined || data == null) {
+		return data;
+	}
+	return createHashWithSharedSecret(data);
```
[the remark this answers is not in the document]
joshkay10: Would be nice, then I can drop the hashserializer and call this directly!! However the `createHashWithSharedSecret` is expected to return a string, so It gets a bit messy trying to get around that 😒

gxueatlassian (approved):

PR #1310, opened 2022-06-29
Updated cron expression to 1 hour for auto deployment

Harminder84:


atrigueiro (approved):

mboudreau:
Why are we doing it every hour?

atrigueiro:
> Why are we doing it every hour?

Because of [ARC-1462]. It's just a temporary solution.

Harminder84:
> 

The issue is pipeline took around 40 mins to deploy to prod and lambda again runs in that timeframe and triggers the pipeline again.  Need to update the lambda to add check for triggering pipeline only if no pipeline is running for prod deployment.
Created [ticket](https://softwareteams.atlassian.net/browse/ARC-1462) for this issue and  will work on it next week.

PR #1319, opened 2022-07-01
ARC-1438 : Update the JiraConfiguration view to show list of GH cloud and GHE organizations separately

krazziekay:
**Changes**
- Frontend
  - Updated the list of connections as the list of GH Cloud apps.
  - Created a new section for the list of GH servers.
- Backend
  - Added a new method to fetch the list of GH server apps for a given `installationId`
  - Used this method to fetch the list of GH server apps.
  - Modified the list of GH servers to get the desired output

**Video**

https://www.loom.com/share/b33c06f6fa4149d4be38deb488be7702


src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -113,20 +112,60 @@ export const JiraConfigurationGet = async (
...
 			? "select-github-version-link"
 			: "add-organization-link";
 
 		res.render("jira-configuration.hbs", {
 			host: jiraHost,
-			successfulConnections,
-			failedConnections,
-			hasConnections: !!installations.total,
+			gheServers,
+			gheCloud: { successfulConnections, failedConnections },
```
gxueatlassian, line 163: The `e` in `gheCloud`, does it refer to `enterprise`? Is it intended or should it be `ghCloud`?

views/jira-configuration.hbs
```diff
@@ -1,143 +1,193 @@
...
-          <div class='jiraConfiguration__syncRetryModal'  id="sync-status-modal">
-            <div class='jiraConfiguration__syncRetryModalContent'>
-              <div class="jiraConfiguration__syncRetryModal__headerContainer">
-                <h4 class="jiraConfiguration__syncRetryModal__header">Backfill status</h4>
-                <span id="status-close" class="jiraConfiguration__syncRetryModal__closeBtn aui-icon aui-iconfont-close-dialog"></span>
-              </div>
+  <section class="jiraConfiguration_content">
+    {{#if hasConnections}}
+      <p>
+        After you connect your GitHub account or organization to Jira and include issue keys in branch names, commit
```
gxueatlassian, line 70: Hi this file has big differences from `git` 's point of view, do you mind re-structure the whitespace and align to main branch as much as possible? 

--------

I think generally, even there's formatting issue in the origin code, if we could not touch it, the PR (git diff) can more focus on the change.

src/util/handlebars/handlebar-helpers.ts
```diff
@@ -7,6 +7,7 @@ export const replaceSpaceWithHyphenHelper = (str?: string) => !isPlainObject(str
 
 export const registerHandlebarsHelpers = () => {
 	hbs.registerHelper("toLowerCase", toLowercaseHelper);
+	hbs.registerHelper("or", (...args: unknown[]) => Array.prototype.slice.call(args, 0, -1).some(Boolean));
```
gxueatlassian, line 10: I am not following what is the purpose of this one. Maybe write a unit test to demo the purpose?
And how is it different to `[...args].some(Boolean)`? why `-1`?

views/jira-configuration.hbs
```diff
@@ -221,128 +273,373 @@
...
+                  Dismiss
+                </button>
+              </div>
+            </div>
+          {{/each}}
+        </div>
+      </div>
+      {{/if}}
+
+      {{#if gheServers.length}}
```
gxueatlassian, line 358: Okay, this is getting big, maybe some point of time, we need to split this into separate file. (if possible in handlebars)

src/util/handlebars/handlebar-helpers.ts
krazziekay: - Helper method re-worked.
- Test cases added for this method and also for `concatStringHelper`, which I had created in the last PR.

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: Nice catch! Renamed to `ghCloud`.

views/jira-configuration.hbs
krazziekay: 2 Partials created to shorten the changes on this file

views/jira-configuration.hbs
```diff
@@ -1,348 +1,254 @@
 <html>
-  <head>
```
gxueatlassian, line 2: I see a lot of git diff for formatting only here. Do you mind update it so that here's no change in these whitespaces?
image

views/jira-configuration.hbs
```diff
@@ -1,348 +1,254 @@
...
-                      The backfill has finished. Information from selected repositories will be shown in Jira’s development information panel.
-                    </td>
-                  </tr>
-                </tbody>
-              </table>
+  />
+  <script src="/public/js/jquery.min.js" nonce="{{nonce}}"></script>
+</head>
+
+<body class="aui-page-hybrid">
```
gxueatlassian, line 30: Same here, quite some changes in the html only because of whitespaces. I guess if you can revert the changes to the whitespaces from main branch, these change will be disappear.
image

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -113,20 +112,60 @@ export const JiraConfigurationGet = async (
...
 		const handleNavigationClassName = await booleanFlag(BooleanFlags.GHE_SERVER, false, jiraHost)
 			? "select-github-version-link"
 			: "add-organization-link";
 
 		res.render("jira-configuration.hbs", {
 			host: jiraHost,
-			successfulConnections,
-			failedConnections,
-			hasConnections: !!installations.total,
+			gheServers,
```
rachellerathbone, line 162: We shouldn't be touching production code here. To be on the safer side, until the FF has been rolled out 100% in prof, I suggest that we create 2 jira-configuration views and pass data to each separately:

- the first would keep the current naming: jira-configuration.hbs - this would take in the data that currently exists in prod
- the second would be the new version: something like jira-configuration-new.hbs or jira-configuration-ghe.hbs - this would take in the data you are specifying in this PR.

Then set the page to be rendered based on whether the FF is on or not and pass the data

rachellerathbone (changes requested):

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -113,20 +112,60 @@ export const JiraConfigurationGet = async (
 				isGlobalInstall: installation.repository_selection === "all"
 			}));
 
+		/**
```
rachellerathbone, line 115: I don't think we should leave this huge TODO in here. Better to add details to [ARC 1471](https://softwareteams.atlassian.net/jira/software/projects/ARC/boards/385/backlog?selectedIssue=ARC-1471&text=jira-con) to outline how we need to munge the data from the backend so it renders correctly.

krazziekay: This has been removed.
But, I add added here to provide a data sample, for anyone who wants to test the UI. What would be the best place to add such data?

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: The new view and css have been separated, which can be enabled by the feature flag `ghe_server`

views/jira-configuration.hbs
```diff
@@ -106,226 +130,107 @@
...
-                  <th class="jiraConfiguration__table__head__title">Repository access</th>
-                  <th class="jiraConfiguration__table__head__title">Backfill status
-                    <span class="jiraConfiguration__table__infoIcon aui-icon aui-iconfont-info-filled" id="sync-status-modal-btn">
-                      Information
-                    </span>
-                  </th>
-                  <th class="jiraConfiguration__table__head__title">Settings</th>
-                </tr>
-              </thead>
+          <div class="jiraConfiguration__optionContainer">
```
rachellerathbone: We only want to show this if a user has both cloud and server connected. It doesn't make much sense to me to have the option to click 'All' or 'Github Cloud'/'GitHub Enterprise Server' if I'm only connected to one version type.

- [ ] hide the lozenges unless a user is connected to both cloud and server

views/jira-configuration-new.hbs
```diff
@@ -0,0 +1,253 @@
...
+  <body class="aui-page-hybrid">
+    <div class="jiraConfiguration">
+      <header class="jiraConfiguration__header">
+        <h1 class="jiraConfiguration__header__title">GitHub configuration</h1>
+
+        {{#if hasConnections}}
+          <button
+              class="jiraConfiguration__header__title__cta ak-button ak-button__appearance-primary aui-button aui-dropdown2-trigger"
+              aria-controls="connectBtnDropdown"
+          >Connect GitHub organization
```
rachellerathbone, line 39: Looks like the label for the button has been updated in the designs to `Connect a GitHub organization`

views/jira-configuration-new.hbs
```diff
@@ -0,0 +1,253 @@
...
+                 target="_blank">
+                <span>Connect GitHub Cloud</span>
+                <i class="aui-icon aui-iconfont-question-filled"></i>
+                <div class="jiraConfiguration__connectGHCloudTooltip">Select this option if the URL of your GitHub
+                  organization starts with "http://github.com".
+                  This includes “Free”, “Team” and “GitHub Cloud Enterprise” accounts.
+                </div>
+              </a>
+              <a class="jiraConfiguration__connectBtnDropdownOption select-github-version-link"
+                 target="_blank">
```
rachellerathbone, line 53: We shouldn't have target blank here. We need it for cloud because that involves navigating _out_ of Jira and 'into' GH. For server, we'll be navigating to a screen that's still on the Jira side of the flow e.g. jira/ghe-server-list (or whatever we end up calling that path). This should happen in the same tab.

views/jira-configuration.hbs
krazziekay: Logic has been corrected. This `optionContainer` will be shown only if there are both GH cloud and GHE server apps. Otherwise it will remain hidden.

views/jira-configuration-new.hbs
krazziekay: Label corrected

src/util/handlebars/handlebar-helpers.ts
```diff
@@ -4,9 +4,13 @@ import { isPlainObject } from "lodash";
 export const concatStringHelper = (...strings: string[]) => strings.filter((arg: unknown) => typeof arg !== "object").join(" ");
 export const toLowercaseHelper = (str?: string) => !isPlainObject(str) && str?.toString?.().toLowerCase() || "";
 export const replaceSpaceWithHyphenHelper = (str?: string) => !isPlainObject(str) && str?.toString?.().replace(/ /g, "-") || "";
+export const compareOr = (...args: unknown[]) => args.filter((arg: unknown) => typeof arg !== "object").some(Boolean);
```
rachellerathbone, line 7: This needs to be renamed to be more explicit. Because helpers are typically 1-lines, the name should tell us what this is doing without us needing to read the code. Also, all helper functions end with `Helper`

src/util/handlebars/handlebar-helpers.ts
```diff
@@ -4,9 +4,13 @@ import { isPlainObject } from "lodash";
...
+export const compareAnd = (...args: unknown[]) => args.filter((arg: unknown) => typeof arg !== "object").every(Boolean);
```
rachellerathbone, line 8: Same here.

src/util/handlebars/handlebar-helpers.ts
krazziekay: Variable renamed to `logicalOrHelper` and `logicalAndHelper`.

src/routes/jira/configuration/jira-configuration-get.test.ts
```diff
@@ -65,11 +66,13 @@ describe("Jira Configuration Suite", () => {
 			.get(`/app/installations/15`)
 			.reply(200, singleInstallation);
 
+		const gheServerEnabled = await booleanFlag(BooleanFlags.GHE_SERVER, false, jiraHost);
```
gxueatlassian, line 69: I'd reckon we don't use booleanFlag to switch the unitTest result.
If this line always return true, we may just test the `true` path.

krazziekay: Initially, the booleanFlag will be turned on for some instances only, at that time this test case would fail. We can revisit this later once the booleanFlag is true for all instances...

gxueatlassian: I mean it is better if the unit test doesn't rely on booleanFlag at runtime. Maybe we can mock the booleanFlag, so that it always return true in your test case, (or always return false in your test case). 

something like `mock(booleanFlag).mockReturnValue(false)` <--- I may not get the syntax right, but you get the idea. Explicit stating the assumption in the unit test is better than rely on implicit assumption.

krazziekay: Oh, yup, agreed!
booleanFlag has been mocked to always return true

gxueatlassian (dismissed):
I am okay to approve this from my point of view at this stage. But I'll leave it for @rachellerathbone to decide the overall. As I don't have the context for one of the comment related to `huge todo block`. 

let me know if I shouldn't approve this PR please.

src/routes/jira/configuration/jira-configuration-get.ts
rachellerathbone: I would create a ticket for `TODO: fetch the actual list of GHE servers` and add it to that. The auto/manual app creation isn't implemented yet but we can `GitHubServerApp.install(payload)` to create entries in the db. This would allow us to implement rendering data from the database instead of mocking it out and we wouldn't need to save the data blob you had here anywhere as we'll effectively setting it up to take the data and programmatically render it.

krazziekay: I've created a new ticket for this. https://softwareteams.atlassian.net/browse/ARC-1488. Once the new endpoint is done, this part can be then be updated to fetch the actual list.

rachellerathbone: That ticket doesn't need to wait on the endpoint. We already have methods that exist in the GHSA app model to install and retrieve data from that table.

rachellerathbone (dismissed):

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: Updated the blank array to fetch actual GH Server apps data

src/models/github-server-app.ts
```diff
@@ -70,6 +70,26 @@ export class GitHubServerApp extends EncryptedModel {
 		});
 	}
 
+	/**
+	 * Get all GitHubServerApps with installationId
+	 *
+	 * @param {{installationId: number}} installationId
+	 * @returns {GitHubServerApp[]}
+	 */
+	static async getForInstallationId(
```
mboudreau, line 79: To stick with nomenclature of sequelize, we should use `find` not `get`

src/routes/jira/configuration/jira-configuration-get.test.ts
```diff
@@ -64,15 +70,16 @@ describe("Jira Configuration Suite", () => {
 
 	it("should return success message after page is rendered", async () => {
 		const response = mockResponse();
+		mocked(booleanFlag).mockResolvedValue(true);
```
mboudreau, line 73: just an FYI, this means that all boolean flags are true everywhere instead of the specific one you want.  If you look at the other tests, we have a `when` convention so you can specify the FF you want to return true which will remove all potential side effects from this test

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -67,6 +68,29 @@ const getInstallation = async (subscription: Subscription, log: Logger, gitHubAp
 	}
 };
 
+const getConnectionsAndInstallations = async (subscriptions, req, id) => {
```
mboudreau, line 71: type this function please

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -94,30 +118,41 @@ export const JiraConfigurationGet = async (
...
+			.map(async (value, key) => ({
+				gitHubBaseUrl: key,
+				// Fetching the subscriptions for each GH Server App
+				applications: await Promise.all(value.map(async (app) => {
+					const {
+						successfulConnections,
+						failedConnections
+					} = await getConnectionsAndInstallations(subscriptions, req, app.id);
+					app["successfulConnections"] = successfulConnections;
+					app["failedConnections"] = failedConnections;
```
mboudreau, line 136: Don't use string literal accessors, just dot accessors will do

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -94,30 +118,41 @@ export const JiraConfigurationGet = async (
...
+
+					return app;
+				}))
+			})).value());
+
+
+		const gheServerEnabled = await booleanFlag(BooleanFlags.GHE_SERVER, false, jiraHost);
```
mboudreau, line 143: Don't you want to evaluate this FF before doing all the work above?  One of the reason to have an FF is that if something goes wrong in production, you can turn it off and it'll work like it did previously, but you've changed so much logic that if something in that code above doesn't work, it'll break with no way to turn it off other than reverting the PR and redeploying.

src/models/github-server-app.ts
krazziekay: Updated the name to `findForInstallationId`

src/routes/jira/configuration/jira-configuration-get.test.ts
krazziekay: Used `when` to mock only one FF.

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: Types added

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: - Used dot accessors
- Added new type to avoid typescript errors

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -67,15 +86,28 @@ const getInstallation = async (subscription: Subscription, log: Logger, gitHubAp
...
+				id: installation.id,
+				deleted: installation.deleted,
+				orgName: repo?.repoOwner
+			};
+		}));
+
+	const successfulConnections: SuccessfulConnection[] = installations.fulfilled
+		.map((installation) => ({
+			...installation,
+			isGlobalInstall: installation.repository_selection === "all"
```
mboudreau, line 106: /nit what's a global install?

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: The logic for this method has been updated by just one if-else, if FF is true, fetch config from `JiraCloudAndEnterpriseConfiguration`, else fetch from `JiraCloudConfiguration`.

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: Not sure, was like this originally. This part of the code was just moved and not changed.

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -93,42 +178,11 @@ export const JiraConfigurationGet = async (
...
-			nonce: res.locals.nonce,
-			handleNavigationClassName,
-			isGitHubCloudApp: await isGitHubCloudApp(gitHubAppId)
-		});
-
+		if (await booleanFlag(BooleanFlags.GHE_SERVER, false, jiraHost)) {
+			res.render("jira-configuration-new.hbs", await JiraCloudAndEnterpriseConfiguration(res, req));
+		} else {
+			res.render("jira-configuration.hbs", await JiraCloudConfiguration(res, req));
+		}
```
mboudreau, line 185: much better :+1:

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -67,23 +86,89 @@ const getInstallation = async (subscription: Subscription, log: Logger, gitHubAp
...
+	return {
+		host: jiraHost,
+		successfulConnections,
+		failedConnections,
+		hasConnections: !!installations.total,
+		APP_URL: process.env.APP_URL,
+		csrfToken: req.csrfToken(),
+		nonce,
+		handleNavigationClassName: "add-organization-link",
+		isGitHubCloudApp: await isGitHubCloudApp(gitHubAppId)
```
mboudreau, line 126: Shouldn't this be true by default for this function?

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -67,23 +86,89 @@ const getInstallation = async (subscription: Subscription, log: Logger, gitHubAp
...
+			isGlobalInstall: installation.repository_selection === "all"
+		}));
+
+	return { installations, successfulConnections, failedConnections };
+};
 
-interface SuccessfulConnection extends AppInstallation {
-	isGlobalInstall: boolean;
-}
+const JiraCloudConfiguration = async (res: Response, req: Request): Promise<object> => {
```
mboudreau, line 112: `Promise<object>` return type is not acceptable.

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -67,23 +86,89 @@ const getInstallation = async (subscription: Subscription, log: Logger, gitHubAp
...
+	const { jiraHost, gitHubAppId, nonce } = res.locals;
```
mboudreau, line 113: why do you need the githubAppId for the jira cloud configuration only function?

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -67,23 +86,89 @@ const getInstallation = async (subscription: Subscription, log: Logger, gitHubAp
...
+				id: installation.id,
+				deleted: installation.deleted,
+				orgName: repo?.repoOwner
+			};
+		}));
+
+	const successfulConnections: SuccessfulConnection[] = installations.fulfilled
+		.map((installation) => ({
+			...installation,
+			isGlobalInstall: installation.repository_selection === "all"
```
mboudreau, line 106: I'm still not sure what a 'global install' is

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -67,23 +86,89 @@ const getInstallation = async (subscription: Subscription, log: Logger, gitHubAp
 	}
 };
 
-interface FailedConnection {
-	id: number;
-	deleted: boolean;
-	orgName?: string;
-}
+const getConnectionsAndInstallations = async (subscriptions: Subscription[], req: Request, id: number): Promise<ConnectionsAndInstallations> => {
```
mboudreau, line 89: what's `id`?  I'm fairly sure it's optional as well

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -67,23 +86,89 @@ const getInstallation = async (subscription: Subscription, log: Logger, gitHubAp
...
+		hasConnections: !!installations.total,
+		APP_URL: process.env.APP_URL,
+		csrfToken: req.csrfToken(),
+		nonce,
+		handleNavigationClassName: "add-organization-link",
+		isGitHubCloudApp: await isGitHubCloudApp(gitHubAppId)
+	};
+};
+
+const JiraCloudAndEnterpriseConfiguration = async (res: Response, req: Request): Promise<object> => {
```
mboudreau, line 130: better return typing

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -67,23 +86,89 @@ const getInstallation = async (subscription: Subscription, log: Logger, gitHubAp
...
+	const { jiraHost, gitHubAppId, nonce } = res.locals;
+	const subscriptions = await Subscription.getAllForHost(jiraHost);
+	const { installations, successfulConnections, failedConnections } = await getConnectionsAndInstallations(subscriptions, req, gitHubAppId);
+	const gheServers = await GitHubServerApp.findForInstallationId(res.locals.installation.id) as GitHubAppServerAppWithConnections[];
```
mboudreau, line 134: `as GitHubAppServerAppWithConnections[];` doesn't make sense, you're just getting the `GitHubServerApp` model in an array with no other data that matches that type

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -67,23 +86,89 @@ const getInstallation = async (subscription: Subscription, log: Logger, gitHubAp
...
+				const {
+					successfulConnections,
+					failedConnections
+				} = await getConnectionsAndInstallations(subscriptions, req, app.id);
+				app.successfulConnections = successfulConnections;
+				app.failedConnections = failedConnections;
+
+				return app;
+			}))
+		})).value());
```
mboudreau, line 151: This is inefficient and might timeout or fail outright on large customers.  Instead of doing a `Promise.all` within another `Promise.all` you should just get all subscriptions based on GithubServerApps first, then modify those which will drastically reduce your DB connections and won't overload it if the customer has multiple apps with subscriptions in it.

```
const subscriptions = Subscriptions.find({
   where: {
      githubAppId: {
         [Op.in]: gheServers.map(s => s.id)
      }
   }
});
```

This is still not perfect, but it's reduced your DB calls to 2 only (O^2), one for the GHE app table, and another to the subscription table, where before you had O^n+x.

This could be reduced further to a single call once we set the DB schema correctly with foreign keys.

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: This has been removed

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: Type `object` replaced by a new interface `ViewConfiguration`

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: This has been made optional and removed

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: This has been made optional.

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: Type object replaced by a new interface ViewConfiguration

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -67,23 +86,89 @@ const getInstallation = async (subscription: Subscription, log: Logger, gitHubAp
...
+
+	return {
+		host: jiraHost,
+		gheServers: modifiedGHEServers,
+		ghCloud: { successfulConnections, failedConnections },
+		hasConnections: !!(installations.total || gheServers?.length),
+		APP_URL: process.env.APP_URL,
+		csrfToken: req.csrfToken(),
+		nonce,
+		handleNavigationClassName: "select-github-version-link"
```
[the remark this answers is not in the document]
krazziekay: This has been removed

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: The loop within the loop has been separated into 2 exact loops. The times complexity should no longer occur.

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: Type updated

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: This part of the code was just moved and not changed.

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -93,42 +200,11 @@ export const JiraConfigurationGet = async (
...
-			failedConnections,
-			hasConnections: !!installations.total,
-			APP_URL: process.env.APP_URL,
-			csrfToken: req.csrfToken(),
-			nonce: res.locals.nonce,
-			handleNavigationClassName,
-			isGitHubCloudApp: await isGitHubCloudApp(gitHubAppId)
-		});
-
+		if (await booleanFlag(BooleanFlags.GHE_SERVER, true, jiraHost)) {
```
rachellerathbone, line 203: You need to flip this conditional. The second arg being passed here should always be false and rendering the current view. The new view should be rendered in the else block.

krazziekay: Ah yes, forgot to revert this.

rachellerathbone: No worries. I've been guilty of it myself from time to time :badpokerface:

rachellerathbone:
Found a bug. I added an entry to the GHSA table. I was expecting to see the server and gh app with no orgs (as none are connected) but it's rendering the connected orgs from my cloud account.

Screen Shot 2022-07-13 at 2 00 36 pm

krazziekay:
> Found a bug. I added an entry to the GHSA table. I was expecting to see the server and gh app with no orgs (as none are connected) but it's rendering the connected orgs from my cloud account.
> 
> Screen Shot 2022-07-13 at 2 00 36 pm

I don't think its a bug. If you are testing this locally, can you please confirm whether you've enabled the feature flag `GHE_SERVER` in these places:
- `get-github-client-config.ts#createAppClient()`: _line70_
-  `jira-configuration-get.ts#JiraConfigurationGet()`: _line203_

rachellerathbone:
Just turned on the additional flag and firstly couldn't even load the page:

https://user-images.githubusercontent.com/37155488/178658576-1a8a0585-5789-4fb4-b08f-e7abbe1220a1.mov

After a few minutes it finally loaded but now I'm seeing this.

Screen Shot 2022-07-13 at 3 40 01 pm

views/partials/jira-configuration-table.hbs
```diff
@@ -0,0 +1,185 @@
...
+      </tr>
+      </thead>
+
+      <tbody class="jiraConfiguration__table__body">
+      {{#each successfulConnections as |connection|}}
+        <tr class="jiraConfiguration__table__row">
+          {{! Organization }}
+          <td class="jiraConfiguration__table__cell">
+            <img
+                class="jiraConfiguration__table__cell__avatar"
```
rachellerathbone, line 24: The avatar needs to be hidden for all server accounts. We have to do an investigation into our content security policy this week to solve an issue with the auto app creation. Changes we make may end up resolving issues we're facing with GHE avatars but, for now, I say we just hide them.

krazziekay: Avatars hidden for GHE servers

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -8,7 +8,58 @@ import { metricError } from "config/metric-names";
 import { AppInstallation, FailedAppInstallation } from "config/interfaces";
 import { createAppClient } from "~/src/util/get-github-client-config";
 import { booleanFlag, BooleanFlags } from "config/feature-flags";
-import { isGitHubCloudApp } from "~/src/util/jira-utils";
+import { GitHubServerApp } from "models/github-server-app";
+
+interface FailedConnection {
```
rachellerathbone, line 13: I know these were here before but just for future reference: if you want to move something because it makes more sense for the 'flow' that's fine, but please do it in a separate PR. It makes it very hard to tell what is original code and what is new code when everything has been moved around.

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -8,9 +8,60 @@ import { metricError } from "config/metric-names";
...
+
+interface SuccessfulConnection extends AppInstallation {
+	isGlobalInstall: boolean;
+}
+
+interface ConnectionsAndInstallations {
+	successfulConnections: SuccessfulConnection[]
+	failedConnections: FailedConnection[]
+	installations: InstallationResults
+}
```
mboudreau, line 34: Don't duplicate interfaces if there's no need.
```
interface ConnectionsAndInstallations extends GitHubCloudObj {
	installations: InstallationResults
}
```

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -8,9 +8,60 @@ import { metricError } from "config/metric-names";
...
+interface ViewConfigurationForGHE {
+	host: string
+	gheServers: GitHubServerObj[]
+	ghCloud: GitHubCloudObj
+	hasConnections: boolean
+	APP_URL?: string
+	csrfToken: string
+	nonce: string
+}
```
mboudreau, line 63: there's similarities between these 2 that should be it's own interface

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -94,42 +209,11 @@ export const JiraConfigurationGet = async (
...
-			nonce: res.locals.nonce,
-			handleNavigationClassName,
-			isGitHubCloudApp: await isGitHubCloudApp(gitHubAppId)
-		});
-
+		if (await booleanFlag(BooleanFlags.GHE_SERVER, false, jiraHost)) {
+			res.render("jira-configuration-new.hbs", await JiraCloudAndEnterpriseConfiguration(res, req));
+		} else {
+			res.render("jira-configuration.hbs", await JiraCloudConfiguration(res, req));
+		}
```
mboudreau, line 216: this is a weird pattern.  The `res.render` should be within each function, not outside of it.  All this conditional is doing is a custom routing based on an FF, that's it.  

A good test of this is if I were to create a new route (let's say `/jira/configuration/new` and add `JiraCloudAndEnterpriseConfiguration` to be the handler, it should render like it should by itself.  Right now, I don't think it would do anything.

mboudreau: also, btw, adding that path might not be a bad idea to test stuff in prod without having the FF turned on :)

src/util/handlebars/handlebar-helpers.ts
```diff
@@ -4,9 +4,13 @@ import { isPlainObject } from "lodash";
 export const concatStringHelper = (...strings: string[]) => strings.filter((arg: unknown) => typeof arg !== "object").join(" ");
 export const toLowercaseHelper = (str?: string) => !isPlainObject(str) && str?.toString?.().toLowerCase() || "";
 export const replaceSpaceWithHyphenHelper = (str?: string) => !isPlainObject(str) && str?.toString?.().replace(/ /g, "-") || "";
+export const logicalOrHelper = (...args: unknown[]) => args.filter((arg: unknown) => typeof arg !== "object").some(Boolean);
+export const logicalAndHelper = (...args: unknown[]) => args.filter((arg: unknown) => typeof arg !== "object").every(Boolean);
```
mboudreau, line 8: I'm not sure how I feel about these.  I understand what you're trying to do, but you're now adding complex logic in handlebars with no compilation errors, which is what we're trying to prevent with typescript.

I think this will encourage bad coding practices in handlebar files.  Could we please remove them and instead, if you need to do complex logic, do it beforehand hand and pass it in as a config property to the template.  If you type it correctly, we'll be able to catch issues before they're a problem :)

mboudreau (changes requested):
I few small things to change but well done for the most part.  This has been tested thoroughly on staging I imagine?

rachellerathbone:
> I few small things to change but well done for the most part. This has been tested thoroughly on staging I imagine?
@mboudreau until we've completed the auto app flow (should hopefully be in the next couple of days) there's no way for us to test as there GHSA table is currently empty. I think we should aim to get this PR ready to be merged but maybe hold off on doing that until a flow is complete and it can be tested in stg.

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: I'm not sure what you mean by `each function`. 
As discussed earlier in this PR, the FF separates the existing view with the new one. And I do think using FF is a better approach than using a new separate route. That way we can turn on the new flow for any _Jira instance_ by turning on the FF, without changing any code.

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: Interface corrected

src/routes/jira/configuration/jira-configuration-get.ts
krazziekay: Interface corrected

src/util/handlebars/handlebar-helpers.ts
krazziekay: This helpers have been removed and the values needed have been passed within the configuration for the view.

src/routes/jira/configuration/jira-configuration-get.ts
rachellerathbone: I think what @mboudreau is saying is that you should be calling res.render from within JiraCloudAndEnterpriseConfiguration and JiraCloudConfiguration. You can still use feature flags to do this. For example, instead of having `res.render("jira-configuration-new.hbs", await JiraCloudAndEnterpriseConfiguration(res, req))` inside the FF check just call the function `await JiraCloudAndEnterpriseConfiguration(res, req)` (although I'd probably rename it to something like `jiraCloudAndEnterpriseServer` or ` renderJiraCloudAndEnterpriseServerView`) then inside that function have 

```
	res.render("jira-configuration-new.hbs", {
		host: jiraHost,
		gheServers: groupedGheServers,
		ghCloud: { successfulConnections, failedConnections },
		hasCloudAndEnterpriseServers: !!((successfulConnections.length || failedConnections.length) && gheServers.length),
		hasCloudServers: !!(successfulConnections.length || failedConnections.length),
		hasConnections: !!(installations.total || gheServers?.length),
		APP_URL: process.env.APP_URL,
		csrfToken: req.csrfToken(),
		nonce
	});
```

and the conditional check for the FF could be simplified to something like:
```
await booleanFlag(BooleanFlags.GHE_SERVER, false, jiraHost)
			? renderJiraCloudAndEnterpriseServerView(res, req)
			: renderJiraCloudView(res, req);
```

krazziekay: Updated

mboudreau (approved):

rachellerathbone (approved):

PR #1320, opened 2022-07-01
ARC-1350: update headers to be httpOnly

gxueatlassian:
**Reason:**
We put our github app token in the session cookie, only signed but not encrypted. So in case our app exposed to XSS, malicious scripts will be able steal user's github app token and pull all intellectual property from user's github account. 

**Solutions**
By changing to httpOnly, under the case of XSS, malicious scripts won't be able to access the session cookie. 

**Caveats**
Not saying we have XSS vlun in our code, nor we should use cookie or storing those tokens  and transfer it over the internet. But this step is just a bit better in terms of protection.


joshkay10 (approved):

mboudreau:
Can we make sure that this will only affect the session cookie?  We have a dependency on javascript cookies to set the `jwt` and `jiraHost` for the session transfer between the iframe and our new tab when connecting a new github org (the `/session` endpoint).  Pretty sure it should be fine though.

gxueatlassian:
> Can we make sure that this will only affect the session cookie? We have a dependency on javascript cookies to set the `jwt` and `jiraHost` for the session transfer between the iframe and our new tab when connecting a new github org (the `/session` endpoint). Pretty sure it should be fine though.

Yup, this `cookieSessionMiddleware` only affect the cookie named `session` and `session.sig`. These two cookies are used by the dependency `cookie-session`.

mboudreau (approved):

PR #1312, opened 2022-06-29
Arc-1455 server url to app creation

rachellerathbone:
**Rationale:**
Continuing task of wiring up the UI screens. This PR focuses on navigating from the GHE server URL screen to the app creation screen.

**Changes:**

- Added 2 new routes: `JiraVerifyServerUrlRouter` to handle routing a post request (`JiraVerifyServerUrlPost`) to validate the passed url is a valid url, check the db for existing apps and verify we can hit the provided url. `JiraAppCreationRouter` handles the get request after the validation passes.
- New `JiraVerifyServerUrlPost` middleware which takes in the provided url and makes a simple get request to check that we ping the provided url. If customers have opened a hole in their firewall, we should get a successful response back, otherwise, it will fail.
- From here, we either navigate to the app creation page or render the list of GH apps (this part of the flow is yet to be implemented). https://github.com/atlassian/github-for-jira/pull/1312/files#r909455400
- New module in connect and updated snapshots of the connect object.
- Modifications to static/js/jira-server-url.js to handle making requests to our new route. 
- What's not included: styling clean-up. There's a noticeable size difference in the containers between these 2 views but we can clean this up later on.

**Demo:**
Like the last 2 PRs, the new path is hidden behind a FF.

With FF set to false we still navigate from the config screen to the connect an org screen:

https://user-images.githubusercontent.com/37155488/176413772-25add353-3d21-4717-bd71-3408dc6b4368.mov

With the FF set to true (done manually on my local for demo purposes only) we can navigate from the config page to the server url screen. In the scenario where I have 'opened a hole in the firewall' (aka connected to the VPN) and I have no existing GH apps, I'm navigated to the auto/manual app creation page.

https://user-images.githubusercontent.com/37155488/176414370-83b638d5-9f3e-4e7f-aee3-b83d1e7a4210.mov

If I do have existing apps, instead of being navigated to /github-app-creation-page I'll be navigated to `github-list-apps-page` (the view and the module for this don't exist yet so it renders nothing and the error flag pops up but not the path)

https://user-images.githubusercontent.com/37155488/176415235-6f55c824-d179-4e65-8a53-2c314d5df7b6.mov

If we can't verify the url, we'll render an error ([this component hasn't been created yet](https://github.com/atlassian/github-for-jira/pull/1312/files#r909332712) but note the response in my terminal ). For demo purposes, I'm disconnected from my VPN for this one.

[...]

PR #1328, opened 2022-07-04
adding url validation util

rachellerathbone:
Rationale: we need to validate the ghe url on the server side. Once again, created a baby pr for this to keep #1312 on the smaller side.

### DIFF
--- a/src/util/is-valid-url-test.ts
+++ b/src/util/is-valid-url-test.ts
@@ -1,0 +1,14 @@
+import { isValidUrl } from "utils/is-valid-url";
+
+describe("Is valid url", () => {
+	it("should return true if valid url is provided", async () => {
+		expect(isValidUrl("http://myvalidurl.com")).toBeTruthy();
+		expect(isValidUrl("https://myvalidurl.net")).toBeTruthy();
+	});
+
+	it("should return false if invalid url is provided", async () => {
+		expect(isValidUrl("noprotocolortopleveldomain")).toBeFalsy();
+		expect(isValidUrl("https://www.ihavenottopleveldomain")).toBeFalsy();
+		expect(isValidUrl("www.wheresmyprotocol.com")).toBeFalsy();
+	});
+});
--- a/src/util/is-valid-url.ts
+++ b/src/util/is-valid-url.ts
@@ -1,0 +1,10 @@
+const ALLOWED_PROTOCOLS = ["http:", "https:"];
+
+export const isValidUrl = (url: string): boolean => {
+	try {
+		const { protocol, hostname } = new URL(url);
+		return !(!ALLOWED_PROTOCOLS.includes(protocol) || hostname.split(".").length < 2);
+	} catch (err) {
+		return false;
+	}
+};
--- a/src/util/is-valid-url.ts
+++ b/src/util/is-valid-url.ts
@@ -1,9 +1,7 @@
-const ALLOWED_PROTOCOLS = ["http:", "https:"];
-
 export const isValidUrl = (url: string): boolean => {
 	try {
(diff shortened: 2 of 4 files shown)


krazziekay (approved):
The validation method done in `jira-server-url` will need to be replaced by this util method.

src/util/is-valid-url-test.ts
```diff
@@ -0,0 +1,14 @@
+import { isValidUrl } from "utils/is-valid-url";
```
joshkay10, line 1: `src/util/is-valid-url-test.ts` needs to be `src/util/is-valid-url.test.ts` with a `.test` to be picked up by jest

src/util/is-valid-url-test.ts
```diff
@@ -0,0 +1,14 @@
...
+
+describe("Is valid url", () => {
+	it("should return true if valid url is provided", async () => {
+		expect(isValidUrl("http://myvalidurl.com")).toBeTruthy();
+		expect(isValidUrl("https://myvalidurl.net")).toBeTruthy();
+	});
+
+	it("should return false if invalid url is provided", async () => {
+		expect(isValidUrl("noprotocolortopleveldomain")).toBeFalsy();
+		expect(isValidUrl("https://www.ihavenottopleveldomain")).toBeFalsy();
```
joshkay10, line 11: the `www.` will cause the `hostname.split(".").length < 2)` to be tru :(

static/js/jira-server-url.js
```diff
@@ -22,15 +22,11 @@ const checkValidGHEUrl = inputURL => {
 	try {
 		const { protocol, hostname } = new URL(inputURL);
 
-		if (!ALLOWED_PROTOCOLS.includes(protocol)) {
-			setErrorMessage(defaultError);
-			return false;
-		}
-		// This checks whether the hostname whether there is an extension like `.com`, `.net` etc.
-		if (hostname.split('.').length < 2) {
```
rachellerathbone, line 30: Removed this entirely because it doesn't actually validate if there is a TLD.

Uploading Screen Recording 2022-07-04 at 2.06.56 pm.mov…

static/js/jira-server-url.js
```diff
@@ -22,15 +22,11 @@ const checkValidGHEUrl = inputURL => {
...
+		if (!/^https?:$/.test(protocol)) {
```
rachellerathbone, line 25: Changed this so the validation is the same as the backend and tested with the same inputs from the test. 
Screen Shot 2022-07-04 at 2 06 47 pm
4-0b8e-46e1-bda1-214c3ed0a465.png">
Screen Shot 2022-07-04 at 2 06 38 pm

src/util/is-valid-url-test.ts
rachellerathbone: already uopdated

src/util/is-valid-url-test.ts
rachellerathbone: Realised that. Copied the logic from what was implemented on the frontend but realised it doesn't work.

joshkay10 (approved):
I like this lots more :)

gxueatlassian (approved):

src/util/is-valid-url.ts
```diff
@@ -0,0 +1,8 @@
+export const isValidUrl = (url: string): boolean => {
+	try {
+		const { protocol } = new URL(url);
+		return (/^https?:$/.test(protocol));
```
joshkay10, line 4: post approval comment 😆 
could pass isValid with this "http://myvalidurl"   ¯\_(ツ)_/¯

PR #1329, opened 2022-07-04
ARC-1369 ugc clean

joshkay10:
**What's in this PR?**
Removed any payload logs, useful but potentially containing UGC
Moved UGC data from the log message to the log object so that UGC can be hashed with the serializer.

**Why**
UGC in splunk should be a HOT, so gotsta get it outta here.

**Added feature flags**
No FF required.

**Affected issues**  
[ARC-1369](https://softwareteams.atlassian.net/browse/ARC-1369) - Review logs for UGC/PII and remove or hash

**How has this been tested?**  
Tested on stage through splunk inspecting 

**Whats Next?**
Continue to monitor on Splunk, see if I can find any UGC/PII on there and then clean up accordingly.

Move the unsafe logger to the logging stream, so we can toggle the hashing per customer.


src/github/pull-request.ts
```diff
@@ -84,7 +83,7 @@ export const pullRequestWebhookHandler = async (context: WebhookContext, jiraCli
 		return;
 	}
 
-	context.log.info(`Sending pull request update to Jira ${baseUrl}`);
+	context.log.info({ jiraHost : baseUrl }, `Sending pull request update to Jira`);
```
gxueatlassian, line 86: can you plz tell me why move it up front?

gxueatlassian (approved):

src/github/pull-request.ts
joshkay10: Moved it out of the template string and into to the optional fields object. The entries in fields object will be run through the sensitive data serializer, so it will hash jiraHost for us.
Also, its way way way easier on splunk to read if it has the values as attributes outside of the msg.

gxueatlassian (approved):

PR #1331, opened 2022-07-05
ARC-1475 - Create a view for creating Apps manually

krazziekay:
**What's in this PR?**
- This PR includes the form for adding GHE server apps.

**Why**
- To allow the users to add GHE server apps manually.

**Assumptions**
- This view will require `githubBaseUrl` and `uuid` to be passed

**Screenshot**
image

**Video**
- https://www.loom.com/share/333a0f1183a14b37b0befb5a0a3916f2


static/js/jira-manual-app-creation.js
```diff
@@ -0,0 +1,28 @@
...
+  // TODO: Form submission
+});
+
+$(".jiraManualAppCreation__formFileInput")
+  .on("dragenter focus", () => {
+    $(".jiraManualAppCreation__formFileDropArea").addClass("active");
+  }).on("dragleave blur drop", () => {
+    $(".jiraManualAppCreation__formFileDropArea").removeClass("active");
+  }).on("change", (event) => {
+    const fileName = $(event.target).val().split('\\').pop();
```
gxueatlassian, line 13: This could be problematic on non-windows platform. Anyway there's a cleaner way to do this, use the new file object. You can even get the binary date off that object. 

https://developer.mozilla.org/en-US/docs/Web/API/File

static/js/jira-manual-app-creation.js
```diff
@@ -0,0 +1,28 @@
+$("#jiraManualAppCreation__form").submit((event) => {
+  event.preventDefault();
+
+  // TODO: Form submission
```
rachellerathbone, line 4: I would expect the frontend validation to be a part of this PR but happy for you to tackle it separately if you prefer. We obviously won't be making a request to the backend to save the data in the db but we should at least take the data in and make sure all required fields (which is all of them) have been provided and render the `this is a required field message` for any empty fields.

views/jira-manual-app-creation.hbs
```diff
@@ -0,0 +1,142 @@
...
+        <div class="field-group">
+          <label>Private key</label>
+          <div class="jiraManualAppCreation__formFileDropArea">
+            <img class="jiraManualAppCreation__formFileIcon" src="/public/assets/file-upload.svg" alt="Upload">
+            <div>
+              <div class="jiraManualAppCreation__formFileMsg">Upload your private key</div>
+              <div class="jiraManualAppCreation__formFileSubMsg">Upload your generated private key from GitHub</div>
+            </div>
+            <div class="aui-button">Upload</div>
+            <input class="jiraManualAppCreation__formFileInput" id="private-key" required type="file">
```
rachellerathbone, line 116: Can we get a screenshot/recording of what this looks like when a file has been uploaded?

views/jira-manual-app-creation.hbs
```diff
@@ -0,0 +1,142 @@
...
+        <div class="jiraManualAppCreation__formTitle">URLs</div>
+        <div class="jiraManualAppCreation__formSubTitle">
+          Copy and paste the following URLs into your GitHub Enterprise Server configuration. This is unique to your site.
+        </div>
+
+        <div class="field-group">
+          <label for="callback-url">Callback URL</label>
+          <!--TODO: replace the value by actual data-->
+          <input class="text full-width-field" type="text" id="callback-url"
+                 name="callback-url" value="https://github.atlassian.com/github/callback/V1StGXR8_Z5jdHi6B-myT"
```
rachellerathbone, line 69: These IDs need to be programmatically generated. This is another thing you can choose to tackle as part of this PR or separately.

static/js/jira-manual-app-creation.js
krazziekay: The form does include a `required` validation, using the default HTML `required` attribute. Do we need to do other validations too?

static/js/jira-manual-app-creation.js
krazziekay: Used `event.target.files[0].name`

views/jira-manual-app-creation.hbs
```diff
@@ -0,0 +1,142 @@
...
+            <div class="aui-button">Upload</div>
+            <input class="jiraManualAppCreation__formFileInput" id="private-key" required type="file">
+          </div>
+          <div class="jiraManualAppCreation__formFileUploaded">
+            <div id="jiraManualAppCreation__uploadedFile"></div>
+            <span class="aui-icon aui-icon-small aui-iconfont-cross" id="jiraManualAppCreation__clearUploadedFile">Remove</span>
+          </div>
+        </div>
+
+        <button class="jiraManualAppCreation_formSubmitBtn aui-button aui-button-primary">Next</button>
```
rachellerathbone, line 124: Where is the error component? It's included in the design so we should definitely add it to this pr.

views/jira-manual-app-creation.hbs
krazziekay: I thought this would be generated and supplied from the backend. If we do it in the frontend, how do we generate the link?

views/jira-manual-app-creation.hbs
krazziekay: Video has been added in the description

views/jira-manual-app-creation.hbs
krazziekay: Error component has been added

static/css/jira-manual-app-creation.css
```diff
@@ -0,0 +1,105 @@
...
+
+.jiraManualAppCreation__formSubTitle {
+  font-size: 1rem;
+  font-weight: 400;
+  margin: 1.25em 0;
+  color: #344563;
+}
+
+.field-group {
+  margin-bottom: 1.1em !important;
```
rachellerathbone, line 33: Do we need !important here? Can we apply this style with specificity?

views/jira-manual-app-creation.hbs
rachellerathbone: Yes, you have added `required` as a attribute in the form but you are not handling the frontend validation if the form is submitted and those fields are empty,

Screen Shot 2022-07-05 at 12 26 24 pm

There is also this frontend validation.

Screen Shot 2022-07-05 at 12 26 41 pm

static/css/jira-manual-app-creation.css
krazziekay: Replaced important using specificity

static/js/jira-manual-app-creation.js
rachellerathbone: Sorry, realised I replied in the wrong thread. https://github.com/atlassian/github-for-jira/pull/1331/files#r913338483

You can't just rely on the browser to handle this because the styling won't align with design. First of all, it looks nothing like what we have in Figma

Screen Shot 2022-07-05 at 4 35 54 pm

and secondly, it doesn't flag **all** empty fields as required at the same time. If I miss multiple it will flag the first empty one. I then fill then in and when I go to submit it flags the next one. This is a pretty frustrating user experience.

Screen Shot 2022-07-05 at 4 36 03 pm

It also behaves differently in each browser. In the video below, watch what happens in Safari. I click next and see the message at the top but then it's gone when I scroll up. Sure, the input field is highlighted but what if I didn't read the message?

https://user-images.githubusercontent.com/37155488/177266409-3f3a2a07-f7cc-4a2f-95c8-0b8b91198766.mov

[ ] Ditch `required` from all the input fields and handle the frontend check for that yourself (like you did on the server app url screen)

views/jira-manual-app-creation.hbs
```diff
@@ -0,0 +1,158 @@
...
+          <input class="text full-width-field" type="text" id="setup-url"
+                 name="setup-url" value="{{githubBaseUrl}}github/setup/{{uuid}}"
+                 readonly
+          />
+        </div>
+
+        <div class="field-group">
+          <label for="webhook-url">Webhook URL</label>
+          <input class="text full-width-field" type="text" id="webhook-url"
+                 name="webhook-url" value="{{githubBaseUrl}}github/webhook/{{uuid}}"
```
rachellerathbone, line 91: Urls won't be passed or stored with the trailing `/` so it needs to be included here after `{{githubBaseUrl}}`

rachellerathbone (changes requested):

static/js/jira-manual-app-creation.js
krazziekay: Used [AUI Form validation](https://aui.atlassian.com/aui/latest/docs/form-validation.html) to match the design.

views/jira-manual-app-creation.hbs
krazziekay: `/` has been added

views/jira-manual-app-creation.hbs
```diff
@@ -0,0 +1,169 @@
...
+        <div class="field-group">
+          <label>Private key</label>
+          <div class="jiraManualAppCreation__formFileDropArea">
+            <!--
+              Regex pattern for empty string or strings ending with these file extensions.
+              The empty string pattern is added to avoid showing validation pattern error message when this field is empty.
+              File extensions are referenced from https://www.microfocus.com/documentation/visual-cobol/vc60/VS2017/BKCJCJCERTS001.html
+            -->
+            <input class="jiraManualAppCreation__formFileInput" id="private-key"
+                   data-aui-validation-pattern="^\s*$|^.*\.(der|DER|cer|CER|pem|PEM|spc|SPC|p7a|P7A|p7b|P7B|p7c|P7C|p8|P8|pfx|PFX|p12|P12)$"
```
rachellerathbone, line 124: Where did this list of file extensions come from? How does this work? From my experience creating GH apps, the extension is always .pem so I don't think most of those are necessary. 

Also, have you tested this? We should render an error message like 'Incorrect file type' or something when a file with the wrong extension is added.

krazziekay: I wasn't sure of the different types of files for this. So, searched and found a list of private keys file types here: https://www.microfocus.com/documentation/visual-cobol/vc60/VS2017/BKCJCJCERTS001.html

Are we sure that its always gonna be just `.pem` files? If so, then can change the regex.

And as for the error message, I'm not sure if we need a specific message. The design shows this error message for wrong file types.
image

rachellerathbone: Can I get you to add a new module for this page under `generalPages`? Prior to us figuring out how to add new pages in Jira is made sense to test out new pages by hijacking the render of another page but now that's implemented, we should be adding new pages in connect when they are created.

static/js/jira-manual-app-creation.js
rachellerathbone: We have a ticket to implement manual app creation on the backend which will include adding the ajax POST request, which will account for the `// TODO: Form submission` you have here. However, I would expect the payload that needs to be passed to be done as part of this PR. You should be pulling in all the data from the form and the object that will be passed to the backend.

krazziekay: The data has been added.

views/jira-manual-app-creation.hbs
rachellerathbone: Pretty sure it's only .pem but you could do some investigation into private keys in GH apps to make sure.

krazziekay: I'll just update the regex to just `pem`then...

rachellerathbone (approved):

gxueatlassian (approved):

PR #1332, opened 2022-07-05
ARC-1462 Added check to make sure pipeline is not running before triggering new one

Harminder84:
**What's in this PR?**
- Added condition to make sure not other pipeline is running  before trigerring new pipeline.
- Updated cron expression to run auto deployment lambda in every 10 mins

**Why**
We want to run auto-deployment lambda in every 10 mins and starts deployment if any change in master branch. The pipeline takes ~40 mins to deploy to prod. If auto-deployment lambda again runs in that timeframe, it will trigger the pipeline again. This will result in muti pipelines running for same commit in master branch.

**Added feature flags**
NA

**Affected issues**  
NA

**How has this been tested?**  
This is running only in Prod. By looking into lamda invocations timestamp and logs.

**Whats Next?**
NA

### DIFF
--- a/deploy/lambda/auto-deployment.js
+++ b/deploy/lambda/auto-deployment.js
@@ -1,5 +1,7 @@
 const https = require('https');
 
+const pipelinesURL = 'https://api.bitbucket.org/2.0/repositories/atlassian/github-for-jira-deployment/pipelines/';
+
 exports.handler = async function (event, context) {
     const autoDeployUsername = process.env.AUTO_DEPLOY_USERNAME;
     const autoDeployToken = process.env.AUTO_DEPLOY_TOKEN;
@@ -27,7 +29,6 @@ exports.handler = async function (event, context) {
         if (deployedCommitSHA !== mainCommitSHA && ! await isPipelineRunning()) {
             console.log('Changes found, starting deployment...');
 
-            const pipelinesURL = 'https://api.bitbucket.org/2.0/repositories/atlassian/github-for-jira-deployment/pipelines/';
             const body = {
                 "target": {
                     "type": "pipeline_ref_target",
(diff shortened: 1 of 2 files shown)


deploy/lambda/auto-deployment.js
```diff
@@ -33,33 +33,50 @@ exports.handler = async function (event, context) {
...
         console.error('Error: ', e);
     }
 };
 
-function getRequest(url) {
+// Returns true if any pipline is running for deployment to stage/prod
+async function isPipelineRunning() {
+    const url = "https://api.bitbucket.org/2.0/repositories/atlassian/github-for-jira-deployment/pipelines/?page=1&pagelen=20&sort=-created_on"
+    const pipelines = await getRequest(url, {
+        'Authorization': 'Basic ' + new Buffer('fusion-release-bot:2nXHekMCpxc3T8rGPKwP').toString('base64')
```
krazziekay, line 62: The `username` and `password` here needs to be saved in Microscope. This will trigger a vulnerability issue.

deploy/lambda/auto-deployment.js
```diff
@@ -33,33 +33,50 @@ exports.handler = async function (event, context) {
...
         }
     } catch (e) {
         console.error('Error: ', e);
     }
 };
 
-function getRequest(url) {
+// Returns true if any pipline is running for deployment to stage/prod
+async function isPipelineRunning() {
+    const url = "https://api.bitbucket.org/2.0/repositories/atlassian/github-for-jira-deployment/pipelines/?page=1&pagelen=20&sort=-created_on"
```
krazziekay, line 60: Better to separate out the url `https://api.bitbucket.org/2.0/repositories/atlassian/github-for-jira-deployment/pipelines/` as its being used multiple times.

krazziekay (changes requested):

deploy/lambda/auto-deployment.js
Harminder84: > The `username` and `password` here needs to be saved in Microscope. This will trigger a vulnerability issue.

Thanks kay, update it. I put it for local testing.

krazziekay (approved):

PR #1334, opened 2022-07-05
Arc 1462 update lambda

Harminder84:
**What's in this PR?**
Code Refacored to fix variable undefined issue

**Why**
Fix variable undefined issue

**Added feature flags**
NA

**Affected issues**  
NA

**How has this been tested?**  
This is running only in Prod. By looking into lamda invocations timestamp and logs.

**Whats Next?**
NA


deploy/lambda/auto-deployment.js
```diff
@@ -119,3 +137,5 @@ function postRequest(url, body, headers = {}) {
         req.end();
     });
 }
+
+isPipelineRunning()
```
krazziekay, line 141: Why does this function being called here? its already being called in line #29

PR #1337, opened 2022-07-05
Arc 1354 use cryptor for shared secret

gxueatlassian:
**What's in this PR?**

Make changes to code so that we save `sharedSecret` in DB using Cryptor

**Why**

According to Atlassian policy, we need to use Cryptor for encryption.

**Added feature flags**
N/A

**Affected issues**  
ARC-1354

**How has this been tested?**  
[Work in progress]

**Whats Next?**
1. DB migration for old secrets into new secrets column with Cryptor (and point sharedSecret property to new column.


src/models/installation.ts
```diff
@@ -1,22 +1,37 @@
...
 	throw new Error("STORAGE_SECRET is not defined.");
 }
 
-export class Installation extends Model {
+export class Installation extends EncryptedModel {
 	id: number;
 	jiraHost: string;
 	secrets: string;
 	sharedSecret: string;
+	safeSharedSecret: string; //LOL
```
gxueatlassian, line 17: 1. For gradually rollout / migration purpose, favor having a new column instead of reuse existing
2. Like in `1`, favor explicit new property name instead of some magic using same prop name but different db column etc.

src/models/installation.ts
```diff
@@ -1,22 +1,37 @@
...
 	updatedAt: Date;
 	createdAt: Date;
 	githubAppId?: number;
 
+	getEncryptionSecretKey() {
+		return EncryptionSecretKeyEnum.GITHUB_SERVER_APP;
+	}
+
+	async getEncryptContext() {
+		return { clientKey: this.clientKey };
```
gxueatlassian, line 28: This make sense, right?

src/models/installation.ts
```diff
@@ -127,11 +147,26 @@ Installation.init({
 		type: DataTypes.INTEGER,
 		allowNull: true
 	}
-}, { sequelize });
+}, {
+	hooks: {
+		beforeSave: async (app: Installation, opts) => {
+			await app.encryptChangedSecretFields(opts.fields);
+			if (!app["safeSharedSecret"]) throw new Error("Fail saving app as safeSharedSecret is empty");
```
gxueatlassian, line 154: I can't change `allowNull: true` to `allowNull: false` for this new column. I suspect it will be find, but worry sequelize might do something weird on query/select, and fail if this column is empty. So in the definition, set `allowNull: true`, but throw error when save.

src/models/installation.ts
mboudreau: I'd recommend we use `encryptedSharedSecret` for both the field and column to follow convention .  After we move all shared secrets to be encrypted by cryptor, we can change the field name to be `sharedSecrets` again without changing the references :)

As for the migration, I think we should create an admin API that migrates things over so we have control over when things are transferred over.

src/github/branch.test.ts
```diff
@@ -27,6 +27,7 @@ describe("Branch Webhook", () => {
 		await Installation.create({
 			clientKey,
 			sharedSecret: "shared-secret",
+			safeSharedSecret: "shared-secret",
```
mboudreau, line 30: shouldn't the `encryptedSharedSecret` be created automatically using hooks? it'll remove the need to duplicate work.

src/models/installation.ts
```diff
@@ -1,22 +1,37 @@
...
 	secrets: string;
 	sharedSecret: string;
+	safeSharedSecret: string; //LOL
 	clientKey: string;
 	updatedAt: Date;
 	createdAt: Date;
 	githubAppId?: number;
 
+	getEncryptionSecretKey() {
+		return EncryptionSecretKeyEnum.GITHUB_SERVER_APP;
```
mboudreau, line 24: Not the correct key

src/models/installation.ts
mboudreau: Yep

src/models/installation.ts
```diff
@@ -127,11 +147,26 @@ Installation.init({
...
+	},
+	sequelize
+});
 
 export interface InstallationPayload {
 	host: string;
 	clientKey: string;
 	// secret: string;
 	sharedSecret: string;
+	safeSharedSecret: string;
```
mboudreau, line 171: I'd remove this from the payload.  We want to use the same interface for both and we will be saving both from the same field. it'll make it easier to transition over.

src/models/installation.ts
mboudreau: I don't think you need it for the time being. Plus, last thing we need is for things to break just because the column isn't set and you can't save anything.

mboudreau: however, we should add something to set the `encryptedSharedSecret` field when the `sharedSecret` field changes.

src/models/installation.ts
```diff
@@ -67,14 +82,16 @@ export class Installation extends Model {
 			},
 			defaults: {
 				jiraHost: payload.host,
-				sharedSecret: payload.sharedSecret
+				sharedSecret: payload.sharedSecret,
+				safeSharedSecret: payload.safeSharedSecret
```
mboudreau, line 86: remove this line.

github-for-jira.sd.yml
```diff
@@ -367,6 +367,16 @@ environmentOverrides:
...
+
+      - &cryptorJiraInstanceSecrets
+        name: jira-instance-secrets
+        type: cryptor
+        attributes:
+          parameters:
+            encryptingServices:
+              - github-for-jira
+            decryptingServices:
+              - github-for-jira
```
mboudreau, line 379: Wait, shouldn't there only be one 'global' one of this?  it's one per service, not one per key right?

Also, shouldn't this be part of the global resources?  Right now it's only on ddev and prod, which means this could break in prod

github-for-jira.sd.yml
```diff
@@ -218,7 +218,7 @@ config:
 
     GLOBAL_AGENT_NO_PROXY: cryptor # needed because of proxy.ts
     CRYPTOR_URL: http://cryptor:26272
-    CRYPTOR_SIDECAR_ENCRYPTION_KEY_ALIASES: micros/github-for-jira/github-server-app-secrets
+    CRYPTOR_SIDECAR_ENCRYPTION_KEY_ALIASES: micros/github-for-jira/github-server-app-secrets,micros/github-for-jira/jira-instance-secrets
```
mboudreau, line 221: do we really need to specify each key that we want to use?  Is there a way to do wildcards?

github-for-jira.sd.yml
gxueatlassian: 1. Good questions, let me check whether it is per key or per service.
2. Not in global is because the doc recommend dev/stage share the same one.

github-for-jira.sd.yml
gxueatlassian: I can double check again. But in the doc, it just say `comma separated blah blah`

src/models/installation.ts
```diff
@@ -127,7 +147,43 @@ Installation.init({
 		type: DataTypes.INTEGER,
 		allowNull: true
 	}
-}, { sequelize });
+}, {
+	hooks: {
+		beforeSave: async (instance: Installation, opts) => {
+			if (opts.fields?.includes("sharedSecret")) {
+				if (!instance.encryptedSharedSecret) {
```
mboudreau, line 154: no need to check for null.  If the sharedSecret changes, so does the encryptedSecret

src/util/encryption-client.ts
```diff
@@ -45,7 +46,9 @@ export class EncryptionClient {
 			return response.data.cipherText;
 		} catch (e) {
 			e.message = e.message?.replace(plainText, "<censored>");
-			logger.warn(e, "Cryptor request failed");
+			const isErrorOfPlainTextEmpty = (plainText || "").length === 0;
+			const encryptionContextKeys = Object.keys(encryptionContext).join(",");
```
mboudreau, line 50: we probably don't want to log this

src/util/encryption-client.ts
```diff
@@ -58,7 +61,9 @@ export class EncryptionClient {
 			});
 			return response.data.plainText;
 		} catch (e) {
-			logger.warn(e, "Decryption request failed");
+			const isErrorOfCipherTextEmpty = (cipherText || "").length === 0;
+			const encryptionContextKeys = Object.keys(encryptionContext).join(",");
```
mboudreau, line 65: same as above

github-for-jira.sd.yml
gxueatlassian: There's no wildcard

github-for-jira.sd.yml
mboudreau: but instead of using the same string as key alias above, it breaks it down into different parts?  seems weird... I'd expect something like `key: micros/github-for-jira/jira-instance-secrets` or something

mboudreau: 2. but why? what's the advantage?

mboudreau: from our perspective, it's less code (and less change of screwing up) the config if we just add all resources in the global

gxueatlassian: > but instead of using the same string as key alias above, it breaks it down into different parts? seems weird... I'd expect something like `key: micros/github-for-jira/jira-instance-secrets` or something

Well, I didn't invent it at the first place. I think the `key` has to be same as the `name` . And why is it broken down? I don't know.

src/models/installation.ts
mboudreau: alright overwrite `encryptedSharedSecret`.  If you wanted to, you could remove the property from the class (but keep it in init), that way someone can't change the encrypted secret manually.  This will also work better long term as we do the switchover since there should be no reference to `encryptedSharedSecret` directly, only `sharedSecret`

github-for-jira.sd.yml
gxueatlassian: > 2. but why? what's the advantage?
This is from the doc.
```
As far as Cryptor is concerned, dev and staging Micros environments both map to the same resource (ie. KMS keys). Cryptor only distinguishes between production and non-production.
```

src/models/installation.ts
```diff
@@ -127,7 +147,43 @@ Installation.init({
...
+					//if caller only set the sharedSecret, copy it to encryptedSharedSecret;
+					instance.encryptedSharedSecret = instance.sharedSecret;
```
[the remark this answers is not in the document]
mboudreau: `await instance.encryptChangedSecretFields(opts.fields);`, but the field that changed could simply just be `sharedSecret`, which we're saving to `encryptedSharedSecret`.  However, the `encryptChangedSecretFields` is looking for which fields has changed and `encryptedSharedSecret` isn't one of them since the ORM only detected `sharedSecret` has changed.  To make sure that our encryption process is triggered, it need to see that `encryptedSharedSecret` is part of the `opts.fields` that has changed.

Hence, doing something like `opts.fields.push("encryptedSharedSecret")` if you've overwritten that `encryptedSharedSecret` with the value of `sharedSecret`

src/models/installation.ts
```diff
@@ -127,7 +147,43 @@ Installation.init({
...
+				//TODO: monitor the prod behaviour and remove this catch along with other migration rollout in another PR.
+				logger.error(`Fail encrypting sharedSecret using cryptor`);
+			}
+		},
+		beforeBulkCreate: async (instances: Installation[], opts) => {
+			for (const instance of instances) {
+				if (opts.fields?.includes("sharedSecret")) {
+					if (!instance.encryptedSharedSecret) {
+						//if caller only set the sharedSecret, copy it to encryptedSharedSecret;
+						instance.encryptedSharedSecret = instance.sharedSecret;
```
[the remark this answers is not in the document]
mboudreau: and we should probably abstract this function as we have duplication...

src/models/installation.ts
gxueatlassian: 👍 I'll do this `always overwrite encryptedSharedSecret with the value of sharedSecret.`.

`you could remove the property from the class` <-- can't, the typings will complain (about no such property in the keyof this)

src/models/installation.ts
gxueatlassian: done.

db/migrations/20220705052312-add-cryptor-sharesecret-column-to-installation-table.js
```diff
@@ -0,0 +1,16 @@
+"use strict";
+
+const tableName = "Installations";
+
+module.exports = {
+	up: async (queryInterface, Sequelize) => {
+		await queryInterface.addColumn(tableName, "safeSharedSecret", {
```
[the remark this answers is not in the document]
gxueatlassian: done.

src/models/installation.ts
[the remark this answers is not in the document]
gxueatlassian: GOOD PICKS!
Turns out i missed a test case. Have added it.

src/models/installation.ts
mboudreau: yeah, but the great thing about typings is that you can ignore it if you want :)

In this special case, we're trying to not expose a property that shouldn't be accessed directly, so we do it covertly but need to break the typings for it to be saved using `setDataValue`

gxueatlassian: Cool, let me try.

mboudreau: ah right.  okay, then leave it in then.

mboudreau (approved):

PR #1340, opened 2022-07-05
ARC-1483 - UI cleanup

krazziekay:
**What's in this PR?**
- This PR addresses the [UI nitpicks](https://hello.atlassian.net/wiki/spaces/PF/pages/1781209083/GHE+prod+segment+-+Nits+and+frogs)

**Why**
- To improve UX


joshkay10 (approved):

rachellerathbone:
Are all the nitpicks accounted for? Would be great to see that reflected in the doc (update status from TODO to DONE)

static/css/global.css
```diff
@@ -211,12 +211,17 @@ a:focus {
   display: flex;
   justify-content: space-between;
   padding: 2em;
+  align-items: center;
```
rachellerathbone, line 214: What is this actually doing? This class is flexed and you don't use text-align to position items in a flex container. You use `align-items` or `justify-content` the latter of which is already being used to push items to the outer edges.

static/css/global.css
```diff
@@ -211,12 +211,17 @@ a:focus {
...
 }
 
 .navigation__backButton__container {
   background: #fff;
   border: none;
   cursor: pointer;
+  height: 24px;
```
rachellerathbone, line 221: Do we need height here? I don't think this changes the styling at all.

static/css/global.css
krazziekay: This is added here to align the items in center position. 

The position of the left arrow was not properly aligned.
Screen Shot 2022-07-06 at 12 52 03 pm

________________________________________________________________________

After adding this:
Screen Shot 2022-07-06 at 12 52 11 pm

static/css/global.css
krazziekay: The height is given to match the size to 24, otherwise the height of the container is 26, which changes the position of the icon inside.

static/css/global.css
rachellerathbone: Sorry my bad. For some reason my brain read it as text-align

rachellerathbone (approved):

gxueatlassian (approved):

joshkay10 (approved):

PR #1345, opened 2022-07-06
Arc-1473 server url errors

rachellerathbone:
**What's in this PR?**
Mapping errors on the back and frontend when making requests against the ghe server url.

**Why**
So we can surface an error to users.

Screen Shot 2022-07-05 at 1 46 07 pm

Screen Shot 2022-07-05 at 2 00 16 pm

Screen Shot 2022-07-05 at 2 14 08 pm

Screen Shot 2022-07-05 at 2 14 59 pm

**Added feature flags**
You won't see it added in this PR but the entire GHE server path is behind GHE_SERVER

**Affected issues**  
[ARC_1473](https://softwareteams.atlassian.net/jira/software/projects/ARC/boards/385?assignee=604894ffe19f910069497c6a&issueParent=125881&selectedIssue=ARC-1473)

**How has this been tested?**  
A [test file](https://github.com/atlassian/github-for-jira/blob/2a4f7ad76190e0620f39688c5006ae4ad4fe32a0/src/routes/jira/server/jira-server-url-post.test.ts) has been added to make sure all success/error scenarios are working as expected on the backend.


src/routes/jira/server/jira-server-url-post.ts
```diff
@@ -3,11 +3,64 @@ import axios from "axios";
...
+interface GheServerUrlErrors {
+	codeOrStatus: GheServerUrlErrorResponses
+}
+
+const gheServerUrlErrors: GheServerUrlErrors = {
+	codeOrStatus: {
+		invalidUrl: {
+			errorCode: "GHE_ERROR_1",
+			message: "Invalid URL",
+			statusCode: 200
```
rachellerathbone, line 25: Sending a 200 response for invalidUrl, ENOTFOUND, and default as none of these are server errors and don't receive a an error status code in the response. Sending `200` with success `false`

static/js/jira-server-url.js
```diff
@@ -68,18 +68,50 @@ const requestFailed = () => {
...
+}
+
 const verifyGitHubServerUrl = (gheServerURL, installationId) => {
-	const csrf = document.getElementById('_csrf').value;
+	const csrf = document.getElementById("_csrf").value
 
-	window.AP.context.getToken(function(token) {
-		$.post("/jira/server-url", {
+	AP.context.getToken(function(token) {
+		$.ajax({
```
rachellerathbone, line 104: Had to change this from the initial implementation as the 502 error seemed to be getting swallowed. Everything works as expected this way.

src/routes/jira/server/jira-server-url-post.ts
```diff
@@ -3,11 +3,64 @@ import axios from "axios";
...
+}
+
+interface GheServerUrlErrors {
+	codeOrStatus: GheServerUrlErrorResponses
+}
+
+const gheServerUrlErrors: GheServerUrlErrors = {
+	codeOrStatus: {
+		invalidUrl: {
+			errorCode: "GHE_ERROR_1",
```
gxueatlassian, line 23: Is there a reason we don't want to use `GHE_ERROR_INVALID_URL`?

static/js/jira-server-url.js
```diff
@@ -68,18 +68,50 @@ const requestFailed = () => {
...
+			title: "Something went wrong",
+			message: "We ran into a hiccup while verifying your details. Please try again later."
+		}
+	}
+};
+
+const handleGheUrlRequestErrors = (err) => {
+	requestFailed();
+	$(".jiraServerUrl__validationError").show();
+	$(".errorMessageBox__title").append(title);
```
gxueatlassian, line 95: Will `append()` append multi error titles to same dom element if user click the submit button twice and failed twice? Ignore this if it doesn't.

gxueatlassian, line 95: Will `append()` append twice if user submit twice and failed twice?

src/routes/jira/server/jira-server-url-post.ts
rachellerathbone: Happy to discuss alternative options for the error codes. What are you thinking for the other 3?
`GHE_ERROR_ENOTFOUND`
`GHE_ERROR_502` / `GHE_SERVER_BAD_GATEWAY`
`GHE_ERROR_DEFAULT`

?

static/js/jira-server-url.js
rachellerathbone: Good catch. Will fix.

src/routes/jira/server/jira-server-url-post.ts
gxueatlassian: Looks good. `GHE_SERVER_BAD_GATEWAY` is better than `GHE_ERROR_502` in my personal opinions.

rachellerathbone: I agree.

src/github/client/github-client-server.test.ts
```diff
@@ -30,7 +30,7 @@ describe("GitHub Client", () => {
 		perPage: number,
 		page: number,
 		expectedInstallationTokenInHeader?: string,
-		scope: nock.Scope = gheNock
+		scope: nock.Scope = gheApiNock
```
mboudreau, line 33: Any changes with `gheNock` and `gheApiNock` are mine.  We weren't handling non-api nock calls correctly in the unit test setup for what Rachelle was trying to do, which is to test the top level github instance.

src/routes/jira/server/jira-server-url-post.ts
```diff
@@ -3,34 +3,88 @@ import axios from "axios";
...
+		res.status(statusCode).send({ success: false, errorCode, message });
+		req.log.error(`The entered URL is not valid. ${gheServerURL} is not a valid url`);
+	}
+
+	try {
+		const gitHubServerApps = await GitHubServerApp.getAllForGitHubBaseUrl(gheServerURL, installationId);
+
+		if (gitHubServerApps?.length) {
+			req.log.debug(`GitHub apps found for url: ${gheServerURL}. Redirecting to Jira list apps page.`);
+			res.status(200).send({ success: true, moduleKey: "github-list-apps-page" });
```
mboudreau, line 79: for clarity, is `moduleKey: "github-list-apps-page"` there to direct where the frontend needs to go next?  This would be break separation of concerns.  This api is only concerned with if the ghe url is valid, not the flow of the app which should be handled by the frontend.

Instead, I'd "generalize" the return type as something like `{valid:boolean; appExists: boolean; errors?:Error[]}`  which should give you all the information you need for the frontend to make it's own decision on how to move forward :)

src/routes/jira/server/jira-server-url-post.ts
```diff
@@ -3,34 +3,88 @@ import axios from "axios";
...
+	statusCode: number;
+}
+
+interface GheServerUrlErrorResponses {
+	[key: string | number]: MessageAndCode;
+}
+
+interface GheServerUrlErrors {
+	codeOrStatus: GheServerUrlErrorResponses;
+}
```
mboudreau, line 18: Not sure if you need this as there are not other keys at the top level, you can just remove `codeOrStatus` from the object

src/routes/jira/server/jira-server-url-post.ts
```diff
@@ -3,34 +3,88 @@ import axios from "axios";
...
+		},
+		default: {
+			errorCode: "GHE_ERROR_DEFAULT",
+			message: "Something went wrong",
+			statusCode: 200
+		}
+	}
+};
+
+export const getGheErrorMessages = (codeOrStatus: number | string | null) => {
```
mboudreau, line 45: not sure this function is needed, couldn't you just do `const errorMessage = gheServerUrlErrors[codeOrStatus] || gheServerUrlErrors.default;` in the try-catch below?

src/routes/jira/server/jira-server-url-post.ts
```diff
@@ -3,34 +3,88 @@ import axios from "axios";
...
-	} else {
-		req.log.error(`Provided gheServerURL ${gheServerURL} is not a valid url`);
-		res.status(200).send({ success: false, error: "Provided URL is not valid." });
+		req.log.debug(`No existing GitHub apps found for url: ${gheServerURL}. Making request to provided url.`);
+		await axios.get(gheServerURL);
+		res.status(200).send({ success: true, moduleKey: "github-app-creation-page" });
+	} catch (err) {
+		req.log.error({ err, gheServerURL }, `Something went wrong`);
+		const { errorCode, message, statusCode } = getGheErrorMessages(err.code || err.response.status);
+		res.status(statusCode).send({ success: false, errorCode, message });
```
mboudreau, line 88: the usual standard for errors is to return an array of errors, each one with a code and message, so in this case we only return one, but should still be an array:
`{success: false, errors:[{code: errorCode, message}]}`

static/js/jira-server-url.js
```diff
@@ -68,18 +68,53 @@ const requestFailed = () => {
 	$("#gheServerBtnSpinner").hide();
 };
 
+const gheServerUrlErrors = {
+	errorCode: {
```
mboudreau, line 72: why the container object?

static/js/jira-server-url.js
```diff
@@ -68,18 +68,53 @@ const requestFailed = () => {
...
+const handleGheUrlRequestErrors = (err) => {
+	requestFailed();
+	const { title, message } = err;
+	$(".jiraServerUrl__validationError").show();
+	$(".errorMessageBox__title").empty().append(title);
+	$(".errorMessageBox__message").empty().append(message);
+	title === gheServerUrlErrors.errorCode.GHE_ERROR_ENOTFOUND.title && $(".errorMessageBox__link").show();
+}
+
+const mapErrorCode = (errorCode) => gheServerUrlErrors.errorCode[errorCode];
```
mboudreau, line 101: this isn't really needed when you can access the object directly

static/js/jira-server-url.js
```diff
@@ -89,12 +124,17 @@ const verifyGitHubServerUrl = (gheServerURL, installationId) => {
...
-					// TODO - render error component
+					const { errorCode } = data;
+					const errorMessage = mapErrorCode(errorCode);
+					handleGheUrlRequestErrors(errorMessage);
 				}
+			},
+			error: function(err) {
+				const { errorCode } = JSON.parse(err.responseText);
+				const errorMessage = mapErrorCode(errorCode);
+				handleGheUrlRequestErrors(errorMessage);
```
mboudreau, line 135: should probably have a single function to handle errors? code's pretty much the same.

src/routes/jira/server/jira-server-url-post.ts
rachellerathbone: Good point on separation of concerns. Will update.

src/routes/jira/server/jira-server-url-post.ts
rachellerathbone: I was doing that initially and then added the switch. But you're right, it's a little redundant.

src/routes/jira/server/jira-server-url-post.ts
rachellerathbone: How does the frontend define which error to render? All of them?

src/routes/jira/server/jira-server-url-post.test.ts
```diff
@@ -0,0 +1,185 @@
...
+			clientKey: "client-key"
+		});
+		app = express();
+		app.use((req: Request, res: Response, next: NextFunction) => {
+			res.locals = { installation };
+			req.log = getLogger("test");
+			req.session = { jiraHost };
+			next();
+		});
+		app.use(RootRouter);
```
gxueatlassian, line 28: Minor suggestions, **can ignore this in this PR**, since we are testing the `jira-server-url-post`, which:
1. Far behind the rootRouter
2. It is not a router anyway, just a function

we can:
1. Don't use `app: express` and `supertest` at all.
2. Ignore preparing `jwt = encodeSymmetric({`
3. And just call that function and test output.

gxueatlassian: Oh, happy to be disagreed on this one though. Would love to trigger some heathy discussion on testing.

static/js/jira-server-url.js
```diff
@@ -68,33 +68,71 @@ const requestFailed = () => {
...
+		title: "Something went wrong",
+		message: "We ran into a hiccup while verifying your details. Please try again later."
+	}
+};
+
+const handleGheUrlRequestErrors = (err) => {
+	requestFailed();
+	const { title, message } = err;
+	$(".jiraServerUrl__validationError").show();
+	$(".errorMessageBox__title").empty().append(title);
```
gxueatlassian, line 94: Still `append`?

gxueatlassian (dismissed):

src/routes/jira/server/jira-server-url-post.test.ts
mboudreau: Because we don't control the input going into the function, we are testing the whole stack that relates to the route and were able to find issues because of this method.  If we didn't, we could have changed the route to be a GET requests instead, or have a type in the path, or that the authentication doesn't work, etc etc

Using supertest doesn't really add any extra overhead, but it does give us absolute confidence that everything is setup correctly in express and the function is triggered as it should be.

src/routes/jira/server/jira-server-url-post.ts
```diff
@@ -3,34 +3,71 @@ import axios from "axios";
...
+	},
+	ENOTFOUND: {
+		errorCode: "GHE_ERROR_ENOTFOUND",
+		message: "Request to URL failed",
+		statusCode: 200
+	},
+	502: {
+		errorCode: "GHE_SERVER_BAD_GATEWAY",
+		message: "Bad gateway",
+		statusCode: 502
```
mboudreau, line 30: Even a bad gateway to github should probably return a 200 to the frontend.  Again, a 502 being returned from us means that _our api_ is having this issue, not github.  This will also mess with SLOs I think...

mboudreau: I think all errors that can be caught should be 200s being returned

static/js/jira-server-url.js
mboudreau: but she's emptying it before?

static/js/jira-server-url.js
mboudreau: all the github error handling should be done in the success function, the error function is just for our own server issues, which I'm not sure how to handle in this case...

src/routes/jira/server/jira-server-url-post.test.ts
gxueatlassian: `If we didn't, we could have changed the route to be a GET requests instead` <-- that's exactly why this test is not complete (but doesn't need to be).

There's a `JiraServerUrlRouter`, in which it calls 
```
JiraServerUrlRouter.route("/").post(JiraContextJwtTokenMiddleware, JiraServerUrlPost);
```

"We could have changed the route to be a GET requests instead" in above, and this test still green, therefore not a valid test.

static/js/jira-server-url.js
gxueatlassian: Didn't see that , my bad.

src/routes/jira/server/jira-server-url-post.test.ts
mboudreau: I'm not sure what you're arguing?  If you change that function to be a GET request instead of a POST, it will fail the test...

mboudreau (approved):

src/routes/jira/server/jira-server-url-post.test.ts
gxueatlassian: All good then.

gxueatlassian (approved):

mboudreau (approved):

PR #1347, opened 2022-07-07
ARC-1441: remove `this` within static method.

gxueatlassian:
**What's in this PR?**
Remove the `this` usage in a static method.

**Why**
Since it is static method, `this` doesn't make sense. 
But above it is just in theory. When compiled, typescript make `this` same as the class itself. Hence why unit test didn't pick this up. But I think we should still use it like static method.

**Added feature flags**
N/A

**Affected issues**  
ARC-1441

**How has this been tested?**  
Unit test.

**Whats Next?**
N/A

### DIFF
--- a/src/util/encryption-client.ts
+++ b/src/util/encryption-client.ts
@@ -28,7 +28,7 @@ interface DecryptResponse {
  */
 export class EncryptionClient {
 
-	protected static readonly axios: AxiosInstance = axios.create({
+	private static readonly axiosInst: AxiosInstance = axios.create({
 		baseURL: envVars.CRYPTOR_URL,
 		headers: {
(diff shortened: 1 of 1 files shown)


mboudreau (commented):
But why?  This isn't a needed change and work fine?

gxueatlassian:
> But why? This isn't a needed change and work fine?

Not sure why, you reckon we keep this as it is? The `this` looks so weird to me in the static method.

mboudreau:
looks fine to me.  Compiler will catch any issues if someone tries to use the instance of the class (which doesn't exist).  Seems to be more of a preferential thing which you can bring up to coding conventions.

gxueatlassian:
Leave it as it is per comments.

PR #1350, opened 2022-07-07
ARC-1482 - Use AUI Form validation in Jira Server URL and Jira Manual App creation 

krazziekay:
**What's in this PR?**
- Updated the form elements in Jira Server URL screen
  - Replaced own custom jQuery logic for form validation by AUI form and AUI form validation
- Used AUI form submit on Jira Manual App creation .

**How was this tested**
- Tested in Staging

**Why**
- For consistency in form handling.
- For better accessibility.

### DIFF
--- a/static/css/jira-server-url.css
+++ b/static/css/jira-server-url.css
@@ -19,9 +19,9 @@
   margin-bottom: 1.2em;
 }
 
-.jiraServerUrl__actionBtn__container {
+.jiraServerUrl__actionBtn {
   margin-top: 1.25em;
-  text-align: right;
+  float: right;
 }
 
 #gheServerBtnSpinner {
--- a/static/css/jira-server-url.css
+++ b/static/css/jira-server-url.css
@@ -19,32 +19,6 @@
   margin-bottom: 1.2em;
 }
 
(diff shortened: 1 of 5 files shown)


static/css/jira-server-url.css
```diff
@@ -19,35 +19,9 @@
   margin-bottom: 1.2em;
 }
 
-.jiraServerUrl__error {
-  display: none;
```
gxueatlassian, line 23: NICE! Less is more!!!

gxueatlassian (approved):
Sorry I still can't tell how the form validation is activated in file `static/js/jira-manual-app-creation.js`, but I trusted you run it locally and it works.

krazziekay:
> Sorry I still can't tell how the form validation is activated in file `static/js/jira-manual-app-creation.js`, but I trusted you run it locally and it works.

@gxueatlassian  In `static/jira-manual-app-creation.js`, the normal form submit was replaced by AJS.$("#jiraManualAppCreation__form").on("aui-valid-submit", () => {})`, which gets triggered only when the whole form is valid. ([More details](https://aui.atlassian.com/aui/latest/docs/form-validation.html))

PR #1351, opened 2022-07-08
ARC-1481 - Fixing Route for GH Config page

krazziekay:
**What's in this PR?**
- Routing corrected after successfully connecting to a GH cloud
  - After closing the _Github config page_(child window), redirect to _Jira config page_.

**Video**

https://user-images.githubusercontent.com/13076255/178188032-c2947b80-b2d0-4535-94bf-d5144502860c.mov


src/routes/jira/server/jira-server-url-post.ts
```diff
@@ -22,6 +22,8 @@ export const JiraServerUrlPost = async (
 			} else {
 				req.log.debug(`No existing GitHub apps found for url: ${gheServerURL}. Making request to provided url.`);
 				await axios.get(gheServerURL);
+				// TODO: need to install the GithubServerApp and pass that id in the response
+				// This id needs to be passed in the frontend, as it is needed for GH Manual app page
```
mboudreau, line 26: I don't get this.  The app hasn't been created yet, why would there be an id?

static/js/github-configuration.js
```diff
@@ -1,4 +1,14 @@
...
+/**
+ * This method passes a value to the Parent Window,
+ * which is used to change the url of the Parent Window
+ */
+const updateParentWindowURL = () => {
+  if (window.opener != null && !window.opener.closed && window.opener.document.getElementById("redirectToGitHubConfigPage")) {
+    window.opener.document.getElementById("redirectToGitHubConfigPage").value = true;
+  }
+  window.close();
+};
```
mboudreau, line 10: This screams of a hack to me.  What is it you're trying to do?

src/routes/jira/server/jira-server-url-post.ts
krazziekay: When the user adds a URL in the _Jira Server URL_ screen and submits that, that should be saved in the **GitHubServerApps** table, isn't it? 
Once the `GitHubServerApp` has been created, we need its `id` to fetch the other data in the next navigated routes. 
When the user navigates to the _Manual App_ screen, then we need to display the _webhook_, _setup_ and _callback_ URL for the server that was just created. The _webhook_, _setup_ and _callback_ URL need the _githubBaseUrl_ and _uuid_ of the created `GitHubServerApp`. For this reason, we need the `id` of the created `GitHubServerApp`.

static/js/github-configuration.js
krazziekay: I'm trying to pass a value from child window to the parent window, which determines the url of the parent window. Couldn't think of a better way than this...

src/routes/jira/jira-atlassian-connect-get.ts
```diff
@@ -87,6 +87,14 @@ const modules = {
 			},
 			url: "/jira/app-creation",
 			conditions: adminCondition
+		},
+		{
+			key: "github-manual-app-creation-page",
+			name: {
+				value: "GitHub Manual App Creation"
+			},
+			url: "/jira/manual-app-creation/?gitHubServerAppId={ac.ghsaId}",
```
mboudreau, line 96: I'm not sure the id is needed at this stage...

krazziekay: The id is needed here to fetch the `gitHubBaseUrl` and `uuid` of the `GithubServerApp` in order to display the _callback_, _webhook_ and _setup_ URLs in the manual app creation page.

gxueatlassian: Does anyone know the need of 
```
{
			key: "github-manual-app-creation-page",
			name: {
```
Is it more a UX requirement (so that user can jump to here directly) ? Or is it a tech requirement?
Since we are iframing our app, doesn't the `window.location.url=xxx` or `history.back()` work without above?

src/routes/jira/server/jira-server-url-post.ts
rachellerathbone: No, this doesn't happen from that screen. When you submit the url on that screen we do our frontend validation checks then this middleware:

- also verifies the passed value is a valid url
- checks to see if there are any existing GH apps in the table for the provided url
- if there are existing apps, we redirect to the screen you made that lists all the GH apps we found
- if there are no existing apps, we confirm we can hit the provided url (hole in the firewall has been configured) and then redirect to the auto/manual screen.

The auto/manual flow is where we will be saving apps to the GHSA table.

src/routes/jira/jira-atlassian-connect-get.ts
rachellerathbone: We shouldn't be creating the module for the manual app creation screen as part of this PR. Plus, that path wouldn't need an id as it's simply a form that will render a couple of urls and take in user inputs.

rachellerathbone: See this recording for the issue we're trying to fix.

https://user-images.githubusercontent.com/37155488/178183440-efd650c0-b059-4431-b766-e65c9439bd94.mov

In the cloud flow, after I hit connect, instead of being redirected to the GH config page, I land on the select GH version screen because we're simply calling `window.close()`. This worked when we only had one screen on the Jira side but obviously needs to be modified to take the user back to the correct place. 

All the manual app creation changes are not needed to resolve this.

src/routes/jira/server/jira-server-url-post.ts
krazziekay: This has been reverted

src/routes/jira/jira-atlassian-connect-get.ts
krazziekay: This has been reverted.

static/js/github-configuration.js
mboudreau: but why? what's the use case?

krazziekay: We want to change the url of the parent window once the user connects to an app through the child window.

This is our current situation:
https://user-images.githubusercontent.com/37155488/178183440-efd650c0-b059-4431-b766-e65c9439bd94.mov
 
This is what we want:
https://user-images.githubusercontent.com/13076255/178188032-c2947b80-b2d0-4535-94bf-d5144502860c.mov

static/js/jira-select-card-option.js
```diff
@@ -7,7 +7,16 @@ function openChildWindow(url) {
 	const interval = setInterval(function () {
 		if (child.closed) {
 			clearInterval(interval);
```
rachellerathbone, line 9: Is this just checking if the child page closes and then navigating back to the config page? This would work for the scenario where users click 'Connect' (cloud or server path) as at this point we're expecting them to go back to the config page. But have you thought about what will happen when a user clicks the back narrow in the navigation from the /github/configuration page?

2 pages import this js file that would be impacted by this implementation: github version page and the auto/manual screen. 

- If I go down the cloud install path, I'll go from the GH version screen to  /github/configuration but if I click the back button, I expect to end up back on the GH version page. 
- Likewise, if I go down the server install path and choose to automatically create my app, the last page I'll be on before I get to  /github/configuration is the auto/manual app view. I would expect to return to the manual/auto app page, not the GH config page.

static/js/github-configuration.js
mboudreau: yeah, there's a better way of doing that with the `openChildWindow` function in jira-configuration.js file.  ie. when child closes, go back to previous view instead of just reload :)

static/js/jira-select-card-option.js
```diff
@@ -7,7 +7,16 @@ function openChildWindow(url) {
...
-			AP.navigator.reload();
+			if (Boolean($("#redirectToGitHubConfigPage").val())) {
```
mboudreau, line 10: instead of `Boolean` casting, you can just do double !: `if (!!$("#redirectToGitHubConfigPage").val())`

static/js/github-configuration.js
mboudreau: oh wait, you already have that below! you don't need this change at all or need to pass any information

static/js/jira-select-card-option.js
mboudreau: nvm, you don't need the conditional at all.  When the window closes, just go to the github-post-install-page

krazziekay: Removing this condition means doing a redirect to the jira config page everytime. Then what about the case when the user doesn't make any connections. If the user simply closes the child window, the url in the parent window will navigate too. Not sure if thats ok, what do you think @rachellerathbone?

rachellerathbone: We need a solution that works for both scenarios:
1. user clicks connect on connect an org screen. Here they need to be redirected to the GH config page. This should happen for both cloud and server customers
2. user clicks the back button in the nav. In this case they would need to go back to one of 3 places:
a. if they're a cloud customer, the last page would be the select gh server screen
b. if they're a server customer the last page could either be:
i. the manual form screen
ii. the auto/manual selection screen (customer selected auto in this case).

krazziekay: I believe this solution works for both scenarios. Heres my understanding of how this works:
- When user follows the cloud path. 
  - In the GH version selection screen(parent window), there is a hidden input valued null.
  - When the user selects the Cloud option, it will open a new child window.
  - When the user clicks on the connect button of the child window, the child window sets a true value for the hidden input in the parent. 
  - In the parent window, when the child window is closed, it navigates to GH config page only when the value of this hidden input is set to true, which happens only when the connect button is clicked. Otherwise it will just do a normal refresh. 
- When user follows the enterprise path, the user navigates through the same window until he/she reaches the automatic/manual app creation step.
  - If its manual, its still the same window, so nothing changes. 
    - We can set the navigation logic at the successful submission of the manual app form.
  - But if its automatic, it will open a child window. So, if we add a hidden input of the same id, then the rest should function similarly as in the step for the cloud above.
    - However, this input file hasn't been added here for the enterprise server path because the action for the automatic button hasn't been implemented yet.

mboudreau: I'm not sure I follow, this file is only used in the jira select screen and as far as I can tell, anytime the window closes, we want to show the landing page no matter what.

maybe we need to get together tomorrow and figure this out over a screenshare.

mboudreau: either way, even if we need a way to communicate back to the opener window, there's a better way of doing it, but first let's assess if it's needed at all.

krazziekay: - The condition has been removed.
- Updated the logic from a page refresh to a page redirect to GH config Page

mboudreau (approved):

PR #1352, opened 2022-07-08
SAI-12030 Don't send an empty commit association array if there are no commit keys to include

jason-hooker:
**What's in this PR?**

- This PR ensures that if there are no commit keys to include (because all 500 available association values have been taken up by issue ids) then don't include an empty "commit" association object in the associations array.

**Why**

This is a safety control so we don't send an empty association type in the associations array.

**How has this been tested?**  
- A unit test is included to verify this works as expected.


src/transforms/transform-deployment.ts
```diff
@@ -170,12 +170,14 @@ const mapJiraIssueIdsAndCommitsToAssociationArray = (
 					repositoryId: repositoryId
 				};
 			});
-		associations.push(
-			{
-				associationType: "commit",
-				values: commitKeys
-			}
-		);
+		if (commitKeys.length > 0) {
```
mboudreau, line 173: just `if (commitKeys.length) {` will do

src/transforms/transform-deployment.ts
```diff
@@ -170,12 +170,14 @@ const mapJiraIssueIdsAndCommitsToAssociationArray = (
...
-			}
-		);
+		if (commitKeys.length) {
+			associations.push(
+				{
+					associationType: "commit",
+					values: commitKeys
+				}
+			);
+		}
```
mboudreau, line 180: Hm, this code seems hard to read and could be more concise.  Maybe something like this instead?

```
const mapJiraIssueIdsAndCommitsToAssociationArray = (
    issueIds: string[],
    repositoryId: string,
    commitSummaries: CommitSummary[] = []
): JiraAssociation[] | undefined => {
    if (!issueIds?.length) {
        return undefined;
    }
    
    return [
        {
            associationType: "issueIdOrKeys",
            values: issueIds
        },
        {
            associationType: "commit",
            values: commitSummaries
                .slice(0, MAX_ASSOCIATIONS_PER_ENTITY - issueIds.length)
                .map((commitSummary) => ({
                    commitHash: commitSummary.sha,
                    repositoryId: repositoryId
                }));
        }
    ].filter(association => !!association.values?.length)
}
```

jason-hooker: But the commit association type should only exist if there are more than 0 commit keys after the slice has occurred.

jason-hooker: I actually found that harder to understand - but then again I am no expert in javascript.

jason-hooker: I have since tried the above - but the syntax checker in IDEA doesn't like something about the association type strings.

mboudreau: What does it say?

mboudreau: actually, nvm, we can fix that later :)

mboudreau (approved):

PR #1354, opened 2022-07-11
add 403 status to mapping

rachellerathbone:
**What's in this PR?** extra error code in mapping

**Why** prod seemed to pick up the correct error when I tested it earlier but staging is getting a 403 (couldn't test in staging earlier... don't worry, everything is behind GHE_SERVER).


src/routes/jira/server/jira-server-url-post.ts
```diff
@@ -12,15 +12,18 @@ interface GheServerUrlErrors {
...
 	invalidUrl: {
 		errorCode: "GHE_ERROR_INVALID_URL",
 		message: "Invalid URL"
 	},
-	ENOTFOUND: {
-		errorCode: "GHE_ERROR_ENOTFOUND",
-		message: "Request to URL failed"
-	},
+	ENOTFOUND: GHE_ERROR_ENOTFOUND,
+	403: GHE_ERROR_ENOTFOUND,
```
gxueatlassian, line 26: Could be a bit weird, since 404 is NOT_FOUND, while 403 is Forbidden.

gxueatlassian: And is it for UX purpose that we need to tell the difference of 403 and ENOTFOUND?

gxueatlassian, line 26: I worry it may confuse other devs as `GHE_ERROR_ENOTFOUND` looks like `404 NOT FOUND`, but here what we are trying to say is "I can't reach the GHE server url you provided." 
So maybe `GHE_ERROR_UNREACHABLE`?

gxueatlassian (dismissed):
minor comments.

gxueatlassian (approved):

mboudreau (approved):

PR #1356, opened 2022-07-11
ARC-1487 - Update jira connect to hide pages in dropdown

krazziekay:
**What's in this PR?**
- Added `location: none` to all the `generalPages` in the app-connector file.

**Why**
- To hide the pages from showing up in the dropdown(screenshot below). We don’t want our customers to be able to jump ahead in the flow/navigate to any parts of that flow if they’re trying to connect to cloud so we need to figure out a way to prevent them from rendering in the dropdown.
- image

### DIFF
--- a/src/routes/jira/jira-atlassian-connect-get.ts
+++ b/src/routes/jira/jira-atlassian-connect-get.ts
@@ -70,6 +70,7 @@ const modules = {
 				value: "GitHub Select Version"
 			},
 			url: "/jira/select-version",
+			location: "none",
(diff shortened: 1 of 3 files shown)


mboudreau:
Can you provide the documentation for why this is needed?  Does that mean that users can't use the back button on their browser?

krazziekay:
> Can you provide the documentation for why this is needed? Does that mean that users can't use the back button on their browser?

I've updated the `Why` in the description.

PR #1360, opened 2022-07-12
Arc 1354 installations migration api

gxueatlassian:
**What's in this PR?**
An admin API to migrate existing sharedSecret in Installations to new db column encryptedSharedSecret using Cryptor.

**Why**
By Atlassian policy, we should use Cryptor for encryption.

**Added feature flags**
N/A

**Affected issues**  
ARC-1354

**How has this been tested?**  
On dev and stg 
image

**Whats Next?**
Cut-over reading/write traffic once all are migrated.

### DIFF
--- a/src/routes/api/api-router.ts
+++ b/src/routes/api/api-router.ts
@@ -15,6 +15,7 @@ import { ApiInstallationDelete } from "./installation/api-installation-delete";
 import { ApiHashPost } from "./api-hash-post";
 import { EncryptionClient, EncryptionSecretKeyEnum } from "utils/encryption-client";
 import { ApiPingGet } from "routes/api/api-ping-get";
+import { CryptorMigrationInstallationPost } from "./cryptor-migrations/migration-router";
 
 export const ApiRouter = Router();
 
@@ -138,6 +139,7 @@ ApiRouter.use("/cryptor", async (_req: Request, resp: Response) => {
 		resp.status(500).send("fail");
 	}
 });
+ApiRouter.post("/migration-installations-shared-secret", CryptorMigrationInstallationPost);
 
 ApiRouter.use("/jira", ApiJiraRouter);
 ApiRouter.use("/:installationId", param("installationId").isInt(), returnOnValidationError, ApiInstallationRouter);
--- a/src/routes/api/api-router.ts
+++ b/src/routes/api/api-router.ts
@@ -142,5 +142,4 @@ ApiRouter.use("/cryptor", async (_req: Request, resp: Response) => {
 ApiRouter.use("/migration", CryptorMigrationRouter);
 
 ApiRouter.use("/jira", ApiJiraRouter);
-//TODO: Following `use(/:installationId)` seems block any more routers to be used bellow, need fix
 ApiRouter.use("/:installationId", param("installationId").isInt(), returnOnValidationError, ApiInstallationRouter);
--- a/src/routes/api/api-router.ts
+++ b/src/routes/api/api-router.ts
@@ -15,6 +15,7 @@ import { ApiInstallationDelete } from "./installation/api-installation-delete";
 import { ApiHashPost } from "./api-hash-post";
 import { EncryptionClient, EncryptionSecretKeyEnum } from "utils/encryption-client";
 import { CryptorMigrationInstallationPost } from "./cryptor-migrations/migration-router";
+import { CryptorMigrationInstallationVerificationPost } from "./cryptor-migrations/migration-verification-router";
 import { ApiPingPost } from "routes/api/api-ping-post";
 
 export const ApiRouter = Router();
@@ -140,6 +141,7 @@ ApiRouter.use("/cryptor", async (_req: Request, resp: Response) => {
 	}
 });
 ApiRouter.post("/migration-installations-shared-secret", CryptorMigrationInstallationPost);
+ApiRouter.post("/migration-installations-verification", CryptorMigrationInstallationVerificationPost);
 
 ApiRouter.use("/jira", ApiJiraRouter);
 ApiRouter.use("/:installationId", param("installationId").isInt(), returnOnValidationError, ApiInstallationRouter);
--- a/src/routes/api/api-router.ts
+++ b/src/routes/api/api-router.ts
@@ -14,8 +14,8 @@ import { json, urlencoded } from "body-parser";
 import { ApiInstallationDelete } from "./installation/api-installation-delete";
 import { ApiHashPost } from "./api-hash-post";
 import { EncryptionClient, EncryptionSecretKeyEnum } from "utils/encryption-client";
-import { ApiPingGet } from "routes/api/api-ping-get";
 import { CryptorMigrationInstallationPost } from "./cryptor-migrations/migration-router";
+import { ApiPingPost } from "routes/api/api-ping-post";
 
 export const ApiRouter = Router();
 
@@ -101,7 +101,7 @@ ApiRouter.post(
 // Hash incoming values with GLOBAL_HASH_SECRET.
 ApiRouter.post("/hash", ApiHashPost);
 
-ApiRouter.get("/ping", ApiPingGet);
+ApiRouter.post("/ping", ApiPingPost);
 
 // TODO: remove once move to DELETE /:installationId/:jiraHost
 ApiRouter.delete(
--- a/src/routes/api/api-router.ts
+++ b/src/routes/api/api-router.ts
@@ -14,9 +14,8 @@ import { json, urlencoded } from "body-parser";
 import { ApiInstallationDelete } from "./installation/api-installation-delete";
 import { ApiHashPost } from "./api-hash-post";
 import { EncryptionClient, EncryptionSecretKeyEnum } from "utils/encryption-client";
-import { CryptorMigrationInstallationPost } from "./cryptor-migrations/migration-router";
-import { CryptorMigrationInstallationVerificationPost } from "./cryptor-migrations/migration-verification-router";
 import { ApiPingPost } from "routes/api/api-ping-post";
+import { CryptorMigrationRouter } from "./cryptor-migrations/migration-router";
 
 export const ApiRouter = Router();
 
@@ -113,10 +112,6 @@ ApiRouter.delete(
 	ApiInstallationDelete
 );
 
-//migrate secrets
-ApiRouter.post("/migration-installations-shared-secret", CryptorMigrationInstallationPost);
-ApiRouter.post("/migration-installations-verification", CryptorMigrationInstallationVerificationPost);
-
 // TODO: remove the debug endpoint
 /*
 How to invoke:
@@ -147,4 +142,4 @@ ApiRouter.use("/cryptor", async (_req: Request, resp: Response) => {
 
 ApiRouter.use("/jira", ApiJiraRouter);
 ApiRouter.use("/:installationId", param("installationId").isInt(), returnOnValidationError, ApiInstallationRouter);
-
+ApiRouter.use("/migration", CryptorMigrationRouter);
--- a/src/routes/api/api-router.ts
+++ b/src/routes/api/api-router.ts
@@ -139,7 +139,8 @@ ApiRouter.use("/cryptor", async (_req: Request, resp: Response) => {
 		resp.status(500).send("fail");
 	}
 });
+ApiRouter.use("/migration", CryptorMigrationRouter);
 
 ApiRouter.use("/jira", ApiJiraRouter);
+//TODO: Following `use(/:installationId)` seems block any more routers to be used bellow, need fix
 ApiRouter.use("/:installationId", param("installationId").isInt(), returnOnValidationError, ApiInstallationRouter);
-ApiRouter.use("/migration", CryptorMigrationRouter);
--- a/src/routes/api/api-router.ts
+++ b/src/routes/api/api-router.ts
@@ -113,6 +113,10 @@ ApiRouter.delete(
 	ApiInstallationDelete
 );
 
+//migrate secrets
+ApiRouter.post("/migration-installations-shared-secret", CryptorMigrationInstallationPost);
+ApiRouter.post("/migration-installations-verification", CryptorMigrationInstallationVerificationPost);
+
 // TODO: remove the debug endpoint
 /*
 How to invoke:
@@ -140,8 +144,6 @@ ApiRouter.use("/cryptor", async (_req: Request, resp: Response) => {
 		resp.status(500).send("fail");
 	}
 });
-ApiRouter.post("/migration-installations-shared-secret", CryptorMigrationInstallationPost);
-ApiRouter.post("/migration-installations-verification", CryptorMigrationInstallationVerificationPost);
 
 ApiRouter.use("/jira", ApiJiraRouter);
 ApiRouter.use("/:installationId", param("installationId").isInt(), returnOnValidationError, ApiInstallationRouter);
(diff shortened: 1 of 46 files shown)


src/routes/api/cryptor-migrations/migration-router.ts
```diff
@@ -0,0 +1,115 @@
...
+const SAFE_GUARD_LOOP_COUNT = 100;
+
+export const CryptorMigrationInstallationPost = async (req: Request, res: Response): Promise<void> => {
+
+	const startTime = Date.now();
+
+	let batchSize = parseInt(req.query.batchSize as string);
+	let totalLimit = parseInt(req.query.totalLimit as string);
+	const jiraHost = req.query.jiraHost as string;
+
```
mboudreau, line 16: It's a post, you should be using the body, not the querystring.

You also need to validate the input and give sensible defaults if nothing is specified or error out the API call.  You also need to add limits to make sure you don't do too many.

src/routes/api/cryptor-migrations/migration-router.ts
```diff
@@ -0,0 +1,115 @@
...
+		jiraHost
+	});
+
+	req.log.info(`Received to migration installation sharedSecret for`);
+	if (jiraHost) {
+		if (batchSize || totalLimit) {
+			req.log.warn("Invalid params");
+			res.status(400).send("Since 'jiraHost' is provided, we are only migrating one record for a time. Please do not specify 'batchSize' nor 'totalLimit'");
+			return;
+		}
```
mboudreau, line 30: this should be right at the beginning of the function

src/routes/api/cryptor-migrations/migration-router.ts
```diff
@@ -0,0 +1,115 @@
...
+		if (isNaN(batchSize) || batchSize < 0 || batchSize > MAX_BATCH_SIZE) {
+			req.log.warn("Invalid params");
+			res.status(400).send(`Please provide a valid batchSize between 0 and ${MAX_BATCH_SIZE}, got ${batchSize}`);
+			return;
+		}
+		if (isNaN(totalLimit) || totalLimit < 0 || totalLimit > MAX_BATCH_SIZE) {
+			req.log.warn("Invalid params");
+			res.status(400).send(`Please provide a valid totalLimit between 0 and ${MAX_TOTAL_LIMIT}, got ${totalLimit}`);
+			return;
+		}
```
mboudreau, line 47: this is too much.  if bad inputs are given, just use sensible defaults

src/routes/api/cryptor-migrations/migration-router.ts
```diff
@@ -0,0 +1,115 @@
...
+
+	req.log = req.log.child({
+		operation: "migrate-installations-shared-secret",
+		batchSize,
+		totalLimit,
+		jiraHost
+	});
+
+	req.log.info(`Received to migration installation sharedSecret for`);
+	if (jiraHost) {
```
mboudreau, line 25: What's the point of the jiraHost?

src/routes/api/cryptor-migrations/migration-router.ts
```diff
@@ -0,0 +1,115 @@
...
+			return;
+		}
+	}
+
+	const toMigrateWhere: WhereOptions = {};
+	if (jiraHost) {
+		toMigrateWhere.jiraHost = jiraHost;
+	} else {
+		toMigrateWhere.encryptedSharedSecret = null;
+	}
```
mboudreau, line 55: ???

src/routes/api/cryptor-migrations/migration-router.ts
gxueatlassian: Isn't this the place of "beginning"? I mean only `req.log = req.log.child` is above it, and sort-of need that for logging.

src/routes/api/cryptor-migrations/migration-router.ts
gxueatlassian: Agree on the post/body part, will fix it. 

`You also need to validate the input` is done bellow.

src/routes/api/cryptor-migrations/migration-router.ts
```diff
@@ -0,0 +1,115 @@
...
+
+	req.log.info("About to start migrating installation sharedSecret");
+	let count = 0;
+	let safeGuardLoopCount = 0;
+	while (count < totalLimit) {
```
mboudreau, line 60: you should _never_ use a while loop with a DB operation

src/routes/api/cryptor-migrations/migration-router.ts
gxueatlassian: In case we want to:
1. Only migrate for one particular jiraHost
2. Or we want to re-migrate a previous migrated jiraHost.

gxueatlassian: Well, if you serious don't think we need it, I can remove it completely.

src/routes/api/cryptor-migrations/migration-router.ts
gxueatlassian: `this is too much` <-- what is too much? My thinking is this is API route, so better explicit compare to implicit "sensible" defaults.

src/routes/api/cryptor-migrations/migration-router.ts
gxueatlassian: 1. Search for a record that match the jiraHost
2. Or search for records that need to be migrated (null encryptedSharedSecret)

src/routes/api/cryptor-migrations/migration-router.ts
gxueatlassian: What do you suggest for migrating xxx amount of records? Say we want to migrate 1000 records everytime we call this api, you mean we shall pull that 1000 records from db in one go?

mboudreau (changes requested):
Way too complicated.  It should essentially be just 1 input: size.  How many do you want to encrypt the value with a sensible default if none is specified (like 500).

Then the code is pretty simple: 
```
const installations = await Installation.findAll({where: { encryptedSharedSecret: null}, limit: limitGiven});
await Installation.sequelize?.transaction(async (transaction) => {
   installations
});
const updatedInstallations = await Promise.all(installations.map(async (installation) => {
   installation.encryptedSharedSecret = installation.sharedSecret;
   installation.save({transaction});
});

```

src/routes/api/cryptor-migrations/migration-router.ts
```diff
@@ -0,0 +1,66 @@
+import { Request, Response } from "express";
+import { Installation } from "models/installation";
+
+const MAX_BATCH_SIZE = 10_000;
+const DEFAULT_BATCH_SIZE = 10;
+const RESPONSE_BATCH_SIZE = 1000;
```
gxueatlassian, line 6: bad naming :( keen for suggestion. See bellow. Indicate how often do we tell the curl client that xxx has been migrated.

src/routes/api/cryptor-migrations/migration-router.ts
```diff
@@ -0,0 +1,66 @@
...
+	}
+	req.log.debug("Found matching installations to migrate, length " + installations.length);
+
+	//--------- loop and save each one of them --------------
+	let count = 0;
+	for (const inst of installations) {
+		//copy from shared secret to encryptedSharedSecret to enable encryption
+		inst.encryptedSharedSecret = inst.sharedSecret;
+		//update db
+		await inst.save();
```
gxueatlassian, line 37: no transaction yet, still reading.
but a bonus for no transaction is "resume from where you left last time".

src/routes/api/cryptor-migrations/migration-router.ts
gxueatlassian: fixed.

src/routes/api/cryptor-migrations/migration-router.ts
gxueatlassian: fixed.

src/routes/api/cryptor-migrations/migration-router.ts
gxueatlassian: fixed.

src/routes/api/cryptor-migrations/migration-router.ts
gxueatlassian: fixed. will look into testing and see whether we need BB pipeline looping.

src/routes/api/cryptor-migrations/migration-router.ts
gxueatlassian: fixed. No more jira host.

src/routes/api/cryptor-migrations/migration-router.ts
gxueatlassian: @mboudreau let me know if you find out more. 
I just realize the benefit of "resume from where you left last time". Assuming you are running up to 999 out of 1000, we don't want to rollback previous 999 migration as they are independant. Should use transaction as it is designed to be (transaction), not for other like performance purpose.

src/routes/api/cryptor-migrations/migration-installation-verification.ts
```diff
@@ -0,0 +1,39 @@
+import { Request, Response } from "express";
+import { Installation } from "models/installation";
+
+export const CryptorMigrationInstallationVerificationPost = async (req: Request, res: Response): Promise<any> => {
```
mboudreau, line 4: express routes always return `Promise<void>`

gxueatlassian: fixed.

src/routes/api/cryptor-migrations/migration-installation-verification.ts
```diff
@@ -0,0 +1,39 @@
...
+	const jiraHost: string = req.body?.jiraHost;
+
+	req.log = req.log.child({ operation: "migrate-installations-verification" });
+
+	if (!jiraHost) {
+		res.status(200).json({ ok: false, reason: "invalid jiraHost param in body" });
+		return;
+	}
+
+	const inst: Installation = await Installation.findOne({ where: { jiraHost } });
```
mboudreau, line 15: use the function that already exists for Installation: `Installation.getForHost(jiraHost)`

src/routes/api/cryptor-migrations/migration-installation-verification.ts
```diff
@@ -0,0 +1,39 @@
+import { Request, Response } from "express";
+import { Installation } from "models/installation";
+
+export const CryptorMigrationInstallationVerificationPost = async (req: Request, res: Response): Promise<any> => {
+
+	const jiraHost: string = req.body?.jiraHost;
+
+	req.log = req.log.child({ operation: "migrate-installations-verification" });
```
mboudreau, line 8: you're not using `req.log` in this route...

src/routes/api/cryptor-migrations/migration-installation-verification.ts
gxueatlassian: fixed.

src/routes/api/cryptor-migrations/migration-installation-verification.ts
gxueatlassian: fixed.

src/routes/api/cryptor-migrations/migration-installation.ts
```diff
@@ -0,0 +1,65 @@
...
+	const endTime = Date.now();
+	const milliSecondsDelta = endTime - startTime;
+	return `${milliSecondsDelta} milli seconds`;
+};
+
+const getValidBatchSize = (req: Request): number => {
+	const num = parseInt((req.body || {}).batchSize);
+	if (isNaN(num)) return DEFAULT_BATCH_SIZE;
+	if (num < 0) return DEFAULT_BATCH_SIZE;
+	if (num > MAX_BATCH_SIZE) return DEFAULT_BATCH_SIZE;
```
mboudreau, line 63: just combine to a single conditional, no reason to triplicate the same return

src/routes/api/cryptor-migrations/migration-installation.ts
```diff
@@ -0,0 +1,65 @@
...
+	req.log.debug("Found matching installations to migrate, length " + installations.length);
+
+	//--------- loop and save each one of them --------------
+	let count = 0;
+	for (const inst of installations) {
+		//copy from shared secret to encryptedSharedSecret to enable encryption
+		inst.encryptedSharedSecret = inst.sharedSecret;
+		//update db
+		await inst.save();
+		req.log.info("Successfully migrated sharedSecret to encryptedSharedSecret");
```
mboudreau, line 37: don't add this, going to _flood_ the production logs...

src/routes/api/cryptor-migrations/migration-installation.ts
gxueatlassian: Don't like long boolean condition. 2 is the max number for me. 
But I accept your comments, fixed.

src/routes/api/cryptor-migrations/migration-installation.ts
```diff
@@ -0,0 +1,65 @@
...
+		if (++count % RESPONSE_BATCH_SIZE === 0) {
+			res.write(`Successfully migrated ${count} records now, still processing...`);
+		}
```
mboudreau, line 40: not sure how useful this really is

src/routes/api/cryptor-migrations/migration-installation.ts
gxueatlassian: Good point. fixed.
Do you think we should log the jiraHost when `inst.save()` throw an error?

src/routes/api/cryptor-migrations/migration-installation.ts
gxueatlassian: I am sure it will be useful if we want to go nuts with 10,000 a time, it takes 5 minutes. I don't want to stare at the screen for 5minutes and see nothing.

src/routes/api/api-router.ts
```diff
@@ -138,7 +139,8 @@ ApiRouter.use("/cryptor", async (_req: Request, resp: Response) => {
 		resp.status(500).send("fail");
 	}
 });
+ApiRouter.use("/migration", CryptorMigrationRouter);
 
 ApiRouter.use("/jira", ApiJiraRouter);
+//TODO: Following `use(/:installationId)` seems block any more routers to be used bellow, need fix
```
mboudreau, line 145: errant comment?

src/routes/api/cryptor-migrations/migration-installation.ts
mboudreau: in that case, instead of having a count variable, change your for loop to be a regular one? `for(let i = 0; i < installations.length; i++)`

gxueatlassian: This is really a minor personal preference. I am closing this. But please let me know if you think otherwise. I can re-open it.

src/routes/api/api-router.ts
mboudreau: ...yes, that's normal.  that's how routing works

mboudreau (approved):

src/routes/api/api-router.ts
gxueatlassian: Okay if you believe it is normal to prevent more routes underneath. We can always come back to this if need to.
For now I removed the comments per your suggestions. Resolve this convo now.

src/routes/api/cryptor-migrations/migration-installation.ts
gxueatlassian: @mboudreau can you give advice on above?

mboudreau: I mean, you're the one that provided the jiraHost...
I think you're fine.

mboudreau (approved):

PR #1361, opened 2022-07-12
Setting up Brahmos Wargames

rachellerathbone:
**What's in this PR?**

**Why**

**Added feature flags**

**Affected issues**  
_Jira Issues_

**How has this been tested?**  

**Whats Next?**


Dockerfile
```diff
@@ -15,3 +15,5 @@ WORKDIR /app
 RUN yarn install --frozen-lockfile
 
 CMD ["yarn", "start"]
+
+COPY --from=docker.atl-paas.net/sox/brahmos-deps/stress-ng:latest /usr/bin/stress-ng /usr/bin/stress-ng
```
mboudreau, line 19: this will fail outside of atlassian VPN unless this specific image is completely open to the world. 

If not, we might need to create a prod specific dockerfile that extends this one and adds bhramos.

mboudreau (changes requested):
We need to re-evaluate to only add brahmos on deployment

Dockerfile
thombergs: yeah, we'll need to create a new Dockerfile and use that Dockerfile in our deployment repo

thombergs: We're building a custom Docker image in our deployment pipeline, anyway, so I moved these changes to a specific Dockerfile called `Dockerfile.micros`. In our deployment pipeline we can then just use this new Dockerfile. 

It's not nice that we have two Dockerfiles then, but as long as they don't diverge too much, it should be fine.

thombergs (approved):

Dockerfile
mboudreau: It's either we do a intermediary build (build Dockerfile as "base", Dockerfile.prod uses "FROM" instruction) or we use [multistage build](https://docs.docker.com/develop/develop-images/multistage-build/#stop-at-a-specific-build-stage) to have a 'dev' stage and a 'prod' stage so we can keep everything in one dockerfile, but only use the dev stage one locally in docker-compose using the `target` attribute

mboudreau: Actually, we already have a `Dockerfile.prod`, just add Brahmos to that one.

thombergs: @mboudreau using `Dockerfile.prod` doesn't work because `Dockerfile.prod` is used in the GitHub actions and it would fail because it can't see the (Atlassian-internal) Brahmos Dockerfile. 

A multistage build doesn't work because the `target` parameter only supports to do everything **up to** a certain stage. What we want is **to skip** a stage so we don't try to copy from the internal Dockerfile when running in GitHub Actions.

I still think the simplest and most easily maintainable solution is to create a separate `Dockerfile.micros` for the Micros deployment. The only other option would be to create 3 Dockerfiles (one base, one for GitHub actions, and one for Micros) as you suggested, but that makes the scripts in both GitHub Actions and our deployment pipeline more complex and it's harder to understand for the next person looking at it. I'd rather just copy a few lines.

mboudreau: @thombergs this is not the case anymore, please move Dockerfile.micros to Dockerfile.prod :)

PR #1364, opened 2022-07-13
ARC-1372 unsafe stream

joshkay10:
**What's in this PR?**
Created a new stream class `RawLogStream` so that we can log out unsafe data to a secure environment for Splunking.

**Why**
First attempt was to create an unsafe logger and use that instance, which would write out the logs to the secure environment.
This worked but a major drawback was that it relied on devs to know about that option and then manually insert the logs they wanted.
With this method its more usable since we just need to toggle the feature flag and it will stream the logs to safe and unsafe environment.

**The catch**
Since we use a feature flag(**with JiraHost**) to determine whether we send the unsafe logs or not, this means we **must** have the jiraHost value as a log attribute, this value is pulled out of the logs data during the write process.

**Runbooks**
[How to Splunk unsafe data](url)

**Added feature flags**
[LOG_UNSAFE_DATA](https://app.launchdarkly.com/github-for-jira/development/features/log-unsafe-data/targeting)

**Affected issues**  
[ARC-1372](https://softwareteams.atlassian.net/browse/ARC-1372)

**How has this been tested?**  
Unit tests have been created for all new functionality.
Testing in stage to check that sensitive data is serialized or sent to unsafe environment.

**Whats Next?**
Thats all for now. Monitor and make sure this suits our needs.


src/config/logger.test.ts
```diff
@@ -1,64 +1,43 @@
-import { getLogger, getUnsafeLogger } from "config/logger";
```
joshkay10, line 1: This looks like a lot of changes in this file, but really I've just removed a bunch of tests for serialization and moved them to logger-utils.tests

src/config/logger.ts
```diff
@@ -43,45 +27,32 @@ const errorSerializer = (err) => err && {
 	request: requestSerializer(err.request)
 };
 
-const hashSerializer = (data: any): string | undefined => {
```
joshkay10, line 46: We serialize at the stream level now since we need access to async feature flags.

src/config/logger.ts
```diff
@@ -1,25 +1,9 @@
...
-
-// We cannot redefine the stream on middleware level (when we create the child logger),
-// therefore we have to do it here, on global level, for all loggers :(
-// And there's no way to disable those for webhooks, see:
-//   https://github.com/probot/probot/issues/1577
-//   https://github.com/probot/probot/issues/598
-//
 export const FILTERING_FRONTEND_HTTP_LOGS_MIDDLEWARE_NAME = "frontend-log-middleware";
-// add levelInString to include DEBUG | ERROR | INFO | WARN
-const LOG_STREAM = filteringHttpLogsStream(FILTERING_FRONTEND_HTTP_LOGS_MIDDLEWARE_NAME,
```
joshkay10, line 20: Replaced `LOG_STREAM` with `RawLogStream`

src/config/logger.ts
```diff
@@ -43,45 +27,32 @@ const errorSerializer = (err) => err && {
...
-	repoName: hashSerializer,
-	userGroup: hashSerializer,
-	aaid: hashSerializer,
-	username: hashSerializer
-});
-
 const logLevel = process.env.LOG_LEVEL || "info";
 const globalLoggingLevel = levelFromName[logLevel] || INFO;
 
+const loggerStream = (isUnsafe) => ({
```
joshkay10, line 33: Not really needed as a function, but needed to created 4 streams so made a little helper.
When probot goes will only have 2 streams so might remove it then.... maybe.

src/util/filtering-http-logs-stream.ts
```diff
@@ -1,34 +0,0 @@
...
- * @name filteringHttpLogsStream
- * @function
- * @param filteringLoggerName - the name of the logger to watch
- * @param out {Stream} original stream
- * @return {WritableStream} that you can pipe bunyan output into
- */
-export const filteringHttpLogsStream = (filteringLoggerName: string, out) => {
-	//TODO Remove this code when there will be convenient way to do it in Probot.
-	//See https://github.com/probot/probot/issues/1577
-	const shouldBeFiltered = (chunk: any): boolean => {
```
joshkay10, line 21: This has been nom nom nommed into `RawLogStream`

src/util/logger-utils.ts
```diff
@@ -0,0 +1,73 @@
...
+		const recordClone = { ...record };
+
+		// hash sensitive data or tag it unsafe
+		if (this.unsafeStream) {
+			this.tagUnsafeData(recordClone);
+		} else {
+			this.hashSensitiveData(recordClone);
+		}
+
+		const chunk = safeJsonStringify(recordClone) + "\n";
```
joshkay10, line 56: using `safeJsonStringify` because this is what bunyan is using, so trying to keep it samesies

src/util/logger-utils.ts
```diff
@@ -0,0 +1,73 @@
...
+		this.writeStream.write(chunk, encoding);
```
joshkay10, line 57: Originally had
```
this.writeStream.write(chunk, encoding, (err) => {
    console.error(err)
});
```

But.... we override the console logging with our logger, so an error on error would keep calling itself. Apparently infinite loops are frowned upon🤷‍♂️

src/util/logger-utils.ts
```diff
@@ -0,0 +1,72 @@
...
+
+	public constructor(filteredHttpLoggerName: string, isUnsafeStream = false) {
+		super({ objectMode: true });
+		this.filteredHttpLoggerName = filteredHttpLoggerName;
+		this.isUnsafeStream = isUnsafeStream;
+		this.writeStream = bformat({ outputMode, levelInString: true });
+	}
+
+	public async _write(record: any, encoding: BufferEncoding, next): Promise<void> {
+		const { booleanFlag, BooleanFlags } = await import("config/feature-flags");
```
joshkay10, line 34: Not a huge fan of importing this here, but it was the lesser of two evils.
Since feature flags depends on logger and logger on feature flags now... I can't import it up top since they depend on each other.

I can replace feature-flags logger with a standard bunyan logger... which will allow it to run fine and dandy, but it starts complicating tests. Am trying a few things still.

gxueatlassian (commented):
```
This worked but a major drawback was that it relied on devs to know about that option and then manually insert the logs they wanted.
```

Somehow I think that's a plus. So the mindset will be:
1. Never ever write any logs that contains sensitive data.
2. If breaking `1`, do `unsafe_logger.info(...)`.

joshkay10:
> 

@gxueatlassian, yeah ideally!
We have this always updating list `const SENSITIVE_DATA_FIELDS = ["jiraHost", "orgName", "repoName", "userGroup", "userGroup", "aaid", "username"];` from in `logger-utils` that will hash stuff before it makes it to logs as well

src/util/logger-utils.ts
gxueatlassian: But it will means `next()` will be called before callback is triggered since the `writeStream.write` return immediately.

How about 
```
this.writeStream.write(chunk, encoding, (err) => { 
   if(err) { 
     //donothing, just next
     next();
   } else {
      next();
   }
});
```

gxueatlassian: above is like 
```
this.writeStream.write(chunk, encoding, next);
```

but more explicit

gxueatlassian:
I want to confirm the logging logic, is it whenever FF is on, anything containing `jiraHost` will always go to two stream/storage?

src/util/logger-utils.ts
gxueatlassian: Agree that `import()` is not great. I agree having a standard bunyan logger just inside the feature-flag. 
Maybe we don't have to test the logging part within the feature flags, so it is be easier?

joshkay10: I'll have another crack at it and see what I can get.

The logger tests are fine, it breaks the featureflag tests because it tries to access it before its ready but I should be able to wait for launchdarkly... we shall see!!!

src/config/logger.ts
mboudreau: sounds good, just add a typing to `isUnsafe` :)

src/util/logger-utils.ts
mboudreau: Why are we using  a FF here at all?  This is a non-functional change and causes circular dependencies if we add FF.  Seems like a good argument to not have it behind an FF.  Plus, we want to start obfuscating the logs ASAP.

src/util/logger-utils.ts
```diff
@@ -0,0 +1,72 @@
...
+	}
+	return !!msg.match(/(GET|POST|DELETE|PUT|PATCH)/);
+};
+
+export class RawLogStream extends Writable {
+	private readonly  filteredHttpLoggerName: string;
+	private readonly isUnsafeStream: boolean;
+	private writeStream: NodeJS.WritableStream;
+
+	public constructor(filteredHttpLoggerName: string, isUnsafeStream = false) {
```
mboudreau, line 26: I'm not sure I understand the point of `isUnsafeStream`, shouldn't all streams be filtered for sensitive information?

src/util/logger-utils.ts
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	private tagUnsafeData(record: Record<string, any>): void {
+		record.env_suffix = "unsafe";
+	}
+
+	private hashSensitiveData(record: Record<string, any>): void {
+		Object.keys(record).forEach(key => {
+			if (SENSITIVE_DATA_FIELDS.includes(key)) {
+				record[key] = createHashWithSharedSecret(record[key]);
```
mboudreau, line 67: I think we're moving away from shared secret and instead using cryptor which is already in the codebase.  maybe encrypt it using that instead?

What do you think @gxueatlassian?

gxueatlassian: I think above code is okay in terms of how it use the "sharedSecret". This "sharedSecret" is not the one in `models/Installations`. This is using the `envVars.GLOBAL_HASH_SECRET`, which will be in `Vault` (not `Cryptor`).

mboudreau:
Sorry, but my logger changes have conflicted with this PR.  We should get together to work through it.

joshkay10:
> Sorry, but my logger changes have conflicted with this PR. We should get together to work through it.

Conflicts were pretty minor so not a lot of trouble fortunately.

Will still set up a pair to see this one through tho!

joshkay10:
Closing as changed the approach slightly, using logging level instead of feature flag. Was simpler to replay in a fresh branch than to unwind then make changes with the many conflicts.

PR #1365, opened 2022-07-13
ARC-1347 Read webhook secret from env. variable for GitHub cloud

Harminder84:
**What's in this PR?**
Pull webhook secret from env. variable for Github cloud and `GitHubServerApps` table for GHE

**Why**
GHE config data stored in `GitHubServerApps` where as Github cloud config in environment variables. 

**Added feature flags**
use the existing `ghe_server` flag

**Affected issues**  
ARC-1347

**How has this been tested?**

### DIFF
--- a/src/routes/github/webhook/webhook-receiver-post.ts
+++ b/src/routes/github/webhook/webhook-receiver-post.ts
@@ -8,6 +8,7 @@ import { WebhookContext } from "./webhook-context";
 import { webhookTimeout } from "~/src/util/webhook-timeout";
 import { issueCommentWebhookHandler } from "~/src/github/issue-comment";
 import { issueWebhookHandler } from "~/src/github/issue";
+import { envVars } from "~/src/config/env";
 
 export const WebhookReceiverPost = async (request: Request, response: Response): Promise<void> => {
 	const logger = getLogger("webhook.receiver");
@@ -18,14 +19,16 @@ export const WebhookReceiverPost = async (request: Request, response: Response):
 	const payload = request.body;
 	let webhookSecret: string;
 	try {
-		const gitHubServerApp = await GitHubServerApp.findForUuid(uuid);
-		if (!gitHubServerApp) {
-			response.status(400).send("GitHub app not found");
-			return;
+		if (uuid) {
+			const gitHubServerApp = await GitHubServerApp.findForUuid(uuid);
+			if (!gitHubServerApp) {
+				response.status(400).send("GitHub app not found");
+				return;
+			}
+			webhookSecret = await gitHubServerApp.decrypt("webhookSecret");
+		} else {
+			webhookSecret = envVars.WEBHOOK_SECRET;
 		}
-
-		webhookSecret = await gitHubServerApp.decrypt("webhookSecret");
-
 		const verification = createHash(JSON.stringify(payload), webhookSecret);
 		if (verification != signatureSHA256) {
 			response.status(400).send("signature does not match event payload and secret");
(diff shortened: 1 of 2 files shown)


src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -18,14 +19,16 @@ export const WebhookReceiverPost = async (request: Request, response: Response):
 	const payload = request.body;
 	let webhookSecret: string;
```
joshkay10, line 20: /nit - This is more personal preference than anything for me, but I would split the getWebhookSecret into a seperate function.. Couple reasons.

Its prettier  ¯\_(ツ)_/¯.... maybe, but its much easier for me to read this way
Single responsibility, if we need it again somewhere else(unlikely since its specifically webhook) we can reuse it pretty easy.
This would catch the undefined envVars.WEBHOOK_SECRET, non-existant value on the gitHubServerApp table and gitHubServerApp not found.


```
const getWebhookSecret = async (uuid?: string): string => {
      // Return GitHub cloud secret
      if(!uuid) {
          return envVars.WEBHOOK_SECRET;
      }
      
      // Return GitHub enterprise secret
			const gitHubServerApp = await GitHubServerApp.findForUuid(uuid);
			return gitHubServerApp && await gitHubServerApp.decrypt("webhookSecret");
}
```

```
export const WebhookReceiverPost = async (request: Request, response: Response): Promise<void> => {
....
const webhookSecret = await getWebhookSecret(uuid);

if (!webhookSecret) {
    response.status(400).send("guts");
    return;
}


```

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -71,3 +68,11 @@ const createHash = (data: BinaryLike, secret: string): string => {
 		.update(data)
 		.digest("hex")}`;
 };
+
+const getWebhookSecret = async (uuid?: string) => {
+	if (uuid) {
+		const gitHubServerApp = await GitHubServerApp.findForUuid(uuid);
+		return gitHubServerApp && await gitHubServerApp.decrypt("webhookSecret");
```
mboudreau, line 75: This should probably throw an error if `GitHubServerApp` isn't found.

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -71,3 +68,11 @@ const createHash = (data: BinaryLike, secret: string): string => {
...
+	}
+	return envVars.WEBHOOK_SECRET;
```
mboudreau, line 77: Should probably do a check here as well to see if webhook secret is defined.  if not, throw error.

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -16,16 +17,12 @@ export const WebhookReceiverPost = async (request: Request, response: Response):
 	const id = request.headers["x-github-delivery"] as string;
 	const uuid = request.params.uuid;
 	const payload = request.body;
-	let webhookSecret: string;
 	try {
-		const gitHubServerApp = await GitHubServerApp.findForUuid(uuid);
-		if (!gitHubServerApp) {
+		const webhookSecret = await getWebhookSecret(uuid);
+		if (!webhookSecret) {
```
mboudreau, line 22: instead of checking if webhookSecret is defined, the function should throw if it doesn't, which will be caught in this try catch. However, I don't think we should return a 500.  Instead, it should return a 400 or a 404

PR #1366, opened 2022-07-13
ARC-1458 - GHE UI: Navigate from server url screen to select GH app screen

krazziekay:
**What's in this PR?**
- New module added in app connector.
- New route and methods added to fetch the list of GHE servers for a JIRA instance.
- Fetching the actual list of GHE servers in the view.
- Added `pathNameForAddNew` for navigating back to jira-server-url page.

**Video**

https://user-images.githubusercontent.com/13076255/178638525-597c6512-a985-492b-b635-668062922aaa.mov

### DIFF
--- a/src/routes/jira/jira-router.ts
+++ b/src/routes/jira/jira-router.ts
@@ -7,6 +7,7 @@ import { JiraSelectVersionRouter } from "./server/jira-select-version-router";
 import { JiraContextJwtTokenMiddleware } from "middleware/jira-jwt-middleware";
 import { JiraServerUrlRouter } from "routes/jira/server/jira-server-url-router";
 import { JiraAppCreationRouter } from "./server/jira-app-creation-router";
+import { JiraGheServersRouter } from "routes/jira/server/jira-ghe-servers-router";
 
 export const JiraRouter = Router();
 
@@ -18,3 +19,4 @@ JiraRouter.use("/events", JiraEventsRouter);
 JiraRouter.use("/select-version", JiraSelectVersionRouter);
 JiraRouter.use("/server-url", JiraServerUrlRouter);
 JiraRouter.use("/app-creation", JiraAppCreationRouter);
+JiraRouter.use("/ghe-servers", JiraGheServersRouter);
--- a/src/routes/jira/jira-router.ts
+++ b/src/routes/jira/jira-router.ts
@@ -4,14 +4,17 @@ import { JiraSyncPost } from "./jira-sync-post";
 import { JiraAtlassianConnectGet } from "./jira-atlassian-connect-get";
 import { JiraEventsRouter } from "./events/jira-events-router";
 import { JiraSelectVersionRouter } from "./server/jira-select-version-router";
-import { JiraContextJwtTokenMiddleware } from "middleware/jira-jwt-middleware";
+import { JiraContextJwtTokenMiddleware, JiraJwtTokenMiddleware } from "middleware/jira-jwt-middleware";
 import { JiraServerUrlRouter } from "routes/jira/server/jira-server-url-router";
 import { JiraAppCreationRouter } from "./server/jira-app-creation-router";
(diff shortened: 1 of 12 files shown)


src/routes/jira/server/jira-ghe-servers-router.ts
```diff
@@ -0,0 +1,9 @@
+import { Router } from "express";
+import { JiraJwtTokenMiddleware } from "middleware/jira-jwt-middleware";
+import { csrfMiddleware } from "middleware/csrf-middleware";
+import { JiraGheServers } from "routes/jira/server/jira-ghe-servers";
+
+export const JiraGheServersRouter = Router();
+
```
joshkay10, line 7: Worth moving these to a .use i reckon
 ```JiraGheServersRouter.use(csrfMiddleware)```
 ```JiraGheServersRouter.use(JiraJwtTokenMiddleware)```

src/routes/jira/server/jira-ghe-servers.ts
```diff
@@ -0,0 +1,24 @@
...
+export const JiraGheServers = async (
+	req: Request,
+	res: Response,
+	next: NextFunction
+): Promise<void> => {
+	try {
+		req.log.debug("Received Jira GHE servers page request");
+
+		const allServers = await GitHubServerApp.findForInstallationId(res.locals.installation.id) || [];
+		const gheServers = chain(groupBy(allServers, "gitHubBaseUrl")).map((_, key) => ({ displayName: key })).value();
```
joshkay10, line 14: Theres a lot going on. I would consider simplifying this, even if its extra code.

Are you using the index as the displayName intentionally?

krazziekay: Firstly, I am grouping the list of GHE servers based on their `gitHubBaseUrls`, since there could be multiple GHE server apps with same gitHubBaseUrl for a single instance. Then doing a map to just fetch the urls. Are you referring to break this into multiple lines?
Yes, I am mapping the index to `displayName` intentionally because the UI component that will receive this data is reusable component. It will be later used again for similar purpose. So, I picked `displayName` as a general index, to make it easy to understand.

src/routes/jira/server/jira-ghe-servers-router.ts
krazziekay: For this new route, I'm simply following the structure of the other existing routes. So, I'm not sure which ones the correct way.

src/routes/jira/jira-router.ts
```diff
@@ -18,3 +19,4 @@ JiraRouter.use("/events", JiraEventsRouter);
 JiraRouter.use("/select-version", JiraSelectVersionRouter);
 JiraRouter.use("/server-url", JiraServerUrlRouter);
 JiraRouter.use("/app-creation", JiraAppCreationRouter);
+JiraRouter.use("/ghe-servers", JiraGheServersRouter);
```
mboudreau, line 22: we don't have a create a new router for every new path that only has a `GET` call on it.  We can just do `JiraRouter.get("/ghe-servers", csrfMiddleware, JiraJwtTokenMiddleware, JiraGheServers)` here

krazziekay: Updated to this and removed the router file.

src/routes/jira/server/jira-ghe-servers-router.ts
krazziekay: The router file has been deleted and used `get` in the JiraRouter.

src/routes/jira/server/jira-ghe-servers.ts
mboudreau: @joshkay10 it was already complicated before he started. I went through it with him and simplifying this function would introduce more risk than needed.

We definitely need to simplify, but I think should be done as a separate PR after we release GHE support

mboudreau (approved):
looking good :+1:

joshkay10 (approved):

PR #1367, opened 2022-07-14
ARC-1511: GHE UI: Update manual form to allow for editing

krazziekay:
**What's in this PR?**
- New page added in connect-app descriptor.
- New route added for manual app page.
- Values displayed in Manual app form.
- Snapshots updated.

**Whats Next?**
- The private key file needs to be pulled into the manual app form.
- The link to this page needs to be added next. Once ARC-1438 is done, the edit in the dropdown for GHE server will need to be linked with this page. 
  - Link can be added by using the following code:
````
AP.navigator.go("addonmodule", {
  moduleKey: "github-edit-app-page",
  customData: { ghsaId: 'XXXXX' }
});
````

**Screenshot**
image

### DIFF
--- a/src/routes/jira/jira-atlassian-connect-get.ts
+++ b/src/routes/jira/jira-atlassian-connect-get.ts
@@ -99,6 +99,15 @@ const modules = {
 			url: "/jira/manual-app?ghsaId={ac.ghsaId}",
 			location: "none",
 			conditions: adminCondition
+		},
+		{
+			key: "github-list-servers-page",
+			name: {
+				value: "GitHub Enterprise Servers"
+			},
+			url: "/jira/ghe-servers",
+			location: "none",
+			conditions: adminCondition
 		}
 	],
 	webSections: [
--- a/src/routes/jira/jira-atlassian-connect-get.ts
+++ b/src/routes/jira/jira-atlassian-connect-get.ts
@@ -92,11 +92,11 @@ const modules = {
 			conditions: adminCondition
 		},
 		{
-			key: "github-manual-app-creation-page",
+			key: "github-manual-app-page",
 			name: {
-				value: "GitHub Manual App Creation"
+				value: "GitHub Manual App"
 			},
-			url: "/jira/manual-app-creation?ghsaId={ac.ghsaId}",
+			url: "/jira/manual-app?ghsaId={ac.ghsaId}",
 			location: "none",
 			conditions: adminCondition
 		}
--- a/src/routes/jira/jira-atlassian-connect-get.ts
+++ b/src/routes/jira/jira-atlassian-connect-get.ts
@@ -90,6 +90,15 @@ const modules = {
 			url: "/jira/app-creation",
 			location: "none",
 			conditions: adminCondition
+		},
+		{
+			key: "github-manual-app-creation-page",
+			name: {
+				value: "GitHub Manual App Creation"
+			},
+			url: "/jira/manual-app-creation?ghsaId={ac.ghsaId}",
+			location: "none",
+			conditions: adminCondition
 		}
 	],
 	webSections: [
--- a/src/routes/jira/jira-atlassian-connect-get.ts
+++ b/src/routes/jira/jira-atlassian-connect-get.ts
@@ -92,11 +92,11 @@ const modules = {
 			conditions: adminCondition
 		},
 		{
-			key: "github-manual-app-page",
+			key: "github-edit-app-page",
 			name: {
-				value: "GitHub Manual App"
+				value: "GitHub Edit App"
 			},
-			url: "/jira/manual-app?ghsaId={ac.ghsaId}",
+			url: "/jira/edit-app/{ac.ghsaId}",
 			location: "none",
 			conditions: adminCondition
 		},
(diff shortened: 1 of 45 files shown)

### CODE REVIEW

File: src/routes/jira/jira-atlassian-connect-get.ts [hunk 1]
```diff
@@ -90,6 +90,15 @@ const modules = {
 			url: "/jira/app-creation",
 			location: "none",
 			conditions: adminCondition
+		},
+		{
+			key: "github-manual-app-creation-page",
+			name: {
+				value: "GitHub Manual App Creation"
+			},
+			url: "/jira/manual-app-creation?ghsaId={ac.ghsaId}",
```
mboudreau, line 99: if we're creating the app, why is there an id already generated for it?

krazziekay: This page is for both creating and editing. Updating the name to just `manual-app`.

src/routes/jira/jira-router.ts
```diff
@@ -4,13 +4,17 @@ import { JiraSyncPost } from "./jira-sync-post";
...
+import { JiraContextJwtTokenMiddleware, JiraJwtTokenMiddleware } from "middleware/jira-jwt-middleware";
 import { JiraServerUrlRouter } from "routes/jira/server/jira-server-url-router";
 import { JiraAppCreationRouter } from "./server/jira-app-creation-router";
+import { csrfMiddleware } from "middleware/csrf-middleware";
+import { JiraManualAppCreationGet } from "routes/jira/server/jira-manual-app-creation-get";
 
 export const JiraRouter = Router();
 
 JiraRouter.get("/atlassian-connect.json", JiraAtlassianConnectGet);
+JiraRouter.get("/manual-app-creation", csrfMiddleware, JiraJwtTokenMiddleware, JiraManualAppCreationGet);
```
mboudreau, line 16: we need to have a chat around all these new routes as they're related but not associated through routing and using a few anti-patterns

krazziekay: We can setup a meeting for this I suppose? Do we hold this PR until then?

mboudreau: yeah, maybe just a quick one over huddle with whoever is interested in this in the ghe channel?

src/routes/jira/jira-atlassian-connect-get.ts
```diff
@@ -90,6 +90,15 @@ const modules = {
 			url: "/jira/app-creation",
 			location: "none",
 			conditions: adminCondition
+		},
+		{
+			key: "github-manual-app-page",
```
rachellerathbone, line 95: We shouldn't have the word 'manual' in this path at all as this editing page is for any GHES app regardless of how it was created. If a customer automatically created an app but rotates their GH client secret, they are going to land on this page when they click the 'Edit' option from the dropdown. We are simply reusing the form but it has nothing to do with 'manual' in this part of the flow. 

- [ ] remove all references to the word 'manual' and replace with 'edit'

views/jira-manual-app-creation.hbs
```diff
@@ -35,7 +35,14 @@
 
       <h1>Manual app Creation</h1>
```
rachellerathbone, line 36: The title isn't 'Manual app creation' for the editing page. Per ticket description: `Manual app creation > Edit <github app name>` which comes from the Figma design.

Screen Shot 2022-07-14 at 4 22 43 pm

rachellerathbone (changes requested):

src/routes/jira/jira-atlassian-connect-get.ts
krazziekay: Updated the names from manual to edit

views/jira-manual-app-creation.hbs
krazziekay: Corrected

src/routes/jira/jira-router.ts
krazziekay: The URL has been updated this one only...

views/jira-manual-app-creation.hbs
```diff
@@ -33,9 +33,22 @@
...
   <div class="jiraManualAppCreation">
 
-      <h1>Manual app Creation</h1>
-
-      <div>Enter all the fields to create your app. <a href="#">Learn more about manual GitHub app creation.</a></div>
+      <h1>
+      {{#if app.id}}
+        Edit {{app.gitHubAppName}}
+      {{else}}
+        Manual app Creation
```
rachellerathbone, line 40: Nit: should be a lowercase c

krazziekay: C corrected

src/routes/jira/server/jira-edit-app-get.ts
```diff
@@ -0,0 +1,24 @@
...
+export const JiraEditAppGet = async (
+	req: Request,
+	res: Response,
+	next: NextFunction
+): Promise<void> => {
+	try {
+		req.log.debug("Received Jira edit app page request");
+
+		const gitHubServerAppId = req.params.id as string;
+		const app = await GitHubServerApp.getForGitHubServerAppId(parseInt(gitHubServerAppId));
```
gxueatlassian, line 13: @mboudreau , so there's line here for looks for `GitHubServerApp`. Tagging you because we had a chat this morning. Do you think we should preserve an app instance in `res.locals.gitHubServerApp`? So that we don't need to call to db multi times. Any route under `/github` will most likely need it anyway.

As for the code inconsistency, I believer only place will be when you create the `gitHubAppClient`, which is only few places.

gxueatlassian: And it is NOT related to this PR, sorry K, closing this comment now.

rachellerathbone (approved):

PR #1368, opened 2022-07-14
Arc 1515 refactor GitHub app middleware

gxueatlassian:
**What's in this PR?**
Having a single middleware to extract github xxx id

**Why**
Stay DRY, not having to repeat to import the GitHubServerAppMiddlware every wher.e

**Added feature flags**
N/A

**Affected issues**  
ARC-1515

**How has this been tested?**  
N/A

**Whats Next?**
N/A

### DIFF
--- a/src/middleware/github-server-app-middleware.ts
+++ b/src/middleware/github-server-app-middleware.ts
@@ -2,12 +2,23 @@ import { NextFunction, Request, Response } from "express";
 import { GitHubServerApp } from "models/github-server-app";
 import { Installation } from "models/installation";
 
+type GitHubAppConfig = {
+	gitHubAppId: string;
+}
+
+export type GitHubAppReqLocals<T = Record<string, any>> = T & {
+	gitHubAppConfig: {
+		gitHubAppId: number | undefined;
+	}
+}
+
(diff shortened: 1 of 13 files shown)


src/middleware/github-server-app-middleware.ts
```diff
@@ -2,12 +2,20 @@ import { NextFunction, Request, Response } from "express";
 import { GitHubServerApp } from "models/github-server-app";
 import { Installation } from "models/installation";
 
+export type GitHubAppReqLocals<T = Record<string, any>> = T & {
+	gitHubAppConfig: {
+		gitHubAppId: number | undefined;
```
gxueatlassian, line 7: First thing first, which one are we using?

mboudreau: using for what?

src/middleware/github-server-app-middleware.ts
```diff
@@ -2,12 +2,20 @@ import { NextFunction, Request, Response } from "express";
...
+		uuid: string;
```
mboudreau, line 8: uuid doesn't exist for cloud app

PR #1370, opened 2022-07-15
gArc 1515 share GitHub server app middleware

gxueatlassian:
**What's in this PR?**
Move the GithubServerAppMiddleware to above to support GHE use case.

**Why**
Many github routes need the gitHubAppId (for now), need to use a shared route

**Added feature flags**
N/A

**Affected issues**  
ARC-1515

**How has this been tested?**  
Unit test.

**Whats Next?**
1. Figure out how to make use of it in the frontend and create PR.
2. Refactor once db schema is done, to use uuid instead of app id.

### DIFF
--- a/src/middleware/github-server-app-middleware.ts
+++ b/src/middleware/github-server-app-middleware.ts
@@ -6,33 +6,27 @@ export const GithubServerAppMiddleware = async (req: Request, res: Response, nex
 	const { jiraHost } = res.locals;
 	const { ghaid: id } = req.query;
 
-	if (!id) return next();
+	if (id) {
+		req.log.debug(`Retrieving GitHub app with id ${id}`);
+		const gitHubServerApp = await GitHubServerApp.getForGitHubServerAppId(Number(id));
 
-	//TODO: ARC-1515 fix it to parse uuid xxx-xxx-xxx
-	const isNonRecognizableId = isNaN(parseInt(id as string));
-	if (isNonRecognizableId) {
-		console.log("wrong path, just return;");
-		next();
-		return;
+		if (!gitHubServerApp) {
+			req.log.error({ id, jiraHost }, "No GitHub app found for provided id.");
+			throw new Error("No GitHub app found for provided id.");
+		}
+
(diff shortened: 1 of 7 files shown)


src/middleware/github-server-app-middleware.ts
```diff
@@ -4,11 +4,17 @@ import { Installation } from "models/installation";
 
 export const GithubServerAppMiddleware = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
 	const { jiraHost } = res.locals;
-	const { id } = req.params;
+	const { id: idStr } = req.query;
```
mboudreau, line 7: This has to be a param

src/middleware/github-server-app-middleware.ts
```diff
@@ -4,11 +4,17 @@ import { Installation } from "models/installation";
...
+	const id = parseInt(idStr as string);
+
+	if (id && isNaN(id)) {
```
mboudreau, line 10: this should probably be uuid, but I guess we could use githubserverappid

src/middleware/github-server-app-middleware.ts
```diff
@@ -4,11 +4,17 @@ import { Installation } from "models/installation";
...
+		req.log.warn({ id, jiraHost }, "Provided id found but not a number " + id);
```
mboudreau, line 11: Don't add dynamic vars to the log, makes aggregation impossible. It's already part of the data.

src/routes/github/github-oauth-router.ts
```diff
@@ -172,5 +172,6 @@ export const GithubAuthMiddleware = async (req: Request, res: Response, next: Ne
 // IMPORTANT: We need to keep the login/callback/middleware functions
 // in the same file as they reference each other
 export const GithubOAuthRouter = Router();
+//TODO: ARC-1515 confirm this is not used at all.
```
mboudreau, line 175: ?

mboudreau (changes requested):
Not the agreed upon URL convention we wanted

PR #1371, opened 2022-07-15
Arc 1515 share GitHub server app middleware  params

gxueatlassian:
**What's in this PR?**

Adding an optional /:uuid to all github route, so that:
1. For cloud, it still can be /github/configuration
2. For GHE, it will be /github/:uuid/configuration

**What's next?**
1. Update many places to use uuid to look for GitHubServerApp
2. Fix up the frontend pages/urls to have :uuid in the path during navigation.


src/routes/github/github-router.ts
```diff
@@ -7,27 +7,35 @@ import { GithubConfigurationRouter } from "routes/github/configuration/github-co
 import { returnOnValidationError } from "../api/api-utils";
 import { header } from "express-validator";
 import { WebhookReceiverPost } from "./webhook/webhook-receiver-post";
+import { GithubServerAppMiddleware } from "middleware/github-server-app-middleware";
 
 export const GithubRouter = Router();
 
+const GithubAppProvidedRouter = Router({ mergeParams: true });
+GithubRouter.use("/:appId?", GithubAppProvidedRouter);
```
gxueatlassian, line 15: With this optional path params, when test the cloud scenario, express will match the `:appId` with `/configuration` from the original url `/github/configuration`.

But since it is matching the appId with 'configuration', the 'configuration' is remove from the `req.path`, therefore it wont' match the following `GitHubConfigurationGet` route.

image

mboudreau: Added a commit with an example of how it should work.  I didn't test it at all.

src/middleware/github-server-app-middleware.ts
```diff
@@ -1,31 +1,52 @@
...
 
 		if (!gitHubServerApp) {
-			req.log.error({ id, jiraHost }, "No GitHub app found for provided id.");
+			req.log.error("No GitHub app found for provided uuid.");
 			throw new Error("No GitHub app found for provided id.");
 		}
-
 		const installation = await Installation.findByPk(gitHubServerApp.installationId);
-
 		if (installation?.jiraHost !== jiraHost) {
```
gxueatlassian, line 24: This already does the installationId check for jiraHost. We can trust the uuid.

src/middleware/github-server-app-middleware.ts
```diff
@@ -1,31 +1,52 @@
...
-			req.log.error({ id, jiraHost }, "Jira hosts do not match");
+			req.log.error({ uuid, jiraHost }, "Jira hosts do not match");
 			throw new Error("Jira hosts do not match.");
 		}
-
 		req.log.info("Found GitHub server app for installation");
-		res.locals.gitHubAppId = id;
-		return next();
+		res.locals.gitHubAppId = gitHubServerApp.appId;
```
gxueatlassian, line 29: This is for compatibility for GHE pages. Even though they are not available for public yet, I don't want to break those pages or have a huge PR to refactor those.

src/routes/github/github-router.ts
```diff
@@ -7,27 +7,37 @@ import { GithubConfigurationRouter } from "routes/github/configuration/github-co
...
 import { header } from "express-validator";
 import { WebhookReceiverPost } from "./webhook/webhook-receiver-post";
+import { GithubServerAppMiddleware } from "middleware/github-server-app-middleware";
 
-export const GithubRouter = Router();
+const UUID_REGEX = "[0-9a-fA-F]{8}\\b-[0-9a-fA-F]{4}\\b-[0-9a-fA-F]{4}\\b-[0-9a-fA-F]{4}\\b-[0-9a-fA-F]{12}";
+
+export const GithubRouter = Router({ mergeParams: true });
+const subRouter = Router({ mergeParams: true });
+GithubRouter.use(`/:uuid(${UUID_REGEX})?`, subRouter);
```
gxueatlassian, line 16: better put the `/:uuid(regex)?` here instead of in the rooter since this logic is specific to the github router.

src/middleware/github-server-app-middleware.ts
```diff
@@ -1,31 +1,54 @@
...
 	const { jiraHost } = res.locals;
-	const { id } = req.params;
+	const { uuid } = req.params;
+
+	req.log = req.log.child({ uuid, jiraHost });
 
-	if (id) {
-		req.log.debug(`Retrieving GitHub app with id ${id}`);
-		const gitHubServerApp = await GitHubServerApp.getForGitHubServerAppId(Number(id));
+	if (uuid && validationResult(req).isEmpty()) {
```
joshkay10, line 14: Could `validationResult(req).isEmpty()` include other validation errors from up the chain?

gxueatlassian: Good call! 
I suspect we can simply remove the `validationResult` here since this middleware doesn't specific any validation rules, so nor should it validate the result. 
Maybe the upper router should do the check.
Let me come back to you.

gxueatlassian: @joshkay10 , I've removed the `validationResult.isEmtpy()` here because:
1. There's no `param("uuid").isXXX()` settings in this middleware, therefore I don't think we need any validation in this middleware. Not its job.
2. In fact the validation is set one level up in the router via regex. Express will direct the traffic to 404 if uuid not valid. I think it make sense in this case, so nothing else we need to do.
3. I've added one test to test the uuid invalid in upper level router. 

Please let me know what you thinks. Thanks :)

joshkay10: Yeah nice, I agree, bit tidier.

3) test. nice!

src/middleware/github-server-app-middleware.ts
```diff
@@ -1,31 +1,53 @@
...
+	} else {
+		// is cloud app
+		res.locals.gitHubAppConfig = {
+			appId: envVars.APP_ID,
+			uuid: undefined, //undefined for cloud
+			clientId: envVars.GITHUB_CLIENT_ID,
+			gitHubClientSecret: envVars.GITHUB_CLIENT_SECRET,
+			webhookSecret: envVars.WEBHOOK_SECRET
+			// once keylocator is merged
+			// privateKey: keyLocator
```
mboudreau, line 47: keylocator has been merged.

src/routes/github/configuration/github-configuration-router.route.test.ts
```diff
@@ -0,0 +1,30 @@
...
+			jest.mocked(GithubConfigurationGet).mockImplementation(async (_req, res) => {res.end("ok");});
+			await supertest(app).get("/configuration");
+			expect(GithubConfigurationGet).toHaveBeenCalled();
+		});
+		it("should hit the configuration on POST", async () => {
+			jest.mocked(GithubConfigurationPost).mockImplementation(async (_req, res) => {res.end("ok");});
+			await supertest(app).post("/configuration");
+			expect(GithubConfigurationPost).toHaveBeenCalled();
+		});
+	});
```
mboudreau, line 28: What are you trying to test here?

src/routes/github/github-router.test.ts
```diff
@@ -0,0 +1,173 @@
...
+import { v4 as v4uuid } from "uuid";
+import { envVars } from "config/env";
+
+jest.mock("./configuration/github-configuration-get");
+jest.mock("utils/get-github-client-config");
+jest.mock("axios", () => ({ get: jest.fn(), create: jest.fn() }));
+jest.mock("models/github-server-app");
+jest.mock("models/installation");
+
+const JIRA_HOST = "https://gh4j.test.atlassian.net";
```
mboudreau, line 20: use `jiraHost` global unless you have a need to use something different.

src/routes/github/github-router.test.ts
```diff
@@ -0,0 +1,173 @@
...
+const VALID_TOKEN = "valid-token";
+const GITHUB_SERVER_APP_UUID: string = v4uuid();
+const GITHUB_SERVER_APP_ID = 789;
+const GITHUB_SERVER_APP_URL = "https://default.internal.github.atlassian.net";
```
mboudreau, line 24: use global `gheUrl` unless there's a reason not to

src/routes/github/github-router.test.ts
```diff
@@ -0,0 +1,173 @@
...
+};
+
+const mockConfigurationGetProceed = ()=>{
+	jest.mocked(GithubConfigurationGet).mockClear();
+	jest.mocked(GithubConfigurationGet).mockImplementation(async (_req, res) => {
+		res.end("ok");
+	});
+};
+
+describe("GitHub router", () => {
```
mboudreau, line 101: this whole suite of test is only testing the middleware from what I can tell and not the route handler, which means that the route handler dependency to `res.locals.githubAppId` or `res.locals.githubAppConfig` is not tested.

src/routes/github/github-router.ts
mboudreau: personally, I'm not a fan of using multiple routers within a router. The argument you're making can be said about specifying `/github` outside of the github router

mboudreau (changes requested):
The tests needs improvement

src/routes/github/configuration/github-configuration-router.route.test.ts
gxueatlassian: The existing `github-configuration-router.test.ts` is not testing specific to this `github-configuration-router`. It is testing the `getFrontendApp()` from the very root level. I don't think that test is appropriate to put any more unit test inside.

So in this file, I purely focus on the unit test, which is just the route.

src/routes/github/github-router.test.ts
gxueatlassian: So the changes to the `github-route.ts` is just added an optional `/:uuid`, therefore this unit test is to test that `/:uuid` is:
1. Optional --> the first "Cloud scenario"
2. When present, uuid is picked up and parsed --> I use one `GithubConfigurationGet` route to test it and assert the locals are expected.

src/routes/github/github-router.ts
gxueatlassian: I agree this can be interpreted in many way. I am closing this thread, since:
1. It doesn't have bugs.
2. It doesn't have performance/security/etc issue.
3. It doesn't harm code clarity ( in my view improve though)

Let me know if you wish to re-open it.

src/middleware/github-server-app-middleware.ts
gxueatlassian: fixed.

src/routes/github/configuration/github-configuration-router.route.test.ts
mboudreau: except, you're not testing anything except how you configured the express server _in your own test_.  That's unnecessary.  The reason we test the `getFrontendApp()` is because it contains all routes.  If someone were to change the route you're trying to test, it would actually break the test.  In this one, if you were to change the configuration route to `/configuration2` in the router, this test wouldn't break.

src/middleware/github-server-app-middleware.test.ts
```diff
@@ -25,23 +29,23 @@ describe("github-server-app-middleware", () => {
...
 
-	it("should call next() when no gitHupAppId is provided",  async() => {
+	it("should call next() when no uuid is provided",  async() => {
 		req = {
 			log: getLogger("request"),
 			params: {
-				id: undefined
+				uuid: undefined
 			}
 		};
```
mboudreau, line 38: this could be made into a function to reduce duplication

src/routes/github/configuration/github-configuration-router.route.test.ts
gxueatlassian: 1. This is test is just a "unit test", not an "integration test". 
2. This unit test test ONLY the file `github-configuration-router.ts`.
3. `github-configuration-router.ts` doesn't contains any route like `/configuration`, it just contains "/" and points to the get and post handler.

I think this test is a valid unit test. 

I agree this test is not an integration test. For example, to test the path `/configuration`, it should be somewhere else, and it will be in another PR.

src/middleware/github-server-app-middleware.test.ts
gxueatlassian: It is a trade-off, either make refactoring on the code to make it better, or keep the code as it is to make pull request more readable. I chose the later in this case.

But if you strongly suggest I should refactor the code in this case, i can do that as well.

src/routes/github/configuration/github-configuration-router.route.test.ts
mboudreau: Testing the router _is_ an integration test as you're testing your config in the framework and not a single function's input/output. This test doesn't give you any extra information and if anything, gives you false confidence that everything works fine when it shouldn't.

Lean into testing the whole router without mocking and then unit test the route handler function itself for full confidence.

src/routes/github/github-router.test.ts
mboudreau: I understand that, what I'm saying is that this is a test for the router which should be tested thoroughly, which includes the route handler itself.  What you seem to be doing is testing the router and route handler  separately, which means that there's no test for them working correctly together.

When it comes to testing routes, integration tests are the only way to get full confidence.  You can unit test route handlers for input/output, but you cannot do this with routers without mocking important parts of the flow which may change in the future and break in production.

src/models/github-server-app.ts
```diff
@@ -65,7 +65,7 @@ export class GitHubServerApp extends EncryptedModel {
 
 		return this.findOne({
 			where: {
-				id: gitHubServerAppId
+				appId: gitHubServerAppId
```
gxueatlassian, line 68: @mboudreau 
I got a feeling the previous usage of `id` column is incorrect. We should be using `appId`, as it is the one that github recognised. 

Has to change this to make the `github-route.test.ts` works anyway.

src/routes/github/github-router.test.ts
gxueatlassian: Please review, this is the discussed integration test. But only testing new features in this PR.

gxueatlassian: And I didn't use the real `GitHubConfigurationGet` and assert on the handlebars output, as I think that would be too much for the test.

src/models/github-server-app.ts
rachellerathbone: No it's meant to be id. When we call `await GitHubServerApp.getForGitHubServerAppId(gitHubAppId);` we are after the unique `id` not the `appId` as appIds may be repeated across servers.

mboudreau: I know it's the one that github recognizes, but it's not unique in the GHE table as you can have collisions between appIds for each server. We need to use the PK for this.

mboudreau: it's just to get the object, then when you do a call out to the github api, you get the appId, but you're also calling a specific server

rachellerathbone: No, it's meant to be `id`. When we call `await GitHubServerApp.getForGitHubServerAppId(gitHubAppId);` we are passing in the unique `id` that is created when an entry in added to the table. `appId` is not unique as it may be repeated across servers. Happy for you to rename that param if you think it will make it clear but it's meant to be referencing the id of the target entry in the GitHub Server App table (hence gitHubServerAppId).

src/github/client/key-locator.test.ts
```diff
@@ -20,7 +20,7 @@ describe("key-locator", () => {
 			installationId: 1
 		});
 
-		const privateKey = await keyLocator(gitHubServerApp.id);
+		const privateKey = await keyLocator(gitHubServerApp.appId);
```
mboudreau, line 23: This is the wrong property, the old one was correct

rachellerathbone, line 23: Again, this needs to be the unique id, not the potentially repeated appId.

src/models/github-server-app.ts
gxueatlassian: Okay, thanks guys, yeah, I was confused about the `appId` in `GitHubServerApp` table is not the `gitHubAppId` everywhere. I'll revert the change of this bit, think about the naming in another PR.

gxueatlassian: Reverted and fixed. Closing this thread.

src/github/client/key-locator.test.ts
gxueatlassian: Reverted and fixed. Closing this thread.

gxueatlassian: Reverted and fixed. Closing this thread.

mboudreau (approved):

PR #1372, opened 2022-07-15
ARC-736 improving .env file structure

mboudreau:
**What's in this PR?**
Improving the dotenv structure to allow environment based env files and local files for user only variables to make development easier.

**Why**
Needed to create certain variables that can be used across multiple environments and also have .local files so that developers can create their own overwrites so they never have to change files that are to be committed.  It also reduces the amount of variables new devs have to set before they have a working setup.

**Added feature flags**
None

**Affected issues**  
ARC-736

**How has this been tested?**  
`docker-compose up` still works and tests pass

**Whats Next?**
nothing?


.husky/post-merge
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
```
joshkay10, line 1: This file is very good news

joshkay10 (approved):
Ran it using atlas tunnel version and works a treat still.

.env.development
```diff
@@ -1,11 +1,6 @@
 # APP VARIABLES - NEEDED FOR PRODUCTION
```
gxueatlassian, line 1: Do we still need some `GITHUB_CLIENT_ID` somewhere like `.env` or `.env.local`? 
If so, we still need an .env.example right?

mboudreau: yes we do.  I need to update the docs to reflect this new structure.  It would be under `.env.development.local`

.husky/post-merge
mboudreau: :+1: wouldn't want to waste anybody's time ;)

rachellerathbone (dismissed):

src/config/env.ts
```diff
@@ -26,15 +27,16 @@ const requiredEnvVars = [
...
-if (env.error && nodeEnv !== EnvironmentEnum.production) {
-	throw env.error;
-}
+// Load environment files
+const envFile = ".env";
+[
+	`${envFile}.${nodeEnv}.local`,
+	`${envFile}.local`,
+	`${envFile}.${nodeEnv}`,
+	envFile
```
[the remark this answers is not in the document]
mboudreau: we do sadly.  We need the .env for the env vars for docker-compose which we use for ngrok_token.  This could be improved upon later with a wrapper or our own ngrok docker service

gxueatlassian (approved):

joshkay10 (approved):

yvonnevictoria (approved):

Harminder84 (approved):

PR #1373, opened 2022-07-17
remove build deploy backfill flag

joshkay10:
**What's in this PR?**
Dropping the feature flag for build and deployments in backfill, has been running at 100% for over a month.

**Why**
Its time, but also about to make some changes here and it was complicating my life.

**Removed feature flags**
run-backfill-for-builds-and-deployments

**How has this been tested?**  
Tested locally and soaked on stage

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -62,10 +62,8 @@ export const sortedRepos = (repos: Repositories): [string, RepositoryData][] =>
 			new Date(a[1].repository?.updated_at || 0).getTime()
 	);
 
-const getNextTask = async (subscription: Subscription, jiraHost: string): Promise<Task | undefined> => {
-
-	const includeBuildAndDeployments = await booleanFlag(BooleanFlags.BACKFILL_FOR_BUILDS_AND_DEPLOYMENTS, false, jiraHost);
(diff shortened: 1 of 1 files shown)


src/sync/installation.ts
```diff
@@ -62,10 +62,8 @@ export const sortedRepos = (repos: Repositories): [string, RepositoryData][] =>
 			new Date(a[1].repository?.updated_at || 0).getTime()
 	);
 
-const getNextTask = async (subscription: Subscription, jiraHost: string): Promise<Task | undefined> => {
-
-	const includeBuildAndDeployments = await booleanFlag(BooleanFlags.BACKFILL_FOR_BUILDS_AND_DEPLOYMENTS, false, jiraHost);
-	const tasks = taskTypes.filter(task => includeBuildAndDeployments || (task !== "build" && task !== "deployment"));
+const getNextTask = async (subscription: Subscription): Promise<Task | undefined> => {
+	const tasks = taskTypes;
```
joshkay10, line 66: Why not just use taskTypes instead of creating a new variable???
I have a secondary branch i'm working at the moment which uses `const tasks` and it was just easier to leave it!

Harminder84 (approved):

PR #1374, opened 2022-07-18
Arc 1502 add gh version tag

rachellerathbone:
**What's in this PR?**
Adding tags to some of our metrics.

**Why**
So we can compare metrics for GitHub versions (cloud/server)

**Affected issues**  
ARC-1502

**How has this been tested?**  
Updated tests to pass to make sure the payload looks correct. A little hard to test most of these metrics in the app because they cover error cases. Will monitor once this goes into prod to see if we can see the 'cloud' tag showing up.

**Whats Next?**
Continuing updating metrics with tag. Only did about half on this pass to keep the number of changes down.

### DIFF
--- a/src/github/client/github-app-client.ts
+++ b/src/github/client/github-app-client.ts
@@ -47,8 +47,8 @@ export class GitHubAppClient extends GitHubClient {
 			handleFailedRequest(this.logger)
 		);
 		this.axios.interceptors.response.use(
-			instrumentRequest(metricHttpRequest.github),
-			instrumentFailedRequest(metricHttpRequest.github)
+			instrumentRequest(metricHttpRequest.github, this.restApiUrl),
+			instrumentFailedRequest(metricHttpRequest.github, this.restApiUrl)
 		);
 
 		this.axios.interceptors.request.use((config: AxiosRequestConfig) => {
--- a/src/util/get-cloud-or-server.ts
+++ b/src/util/get-cloud-or-server.ts
@@ -1,0 +1,4 @@
+import { GITHUB_CLOUD_API_BASEURL } from "utils/get-github-client-config";
+
+export const getCloudOrServerFromSubscription = (gitHubAppId: number | undefined) => gitHubAppId ? "server" : "cloud";
+export const getCloudOrServerFromHost = (host: string) => host === GITHUB_CLOUD_API_BASEURL ? "cloud" : "server";
--- a/src/util/get-cloud-or-server.ts
+++ b/src/util/get-cloud-or-server.ts
@@ -1,4 +1,4 @@
 import { GITHUB_CLOUD_API_BASEURL } from "utils/get-github-client-config";
 
-export const getCloudOrServerFromSubscription = (gitHubAppId: number | undefined) => gitHubAppId ? "server" : "cloud";
+export const getCloudOrServerFromGitHubAppId = (gitHubAppId: number | undefined) => gitHubAppId ? "server" : "cloud";
 export const getCloudOrServerFromHost = (host: string) => host === GITHUB_CLOUD_API_BASEURL ? "cloud" : "server";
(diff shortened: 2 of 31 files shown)


src/util/is-cloud-or-server.ts
```diff
@@ -0,0 +1,4 @@
+import { GITHUB_CLOUD_API_BASEURL } from "utils/get-github-client-config";
+
+export const isCloudOrServerSubscription = (gitHubAppId: number | undefined) => gitHubAppId ? "server" : "cloud";
+export const isCloudOrServerHost = (host: string) => host === GITHUB_CLOUD_API_BASEURL ? "cloud" : "server";
```
gxueatlassian, line 4: /nit
Usually method `isXXX` return `boolean`. We could use `getXXXFromHost` like ( `getGitHubTagFromHost(host)` or `gitHubTagFromHost()` )

gxueatlassian: /nit
And the var name `gitHubVersion`, should it be like `gitHubType`? Isn't `xxxVersion` usually refer to `x.x.x`?

src/github/client/github-app-client.ts
```diff
@@ -47,8 +47,8 @@ export class GitHubAppClient extends GitHubClient {
 			handleFailedRequest(this.logger)
 		);
 		this.axios.interceptors.response.use(
-			instrumentRequest(metricHttpRequest.github),
-			instrumentFailedRequest(metricHttpRequest.github)
+			instrumentRequest(metricHttpRequest.github, this.restApiUrl),
```
gxueatlassian, line 50: Shall we move the `this.restApiUrl` to above, and only passing generic tags in the `instrucmentRequest`? Like


```javascript
const gitHubVersion = isCloudOrServerHost(url);
const extraTags = { gitHubVersion};
this.axios.interceptors.response.use(
			instrumentRequest(metricHttpRequest.github, extraTags),
			instrumentFailedRequest(metricHttpRequest.github, extraTags)
		);
```

src/util/is-cloud-or-server.ts
rachellerathbone: I'll update the method name. Regarding version, however, yes it _can_ mean that if we look at it through a narrow 'developer lense' but the literal definition of the word is `a particular form of something differing in certain respects from an earlier form`. Happy to discuss further (especially seeing I can guarantee others on the team would most definitely want to have input on this... @thombergs * cough *) but would say we do it separately to this pr (especially seeing that wording has already been used elsewhere).

src/github/client/github-app-client.ts
gxueatlassian: "What's the benefit of..."
     <--- I am thinking from `instrumentRequest` point of view, what does it really need from consumer? Does it need the full url? Or does it just need a tag provided by consumer? And what if there's future need to add more tags from consumer? 

"Also, the metrics handle the value for gitHubVersion a little differently..."
    <--- I would say something like ` { client: "axios", status: status, ...extraTags }` should do the trick in both cases. 

--------------

But having say that, it seems you have thought about the both cases thoroughly, so I am happy to follow your idea, and close this thread. Please feel free to close this thread.

src/util/get-cloud-or-server.ts
```diff
@@ -0,0 +1,4 @@
+import { GITHUB_CLOUD_API_BASEURL } from "utils/get-github-client-config";
+
+export const getCloudOrServerFromSubscription = (gitHubAppId: number | undefined) => gitHubAppId ? "server" : "cloud";
```
gxueatlassian, line 3: /nit

Maybe `getCloudOrServerFromGitHubAppId` is more accurate than `getCloudOrServerFromSubscription`

gxueatlassian (approved):

gxueatlassian (approved):

PR #1375, opened 2022-07-18
ARC-611 backfill from date

joshkay10:
**What's in this PR?**
Added a new endpoint argument for the  `/resync` call.
~~`commitHistoryDepth`~~ `commitsFromDate` which allows you to pass the date to fetch commits (currently doubles for main and branch commits, instead of two seperate depth values)
Example 
{...`"commitsFromDate": 2022-07-26`} or with timezone
{...`"commitsFromDate": 2022-07-26T23:47:22.393Z'`}
It has higher importance than feature flag.

**Why**
If we want to run this from the command line we can!!
Also is groundwork so we can call it from the UI

**Runbook**
[HOW TO run backfill](https://hello.atlassian.net/wiki/spaces/PF/pages/1824654901/HOW+TO+run+backfill+resync+endpoint)

**Affected issues**  
ARC-611

**How has this been tested?**  
Updated the branch and commits unit tests in sync to make sure this value is used and used before ff value.

Tested in stage, tried install and `/resync` with combinations of values and feature flag settings.
Results as expected. 

**Whats Next?**
Create the UI which will need to pass a similar value


src/routes/api/api-router.ts
```diff
@@ -82,16 +82,23 @@ ApiRouter.post(
 		const offset = Number(req.body.offset) || 0;
 		// only resync installations whose "updatedAt" date is older than x seconds
 		const inactiveForSeconds = Number(req.body.inactiveForSeconds) || undefined;
+		// How far back to fetch commit history(main and branch) in milliseconds.
+		const commitHistoryDepth = Number(req.body.commitHistoryDepth) || undefined;
```
joshkay10, line 86: Went back and forth with commitHistoryDepth by itself or commitHistoryDepth + branchCommitHistoryDepth

Since there are two commit queries one for the main branch and one for other branches.

I ended up with one thats shared, because I was trying to limit the noise, but also the UI is only going to have 1 date picker(at this point 😂) so there was a consistency card to play there too....

src/sync/installation.ts
```diff
@@ -259,7 +260,7 @@ async function doProcessInstallation(app, data: BackfillMessagePayload, sentry:
 		for (const perPage of [20, 10, 5, 1]) {
 			// try for decreasing page sizes in case GitHub returns errors that should be retryable with smaller requests
 			try {
-				return await processor(logger, github, gitHubInstallationClient, jiraHost, repository, cursor, perPage);
+				return await processor(logger, github, gitHubInstallationClient, jiraHost, repository, cursor, perPage, data);
```
joshkay10, line 263: This was the least evil way I could think of..... passing around the message payload, because of the cute way that this code exists all task processors being shared etc this was the most generic solution I found, basically just  goodies bag for everything now. 😂

src/routes/api/api-router.ts
gxueatlassian: /nit

I am okay with it not too fuzzy, but personally find it `xxxDepth` is a number like indicate much layer was nested inside a directory.

If the `commitHistoryDepth` means "xxx in milliseconds", I find it easier to understand if named `commitHistoryDepthInMilliseconds`

gxueatlassian (approved):
LGTM, but maybe wait for one more approval?

src/routes/api/api-router.ts
joshkay10: Fair, adding `milliseconds` os clear. naming is hard. lolol

The other option is align with the feature flag names `commitTimeLimit`

src/routes/api/api-router.ts
```diff
@@ -82,16 +82,23 @@ ApiRouter.post(
 		const offset = Number(req.body.offset) || 0;
 		// only resync installations whose "updatedAt" date is older than x seconds
 		const inactiveForSeconds = Number(req.body.inactiveForSeconds) || undefined;
+		// How far back to fetch commit history(main and branch) in milliseconds.
+		const commitTimeLimit = Number(req.body.commitTimeLimit) || undefined;
```
mboudreau, line 86: Why `|| undefined` here?

src/routes/api/api-router.ts
```diff
@@ -82,16 +82,23 @@ ApiRouter.post(
...
 
 		if (!statusTypes && !installationIds && !limit && !inactiveForSeconds){
 			res.status(400).send("please provide at least one of the filter parameters!");
 			return;
 		}
 
+		if (commitTimeLimit !== undefined && (commitTimeLimit <= 0 || isNaN(commitTimeLimit))){
```
mboudreau, line 93: if you don't do `|| undefined` above, you can just do `if(!commitTimeLimit || commitTimeLimit <=0)`

src/routes/api/api-router.ts
mboudreau: ah, nvm, I can see why now as it's being passed.

src/routes/api/api-router.ts
mboudreau: actually, you can do the same even with `|| undefined` above

src/routes/api/api-router.ts
```diff
@@ -82,16 +82,23 @@ ApiRouter.post(
...
+			res.status(400).send("Invalid commitTimeLimit value, please enter a natural number.");
+			return;
+		}
+
 		const subscriptions = await Subscription.getAllFiltered(installationIds, statusTypes, offset, limit, inactiveForSeconds);
 
 		await Promise.all(subscriptions.map((subscription) =>
-			findOrStartSync(subscription, req.log, syncType)
+			findOrStartSync(subscription, req.log, syncType, commitTimeLimit)
```
mboudreau, line 101: This I'm not sure I agree with, now that I think about it.  In the UI, it'll be something like "backfill from 7 days ago", which is fine, but I think that should be a date coming in specifically as it should be based on the user's current timezone and not the server's.

That means that the 'timeLimit' in this case is specifically a date which should be filtered down to the graphql query which is a GitTimestamp (a version of date).

src/sqs/backfill.ts
```diff
@@ -8,7 +8,8 @@ import { SentryScopeProxy } from "models/sentry-scope-proxy";
 export type BackfillMessagePayload = {
 	installationId: number,
 	jiraHost: string,
-	startTime?: string
+	startTime?: string,
+	commitTimeLimit?: number
```
mboudreau, line 12: This should be date

src/sync/installation.ts
mboudreau: yeah, this is an acceptable solution.  This whole sync functionality needs to be redone...

src/sync/installation.ts
```diff
@@ -43,7 +43,8 @@ interface TaskProcessors {
 		jiraHost: string,
 		repository: Repository,
 		cursor?: string | number,
-		perPage?: number
+		perPage?: number,
+		messagePayload?: Record<string, any>
```
mboudreau, line 47: Don't we have types for these message payloads?  Pretty sure the queues use a generic, but there's no reason we can't do the same here and let each processor have their own typings to make it a bit safer

mboudreau (changes requested):
I think there's a core issue here that needs addressing but good first pass.  Happy to chat some more through it

src/routes/api/api-router.ts
joshkay10: change to be even even more explicit with commitsFromDate string

src/sync/installation.ts
joshkay10: Used correct types

src/sync/branches.ts
```diff
@@ -3,7 +3,9 @@ import { GitHubAPI } from "probot";
 import { Repository } from "models/subscription";
 import { GitHubInstallationClient } from "../github/client/github-installation-client";
 import Logger from "bunyan";
-import { numberFlag, NumberFlags } from "config/feature-flags";
+import { BackfillMessagePayload } from "~/src/sqs/backfill";
+import { NumberFlags } from "config/feature-flags";
+import { getCommitSinceDate } from "~/src/sync/sync-utils";
```
gxueatlassian, line 8: /nit
Is `~/src/blah` a pattern we prefer? Or team prefer `./blah` (or even `../../blah` if is another folder?)

src/sync/sync-utils.ts
```diff
@@ -29,5 +31,16 @@ export async function findOrStartSync(
 	}
 
 	// Start sync
-	await sqsQueues.backfill.sendMessage({ installationId, jiraHost, startTime: fullSyncStartTime }, 0, logger);
+	await sqsQueues.backfill.sendMessage({ installationId, jiraHost, startTime: fullSyncStartTime, commitsFromDate }, 0, logger);
 }
+
+export const getCommitSinceDate = async (jiraHost: string, flagName: NumberFlags, commitFromDate?: string): Promise<string | undefined> => {
```
gxueatlassian, line 37: /nit
`flagName` could be more strict typed as 
`SYNC_MAIN_COMMIT_TIME_LIMIT  | SYNC_BRANCH_COMMIT_TIME_LIMIT`

gxueatlassian (dismissed):

src/sync/branches.ts
mboudreau: if it's in the same folder, use `./blah`, if it's in a shortcut folder (like models) use `models/blah`, for anything else, use `~/src/some-dir/blah`

src/sync/sync-utils.ts
```diff
@@ -2,11 +2,13 @@ import { RepoSyncState } from "models/reposyncstate";
...
 import { Subscription, SyncStatus } from "models/subscription";
 import Logger from "bunyan";
+import { numberFlag, NumberFlags } from "config/feature-flags";
 
 export async function findOrStartSync(
 	subscription: Subscription,
 	logger: Logger,
-	syncType?: "full" | "partial"
+	syncType?: "full" | "partial",
+	commitsFromDate?: string
```
mboudreau, line 11: I'd made this a date, easier typing to deal with.

src/routes/api/api-router.ts
```diff
@@ -82,16 +82,23 @@ ApiRouter.post(
...
 		const inactiveForSeconds = Number(req.body.inactiveForSeconds) || undefined;
+		// A date to start fetching commit history(main and branch) from.
+		const commitsFromDate = req.body.commitsFromDate;
 
 		if (!statusTypes && !installationIds && !limit && !inactiveForSeconds){
 			res.status(400).send("please provide at least one of the filter parameters!");
 			return;
 		}
 
+		if (commitsFromDate && !Date.parse(commitsFromDate)) {
```
mboudreau, line 93: Why aren't you keeping the parsed date to pass it on to the next function?  Better typings than just a string :)

mboudreau: also, this should check to see that the date is valid (like, not be in the future with `date < new Date()`)

src/sqs/backfill.ts
```diff
@@ -8,7 +8,8 @@ import { SentryScopeProxy } from "models/sentry-scope-proxy";
 export type BackfillMessagePayload = {
 	installationId: number,
 	jiraHost: string,
-	startTime?: string
+	startTime?: string,
+	commitsFromDate?: string
```
mboudreau, line 12: Date here

src/sync/sync-utils.ts
mboudreau: yeah, agreed, it should just be the flags we do care about for this, which is the 2 in `NumberFlags` but you don't want your typings to include others if another one is added in the future.

src/routes/api/api-router.ts
joshkay10: Have converted to Dates when appropriate 🤟

src/github/client/github-installation-client.ts
```diff
@@ -213,12 +213,12 @@ export class GitHubInstallationClient extends GitHubClient {
...
 	}
 
-	public async getBranchesPage(owner: string, repoName: string, perPage = 1, timeCutoffMsecs?: number, cursor?: string): Promise<getBranchesResponse> {
+	public async getBranchesPage(owner: string, repoName: string, perPage = 1, commitSince?: Date, cursor?: string): Promise<getBranchesResponse> {
 		const variables = {
 			owner,
 			repo: repoName,
 			per_page: perPage,
-			commitSince: timeCutoffMsecs ? new Date(Date.now() - timeCutoffMsecs).toISOString() : undefined,
+			commitSince: commitSince?.toISOString(),
```
joshkay10, line 221: A date right until the string is required 😊

src/sync/sync-utils.ts
joshkay10: typed!

src/routes/api/api-router.ts
```diff
@@ -82,16 +84,27 @@ ApiRouter.post(
 		const offset = Number(req.body.offset) || 0;
 		// only resync installations whose "updatedAt" date is older than x seconds
 		const inactiveForSeconds = Number(req.body.inactiveForSeconds) || undefined;
+		// A date to start fetching commit history(main and branch) from.
+		let commitsFromDate = req.body.commitsFromDate;
```
mboudreau, line 88: you're changing the type of the object here from unknown to date.  I'd probably just stick to Date to make sure there's no confusion.

mboudreau (approved):

gxueatlassian (approved):

PR #1376, opened 2022-07-18
ARC-1469 - GHE UI - navigate from select a server screen to app creation/gh apps list screen

krazziekay:
**What's in this PR?**
- New Page added in connect app descriptor file.
- New route created for displaying the list of server apps.
- Navigation logic updated for `jira-select-server`
- Snapshots updated

**Why**
- Connecting the navigation between Github Server list and GitHub Server apps list.

**Video**

- When the user tries to add a GH server which already exists, it will navigate the user to the list of GH server apps

https://user-images.githubusercontent.com/13076255/179458285-b854f5ad-8bf2-4477-bfa8-084c42530d4e.mov

- When the user tries to connect from the list of GHE servers

https://user-images.githubusercontent.com/13076255/179459937-921800a7-015a-462f-b81e-d6c61880b031.mov


src/routes/jira/server/jira-ghe-server-apps.ts
```diff
@@ -0,0 +1,37 @@
...
+export const JiraGheServerApps = async (
+	req: Request,
+	res: Response,
+	next: NextFunction
+): Promise<void> => {
+	try {
+		req.log.debug("Received Jira GHE server apps page request");
+
+		const { id: installationId } = res.locals.installation;
+		const githubServerBaseUrl = req?.query.serverUrl as string;
```
gxueatlassian, line 13: Do we need a `?` in the `req?`

static/js/jira-select-server.js
```diff
@@ -1,5 +1,8 @@
 $('.select-server').click(function (event) {
   event.preventDefault();
 
-  // TODO: Need to add the logic for navigation
+  AP.navigator.go("addonmodule", {
+    moduleKey: "github-list-server-apps-page",
+    customData: { serverUrl: $(event.target).data("server-name") }
```
gxueatlassian, line 6: is it supposed to be `url` or `name?

src/routes/jira/server/jira-ghe-server-apps.ts
krazziekay: Not anymore, will remove it...

static/js/jira-select-server.js
krazziekay: The component that has this attribute is used by both GHE Servers and GH Server apps, for GHE Server, it is serverUrl, but for GH Server apps, it will be app name.
So, `server-name` is an attempt to fit for both cases..

krazziekay: Updated the name to `identifier`..

src/routes/jira/server/jira-ghe-server-apps.ts
```diff
@@ -0,0 +1,37 @@
...
+export const JiraGheServerApps = async (
+	req: Request,
+	res: Response,
+	next: NextFunction
+): Promise<void> => {
+	try {
+		req.log.debug("Received Jira GHE server apps page request");
+
+		const { id: installationId } = res.locals.installation;
+		const githubServerBaseUrl = req.query.serverUrl as string;
```
gxueatlassian, line 13: /nit
not sure how url is passed in connect app, will it do things like `encodeURIComponent`? But should be minor, should works on most browser.

views/partials/select-server-table.hbs
```diff
@@ -16,7 +16,7 @@
         <div class="selectServer__serverURL">{{server.displayName}}</div>
         <button
           class="aui-button aui-button-label select-server"
-          data-installation-id="{{installationId}}"
+          data-identifier="{{server.displayName}}"
```
gxueatlassian, line 19: Hey sorry I still don't get it, above apps listing page is expecting an url
```
{
			key: "github-list-server-apps-page",
			name: {
				value: "GitHub Enterprise Server Apps"
			},
			url: "/jira/ghe-servers/apps?serverUrl={ac.serverUrl}",
```

But here we are supplying a name? How does that work?

krazziekay: In the frontend, for GHE server, we are displaying the `baseUrl` as the identifier, whereas for GH server apps, we are displaying the `gitHubAppName` as the identifier.
So, the `name` in this case is the `baseUrl`.

views/partials/select-server-table.hbs
```diff
@@ -16,7 +16,7 @@
         <div class="selectServer__serverURL">{{server.displayName}}</div>
```
gxueatlassian, line 16: /nit Happy to proceed on this PR without it. 

------------

Since this is for both apps and servers, the naming could be confusing.

For better readability, we can have different options:

1. Rename this to `select-table`, try to name things inside as generic as possible. 
2. Or Duplicate this file, so we have `select-server-table.hbs` and `select-apps-table.hbs`. Sometimes duplication is not a bad things. Consider in the future the server listing page and apps listing page may have more differences. --> And also remove the mapping `const gheServers = chain(groupBy(allServers, "gitHubBaseUrl")).map((_, key) => ({ displayName: key })).value();` into `displayName`. Map it to `xxxUrl`.
3. And/Or Rename the bit to be `{{item.identifier}}`, as neither it is always a `server` nor the field inside is always a `displayName` (it can be url)

gxueatlassian (dismissed):

src/routes/jira/server/jira-ghe-server-apps.ts
```diff
@@ -0,0 +1,37 @@
+import { NextFunction, Request, Response } from "express";
+import { GitHubServerApp } from "models/github-server-app";
+
+export const JiraGheServerApps = async (
```
rachellerathbone, line 4: Naming convention is to use the method in the name of the function and in the file name. Just noticed that you've previously created `JiraGheServers`. That will need to be updated as well but that one can be done in a separate PR.

src/routes/jira/server/jira-ghe-server-apps.ts
```diff
@@ -0,0 +1,37 @@
...
+		const serverApps = servers.map(server => ({ displayName: server.gitHubAppName }));
+
+		if (serverApps.length) {
+			res.render("jira-select-github-cloud-app.hbs", {
+				servers: serverApps,
+				previousPagePath: "github-list-servers-page"
+			});
+		} else {
+			res.render("jira-select-app-creation.hbs", {
+				previousPagePath: "github-list-servers-page"
```
rachellerathbone, line 29: Seeing we're using this in 2 places can we set it to a variable and pass it in?

views/partials/select-server-table.hbs
krazziekay: I'll do these changes in a separate PR..

src/routes/jira/server/jira-ghe-server-apps.ts
krazziekay: Just realized that this path is incorrect and will be different in different case scenarios, so removing it.

src/routes/jira/server/jira-ghe-server-apps.ts
krazziekay: Will do it in a separate PR.

views/partials/select-server-table.hbs
krazziekay: https://github.com/atlassian/github-for-jira/pull/1377

gxueatlassian (approved):

views/jira-select-server.hbs
```diff
@@ -45,7 +45,7 @@
             subHeader="We found the following GitHub servers"
         }}
         <!-- servers refer to the array of GHE apps-->
-        {{> select-server-table
+        {{> select-table
```
gxueatlassian, line 48: Still have `servers=servers` , it should be `list=list` right?

views/jira-select-github-cloud-app.hbs
```diff
@@ -44,7 +44,7 @@
             subHeader=(concat "We found the following GitHub apps for your server, " "<b>" serverUrl "</b>")
         }}
         <!-- servers refer to the array of GHE apps/servers-->
-        {{> select-server-table
+        {{> select-table
```
gxueatlassian, line 47: Still have `servers=servers` , it should be `list=list` right?

gxueatlassian, line 47: ** No need to action on this in this PR **
This kind of bug is annoying, not sure what's better way we can test it.

Feel free to close this thread.

src/routes/jira/server/jira-ghe-server-apps.ts
krazziekay: This will be taken care in https://softwareteams.atlassian.net/browse/ARC-1519

gxueatlassian (approved):

gxueatlassian (approved):

src/routes/jira/server/jira-ghe-server-apps.ts
```diff
@@ -0,0 +1,35 @@
...
+		const { id: installationId } = res.locals.installation;
+		const githubServerBaseUrl = req.query.serverUrl as string;
+
+		if (!githubServerBaseUrl) {
+			throw new Error("No server URL passed!");
+		}
+
+		const servers = await GitHubServerApp.getAllForGitHubBaseUrl(githubServerBaseUrl, installationId) || [];
+		// `identifier` is the githubAppName for the GH server app
+		const serverApps = servers.map(server => ({ identifier: server.gitHubAppName }));
```
rachellerathbone, line 21: Wouldn't we just call it `gitHubAppName`? Imo, if you have to explain why you gave a variable/func a name, then you need to rename it. The name itself should provide the explanation.

krazziekay: This data is to be passed to the `select-table` component. And this component displays the list provided to it, which could be GHE servers or server apps.
For GHE server, we are displaying the _baseUrl_ as the identifier, whereas for GH server apps, we are displaying the _gitHubAppName_ as the identifier.
This property had to be named something generic, so named it `identifier`. Confirmed with @gxueatlassian.

PR #1380, opened 2022-07-19
Arc-1502 add gh version tag metric http request

rachellerathbone:
**What's in this PR?**
Updated one more existing metric and added some new ones. I looked at updating the metrics for `elapsedTimeMetrics` and `sqsQueueMetrics` but it looks like that would require a lot of work for little value. We can come back to passing a cloud/server tag these at a later time if we determine doing so would be valuable.

**Why**
So we can update our charts to distinguish between cloud/server customers.

**Affected issues**  
ARC-1502

**Whats Next?**

- Updating/adding charts in our module in fusion-monitoring
- Adding more metrics when the rest of the flow is complete for GHE server


mboudreau (approved):

src/routes/jira/server/jira-server-url-post.ts
```diff
@@ -67,5 +69,6 @@ export const JiraServerUrlPost = async (
 		const codeOrStatus = err.code || err.response.status;
 		const { errorCode, message } = gheServerUrlErrors[codeOrStatus] || gheServerUrlErrors.default;
 		res.status(200).send({ success: false, errors: [{ code: errorCode, message }] });
+		statsd.increment(metricError.gheServerUrlError, { errorCode, status: err.response.status	 });
```
joshkay10, line 72: /lint `.status	 });`

src/sync/pull-request.ts
```diff
@@ -66,7 +68,12 @@ export const getPullRequestTask = async (
 				direction: SortDirection.DES
 			});
 
-	statsd.timing(metricHttpRequest.syncPullRequest, Date.now() - startTime, 1, [`status:${status}`]);
+	const gitHubVersion = getCloudOrServerFromHost(request.host);
+	statsd.timing(
+		metricHttpRequest.syncPullRequest,
+		Date.now() - startTime,
+		1,
```
joshkay10, line 75: /nit /notSureIfICare Is it worth adding a comment beside `,1` like `Sample rate is set to 100%`

joshkay10 (approved):

PR #1384, opened 2022-07-20
Arc 810 analytics take 2

rachellerathbone:
**What's in this PR?**
Set up of the Node analytics client + emitting of 1 analytic from GH config. **Related PR:** https://bitbucket.org/atlassian/github-for-jira-deployment/pull-requests/45/update-publish-script-to-copy-npmrc We'll probably want to approve both, merge the change to the publish script first, and then this one shortly after.

**Why**
So we can finally omit bloody analytics.

**Affected issues**  
ARC-810

**How has this been tested?**  
Confirmed that we are seeing events in staging. Query `micros_analytics-service` env="prod-*" m.t="application" ext.event.name=githubConfigurationScreen`

Screen Shot 2022-07-20 at 12 00 19 pm

**Whats Next?**
ARC-1506: emitting must-have analytics


.github/workflows/on-push.yml
```diff
@@ -55,7 +55,3 @@ jobs:
         uses: docker/build-push-action@v2.7.0
         with:
           file: Dockerfile
-      - name: Build and push Docker image
```
rachellerathbone, line 58: @mboudreau remind me why we deleted this? Seems like a lifetime ago that we talked about it.

.npmrc
```diff
@@ -0,0 +1 @@
+# IMPORTANT: DO NOT DELETE THIS. IT IS REQUIRED TO PASS THE BUILD FOR SOX COMPLIANT PACKAGES IN THE BB PIPELINE
```
mboudreau, line 1: Of course it is...
image

.github/workflows/on-push.yml
mboudreau: Because in the Dockerfile.prod, we will try to install the analytics package from github, which simply won't work because it doesn't have access to PAC and will break the build because of it.

I think this was needed as we'll have the same issue with Brahmos and since we're now have a decent CD system going, the feedback loop should be good enough (I think).  It would be nice to have BB run when we have a PR to make sure the dockerfile.prod build works as expected...

.npmrc
rachellerathbone: lol... I know right. I can't believe this is what I needed to do to get this to work.

mboudreau: Actually, if we need to have it, just copy/paste the contents of `.yarnrc` into here and delete that file.  No need for duplication.  Yarn should read from `.npmrc` if `.yarnrc` is not available :)

src/util/analytics-client.ts
```diff
@@ -0,0 +1,94 @@
...
+	}
+}
+
+function wrapPromise(promise: Promise<unknown>) {
+	promise.catch((error) => {
+		logger.warn(`Cannot sendAnalytics event: ${error}`);
+	});
+}
+
+export { sendAnalytics };
```
mboudreau, line 94: just do an export on the function directly

.npmrc
rachellerathbone: Just tried doing this and it looks like the BB pipeline is going to fail. It's hanging on the deploy step. So not sure this is a viable solution (the implemented solution was suggested by 2 engineers from the build team during a zoom call yesterday).

.github/workflows/on-push.yml
rachellerathbone: It would wouldn't it? :laughcry:

rachellerathbone:
@mboudreau made the changes you suggested and confirmed it still passes in the pipeline and I can still see analytics coming through 🎉

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -130,7 +132,13 @@ export const JiraConfigurationGet = async (
 			isGitHubCloudApp: await isGitHubCloudApp(gitHubAppId)
 		});
 
-		req.log.debug("Jira configuration rendered successfully.");
+		sendAnalytics(AnalyticsEventTypesEnum.ScreenEvent, {
+			name: AnalyticsScreenEventsEnum.GitHubConfigScreenEventName,
+			jiraHost,
```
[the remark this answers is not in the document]
rachellerathbone: I already chatter with Josh about this and he has said he's going to revert the hashing changes on the jiraHost because it makes it impossible to look through the logs.

.npmrc
```diff
@@ -0,0 +1,2 @@
+registry=https://registry.npmjs.org/
+@atlassian=registry:https://registry.npmjs.org/
```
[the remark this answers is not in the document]
rachellerathbone: I thought that too and it was how I originally had it but that formatting results in the following. The above passes all builds and I've tested in dev and stg.

Screen Shot 2022-07-21 at 9 50 01 am

src/routes/jira/configuration/jira-configuration-get.ts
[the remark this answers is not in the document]
gxueatlassian: lol, all good then.

src/util/analytics-client.ts
```diff
@@ -0,0 +1,92 @@
...
+				}
+			}));
+			break;
+		default:
+			logger.warn(`Cannot sendAnalytics: unknown eventType`);
+			break;
+	}
+}
+
+function wrapPromise(promise: Promise<unknown>) {
```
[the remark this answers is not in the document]
rachellerathbone: The analytics client was created by @atrigueiro and merged back in March https://github.com/atlassian/github-for-jira/pull/1093 but was reverted because it was causing builds to fail. I've just copied her code over and fixed the build issues. Happy to make changes we think are really important but would otherwise prefer to leave as is. 

That said, I would be happy to potentially change the name of this function but don't think`logWarnOnError` makes sense. The calls to `wrapPromise` send analytics e.g.
```
wrapPromise(analyticsNodeClient.sendTraitEvent({
				entityType: attributes?.entityType,
				entityId: attributes?.entityId,
				entityTraits: attributes?.entityTraits
			}));
```
It falls into the catch block and sends a warning message if something goes wrong but this isn't the purpose of the function. Something like `sendEvent` or `sendAnalyticEvent` are possible options.

.npmrc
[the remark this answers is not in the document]
mboudreau: Is this just on your local?  trying doing `yarn cache clean`  before

mboudreau (approved):

PR #1385, opened 2022-07-20
Arc 1533 conditional proxy

thombergs:
This PR removes the global proxy configuration we've been using and replaces it with a more targeted proxy configuration on the level of the GitHub clients.

I have also added a special case so we're not using the proxy for our internal GitHub Server (which was the original reason to change this). Access to the GitHub Server has previously not worked, because the outbound proxy doesn't allow traffic back into the Atlassian network.

**End state:**
- all calls to GitHub APIs (both cloud and on-premise) go through the outbound proxy
- exception: calls to github.internal.atlassian.com don't use any proxy
- calls to the Jira API will no longer go through the outbound proxy, because I removed the global configuration.  This doesn't matter, however, since calls to Jira should never really have gone through the outbound proxy in the first place. I would have liked to keep the global proxy config behind a feature flag, but I'm not sure about how the global config and the local Axios config I added interfere.

This change is not feature flagged because I didn't find a good way to feature flag it. I tried a few things, but suggestions are welcome. I will test in staging if it still works with an IP allowlist (which is the only feature that might break due to proxy changes).


src/config/proxy.ts
```diff
@@ -1,15 +1,27 @@
...
-
-const logger = getLogger("config.proxy");
-
-if (envVars.PROXY) {
-	logger.info(`configuring proxy: ${envVars.PROXY} for outbound calls`);
-	process.env.GLOBAL_AGENT_HTTP_PROXY = envVars.PROXY;
-	bootstrap();
-} else {
-	logger.info("configuring no proxy for outbound calls");
-}
```
thombergs, line 13: This was the global proxy config that would be used for ALL http calls. We had to do this globally, because we couldn't teach Probot to use a specific proxy. Now that Probot is replaced with our own GitHub clients, we can use a more targeted proxy config.

src/config/proxy.ts
```diff
@@ -1,15 +1,27 @@
...
+/**
+ * Use this Axios config to make HTTP(S) requests via the configured outbound proxy.
+ */
+export const OUTBOUND_PROXY_CONFIG: Partial<AxiosRequestConfig> = {
+	// Even though Axios provides the `proxy` option to configure a proxy, this doesn't work and will
+	// always cause an HTTP 501 (see https://github.com/axios/axios/issues/3459). The workaround is to
+	// create an HttpsProxyAgent and set the `proxy` option to false.
+	httpsAgent: outboundProxyHttpsAgent,
+	proxy: false
+};
```
thombergs, line 16: This Axios issue drove me nuts. Why is there a `proxy` option to configure a proxy host and port when it ALWAYS returns a 501?

mboudreau:
@thombergs I think this needs to wait until we've completely removed probot/octokit.  @joshkay10 how's that going btw?  I think we're using the new clients everywhere, but still have some code for the old way.

src/github/client/github-app-client.ts
```diff
@@ -31,13 +30,6 @@ export class GitHubAppClient extends GitHubClient {
 		super(logger, baseUrl);
 		this.appToken = AppTokenHolder.createAppJwt(privateKey, appId);
 
-		this.axios = axios.create({
-			baseURL: this.restApiUrl,
-			transitional: {
-				clarifyTimeoutError: true
-			}
-		});
-
```
thombergs, line 40: I moved the axios client out of the concrete GitHub clients and into the `GitHubClient` parent class. This was duplicate code and made it harder to configure the clients in a central place (for example to configure a proxy).

src/github/client/github-client-interceptors.ts
```diff
@@ -104,7 +105,7 @@ export const handleFailedRequest = (logger: Logger) =>
 
 		if (response) {
 			const status = response?.status;
-			const errorMessage = `Error executing Axios Request ` + error.message;
+			const errorMessage = `Error executing Axios Request (${toUpper(config.method)} ${config.baseURL}${config.url}): ` + error.message;
```
thombergs, line 108: At one point when I was investigating the 501 that Axios always returned I added some extra information to the error logs so I didn't have to guess which call went wrong.

src/github/client/github-client.ts
```diff
@@ -26,5 +29,22 @@ export class GitHubClient {
...
+			... proxyConfig
+		});
 	}
+
+	private getProxyConfig = (baseUrl: string): Partial<AxiosRequestConfig> => {
+		if (baseUrl.includes("github.internal.atlassian.com")) {
+			return NO_PROXY_CONFIG;
+		}
+		return OUTBOUND_PROXY_CONFIG;
+	};
```
thombergs, line 49: Moved the axios client to the `GitHubClient` parent class and added the proxy config (or not, depending on the `baseUrl`).

src/github/client/github-user-client.ts
```diff
@@ -11,20 +11,12 @@ import { GitHubClient } from "./github-client";
...
 		super(logger, baseUrl);
 		this.userToken = userToken;
 
-		this.axios = axios.create({
-			baseURL: this.restApiUrl,
-			transitional: {
-				clarifyTimeoutError: true
-			}
-		});
-
```
thombergs, line 27: This is using the parent class's `axios` instance now.

src/github/client/github-installation-client.ts
```diff
@@ -45,12 +44,6 @@ export class GitHubInstallationClient extends GitHubClient {
 		appTokenHolder: AppTokenHolder = AppTokenHolder.getInstance()
 	) {
 		super(logger, baseUrl);
-		this.axios = axios.create({
-			baseURL: this.restApiUrl,
-			transitional: {
-				clarifyTimeoutError: true
-			}
-		});
```
thombergs, line 53: This is using the parent class's `axios` instance now.

src/config/proxy.ts
```diff
@@ -1,15 +1,27 @@
...
+import { HttpsProxyAgent } from "https-proxy-agent";
+
+const outboundProxyHttpsAgent = envVars.PROXY ? new HttpsProxyAgent(envVars.PROXY) : undefined;
+
+/**
+ * Use this Axios config to make HTTP(S) requests via the configured outbound proxy.
+ */
+export const OUTBOUND_PROXY_CONFIG: Partial<AxiosRequestConfig> = {
+	// Even though Axios provides the `proxy` option to configure a proxy, this doesn't work and will
+	// always cause an HTTP 501 (see https://github.com/axios/axios/issues/3459). The workaround is to
```
mboudreau, line 12: that thread at the end said to update to the latest version as it should be fixed now?

src/config/proxy.ts
```diff
@@ -1,15 +1,27 @@
...
-}
+import { AxiosRequestConfig } from "axios";
+import { HttpsProxyAgent } from "https-proxy-agent";
+
+const outboundProxyHttpsAgent = envVars.PROXY ? new HttpsProxyAgent(envVars.PROXY) : undefined;
+
+/**
+ * Use this Axios config to make HTTP(S) requests via the configured outbound proxy.
+ */
+export const OUTBOUND_PROXY_CONFIG: Partial<AxiosRequestConfig> = {
```
mboudreau, line 10: I don't like the idea that this util knows about AxiosRequestConfig data type.  

Plus, this is only being used in a single place right? do we even need this file at all? We could put all of it in `github-client`

src/github/client/github-client.ts
```diff
@@ -26,5 +29,22 @@ export class GitHubClient {
...
+			baseURL: this.restApiUrl,
+			transitional: {
+				clarifyTimeoutError: true
+			},
+			... proxyConfig
+		});
 	}
+
+	private getProxyConfig = (baseUrl: string): Partial<AxiosRequestConfig> => {
+		if (baseUrl.includes("github.internal.atlassian.com")) {
```
mboudreau, line 45: I feel a great disturbance in the force...

should we be ignoring the proxy altogether for all `atlassian.com` or even `atlassian.net` urls so that it gets routed internally instead of going to the internet first?

rachellerathbone:
> @thombergs I think this needs to wait until we've completely removed probot/octokit. @joshkay10 how's that going btw? I think we're using the new clients everywhere, but still have some code for the old way.

Probot? We're still in the process of implementing the new webhook receiver. We need this change so we can test in environments outside of our local so we can feel confident that all the GHE changes and the new webhook receiver work as expected. This needs to happen before we officially start ripping Probot out.

yarn.lock
```diff
@@ -3897,6 +3897,14 @@ http-proxy-agent@^4.0.1:
     agent-base "6"
     debug "4"
 
+https-proxy-agent@5.0.1:
+  version "5.0.1"
+  resolved "https://packages.atlassian.com/api/npm/npm-remote/https-proxy-agent/-/https-proxy-agent-5.0.1.tgz#c59ef224a04fe8b754f3db0063a25ea30d0005d6"
```
mboudreau, line 3902: this is wrong... it should be pointing to npmjs.org not PAC for this library

src/config/proxy.ts
mboudreau: apparently, it has something to do with a combination of sub package node minimum TLS versions. This might be fixed in the latest version of Axios (0.27.2).  Maybe do an update while you're at it?

src/github/client/github-app-client.ts
mboudreau: pretty sure you could move the interceptors with it as well to the superclass one

mboudreau: while you're at it, can you also move the `get`, `post`, `delete`, `put` and `graphql` functions to the superclass?  fanx :)

yarn.lock
thombergs: Someday, I'll learn not to stumble over this stupid file....

thombergs:
Tests in staging have been successful:

Enabled the IP allowlist in our org:
image

It does show the "blocked" error message when the IP range is not in the list:
image

Logs say this:
```
Although you appear to have the correct authorization credentials, the `Fusion-Arc` organization has an IP allow list enabled, and 13.52.5.107 is not permitted to access this resource.
```
which means that the calls have been made through the proxy.

And it doesn't show the error message when the IP range IS in the list:
image

src/github/client/github-app-client.ts
thombergs: I'm happy to do that (and some other cleanup around the GitHub clients that I think should be done), but I'd prefer to do that in a second PR.

src/config/proxy.ts
thombergs: I'll try that.

thombergs: The comment "Please could you retry with the latest version and open a new issue should this error still be relevant?" combined with the status "possible bug" don't spark confidence in me, though :)

thombergs: Nope, not working:
image

src/config/proxy.ts
thombergs: moved the proxy config into the GitHubClient

src/github/client/github-client.ts
thombergs: made it so that it matches everything that ends with atlassian.com, now

joshkay10:
> > @thombergs I think this needs to wait until we've completely removed probot/octokit. @joshkay10 how's that going btw? I think we're using the new clients everywhere, but still have some code for the old way.
> 
> Probot? We're still in the process of implementing the new webhook receiver. We need this change so we can test in environments outside of our local so we can feel confident that all the GHE changes and the new webhook receiver work as expected. This needs to happen before we officially start ripping Probot out.

@mboudreau @rachellerathbone @thombergs  Everything is behind a feature flag and no Probot should have been called in at least a month.... however, I spend today pulling them all out so we can be sure!

src/github/client/github-client.ts
```diff
@@ -26,5 +30,42 @@ export class GitHubClient {
...
+		this.axios = axios.create({
+			baseURL: this.restApiUrl,
+			transitional: {
+				clarifyTimeoutError: true
+			},
+			... this.getProxyConfig(this.restApiUrl)
+		});
 	}
+
+	private getProxyConfig = (baseUrl: string): Partial<AxiosRequestConfig> => {
```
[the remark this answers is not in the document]
thombergs: Added the tests now. Wanted to get some feedback on the PR before locking it in with tests.

thombergs:
Alright, thanks for the update @joshkay10.

That means we should probably merge this only once the feature flags have been removed, i.e. when these PRs have been merged: 

- https://github.com/atlassian/github-for-jira/pull/1394
- https://github.com/atlassian/github-for-jira/pull/1393
- https://github.com/atlassian/github-for-jira/pull/1392
- https://github.com/atlassian/github-for-jira/pull/1391
- https://github.com/atlassian/github-for-jira/pull/1390
- https://github.com/atlassian/github-for-jira/pull/1389

src/github/client/github-client.ts
```diff
@@ -26,5 +30,42 @@ export class GitHubClient {
...
+		this.axios = axios.create({
+			baseURL: this.restApiUrl,
+			transitional: {
+				clarifyTimeoutError: true
+			},
+			... this.getProxyConfig(this.restApiUrl)
+		});
 	}
+
+	public getProxyConfig = (baseUrl: string): Partial<AxiosRequestConfig> => {
```
gxueatlassian, line 43: /nit
Happy to leave as it is, just minor suggestion.

-------------

I know it is tricky to test, and therefore change from `private` to `public`.

In this case, I'd usually do bellow in the test file, so instead of testing private method, assert on the dependencies. 
```javascript
import axios from "axios";
it('should ...', async ()=>{
   jest.spyOn(axios, 'create');
   ... do stuff
   expect(axios.get).toBeCalledWith(expect.objectContaining({
          proxy: { agent: blah blah } 
   });
});
```

Another option will be using nock to assert the http call is going through proxy I guess, but seems much more code.

gxueatlassian (approved):

gxueatlassian:
Approved, but please wait for more approvals.

src/github/client/github-client.ts
```diff
@@ -26,5 +30,42 @@ export class GitHubClient {
...
+		if (baseUrl
+			.replace(RegExp("/$"), "")
```
mboudreau, line 45: I'm not sure this is going to work as you expect it to.  GHE needs to have `/api/` in baseUrl. Instead, I'd probably do something like `new URL(baseUrl).host.endsWith("atlassian.com")` .  Do we also want to include atlassian.net?

src/github/client/github-client.ts
```diff
@@ -26,5 +30,42 @@ export class GitHubClient {
...
+			.endsWith("atlassian.com")) {
+			return this.noProxyConfig();
+		} else {
```
mboudreau, line 48: don't need the else if you're returning in the if

mboudreau (changes requested):
just one small thing to change for the baseUrl so that it works with GHE

src/github/client/github-client.ts
thombergs: good catch. Although, it worked in my tests. 

I made it more robust now ... just atlassian.com for now because I hope this is the only case we'll ever need (before we rip it out again)

src/github/client/github-client.ts
thombergs: ✅

src/github/client/github-client.ts
thombergs: It doesn't work with Nock, I had researched that. 

I had thought about mocking the axios client, but the current test is so much easier to read and maintain so I opted for that (even though it doesn't test the ultimate axios call).

mboudreau (approved):

PR #1387, opened 2022-07-20
Arc 1354 use cryptor for installation secrets

gxueatlassian:
**What's in this PR?**
Using an FF to enable reading from new encryptedSharedSecret field for models/Installations. 

**Why**
By Atlassian policy, all encryption should use Cryptor. Previous PR has enable parallel write `sharedSecret` to both column. And ALL prod data has been migrated. Now is good time to use FF to cut-over the read traffic. 

**Added feature flags**
`read-shared-secret-from-cryptor`

**Affected issues**  
[ARC-1354](https://softwareteams.atlassian.net/browse/ARC-1354)

**How has this been tested?**  
Unit test. 
Stage:

FF On
image

Pollinator green
image

Did a restart backfill and green
image

**Whats Next?**
1. ~~Save sharedSecret to both old and new column~~ [DONE]
2. ~~Migrate existing PROD data from old column to new column~~ [DONE]
3. **Cut-over read traffic using FF** [this pr]
4. Cut-over write traffic to only new column using FF
5. Remove all code that read/write from/to old column
6. Add db migration scripts to remove old column.


src/jira/client/jira-client.test.ts
```diff
@@ -1,6 +1,15 @@
...
 import { getJiraClient } from "./jira-client";
+import { getHashedKey } from "models/sequelize";
+import * as Axios from "./axios";
+import { BooleanFlags, booleanFlag } from "config/feature-flags";
+import { when } from "jest-when";
+
+jest.mock("config/feature-flags", ()=>({
+	...jest.requireActual("config/feature-flags"),
+	booleanFlag: jest.fn()
+}));
```
mboudreau, line 12: just use `when` instead of requiring the real file?

src/jira/verify-installation.test.ts
```diff
@@ -3,25 +3,34 @@ import { verifyJiraInstallation } from "./verify-installation";
...
 import { getAxiosInstance } from "./client/axios";
-import { mocked } from "ts-jest/utils";
+import { BooleanFlags, booleanFlag } from "config/feature-flags";
+import { when } from "jest-when";
 
 jest.mock("./client/axios");
+jest.mock("config/feature-flags", ()=>({
+	...jest.requireActual("config/feature-flags"),
+	booleanFlag: jest.fn()
+}));
```
mboudreau, line 13: just `jest.mock("config/feature-flags")` will do

src/models/jira-client.test.ts
```diff
@@ -6,13 +6,13 @@ describe("JiraClient", () => {
...
 
-		beforeEach(() => {
+		beforeEach(async () => {
 			const installation: any = {
 				jiraHost,
 				sharedSecret: "secret"
 			};
 
-			jiraClient = new JiraClient(installation, logger);
+			jiraClient = await JiraClient.getNewClient(installation, logger);
```
mboudreau, line 15: Why this pattern?

gxueatlassian: I couldn't find a way to make constructor async (to decrypt)

mboudreau: ah right. I guess that's acceptable for now until we don't use the FF anymore.  In the future, I think we'll just send the whole installation object and we'd only decrypt the secret when a request is made.

mboudreau (approved):

PR #1389, opened 2022-07-20
removed feature flag use-new-github-client-for-github-config-post

joshkay10:
**What's in this PR?**
Removing feature flag use-new-github-client-for-github-config-post

**Why**
It is overdue

**How has this been tested?**  
Updated tests to be only ghclient will run through stage

**Whats Next?**
More Feature flagss


thombergs (dismissed):

mboudreau (approved):

thombergs (approved):

PR #1390, opened 2022-07-20
remove ff client-for-github-config

joshkay10:
**What's in this PR?**
Removing ff use-new-github-client-for-github-config

**How has this been tested?**  
Updated tests to drop probot version and only test gh client
Will soak in stage


thombergs (approved):

mboudreau (approved):

PR #1391, opened 2022-07-20
remove ff for USE_NEW_GITHUB_CLIENT_FOR_GET_REPO

joshkay10:
What's in this PR?
Removing ff use-new-github-client-for-get-repo

How has this been tested?
Will soak in stage


thombergs (approved):

PR #1393, opened 2022-07-20
remove ff use-new-github-client-for-pr-title

joshkay10:
**What's in this PR?**
Removing ff use-new-github-client-for-pr-title

**How has this been tested?**  
Soaking in stage


thombergs (approved):
I checked that the FF is currently serving `true` in both staging and prod, so there will be no change in logic when we merge this PR

PR #1400, opened 2022-07-22
Sai 12053 remove ff from git hub app

jason-hooker:
**What's in this PR?**
- Removes the feature flag controlling the sending of related commits with deployments and permanently enables this feature

**Why**
- The feature for sending of related commits with deployments is fully rolled out in production

**How has this been tested?**  
- Unit tests


jason-hooker:
Closing because something went wrong with the sync with master.

PR #1401, opened 2022-07-22
Arc 1456 auto app creation

Harminder84:
**What's in this PR?**
Automatic app creation has been implemented using App manifest flow.

Sequence diagram 

App manifest flow

Video demonstrating the feature

https://user-images.githubusercontent.com/14277763/180709757-9ad38975-0db3-4289-bc50-2fd8af1c7e3b.mov

**Why**
This is one of the options to create GitHub app for enterpise

**Added feature flags**
Existing flag `GHE_SERVER`

**Affected issues**  
[ARC-1456]

**How has this been tested?**  
Unit test cases added for new routes

[ARC-1456]: https://softwareteams.atlassian.net/browse/ARC-1456


ghe-app-manifest.template.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "ghe-app-for-jira",
```
rachellerathbone, line 2: We probably want this to be an empty string right? Otherwise this will auto-fill the input for every customer which they will then have to delete.

rachellerathbone:
Can you add a video recording of the implementation now that you've included the whole flow (from clicking 'create' on auto/manual through to the redirect screen).

ghe-app-manifest.template.json
Harminder84: Don't we want to give any relevant default name?

views/jira-redirect.hbs
```diff
@@ -0,0 +1,15 @@
+<!doctype html>
```
Harminder84, line 1: This is a temporary page to complete the flow. Actual page will be developed as part of another ticket.

static/js/jira-server-url.js
```diff
@@ -81,7 +81,8 @@ const verifyGitHubServerUrl = (gheServerURL) => {
 			},
 			function(data) {
 				if (data.success) {
-					const pagePath = data.appExists ? "github-list-servers-page" : "github-app-creation-page";
+					//const pagePath = data.appExists ? "github-list-servers-page" : "github-app-creation-page";
```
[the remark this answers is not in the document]
Harminder84: I did it for local testing, uncommented.

krazziekay (commented):
I'd suggest to add test cases for the new routes.

src/routes/github/manifest/github-manifest-complete-get.ts
```diff
@@ -0,0 +1,36 @@
...
+		gitHubAppName: gitHubAppConfig.name,
+		gitHubBaseUrl: gheHost,
+		gitHubClientId: gitHubAppConfig.client_id,
+		gitHubClientSecret: gitHubAppConfig.client_secret,
+		webhookSecret: gitHubAppConfig.webhook_secret,
+		privateKey:  gitHubAppConfig.pem,
+		installationId: installation.id
+	});
+	req.session.temp = undefined;
+	res.json({ success:true });
```
Harminder84, line 35: Need to render app installation page which is part of another ticket and is in progress.

ghe-app-manifest.template.json
mboudreau: it's not a bad idea, however, something to consider is that the name has to be unique per server...

rachellerathbone:
Just pulled this branch to take a look locally and am running into this error when I try to run the app. Any thoughts?

Screen Shot 2022-07-26 at 1 39 20 pm

src/routes/github/manifest/github-manifest-complete-get.ts
```diff
@@ -0,0 +1,36 @@
+import { Request, Response } from "express";
+import axios from "axios";
+import { GitHubServerApp } from "~/src/models/github-server-app";
+import { Installation } from "~/src/models/installation";
+
+export const GithubManifestCompleteGet = async (req: Request, res: Response) => {
+	const uuid = req.params.uuid;
```
mboudreau, line 7: where is this coming from?

ghe-app-manifest.template.json
```diff
@@ -0,0 +1,34 @@
...
+    "deployment_status",
+    "issue_comment",
+    "issues",
+    "pull_request",
+    "pull_request_review",
+    "push",
+    "repository",
+    "workflow_run"
+  ]
+}
```
mboudreau, line 34: is there a reason we're not using handlebars to generate this JSON or even just using a function in the code with string interpolation?

static/js/jira-redirect.js
```diff
@@ -0,0 +1,22 @@
+const params = new URLSearchParams(window.location.search.substring(1));
+const gheHost = params.get("serverUrl");
```
rachellerathbone, line 2: Where are we passing this as a param? In `jira-server-url.js` it looks like we're setting `customData` to an empty object if no app exists `const customData = data.appExists ?  { serverUrl: gheServerURL } : {};` so from empty state we're never passing this value to the URL. And then in `github-app-creation.js` the only thing we're pulling from params is the jiraHost. So my guess here is that `gheHost` will be null.

We need to fix this and also add error handling in the event there isn't a gheHost.

rachellerathbone (changes requested):

ghe-app-manifest.template.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "ghe-app-for-jira",
+  "url": "https://github.com/marketplace/jira-software-github",
+  "redirect_url": "<<APP_HOST>>/github/manifest/<<UUID>>/complete",
+  "hook_attributes": {
+    "url": "<<APP_HOST>>/github/<<UUID>>/events"
+  },
+  "setup_url": "<<APP_HOST>>/github/<<UUID>>/setup",
```
mboudreau, line 8: Can you explain the flow here?  UUID is setup beforehand but the app is not installed yet, so what's the point of having the UUID there?

static/js/github-app-creation.js
```diff
@@ -0,0 +1,40 @@
+/* globals $, AP */
```
rachellerathbone, line 1: This file needs to be renamed to jira-app-creation. At this point in the flow we are still 'in jira'. We switch to prepending GH when we open a new tab to navigate our of Jira.

ghe-app-manifest.template.json
mboudreau: Also, keep in mind that we might use this app manifest for our own dev flow.  As might want to keep it a bit more abstracted?

mboudreau (changes requested):
a few issues that needs addressing

ghe-app-manifest.template.json
rachellerathbone: Caught up with Jade on this one and she agrees it should be blank. From a UX standpoint, we'd only pre-populate an input field if it wasn't clear what a user is meant to enter. However, is this case, app name is pretty straightforward. Plus, to @mboudreau's point: GH app names need to be unique per server so they second a customer tries to create a second app, this would fail. 

Just out of curiosity, what happens when you try and add an existing app name in the input? Guessing GH renders some kind of error but want to be sure.

ghe-app-manifest.template.json
rachellerathbone: @mboudreau do you have thoughts on how we should handle the url value here? This is required and is meant to link to the apps homepage. However, we don't have the app name at this point (even if we prepopulate, people can edit it). The best I can see us doing is entering `<<APP_HOST>>/apps/` and adding messaging somewhere in the auto flow to let customers know they will need to update this in their settings manually?

static/js/jira-app-creation.js
```diff
@@ -0,0 +1,40 @@
...
+const jiraHost = params.get("xdm_e");
+
+function openChildWindow(url) {
+	const child = window.open(url);
+	const interval = setInterval(function () {
+		if (child.closed) {
+			clearInterval(interval);
+			AP.navigator.reload();
+		}
+	}, 100);
```
gxueatlassian, line 12: i suggest we can do 1 or 2 seconds

static/js/jira-redirect.js
```diff
@@ -0,0 +1,22 @@
...
+$(document).ready(() => {
+  $.ajax({
+    type: "GET",
+    url: `/github/manifest?gheHost=${gheHost}`,
+    success: function (appManifest) {
+      const newForm = jQuery("<form>", {
+        "action": `${gheHost}/settings/apps/new`,
+        "method": "post",
+        "target": "_top"
```
gxueatlassian, line 11: change it to `_self`

ghe-app-manifest.template.json
mboudreau: Maybe we should get together on this, but my understanding is that we shouldn't create any apps on our end until the github flow is done, which will provide us with everything we need to create the app in our DB.  Creating the app beforehand means that the process might not finish and we'll have apps that were never created...

src/routes/github/manifest/github-manifest-complete-get.ts
```diff
@@ -0,0 +1,36 @@
...
+		throw new Error("Jira Host not found");
+	}
+	const gheHost = req.session.temp?.gheHost;
+	if (!gheHost) {
+		throw new Error("GitHub Enterprise Host not found");
+	}
+	// complete GitHub app manifest flow
+	const apiUrl = `${gheHost}/api/v3/app-manifests/${req.query.code}/conversions`;
+	const gitHubAppConfig = (await axios.post(apiUrl, {}, { headers: { Accept: "application/vnd.github.v3+json" } })).data;
+	const installation = await Installation.getForHost(jiraHost);
```
gxueatlassian, line 19: validation of jiraHost should be done before the code exchange.

src/routes/github/manifest/github-manifest-complete-get.ts
gxueatlassian: maybe we can put both uuid and gheHost in the session cookie, as sessio cookie is signed cannot be tempered.

src/routes/github/manifest/github-manifest-complete-get.test.ts
```diff
@@ -0,0 +1,95 @@
...
+			webhook_secret: "webhook_secret_test",
+			pem: "private_key_test"
+		});
+		await Installation.install({
+			host: res.locals.jiraHost,
+			sharedSecret: "new-encrypted-shared-secret",
+			clientKey: "client-key"
+		});
+		await GithubManifestCompleteGet(req, res);
+		expect(res.json).toBeCalled();
```
gxueatlassian, line 92: 1. assert the record is created in db ( either `GitHubServerApp.install` is called or checking db direclty or use `GitServerApp.find` ) and assert the vlaue is correc.t

static/js/jira-app-creation.js
```diff
@@ -0,0 +1,40 @@
+/* globals $, AP */
```
krazziekay, line 1: This is a duplicate of `jira-select-card-option.js`. All we need here is the logic for click event on `optionBtn`. 
I think we can simply move that logic inside `jira-select-card-option.js, using the `id` selector of the button instead of the using `.optionBtn`(`class`).

krazziekay: I've added a _TODO_ in my PR for this part. Can check it out here:
https://github.com/atlassian/github-for-jira/blob/ef01dae8ae0c1451191dc05f1cd0cb2aabad4fba/static/js/jira-select-card-option.js

ghe-app-manifest.template.json
Harminder84: @mboudreau, updated with function with string interpolation

mboudreau (approved):

rachellerathbone (approved):

PR #1405, opened 2022-07-24
Probot sweep 220725

joshkay10:
**What's in this PR?**
Now the Flags are gone, I am ripping out Octokit/probot references that are no longer used.
Have broken it up into smaller PRs becuase it touches quite a lot of areas and was too hard to review as one big bang.
This one is mainly focused on the backfill codes.

**Why**
No longer used so pulling it out

**Testing**
Validated it still works and has been soaking on test > 4 hours

**Whats Next?**
GitHubApi and github middleware. remove the res.local values for github


src/config/github-api.ts
```diff
@@ -1,12 +1,11 @@
...
-import { enhanceOctokit } from "./enhance-octokit";
 import { Options } from "probot/lib/github";
 
 export const GithubAPI = (options: Partial<Options> = {}): GitHubAPI => {
 	options.logger = options.logger || getLogger("config.github-api");
 	options.throttle = {
 		enabled: false
 	};
-	return enhanceOctokit(GitHubAPI(options as Options));
+	return GitHubAPI(options as Options);
```
joshkay10, line 10: will be removing GitHubAPI in another PR since it would blow this one out.
Focus was `enhanceOctokit`

src/config/rate-limiting-error.ts
```diff
@@ -0,0 +1,12 @@
+export class RateLimitingError extends Error {
```
joshkay10, line 1: This code is as is, it was in `enhance-octokit.ts` just split it out into seperate file.
Should be able to remove it soon, couple of tests rely on it so will do it seperately.

src/sync/installation.ts
```diff
@@ -446,65 +439,62 @@ export async function maybeScheduleNextTask(
...
+			const result = await deduplicator.executeWithDeduplication(
+				"i-" + installationId + "-" + jiraHost,
+				() => doProcessInstallation(data, sentry, installationId, jiraHost, logger, (delay: number) =>
+					nextTaskDelaysMs.push(delay)
+				));
+
+			switch (result) {
+				case DeduplicatorResult.E_OK:
+					logger.info("Job was executed by deduplicator");
+					await maybeScheduleNextTask(data, nextTaskDelaysMs, logger);
```
joshkay10, line 473: This await was missing.

src/config/rate-limiting-error.ts
```diff
@@ -0,0 +1,12 @@
...
+	/**
+	 * The value of the x-ratelimit-reset header, i.e. the epoch seconds when the rate limit is refreshed.
+	 */
+	rateLimitReset: number;
+
+	constructor(resetEpochSeconds: number) {
+		super("rate limiting error");
+		this.rateLimitReset = resetEpochSeconds;
+		Object.setPrototypeOf(this, RateLimitingError.prototype);
```
gxueatlassian, line 10: I am pretty sure this `setPrototypeOf` is pre-class js era stuff. 
We are using `class`, so we probably don't need this one.

------------

I do understand this is from existing code, not a new change in this PR, so I am happy to leave it for another PR.

src/sync/installation.ts
```diff
@@ -446,65 +439,62 @@ export async function maybeScheduleNextTask(
 const redis = new IORedis(getRedisInfo("installations-in-progress"));
 
 const RETRY_DELAY_BASE_SEC = 60;
-export const processInstallation =
```
gxueatlassian, line 449: /nit
There're quite a bit of `git diff` in this section, do you think it is possible to just remove the `(app: Appliaction) => {` bit, but still keep the original format/indent? 

My personal preference will be try to have the `git diff` show as less as possible, but happy to follow other opinions.

src/config/enhance-octokit.ts
```diff
@@ -1,76 +0,0 @@
...
-	rateLimitReset: number;
-
-	constructor(resetEpochSeconds: number) {
-		super("rate limiting error");
-		this.rateLimitReset = resetEpochSeconds;
-		Object.setPrototypeOf(this, RateLimitingError.prototype);
-	}
-}
-
-const instrumentRequests = (octokit: GitHubAPI) => {
```
gxueatlassian, line 23: So is it:
1. In this PR, that feature of `instrumentRequests` are lost in code?
2. And this feature in prod, is actually already lost due to FF turned on?

src/config/rate-limiting-error.ts
joshkay10: Made it simpler for all and removed it now. the tests that was relying on it had a test for old(this one) and new so was a pretty straight forward remove.

src/sync/installation.ts
joshkay10: fair, have reverted the formatting... for now. you are right, its way easier to read without the new indent.

Will probably re-format it again after approval. The current indent just makes me feel too uncomfortable reading it lol

src/config/enhance-octokit.ts
joshkay10: on friday we removed about 6 feature flags, the code no longer uses this(hasn't for over a month with flags on). It is run in prod but never accessed since we use the GitHub clients instead of Octokits offering

gxueatlassian: Never mind, the instrument is added somewhere else already. All good 👍

src/config/rate-limiting-error.ts
gxueatlassian: even better.

src/config/github-api.ts
gxueatlassian, line 10: wait 😱 , will it be infinite recursion? Maybe try bellow for safety sake?
```javascript
import { GitHubAPI as ProbotGitHubAPI } from "probot";
...
	return ProbotGitHubAPI(options as Options);
```

gxueatlassian: LOL, it wont' be infinite recursion, the `h` is `H` in probot! But still ...

joshkay10: haha yeah its not ideal is it, only use on one class at the moment at least. will update it anyway with `ProbotGitHubAPI`

src/config/github-api.ts
```diff
@@ -1,12 +1,11 @@
-import { GitHubAPI } from "probot";
+import { GitHubAPI as ProbotGitHubAPI } from "probot";
 import { getLogger } from "./logger";
-import { enhanceOctokit } from "./enhance-octokit";
 import { Options } from "probot/lib/github";
 
-export const GithubAPI = (options: Partial<Options> = {}): GitHubAPI => {
+export const GithubAPI = (options: Partial<Options> = {}): ProbotGitHubAPI => {
```
mboudreau, line 5: Is anything still using this reference?  I thought we were only using our github clients from now on?

mboudreau (dismissed):
image

src/config/github-api.ts
joshkay10: in the webhooks theres reference to it, shouldn't be in use but if i remove this the thread it pulls makes the PR way bigger so i split it in to another PR.

gxueatlassian (dismissed):

mboudreau (approved):

PR #1407, opened 2022-07-25
Removing deployment FF

mboudreau:
**What's in this PR?**
Removal of feature flag that's been turned on for a while

**Why**
Not needed anymore

**Removed feature flags**
SUPPORT_BRANCH_AND_MERGE_WORKFLOWS_FOR_DEPLOYMENTS

**Affected issues**  
None

**How has this been tested?**  
Same unit tests as before minus the FF

**Whats Next?**
Archive the FF


joshkay10 (dismissed):

package.json
```diff
@@ -26,7 +26,7 @@
 		"lint:ts": "eslint . --ext .ts,.tsx",
 		"lint:yaml": "yamllint github-for-jira.sd.yml",
 		"pretest": "NODE_ENV=test run-p clean setup db:test",
-		"test": "jest --runInBand --forceExit --coverage",
+		"test": "jest --runInBand --forceExit --coverage --detectOpenHandles",
```
mboudreau, line 29: This will give a breakdown of why jest doesn't close cleanly.  We can use it as a todo list of things that we need to fix as part of the tests.

src/sync/installation.ts
```diff
@@ -446,65 +446,64 @@ export async function maybeScheduleNextTask(
...
+			const result = await deduplicator.executeWithDeduplication(
+				"i-" + installationId + "-" + jiraHost,
+				() => doProcessInstallation(app, data, sentry, installationId, jiraHost, logger, (delay: number) =>
+					nextTaskDelaysMs.push(delay)
+				));
+
+			switch (result) {
+				case DeduplicatorResult.E_OK:
+					logger.info("Job was executed by deduplicator");
+					maybeScheduleNextTask(data, nextTaskDelaysMs, logger);
```
joshkay10, line 482: Do you want to chuck an `await` here while your in the neighbourhood `maybeScheduleNextTask(data, nextTaskDelaysMs, logger);`

test/setup/setup.ts
```diff
@@ -142,7 +141,7 @@ beforeEach(() => {
 	global.gheUserTokenNock = githubUserToken(gheApiNock);
 	global.gheAppTokenNock = githubAppToken(gheApiNock);
 	global.mockSystemTime = (time: number | string | Date) => {
-		const mock = jest.isMockFunction(Date.now) ? mocked(Date.now) : jest.spyOn(Date, "now");
+		const mock = jest.isMockFunction(Date.now) ? jest.mocked(Date.now) : jest.spyOn(Date, "now");
```
joshkay10, line 144: Just highlighting this for other reviewers, `mocked` from ts-jest is deprecated and gone in the next major update 28(which is already released, we are on 27).

joshkay10 (dismissed):

src/sync/installation.ts
mboudreau: actually, this is done on purpose.  We don't want to wait as it's just putting another message on the queue

src/jira/client/jira-client.ts
```diff
@@ -475,6 +479,14 @@ export const getTruncatedIssuekeys = (data: IssueKeyObject[] = []): IssueKeyObje
 				issueKeys: value.issueKeys.slice(ISSUE_KEY_API_LIMIT)
 			});
 		}
+		const association = findIssueKeyAssociation(value);
+		if (association && association.values.length  > ISSUE_KEY_API_LIMIT) {
```
joshkay10, line 483: Could the `association.values` be undefined?
e.g. `association.values?.length` instead

mboudreau: good point

src/jira/client/jira-client.ts
```diff
@@ -414,13 +414,16 @@ const batchedBulkUpdate = async (
 	return Promise.all(batchedUpdates);
 };
 
+const findIssueKeyAssociation = (resource: IssueKeyObject): JiraAssociation | undefined =>
+	resource.associations?.find(a => a.associationType == "issueIdOrKeys");
```
gxueatlassian, line 418: Will it be possible to multiple entries in the `associations` that have type `issueIdOrKeys`?
Not sure what's the spec for this api say.  Couldn't find anything on doc.

src/sync/installation.ts
```diff
@@ -454,65 +455,64 @@ export async function maybeScheduleNextTask(
...
-			const { installationId, jiraHost } = data;
-
-			logger.child({ gitHubInstallationId: installationId, jiraHost });
-
-			try {
-				if (await isBlocked(installationId, logger)) {
-					logger.warn("blocking installation job");
-					return;
-				}
+export const processInstallation = (app: Application) => {
```
gxueatlassian, line 458: There's a big `git diff` here. It'd be great if you can stick to original format and show a cleaner `git diff` .

src/jira/client/jira-client.ts
```diff
@@ -463,18 +466,27 @@ const truncateIssueKeys = (repositoryObj) => {
...
 interface IssueKeyObject {
 	issueKeys?: string[];
+	associations?: JiraAssociation[];
 }
 
 export const getTruncatedIssuekeys = (data: IssueKeyObject[] = []): IssueKeyObject[] =>
 	data.reduce((acc: IssueKeyObject[], value: IssueKeyObject) => {
 		// Filter out anything that doesn't have issue keys or are not over the limit
-		if (value.issueKeys && value.issueKeys.length > ISSUE_KEY_API_LIMIT) {
+		if (value?.issueKeys && value.issueKeys.length > ISSUE_KEY_API_LIMIT) {
```
gxueatlassian, line 475: Actually since the FF is removed, will it means that there's no `issueKeys` at all? All we have are just `associations`?
Or are you just concerning about existing msg in the queue?

src/jira/client/jira-client.ts
mboudreau: There should only be 1 association per type, but also, we're the ones creating it in the first place.  Having to support multiple is much more difficult and unnecessary at this moment.

src/jira/client/jira-client.ts
```diff
@@ -505,8 +517,14 @@ const updateRepositoryIssueKeys = (repositoryObj, mutatingFunc) => {
  * Runs the mutatingFunc on the issue keys field for each branch, commit or PR
  */
 const updateIssueKeysFor = (resources, func) => {
```
joshkay10, line 519: You're the Typescript expert and this looks like the end of a long line of non - typed events, is there any value adding :any types to `resources` and `func`?

src/github/deployment.ts
```diff
@@ -44,7 +44,7 @@ export const processDeployment = async (
 
 	logger.info("processing deployment message!");
 
-	const jiraPayload: JiraDeploymentData | undefined = await transformDeployment(newGitHubClient, webhookPayload, jiraHost, logger);
+	const jiraPayload: JiraDeploymentData | undefined = await transformDeployment(newGitHubClient, webhookPayload, logger);
```
joshkay10, line 47: /nit existing code so 🤷‍♂️, types will be inferred from transformDeployment.
`const jiraPayload = await transformDeployment(newGitHubClient, webhookPayload, logger);`

joshkay10 (approved):

src/jira/client/jira-client.ts
mboudreau: there are issueKeys for other jira entities, not just deployments.  Not all of them uses associations yet.

src/jira/client/jira-client.ts
mboudreau: yeah, agreed, this needs to be fixed, but the whole files is bad and there's a ticket to replace it/fix it already.  There's no reason to add any as it's already assumed.

PR #1411, opened 2022-07-25
Arc 1506 analytics

rachellerathbone:
**What's in this PR?**
[Must-have](https://hello.atlassian.net/wiki/spaces/PF/pages/1807205537/Decision+What+analytics+do+we+need#Must-have-analytics) analytics.

**Why**
These are the analytics that were determined essential to gain insight into customer behaviour and potential 'sticking points' in the app.

**Affected issues**  
ARC-1506.

**How has this been tested?**  
Checking for logs in [splunk](https://splunk.paas-inf.net/en-GB/app/search/search?q=search%20%60micros_analytics-service%60%20env%3D%22prod-*%22%20m.t%3D%22application%22%20ext.event.name%3DgithubConfigurationScreen&display.page.search.mode=verbose&dispatch.sample_ratio=1&earliest=-24h%40h&latest=now&display.page.search.tab=events&display.general.type=events&sid=1658811075.206652_08994721-1899-43E7-8F3A-864FED18C26C) 

**Whats Next?**
Setting up a dashboard in Redash


src/interfaces/common.ts
```diff
@@ -11,17 +11,22 @@ export enum BooleanEnum {
 
 // All variables below were defined by DataPortal. Do not change their values as it will affect our metrics logs and dashboards.
 export enum AnalyticsEventTypes {
-	ScreenEvent = "screen",
-	UiEvent = "ui",
-	TrackEvent = "track",
-	OperationalEvent = "operational",
-	TraitEvent = "trait",
```
rachellerathbone, line 18: Removed this as the trait event is primarily used by the personalization team (major use case is for in-product messaging and onboarding experiences). We have been advised by the analytics team to only use the above 4 event types.

src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -133,6 +135,14 @@ export const GithubConfigurationGet = async (req: Request, res: Response, next:
 	gitHubAppId ? req.log.debug(`Displaying orgs that have GitHub Enterprise app ${gitHubAppId} installed.`)
 		: req.log.debug("Displaying orgs that have GitHub Cloud app installed.");
 
+	const gitHubProduct = gitHubAppId ? "server" : "cloud";
+
+	sendAnalytics(AnalyticsEventTypes.ScreenEvent, {
+		name: AnalyticsScreenEventsEnum.ConnectAnOrgScreenEventName,
+		jiraHost,
+		gitHubProduct
```
rachellerathbone, line 143: This will help us determine how many customers (both cloud and server) are making it to the point where they can connect an org.

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -166,18 +182,32 @@ const renderJiraCloudAndEnterpriseServer = async (res: Response, req: Request):
...
 		jiraHost,
-		connectedOrgCount: installations.total + gheServers.length
+		connectedOrgCountCloudCount: successfulCloudConnections.length,
+		connectedOrgCountServerCount: successfulServerConnections,
+		totalOrgCount: successfulCloudConnections.length + successfulServerConnections,
+		failedCloudBackfillCount: countStatus(failedCloudConnections, "FAILED"),
+		failedServerBackfillCount: countStatus(failedServerConnections, "FAILED"),
+		successfulCloudBackfillCount: countStatus(successfulCloudConnections, "FINISHED"),
+		successfulServerBackfillCount: countStatus(successfulServerConnections, "FINISHED"),
+		numberOfSkippedRepos: countNumberSkippedRepos(allSuccessfulConnections)
```
rachellerathbone, line 210: Metrics added here:
- number of successful and failed connections per product (cloud | server)
- number of connected orgs per product
- total number of connected orgs (easy way to create a chart with total number but may not be needed as we may be able to add these numbers in redash... thoughts?)
- the number of repos skipped: will give us insight how frequently we're skipping backfilling repos for each customer

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -142,7 +155,11 @@ const renderJiraCloudAndEnterpriseServer = async (res: Response, req: Request):
 	const gheServerSubscriptions = difference(subscriptions, ghCloudSubscriptions);
 
 	// Connections for GHCloud
-	const { installations, successfulConnections, failedConnections } = await getConnectionsAndInstallations(ghCloudSubscriptions, req);
+	const {
+		installations,
+		successfulConnections: successfulCloudConnections,
```
rachellerathbone, line 160: Renamed these so there is a clear distinction between successful/failed connections for cloud vs server.

src/util/analytics-client.ts
```diff
@@ -10,11 +10,10 @@ const appKey = `com.github.integration${instance ? `.${instance}` : ""}`;
...
 let analyticsNodeClient;
 
-export function sendAnalytics(eventType: "trait", attributes: Record<string, unknown>)
-export function sendAnalytics(eventType: "screen", attributes: { name: string } & Record<string, unknown>)
+export function sendAnalytics(eventType: "screen", attributes?: { name: string } & Record<string, unknown>)
 export function sendAnalytics(eventType: "ui" | "track" | "operational", attributes: Record<string, unknown>)
-export function sendAnalytics(eventType: string, attributes: Record<string, unknown>): void {
-	if (!analyticsClient || !isNodeProd()){
+export function sendAnalytics(eventType: string, attributes?: Record<string, unknown>): void {
+	if (!analyticsClient && !isNodeProd()){
```
rachellerathbone, line 16: @gxueatlassian turns out changing this to || per your suggestion in my last pr broke analytics (currently not being sent).

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -111,6 +111,13 @@ const getConnectionsAndInstallations = async (subscriptions: Subscription[], req
 	return { installations, successfulConnections, failedConnections };
 };
 
+const countStatus = (connections, syncStatus: string): number =>
+	connections && connections.filter(org => org?.syncStatus === syncStatus).length || 0;
+
```
joshkay10, line 116: Can we add a type to `connections` for `countStatus` and `countNumberSkippedRepos` they look like `Subscription[]`?

gxueatlassian: +1

src/util/analytics-client.ts
gxueatlassian: Tbh it is always hard to wrap my head around multi conditions with `!`. 
Just out of curious, are we sending analytics to stage as well?

----------

Anyway, this is not a blocker, I'll leave this to your decision. Good to me. You can close this thread.

src/util/analytics-client.ts
```diff
@@ -33,7 +32,7 @@ export function sendAnalytics(eventType: string, attributes: Record<string, unkn
 		tenantId: "NONE"
 	};
 
-	attributes.appKey = appKey;
+	attributes!.appKey = appKey;
```
gxueatlassian, line 35: Some deja vu here, is this `attributes` optional or not? Why the `!` here?

src/util/analytics-client.ts
mboudreau: this is a weird conditional though.  It should _always_ return if the analytics client is undefined no matter the env or else it'll create an error. Best way IMO is to do `!analyticsClient || isNodeDev()` so it never sends anything while in development and only sends it in production mode, but only if the client is available.

src/util/analytics-client.ts
mboudreau: yeah, agreed.  Might be easier to have the parameter be `attributes: Record<string, unknown> = {}` so that it's never undefined

src/util/analytics-client.ts
rachellerathbone: This is the original code from @atrigueiro branch that got merged but happy to update.

src/util/analytics-client.ts
```diff
@@ -12,8 +12,8 @@ let analyticsNodeClient;
 
 export function sendAnalytics(eventType: "screen", attributes?: { name: string } & Record<string, unknown>)
```
gxueatlassian, line 13: Should `attributes?: { name: string }` be `attributes: { name: string }`? Remove the `?` to match the removal of `!` in the code bellow?

gxueatlassian (approved):

src/routes/jira/configuration/jira-configuration-get.ts
```diff
@@ -166,18 +183,31 @@ const renderJiraCloudAndEnterpriseServer = async (res: Response, req: Request):
...
+	const completeConnections = allSuccessfulConnections.filter(connection => connection.syncStatus === "FINISHED");
+
 	sendAnalytics(AnalyticsEventTypes.ScreenEvent, {
 		name: AnalyticsScreenEventsEnum.GitHubConfigScreenEventName,
 		jiraHost,
-		connectedOrgCount: installations.total + gheServers.length
+		connectedOrgCountCloudCount: successfulCloudConnections.length,
+		connectedOrgCountServerCount: successfulServerConnections,
+		totalOrgCount: successfulCloudConnections.length + successfulServerConnections,
+		failedCloudBackfillCount: countStatus(successfulCloudConnections, "FAILED"),
```
krazziekay, line 206: This is a bit confusing. Can there be any `successfulCloudConnection` whose status is `FAILED`? 

All these connections come from `installations` in the method `getConnectionsAndInstallations`. Amongst these `installations`, the `successful` and `failed` ones are already separated. So, can there be a status `FAILED` connection in the already filtered out `successfulCloudConnection`? 

Shouldn't this be `failedCloudConnections.length`?

PR #1412, opened 2022-07-25
ARC-1527 - Subset of tasks to backfill

joshkay10:
**What's in this PR?**
New argument  to pass in the `/resync` endpoint to filter what entities to sync.

**Why**
Sometimes we need to only sync certain things, example if deploymeny webhooks are down and gone, we can just since all deployments.

**Runbook**
[WIP runbook](https://hello.atlassian.net/wiki/spaces/PF/pages/1824654901/HOW+TO+run+backfill+resync+endpoint)

**Affected issues**  
ARC-1527

**How has this been tested?**  
Unit tests added to make sure tasks are filtered as expected.
Ran testing on stage to manually check that only expected entities are backfilled on installation and through resync endpoint

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -64,8 +64,8 @@ export const sortedRepos = (repos: Repositories): [string, RepositoryData][] =>
 			new Date(a[1].repository?.updated_at || 0).getTime()
 	);
 
-export const getTargetTasks = async (targetTasks?: TaskType[]): Promise<TaskType[]> => {
-	if (targetTasks) {
+export const getTargetTasks = (targetTasks?: TaskType[]): TaskType[] => {
+	if (targetTasks?.length) {
 		return intersection(allTaskTypes, targetTasks);
 	}
 
@@ -73,7 +73,7 @@ export const getTargetTasks = async (targetTasks?: TaskType[]): Promise<TaskType
 };
 
 const getNextTask = async (subscription: Subscription, targetTasks?: TaskType[]): Promise<Task | undefined> => {
-	const tasks = await getTargetTasks(targetTasks);
+	const tasks = getTargetTasks(targetTasks);
 
 	if (subscription.repositoryStatus !== "complete") {
 		return {
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -55,7 +55,7 @@ export interface TaskPayload<E = any, P = any> {
 	jiraPayload?: P;
 }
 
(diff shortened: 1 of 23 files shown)


src/sync/sync.types.ts
```diff
@@ -0,0 +1,2 @@
+/* valid task types */
+export type TaskType = "repository" | "pull" | "commit" | "branch" | "build" | "deployment";
```
joshkay10, line 2: started using this sucker all over the shop so pulled it out of installation.ts and created new types file...

src/sync/installation.ts
```diff
@@ -63,8 +64,23 @@ export const sortedRepos = (repos: Repositories): [string, RepositoryData][] =>
 			new Date(a[1].repository?.updated_at || 0).getTime()
 	);
 
-const getNextTask = async (subscription: Subscription): Promise<Task | undefined> => {
-	const tasks = taskTypes;
+export const getTargetTasks = async (jiraHost: string, allTasks: TaskType[], targetTasks?: TaskType[]): Promise<TaskType[]> => {
```
joshkay10, line 67: `allTasks` doesn't need to exist since we have access to the allTaskTypes global, one extra argument to make it purer was a small cost.

src/sync/installation.ts
```diff
@@ -1,4 +1,7 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
+import { intersection } from "lodash";
+import IORedis from "ioredis";
```
joshkay10, line 3: I like having node module imports up the top so I moved these here to get opinions....
`import { intersection } from "lodash"; `is new tho

src/sync/installation.ts
```diff
@@ -63,8 +64,23 @@ export const sortedRepos = (repos: Repositories): [string, RepositoryData][] =>
 			new Date(a[1].repository?.updated_at || 0).getTime()
 	);
 
-const getNextTask = async (subscription: Subscription): Promise<Task | undefined> => {
-	const tasks = taskTypes;
+export const getTargetTasks = async (jiraHost: string, allTasks: TaskType[], targetTasks?: TaskType[]): Promise<TaskType[]> => {
+	if (targetTasks) {
+		return intersection(allTasks, targetTasks);
```
joshkay10, line 69: originally just returned targetTasks, but if some jerk passed in "cats, commits" as the value we want to filter out the trash ones.

mboudreau (approved):
noice

PR #1414, opened 2022-07-26
ARC-1552 - Update the `session` middleware for handling loading screens for GHE cloud and GHE Enterprise

krazziekay:
**What's in this PR?**
- Updated `SessionGet` middleware to display the loading screen. The text is based upon the query parameters passed
- Added test Case for `SessionGet`.
- Added the query parameters in all the places using `session/**`

**How has this been tested?**  
- Tested locally 
- Running Test cases

**Videos**

https://user-images.githubusercontent.com/13076255/183339905-91e9e58d-84bb-487b-a225-0dbca52af881.mov


src/routes/session/session-get.ts
```diff
@@ -2,13 +2,39 @@ import { NextFunction, Request, Response } from "express";
...
+		return next(new Error("Missing redirect url for session enterprise. Needs to be in format `/session/enterprise/:redirectUrl`"));
+	}
+
+	res.render("gitHub-loading.hbs", {
+		APP_URL: process.env.APP_URL,
+		redirectUrl: new URL(req.params[0], process.env.APP_URL).href,
+		nonce: res.locals.nonce,
+		isGitHubCloud: false
+	});
+};
```
mboudreau, line 27: Not sure I understand the need to have a different one between cloud and enterprise.  Aren't they doing the same thing?

krazziekay: Both are loading pages, but I guess they will have different text, so separated them

views/gitHub-loading.hbs
```diff
@@ -0,0 +1,51 @@
+<html>
```
rachellerathbone, line 1: We should probably include the name `session` in the name of this file e.g. gitHub-server-session. Also, it probably makes sense to rename the current `session.hbs` file to something like gitHub-cloud-session.

views/gitHub-loading.hbs
```diff
@@ -0,0 +1,51 @@
...
+  <head>
+    <meta charset="utf-8" />
+    <meta http-equiv="X-UA-Compatible" content="IE=edge" />
+    <meta name="ap-local-base-url" content="{{localBaseUrl}}" />
+    <title>{{title}}</title>
+    <link rel="stylesheet" href="/public/css/loading-screen.css" media="all" />
+    <link rel="stylesheet" href="/public/css/ghe-loading.css" media="all" />
```
rachellerathbone, line 8: I don't think we need to create a separate file for one class. Makes more sense just to add it to the loading-screen file and treat that as the stylesheet for both cloud and server loading screens. Maybe just add a comment above the ghe specific styling.

views/gitHub-loading.hbs
rachellerathbone: Oh, I just realised you're rendering both cloud and server in this file. We can probably just delete the session.hbs file and rename this to just session.hbs or github-session.hbs

views/gitHub-loading.hbs
```diff
@@ -0,0 +1,51 @@
...
+      rel="stylesheet"
+      href="/public/aui/aui-prototyping.css"
+      integrity="DTM1Q+8lU7SzJT+FWr0JFisCSZlwfM0GiAKYy7h1s9vIKa/CIh37s9NuOCqIOgK4tmqrjLK4NuWuIPUQNsikHA=="
+      crossorigin="anonymous"
+      referrerpolicy="no-referrer"
+    />
+    <script src="/public/js/jquery.min.js" nonce="{{nonce}}"></script>
+  </head>
+  <body class="gheLoading___body">
+    {{#if isGitHubCloud}}
```
rachellerathbone, line 19: Have you thought about how you are going to tie this session screen into this conditional check? Right now we're basically saying, if it's a cloud app, render content 'a', otherwise render content 'b'. Unfortunately, that doesn't really work in the server flow when you take the 'Retrieving data from GitHub Enterprise Server' into account.

When a server customer goes through the flow what screen/s they see will depend upon whether they choose auto/manual app creation.

- auto: when they click create on the auto/manual screen, they'll see the session view with 'Redirecting to your GitHub Enterprise Server instance' before being redirected to GH to enter a name for their app. Once that is done, they'll be redirected back to our app and see 'Retrieving data from GitHub Enterprise Server' while we're saving their app data to the db.
- manual: these customers will only see the second session screen. From the auto/manual screen they'll land on the form. After they click next, they'll see  'Retrieving data from GitHub Enterprise Server' while we're saving their app data to the db.

Screen Shot 2022-07-27 at 3 33 35 pm

src/routes/session/session-get.test.ts
```diff
@@ -0,0 +1,93 @@
...
+					.then(response => {
+						expect(response.text.includes("Logging you into GitHub")).toBeFalsy();
+						expect(response.text.includes("Redirecting to your GitHub Enterprise Server instance")).toBeTruthy();
+						expect(response.text.includes("Redirecting to your GitHub Cloud instance")).toBeFalsy();
+					}));
+
+			it("should return error asking for redirect url", () =>
+				supertest(app)
+					.get("/session/enterprise")
+					.expect(500)
```
gxueatlassian, line 62: /nit 
I think 4xx will state the problem better in this case, as is the client's error that forgot to supply a redirect url.

src/routes/session/session-get.ts
gxueatlassian: I guess since we have `isGitHubCloud: false` here, we can do the different text in hbs via this flag?

views/gitHub-loading.hbs
krazziekay: `session.hbs` has been deleted and this file has been renamed to `gitHub-session.hbs`.

views/gitHub-loading.hbs
krazziekay: Separate CSS file removed, added the styling in `loading-screen.css`

src/routes/session/session-get.ts
krazziekay: - `isGitHubCloud` removed.
- Replaced by the content `text` instead.

views/gitHub-loading.hbs
krazziekay: Removed the if condition and passed in the `text` from the backend. For the scenario mentioned above, we might have to add some logic to get the desired `text` and then pass that along, and the `gitHub-session.hbs`(loading screen) will simply render it

src/routes/session/session-router.ts
```diff
@@ -0,0 +1,7 @@
+import { Router } from "express";
+import { SessionEnterpriseGet, SessionGet } from "routes/session/session-get";
+
+export const SessionRouter = Router();
+
+SessionRouter.get(["/enterprise", "/enterprise/*"], SessionEnterpriseGet);
+SessionRouter.get(["/", "/*"], SessionGet);
```
mboudreau, line 7: why do we need a separate handler when they both do exactly the same?

krazziekay: I thought it would be better to separate the routes to make it simple. 
The view is the same in both cases, but the message displayed are different. We can use the same route and pass the _enterprise_ part as a `queryString` or a separate `route parameter`. But that would just add a bit of complexity on the handler, as it would have to display different messages based on the  `queryString` or `params`. 
So instead of doing the conditional handling, I went with this approach.

src/routes/session/session-get.test.ts
krazziekay: Status updated to `400`

src/routes/session/session-router.ts
krazziekay: Reverted back to single route. 
Handled by adding query string.

rachellerathbone (dismissed):

rachellerathbone (approved):

rachellerathbone (dismissed):

rachellerathbone (approved):

src/routes/session/session-get.ts
```diff
@@ -1,18 +1,41 @@
...
-	const url = createUrlWithQueryString(req, req.params[0]);
+export const SessionGet = (req: Request, res: Response) => {
+	const url = createUrlWithQueryString(req, req.params[0] || "");
+	const { title, loadAutoAppCreation } = configForPage(req);
 
-	return res.render("session.hbs", {
-		title: "Logging you into GitHub",
-		APP_URL: process.env.APP_URL,
+	res.render("session.hbs", {
 		redirectUrl: new URL(url, process.env.APP_URL).href,
```
rachellerathbone, line 9: What happens here when req.params[0] isn't defined and we set url to an empty string? Is there a reason why you removed the exception here? I think we should still include this.

rachellerathbone (changes requested):

src/routes/session/session-get.ts
krazziekay: If there is no `req.params[0]` then empty string is used as url. This condition is done in line 5. 
There are 1 test case for this case too `Testing the route just `session` when no query params`

rachellerathbone, line 9: What are we expecting to happen here when url is an empty string? Is there a reason we ditched the error that @mboudreau had added here?

rachellerathbone (changes requested):

rachellerathbone (approved):

rachellerathbone (approved):

PR #1417, opened 2022-07-27
ARC-1519-part-2 - Improve Jira Routing for connect/app creation 

krazziekay:
**What's in this PR?**
- Updating route for deleting Jira subscription

**How has this been tested?**  
- In local dev environment
- Test cases


src/routes/jira/jira-delete.ts
```diff
@@ -4,17 +4,26 @@ import { Request, Response } from "express";
...
 	if (!jiraHost) {
 		req.log.error("Missing Jira Host");
 		res.status(401).send("Missing jiraHost in body");
 		return;
 	}
 
-	req.log.info({ installationId }, "Received delete jira configuration request");
+	if (!installationId) {
+		req.log.error("Missing Github Installation ID");
+		res.status(401).send("Missing Github Installation ID in params");
```
joshkay10, line 22: /nit - message will be misleading for old route since id is in the body... but ... meh..

joshkay10 (approved):

src/routes/jira/jira-delete.ts
krazziekay: Updated the message to `Missing Github Installation ID`.

PR #1428, opened 2022-07-28
Arc 926 Webhook handler for pull request, create and delete branch

Harminder84:
**What's in this PR?**
Hooked Pull request, create and delete branch handler with new webhook receiver

**Added feature flags**
Part of new webhook handler which is not used in production.

**Affected issues**  
[ARC-926]
[ARC-929]
[ARC-930]

**How has this been tested?**  
Unit test cases added

[ARC-926]: https://softwareteams.atlassian.net/browse/ARC-926
[ARC-929]: https://softwareteams.atlassian.net/browse/ARC-929
[ARC-930]: https://softwareteams.atlassian.net/browse/ARC-930


src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -56,6 +58,21 @@ const webhookRouter = (context: WebhookContext) => {
 				GithubWebhookMiddleware(issueWebhookHandler)(context);
 			}
 			break;
+		case "pull_request":
+			if (context.action === "opened" || context.action === "closed"
+				|| context.action === "reopened" || context.action === "edited") {
+				GithubWebhookMiddleware(pullRequestWebhookHandler)(context);
```
joshkay10, line 64: /nit do you expect this list to grow?
I'm quite fond of having `const VALID_PR_ACTIONS = [""opened","reopened","closed","edited"]`
Then you can just `VALID_PR_ACTIONS.includes(context.action)`

might be overkill tho

Harminder84: It is final as of now.

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -56,6 +58,21 @@ const webhookRouter = (context: WebhookContext) => {
...
+		case "pull_request_review":
+			GithubWebhookMiddleware(pullRequestWebhookHandler)(context);
+			break;
+		case "create":
+			GithubWebhookMiddleware(createBranchWebhookHandler)(context);
+			break;
+		case "delete":
+			GithubWebhookMiddleware(deleteBranchWebhookHandler)(context);
+			break;
 	}
```
joshkay10, line 76: Is it worth logging out on `default:` the skipped webhooks? Might be more noise than its worth, but as I understand we subscribe to webhooks in the app permissions so we should only receive one we want?

Harminder84: We receive webhook events only for those that we subscribe in app permission but there are certain event actions that we're ignoring. It will more noise in logs if we start logging it.

src/routes/github/webhook/webhook-receiver-post.test.ts
```diff
@@ -132,4 +132,73 @@ describe("webhook-receiver-post", () => {
 		expect(GithubWebhookMiddleware).toBeCalledWith(issueWebhookHandler);
 	});
 
+	it("should call pull handler", async () => {
+		req = {
```
gxueatlassian, line 136: So this is getting longer for each test. Can you see if you extract them into a method and reduce the clutter? 

```
req = {
			headers: {
				"x-hub-signature-256": "sha256=b3c1c0ce21666c0349b8f3010ce81299adcc9c4ee09b904f20f712292c19b792",
				"x-github-event": "pull_request"
			},
			params: {
				uuid
			},
			body: {
				action: "opened"
			}
		};
```

src/routes/github/webhook/webhook-receiver-post.test.ts
```diff
@@ -132,4 +132,73 @@ describe("webhook-receiver-post", () => {
...
+			params: {
+				uuid
+			},
+			body: {
+				action: "opened"
+			}
+		};
+
+		await WebhookReceiverPost(req, res);
+		expect(GithubWebhookMiddleware).toBeCalledWith(pullRequestWebhookHandler);
```
gxueatlassian, line 150: I think there're three layers of assertions:

1. Expect the `(GithubWebhookMiddleware).toBeCalledWith(pullRequestWebhookHandler);` <-- this is good.
2. Expect the return of `GithubWebhookMiddleware(pullRequestWebhookHandler)` is called with a correct `context`
3. Expect `pullRequestWebhookHandler` is called with correct context with extra info.

-----

You have done `1`, which is good. I am looking for `2` and `3` to improve the test. I think `3` could be hard, need more setup on the DB. But maybe try if you can achieve `2`?

src/routes/github/webhook/webhook-receiver-post.test.ts
Harminder84: @gxueatlassian , extracted in a seperate method

src/routes/github/webhook/webhook-receiver-post.test.ts
Harminder84: @gxueatlassian, added assertion for verifying context.

src/routes/github/webhook/webhook-receiver-post.ts
mboudreau: personally, I'd still like to see josh's suggestion.  It's much cleaner

Harminder84: Done!!

rachellerathbone (approved):

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -56,10 +59,24 @@ const webhookRouter = (context: WebhookContext) => {
...
 };
 
-const createHash = (data: BinaryLike, secret: string): string => {
+export const createHash = (data: BinaryLike, secret: string): string => {
```
gxueatlassian, line 79: Personally I wouldn't export this `createHash`, but duplicate it in the test file.

gxueatlassian (approved):

src/routes/github/webhook/webhook-receiver-post.ts
gxueatlassian: We want a valid test, so if we have a bug in the `createHash` function, like `const createHash = () => { return 'always_success'`, the test will still "success", but there's a bug, so test is not a valid test.

---------

Instead, we could just duplicate the method in the test, as that `sha256` is an intention we want to test.

PR #1430, opened 2022-07-28
Fixing logger issue

mboudreau:
**What's in this PR?**
A few fixes to logger to make it log things as it should

**Why**
There was an issue with circular dependencies fixed with the `inspect` outputMode and another around logger references when doing a ternary between `logger.error` and `logger.warn`

**Added feature flags**
None

**Affected issues**  
PCS-107860

**How has this been tested?**  
run the app

**Whats Next?**
Fix all the other things...

### DIFF
--- a/src/config/logger.ts
+++ b/src/config/logger.ts
@@ -33,8 +33,8 @@ const requestSerializer = (req) => req && ({
 	url: req.originalUrl || req.url,
 	path: req.path,
 	headers: req.headers,
-	remoteAddress: req.socket.remoteAddress,
-	remotePort: req.socket.remotePort,
+	remoteAddress: req.socket?.remoteAddress,
+	remotePort: req.socket?.remotePort,
 	body: req.body
 });
 
--- a/src/config/logger.ts
+++ b/src/config/logger.ts
@@ -1,13 +1,13 @@
-import Logger, { createLogger, INFO, levelFromName, stdSerializers } from "bunyan";
+import Logger, { createLogger, INFO, levelFromName, LoggerOptions, stdSerializers } from "bunyan";
 import bformat from "bunyan-format";
 import { filteringHttpLogsStream } from "utils/filtering-http-logs-stream";
-import { Request } from "express";
-import { AxiosResponse } from "axios";
 import { createHashWithSharedSecret } from "utils/encryption";
+import { isArray, isString } from "lodash";
 
 // For any Micros env we want the logs to be in JSON format.
 // Otherwise, if local development, we want human readable logs.
-const outputMode = process.env.MICROS_ENV ? "json" : "short";
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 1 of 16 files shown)


src/config/logger.ts
```diff
@@ -17,34 +17,47 @@ const outputMode = process.env.MICROS_ENV ? "json" : "short";
...
 export const FILTERING_FRONTEND_HTTP_LOGS_MIDDLEWARE_NAME = "frontend-log-middleware";
 // add levelInString to include DEBUG | ERROR | INFO | WARN
-const LOG_STREAM = filteringHttpLogsStream(FILTERING_FRONTEND_HTTP_LOGS_MIDDLEWARE_NAME,
+const LOG_STREAM = filteringHttpLogsStream(
+	FILTERING_FRONTEND_HTTP_LOGS_MIDDLEWARE_NAME,
 	bformat({ outputMode, levelInString: true })
 );
 
-const responseSerializer = (res: AxiosResponse) => ({
+const responseSerializer = (res) => res && ({
```
thombergs, line 25: Why remove the typing here and below?

src/jira/client/axios.ts
```diff
@@ -63,7 +63,7 @@ const getErrorMiddleware = (logger: Logger) =>
 
 		// Log appropriate level depending on status - WARN: 300-499, ERROR: everything else
 		// Log exception only if it is error, because AxiosError contains the request payload
-		(isWarning ? logger.warn : logger.error)({ err: error, res: error?.response }, errorMessage);
+		logger[isWarning ? "warn" : "error"]({ err: error, res: error?.response }, errorMessage);
```
thombergs, line 66: I would find a simple `if/else` much easier on the eyes...

thombergs (approved):

src/config/logger.ts
```diff
@@ -17,34 +17,47 @@ const outputMode = process.env.MICROS_ENV ? "json" : "short";
...
 });
 
-const requestSerializer = (req: Request) => (!req || !req.socket) ? req : {
+const requestSerializer = (req) => req && ({
 	method: req.method,
 	url: req.originalUrl || req.url,
 	path: req.path,
 	headers: req.headers,
 	remoteAddress: req.socket.remoteAddress,
 	remotePort: req.socket.remotePort,
```
thombergs, line 37: won't this run into an `undefined` error when `req.socket` is not there? That was specifically handled in the previous version of the code

src/config/logger.ts
mboudreau: Because not all responses are `AxiosResponse` and didn't want to add that assumption in case someone adds something from AxiosResponse in the code because the type said it was fine.

src/config/logger.ts
mboudreau: yep, good catch.

src/jira/client/axios.ts
mboudreau: done

thombergs (approved):

PR #1179, opened 2022-05-04
User Config

mboudreau:
First pass at reimagining configuring github for jira at the global, subscription and repo level.

PR #1433, opened 2022-07-28
User config (config-as-code)

thombergs:
**What's in this PR?**
Replaces https://github.com/atlassian/github-for-jira/pull/1179 as that PR had a rebase issue. This PR adds the functionality to add a file `.jira/config.yml` to a repository that contains user-defined deployment mappings so that when a deployment with the environment name `Entwicklung` comes in, for example, they can add a mapping so that it shows up as `development` in Jira. 

An example `.jira/config.yml` file that does the above mapping looks like this:
```yaml
deployments:
  environmentMapping:
    development:
      - "Entwicklung"
```

Read the [updated docs](https://github.com/atlassian/github-for-jira/blob/user-config-2/docs/deployments.md#environment-mapping) for more context.

What this PR adds is:

- On a push event, we now check if the config file has been touched. If yes, we fetch it, parse it, and store the config in the database against the repository.
- On a deployment event, we now check if there is a config stored against the repo in the database. If yes, we use that config to do the mapping 

What's missing:

- On initial connection of a GitHub org, we need to look into each repository to check if it has a config file. If yes, save it in the database.

Will add that in a follow-up PR.

**Why**
Because many users have asked for custom environment mappings for deployments and a config-as-code mechanism is a simple way of implementing this.

**Added feature flags**
`CONFIG_AS_CODE`

**How has this been tested?**  
Not yet. WIP.

**Whats Next?**
We can build other cool features on top of the config as code mechanism! :) JSM is waiting to use this mechanism to add another user configuration.


src/github/push.ts
```diff
@@ -28,6 +34,18 @@ export const pushWebhookHandler = async (context: WebhookContext, jiraClient, _u
 		gitHubInstallationId
 	});
 
+	if (await booleanFlag(BooleanFlags.CONFIG_AS_CODE, false, jiraClient.baseURL)) {
+		if (subscription) {
+			const repoSyncState = await RepoSyncState.findByRepoId(subscription, payload.repository.id);
+			const modifiedFiles = context.payload?.commits?.reduce((acc, commit) =>
+				([...acc, ...commit.added, ...commit.modified, ...commit.removed]), []);
+			// TODO: this call must be updated to support GitHub Server events
+			await updateRepoConfig(repoSyncState, getInstallationId(gitHubInstallationId), modifiedFiles);
```
thombergs, line 43: Flagging that this call must be updated to support GitHub Server. How do we get the app ID and GitHub base url in the webhook path @joshkay10 @gxueatlassian?

docs/deployments.md
```diff
@@ -17,3 +19,46 @@ If we use the Jira issue keys (`JIRA-*`) in the commit messages as shown in the
 To create deployments as a part of a GitHub Actions workflow, you can use [this deployment action](https://github.com/chrnorm/deployment-action), for example. Use the action `chrnorm/deployment-action@releases/v1` to create a deployment and don't forget to use the `chrnorm/deployment-status@releases/v1` to update the state of a deployment.
 
 **The GitHub for Jira app only listens to `deployment_status` events**. Just creating a deployment via the [create deployment API](https://docs.github.com/en/rest/reference/repos#create-a-deployment) (or the `chrnorm/deployment-action@releases/v1` action) isn't enough. After creating a deployment you need to call the [create deployment status API](https://docs.github.com/en/rest/reference/repos#create-a-deployment-status) (or the `chrnorm/deployment-status@releases/v1` action) at least once to provide a status for that deployment (success, failure, pending, ...).
+
+
+## Environment mapping
```
thombergs, line 24: Read this before reviewing the rest, as it hopefully clarifies the intent of this feature. 

Link for better reading: https://github.com/atlassian/github-for-jira/blob/user-config-2/docs/deployments.md#environment-mapping

src/github/push.ts
```diff
@@ -28,6 +34,18 @@ export const pushWebhookHandler = async (context: WebhookContext, jiraClient, _u
 		gitHubInstallationId
 	});
 
+	if (await booleanFlag(BooleanFlags.CONFIG_AS_CODE, false, jiraClient.baseURL)) {
```
thombergs, line 37: On a `push` webhook, we're now sending all modified files to the `updateRepoConfig()` function, which determines whether the config file has been changed and then reads the file and stores the config in the database.

src/middleware/github-webhook-middleware.ts
```diff
@@ -71,7 +71,7 @@ function extractWebhookEventNameFromContext(context: WebhookContext): string {
 
 // TODO: fix typings
 export const GithubWebhookMiddleware = (
-	callback: (webhookContext: WebhookContext, jiraClient: any, util: any, githubInstallationId: number) => Promise<void>
+	callback: (webhookContext: WebhookContext, jiraClient: any, util: any, githubInstallationId: number, subscription: Subscription) => Promise<void>
```
thombergs, line 74: Passing in the subscription, because we need it in the push webhook handler to find the `RepoSyncState` for the given subscription.

src/github/push.ts
gxueatlassian: there should be `/git/:uuid/webhook/blah` something like that. And in the `GithubServerAppMiddleware`, it will extract the uuid and put it in side `locals` for use. Me or @Harminder84 is working on it atm.

src/models/reposyncstate.ts
```diff
@@ -30,6 +31,7 @@ export class RepoSyncState extends Model {
 	repoCreatedAt?: Date;
 	syncUpdatedAt?: Date;
 	syncCompletedAt?: Date;
+	config: Config;
```
thombergs, line 34: Later, we might want to add a `config` field to the `Subscription` and `Installation` types as well to act as a fallback when there is no repo-scoped config.

db/migrations/20220803110900-add-config-columns.js
```diff
@@ -0,0 +1,19 @@
+"use strict";
+
+const tableNames = ["RepoSyncStates"];
+module.exports = {
+	up: async (queryInterface, Sequelize) => {
+		await Promise.all([
```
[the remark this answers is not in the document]
thombergs: Yeah, you're right, this could be simplified. This code created the `config` field to 3 different tables before, so I just removed the 2 we don't need for now. I don't really want to touch the database migration code again because it's not covered with tests and bad things happen if I screw it up.

src/github/client/github-client.types.ts
```diff
@@ -36,5 +36,18 @@ export type GraphQlQueryResponse<ResponseData> = {
 };
 
 export type PaginatedAxiosResponse<T> = { hasNextPage: boolean; } & AxiosResponse<T>;
+export interface ReposGetContentsResponse {
```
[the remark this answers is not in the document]
thombergs: Yeah, good call. I don't think there's an absolute answer to that question. 

We currently only need the `content` field. But a developer working on that code might want to know what other fields are available and this way they can see those fields without searching the GitHub API. Typical trade off. I'm not going to die on that hill, but I'd leave the fields in there unless someone really doesn't want me to.

src/github/client/github-installation-client.ts
```diff
@@ -271,6 +277,28 @@ export class GitHubInstallationClient extends GitHubClient {
...
+			// can't pass the path as a path param, because "/"s would be url encoded
+			const response = await this.get<ReposGetContentsResponse>(`/repos/{owner}/{repo}/contents/${path}`, {}, {
+				owner,
+				repo
+			});
+
+			return response.data.content;
+		} catch (err) {
+			if (err.status == 404) {
+				this.logger.info({ err, owner, repo, path }, "could not find file in repo");
```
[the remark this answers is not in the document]
thombergs: Yup, I think it's a good idea to make this more visible by raising the log level to warn. Done.

src/services/user-config-service.ts
```diff
@@ -0,0 +1,120 @@
...
+ * if the config file has changed.
+ *
+ * @param subscription the subscription to which the repository belongs.
+ * @param repositoryId the ID of the repository.
+ * @param githubInstallationId the ID of the installation to which the repository belongs.
+ * @param modifiedFiles list of modified files (added, modified, or removed). The config will only be updated if this list contains
+ * the config file.
+ */
+export const updateRepoConfig = async (
+	subscription: Subscription,
```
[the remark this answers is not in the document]
thombergs: Yeah, technically, we would only need to pass around the `subscriptionId` instead of the whole `subscription`. But the database APIs all take the subscription object as an input and it's a very central object that is pretty much available everywhere, anyways. 

Also, this way it's typed instead of just passing around a generic integer `subscriptionId` :).

src/services/user-config-service.ts
```diff
@@ -0,0 +1,120 @@
...
+const convertYamlToUserConfig = (input?: string): Config => {
+
+	if (!input) {
+		return {};
+	}
+
+	const config: Config = YAML.parse(input);
+
+	if (!config?.deployments?.environmentMapping) {
+		throw new Error(`Invalid .jira/config.yml structure`);
```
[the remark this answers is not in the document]
thombergs: When the shape of the supported config file changes, this validation code has to change so that we're not throwing an error in this case anymore (and so that the rest of the code can handle am empty `environmentMapping`)

db/migrations/20220803110900-add-config-columns.js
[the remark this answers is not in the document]
mboudreau: why remove the other 2 tables?

src/github/client/github-client.types.ts
```diff
@@ -1,6 +1,6 @@
 import { AxiosResponse } from "axios";
-import { Octokit } from "probot";
 import { GraphQLError } from "./github-client-errors";
+import { Octokit } from "probot";
```
mboudreau, line 3: We're trying to remove this. Maybe just copy the pertinent interface over instead?

src/github/push.ts
Harminder84: Won't it make more readable if we add check here to determines whether config file has been changed or not before calling `updateRepoConfig`?

db/migrations/20220803110900-add-config-columns.js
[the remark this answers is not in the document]
thombergs: Because we're not using them, yet. We'd need to build a whole mechanism around when to store config on the level of an installation or subscription. How should users contribute an installation-level or subscription-level config? That needs more thinking.

src/github/client/github-client.types.ts
thombergs: I didn't add this (my formatter just changed the order of the imports)

src/github/push.ts
thombergs: No, because this would mean that the push webhook processor would have to need to know what file to look for. That is information that only the user-config-service should have.

src/middleware/github-webhook-middleware.ts
Harminder84: Do we really need to pass `Subscription`? It can be fetched using `githubInstallationId` and `githubAppId` in push webhook handler if config file has been changed.

src/github/push.ts
Harminder84: The`user-config-service` can provide that information instead of having it in push webhook processor.

thombergs: Yeah, but that would mean we would have more code in the push webhook and less code in the service. I would rather have as little code as possible in the webhook processor because it should stay rather dumb.

That said, it's really a small issue and I wouldn't mind too much doing it the way you proposed if the code wasn't there already :)

src/middleware/github-webhook-middleware.ts
thombergs: Why fetch it again if it's already available here? This way saves one database call.

Harminder84: Yep, it is saving one DB call but only in one case when config file is created/updated which is going to be very rare.  The only point I'm trying to emphasise is why to have pass another argument (already having 4 arguments) that makes code more complex without having any significant advantage.
This is small issue, you can ignore it.

Harminder84 (dismissed):

src/services/user-config-service.ts
```diff
@@ -0,0 +1,120 @@
...
+	}
+};
+
+/**
+ * Returns the config for a given repo.
+ */
+export const getRepoConfig = async (subscription: Subscription, repositoryId: number): Promise<Config | undefined> => {
+	// In the future, we may look in other places for a config than just in the RepoSyncState (for example,
+	// we might fall back to default configs on the level of a subscription or an installation).
+	const repoSyncState = await RepoSyncState.findByRepoId(subscription, repositoryId);
```
mboudreau, line 53: this might be null

src/models/reposyncstate.ts
```diff
@@ -80,6 +82,14 @@ export class RepoSyncState extends Model {
 		}));
 	}
 
+	static async findByRepoId(subscription: Subscription, repoId: number): Promise<RepoSyncState> {
```
mboudreau, line 85: return type should be `| null`

src/services/user-config-service.ts
```diff
@@ -0,0 +1,120 @@
...
+ * Iterates through environment patterns and returns true if any environment contains too many patterns to test against.
+ */
+const hasTooManyPatternsPerEnvironment = (config: Config): boolean => {
+	const environmentMapping = config?.deployments?.environmentMapping;
+	if (!environmentMapping) {
+		return false;
+	}
+	return Object.keys(environmentMapping).some(key => {
+		return environmentMapping[key]?.length > MAX_PATTERNS_PER_ENVIRONMENT;
+	});
```
mboudreau, line 80: What's this for?

src/services/user-config-service.ts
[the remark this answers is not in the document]
mboudreau: we could use a yaml schema validator then...

src/transforms/transform-deployment.ts
```diff
@@ -101,11 +106,28 @@ const mapState = (state: string | undefined): string => {
...
+	return false;
+};
+
+/**
+ * Maps a given environment name to a Jira environment name using the custom mapping defined in a RepoConfig.
+ */
+export const mapEnvironmentWithConfig = (environment: string, config: Config): string | undefined => {
+	return _.keys(config?.deployments?.environmentMapping)
+		.find(jiraEnvironmentType => matchesEnvironment(environment, config.deployments?.environmentMapping?.[jiraEnvironmentType]));
+};
```
mboudreau, line 124: would it be possible to match more than one environment?

src/transforms/transform-deployment.ts
```diff
@@ -120,6 +142,17 @@ export const mapEnvironment = (environment: string): string => {
...
 	};
 
+	// if there is a user-defined config, we use that config for the mapping
+	if (config) {
+		const environmentType = mapEnvironmentWithConfig(environment, config);
+		if (environmentType){
+			return environmentType;
+		}
+	}
+
```
mboudreau, line 152: shouldn't the user config be merge with the default?  

By doing it this way, it means that simply having a config yaml file with a single environment in `environmentMapping` will remove all defaults mappings for other enviroment.  ie. if I add `staging` as an environment only in the yaml, nothing will ever be mapped to production, dev, or testing.  This seems counter intuitive to me and will result in data loss for some customers

mboudreau (changes requested):
I think there's a few issues that needs to be addressed from a customer experience point of view before we release this

src/transforms/transform-deployment.ts
gxueatlassian: Good call, and I want to clarify as well, when the repo has multi branch, which branch "wins" in terms of this config?

src/services/user-config-service.ts
thombergs: just a check to limit user input to not allow any nonsense

src/transforms/transform-deployment.ts
thombergs: No, the first match wins. If people have overlapping patterns, they have a different problem :)

src/transforms/transform-deployment.ts
thombergs: @mboudreau The merge you had previously built failed because the hardcoded fallback is matching regexes whereas the user config matches glob patterns (because we don't want to have user-defined regexes that could be evil). We can't treat them the same, so I split it up.

But the case you mention is still covered by this logic. If the manual config doesn't provide a match, we look in the hardcoded config. I added a test case to showcase this (see test `falls back to hardcoded config when user config doesn't match`).

src/services/user-config-service.ts
```diff
@@ -0,0 +1,129 @@
...
+	repositoryId: number,
+	githubInstallationId: InstallationId,
+	modifiedFiles: string[] = []): Promise<void> => {
+
+	if (modifiedFiles.includes(USER_CONFIG_FILE)) {
+		try {
+			const repoSyncState = await RepoSyncState.findByRepoId(subscription, repositoryId);
+
+			if (!repoSyncState){
+				logger.error({
```
thombergs, line 38:
```suggestion
				logger.warn({
```

src/services/user-config-service.ts
```diff
@@ -0,0 +1,129 @@
...
+	}
+};
+
+/**
+ * Returns the config for a given repo.
+ */
+export const getRepoConfig = async (subscription: Subscription, repositoryId: number): Promise<Config | undefined> => {
+	// In the future, we may look in other places for a config than just in the RepoSyncState (for example,
+	// we might fall back to default configs on the level of a subscription or an installation).
+	const repoSyncState = await RepoSyncState.findByRepoId(subscription, repositoryId);
```
thombergs, line 62: If repoSyncState is null, ask GitHub for the config file.

docs/deployments.md
```diff
@@ -17,3 +19,46 @@ If we use the Jira issue keys (`JIRA-*`) in the commit messages as shown in the
...
+Each deployment event contains the name of the environment your app was deployed to. Jira needs to know whether the environment is a `development`, `testing`, `staging`, or `production` environment to give you the most value from the issue and deployment views.
+
+So, the GitHub integration tries to guess the environment type from your environment name. If it contains the word `prod`, for example, it will be considered a `production` environment. If it contains the word `test`, it will be considered a `testing` environment, and so on.
+
+If you are using custom environment names in your deployments (for example non-english environment names), we might not be able to resolve the correct environment type. You may see the message "undefined environment" in Jira:
+
+![Undefined environment in the Jira issue view](undefined-environment.png)
+
+If this is the case, you can add a file called `.jira/config.yml` to you repository with the following content:
+
```
thombergs, line 35: Add that the file must be in the main branch!

joshkay10 (dismissed):

gxueatlassian (dismissed):

mboudreau (approved):

gxueatlassian (approved):

joshkay10 (approved):

PR #1434, opened 2022-07-28
Fixing logger circular dependencies

mboudreau:
**What's in this PR?**
Adding `safe-json-stringify` to remove possible circular dependencies in log data (common with req/res)

**Why**
It would sometimes cause issues in logs and throw an error

**Added feature flags**
None

**Affected issues**  
None

**How has this been tested?**  
Logs still work, added test for circular dependency logging

### DIFF
--- a/src/config/logger.ts
+++ b/src/config/logger.ts
@@ -57,7 +57,7 @@ const errorSerializer = (err) => {
 };
 
 const hashSerializer = (data?: string): string | undefined => {
-	if (!data || isString(data)) {
+	if (!data || !isString(data)) {
 		return undefined;
 	}
 	return createHashWithSharedSecret(data);
--- a/src/config/logger.ts
+++ b/src/config/logger.ts
@@ -1,13 +1,13 @@
-import Logger, { createLogger, INFO, levelFromName, stdSerializers } from "bunyan";
+import Logger, { createLogger, INFO, levelFromName, LoggerOptions, stdSerializers } from "bunyan";
 import bformat from "bunyan-format";
 import { filteringHttpLogsStream } from "utils/filtering-http-logs-stream";
-import { Request } from "express";
-import { AxiosResponse } from "axios";
 import { createHashWithSharedSecret } from "utils/encryption";
+import { isArray, isString } from "lodash";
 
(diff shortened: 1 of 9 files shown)


src/config/logger.ts
```diff
@@ -117,12 +117,7 @@ export const getUnsafeLogger = (name: string, fields?: Record<string, unknown>):
 	return logger.child({ ...fields });
 };
 
-
-export const cloneAllowedLogFields = (fields: Record<string, unknown>) => {
-	const allowedFields = { ...fields };
-	delete allowedFields.name;
-	return allowedFields;
-};
+export const cloneAllowedLogFields = (fields: Record<string, unknown>) => omit(fields, ["name"]);
```
joshkay10, line 120: havn't used `omit` before, looks rather handy!

joshkay10 (approved):

src/config/logger.ts
```diff
@@ -91,23 +82,32 @@ export const overrideProbotLoggingMethods = (probotLogger: Logger) => {
...
-			...options
-		});
+	return createLogger({
+		name,
+		stream: LOG_STREAM,
+		level: globalLoggingLevel,
+		serializers: {
+			err: errorSerializer,
+			error: errorSerializer,
+			res: responseSerializer,
```
gxueatlassian, line 92: I assume this is the place that fix the `circular` json stuff, as mostly `res`, `req` has circular reference inside and this `serializer` will use that `safe-json-stringify`

-----

But just to double check, this fix wont' fix the error if someone log a circular json with some other names right?, like bellow, or am I missing something?

```
const data = {
				a: { foo: true },
				b: { bar: "string" }
			};
data.a = data as any;
req.log('whatever', { data } );
```

gxueatlassian (approved):

src/config/logger.ts
mboudreau: sure is! lodash is full of goodies like that!

src/config/logger.ts
mboudreau: yes, it will.  bunyan uses it as a fallback if it's own circular dependency check fails to catch it (which I've seen in a few occasions).  I have a test that does exactly this.

PR #1437, opened 2022-07-29
ARC-1514 backfill from date UI

joshkay10:
**Can't merge until Pollinator is updated**

**What's in this PR?**
Restart backfill option now opens a backfill modal where user can select the date to backfill from.

max date is today,
default date is 1 year ago.

**Why**
Since we now limit how far we backfill commits via feature flags, should the user wish to override that value they can

**Added feature flags**
Nope

**Affected issues**  
[ARC-1514](https://softwareteams.atlassian.net/browse/ARC-1514)

**How has this been tested?**  
tested locally and on staging

**Whats Next?**
Pollinator needs to be updated to click the new backfill button

Screen Shot 2022-07-29 at 3 55 02 PM


src/routes/jira/jira-sync-post.ts
```diff
@@ -19,7 +22,13 @@ export const JiraSyncPost = async (req: Request, res: Response, next: NextFuncti
 			res.status(404).send("Subscription not found, cannot resync.");
 			return;
 		}
-		await findOrStartSync(subscription, req.log, syncType);
+
+		if (commitsFromDate && commitsFromDate.valueOf() > Date.now()){
```
gxueatlassian, line 26: Since we are opening this to UI, do you think we shall check the date is not too loooong ago? Like checking the date is within a year for example?

static/js/jira-configuration.js
```diff
@@ -68,38 +68,52 @@ $(".delete-connection-link").click(function(event) {
...
 });
 
-$(".sync-connection-link").click(function(event) {
-	event.preventDefault();
+const initializeBackfillDateInput = function () {
+	const dateElement = document.getElementById('backfill-date-picker');
+	const date = new Date();
+	date.setFullYear(date.getFullYear() - 1);
+	dateElement.value = date.toISOString().split('T')[0];
+	dateElement.max = new Date().toISOString().split('T')[0];
```
gxueatlassian, line 76: Should we set a `dateElement.min` for UX purpose?

src/routes/jira/jira-sync-post.ts
gxueatlassian: Or is it already check somewhere else?

mboudreau: I don't think it's a concern just yet as we are doing all of it right now.

src/routes/jira/jira-sync-post.ts
```diff
@@ -19,7 +22,13 @@ export const JiraSyncPost = async (req: Request, res: Response, next: NextFuncti
...
+			res.status(400).send("Invalid date, please select historical date!");
```
mboudreau, line 27: "historical date" doesn't sound right.  Maybe just say "cannot choose a date in the future"?

static/js/jira-configuration.js
mboudreau: not necessary

mboudreau (approved):
good enough, but could address some small nits

src/routes/jira/jira-sync-post.ts
joshkay10: Exactly, right. Its the correct question to ask, but they can put 10000 BC if they want, this give them the option to have as much or as little backfill as they choose(for commits).

src/routes/jira/jira-sync-post.ts
joshkay10: Haha yea reading that now looks odd, 'date does not have enough dinosaurs alive" kinda vibe

krazziekay (approved):

PR #1439, opened 2022-08-02
ARC-1571 add cryptor dependencies in healthcheck

gxueatlassian:
**What's in this PR?**
Add calling to cryptor in healthcheck.

**Why**
Make sure the server only start after cryptor is ready since cryptor is a hard dependency.

**Added feature flags**
N/A

**Affected issues**  
[_Jira Issues_](https://softwareteams.atlassian.net/browse/ARC-1571)

**How has this been tested?**  
Local

**Whats Next?**
N/A

### DIFF
--- a/src/routes/healthcheck/healthcheck-get.ts
+++ b/src/routes/healthcheck/healthcheck-get.ts
@@ -1,8 +1,20 @@
 import { Request, Response } from "express";
 import { getLogger } from "config/logger";
+import { EncryptionClient } from "utils/encryption-client";
 
 const logger = getLogger("healthcheck");
-export const HealthcheckGet = (_: Request, res: Response): void => {
+export const HealthcheckGet = async (_: Request, res: Response): Promise<void> => {
+
+	try {
+		await EncryptionClient.healthcheck();
+		logger.debug("Cryptor ready in healthcheck");
+	} catch (error) {
+		logger.warn({ reason: error }, "Cryptor not ready in healthcheck");
+		res.status(500).send("NOT OK");
+		return;
+	}
+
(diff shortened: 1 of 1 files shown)


krazziekay (approved):

src/routes/healthcheck/healthcheck-get.ts
```diff
@@ -1,8 +1,20 @@
...
+export const HealthcheckGet = async (_: Request, res: Response): Promise<void> => {
+
+	try {
+		await EncryptionClient.healthcheck();
+		logger.debug("Cryptor ready in healthcheck");
+	} catch (error) {
+		logger.warn({ reason: error }, "Cryptor not ready in healthcheck");
+		res.status(500).send("NOT OK");
+		return;
+	}
```
mboudreau, line 15: pretty sure this is suppose to be in the deepcheck, not the healthcheck

mboudreau (changes requested):
move to deepcheck

src/routes/healthcheck/healthcheck-get.ts
```diff
@@ -1,8 +1,20 @@
 import { Request, Response } from "express";
 import { getLogger } from "config/logger";
+import { EncryptionClient } from "utils/encryption-client";
 
 const logger = getLogger("healthcheck");
-export const HealthcheckGet = (_: Request, res: Response): void => {
+export const HealthcheckGet = async (_: Request, res: Response): Promise<void> => {
```
joshkay10, line 6: We have `/deepcheck` endpoint as well I wonder if that might be a better place for it.
Only because `/healthcheck` has always been super simple, just returning a 200?

PR #1427, opened 2022-07-28
ARC-1519 - Improve Jira Routing for connect/app creation

krazziekay:
**What's in this PR?**
- The combination of the following PRs:
  - https://github.com/atlassian/github-for-jira/pull/1415
  - https://github.com/atlassian/github-for-jira/pull/1417
  - https://github.com/atlassian/github-for-jira/pull/1418
  - https://github.com/atlassian/github-for-jira/pull/1419
  - https://github.com/atlassian/github-for-jira/pull/1420
  - https://github.com/atlassian/github-for-jira/pull/1421
  - https://github.com/atlassian/github-for-jira/pull/1422
  - https://github.com/atlassian/github-for-jira/pull/1423
  - https://github.com/atlassian/github-for-jira/pull/1424
  - https://github.com/atlassian/github-for-jira/pull/1425
  - https://github.com/atlassian/github-for-jira/pull/1449
- Added/Updated routes fixes for 
  - **Connect new server** in _jira-select-server.hbs_
  - **Connect new application** in _jira-select-github-cloud.hbs_
  - Query string corrected in _jira-server-url.hbs_
  - Route added for _Manual app_ in `jira-select-app-creation.hbs`
- Updated the `POCO` `main.json` 

**Why**
- To test all the new routes combined in one branch.

**How has this been tested?**  
- In local environment
- In staging environment - WIP
- Test cases

PR #1442, opened 2022-08-02
Passing the GHE baseUrl for auto app creation

krazziekay:
**What's in this PR?**
- After merging the [Auto App creation PR](https://github.com/atlassian/github-for-jira/pull/1401#event-7106030416) and the [Jira Route Improvement PR](https://github.com/atlassian/github-for-jira/pull/1427/), there were some conflicts and some assumptions that needed to be implemented.
- For `jira-redirect.js` the `gheHost` was assumed to be passed in from the `queryParameter`. This was updated by passing the `baseUrl` from the backend `ts` file.

**Why**
- The assumption in the mentioned tickets were not properly connected. 
- `gheHost` url was not being passed and was throwing errors.

**How has this been tested?**  
- In local environment
- In staging environment


views/jira-redirect.hbs
```diff
@@ -10,6 +10,8 @@
 
 <body>
    Redirecting to your GitHub Enterprise Server Instance
+
+   <input type="hidden" id="baseUrl" value="{{baseUrl}}">
```
Harminder84, line 14: Instead of passing via hidden textfield, pass it via query paramter.

src/routes/session/session-get.ts
```diff
@@ -5,10 +5,15 @@ export const SessionGet = (req: Request, res: Response, next: NextFunction) => {
 		return next(new Error("Missing redirect url for session.  Needs to be in format `/session/:redirectUrl`"));
 	}
 
+	let url = req.params[0];
+	if (req.query.baseUrl) {
+		url += encodeURIComponent("?baseUrl=" + req.query.baseUrl);
+	}
+
```
Harminder84, line 12: This is being used at mutiple places, have a generic logic to pass on query parameters

views/jira-redirect.hbs
krazziekay: This `baseUrl` is passed from `GithubRedirectGet` as an `option`  inside `res.render`. How can we pass this value from the `backend` to the frontend `URL`?

src/routes/session/session-get.ts
krazziekay: The query parameter is being passed in multiple places, not sure how to make this part generic. Can you please ellaborate a lil bit?

views/jira-redirect.hbs
Harminder84: Why to pass it from backend in this case? Grab it directly from the URL on the frontend.

src/routes/session/session-get.ts
Harminder84: Here we're relying on `baseUrl` is being passed as query parameter. Have to write a similar kind of code for another query parametre if added in future. So my observation is to write generic code that pass on all query parameters to the targeted URL. Hope that makes sense.

views/jira-redirect.hbs
krazziekay: It can't be grabbed directly from the frontend because the `queryParameter` is not present when getting to this view. The user gets to this view after getting redirected, during which we don't have the `baseUrl`.
 
From the frontend, the user is firstly redirected to `"/session/github/redirect?baseUrl=" + gitHubServerBaseUrl`, which takes the user to `GithubRedirectGet`. From there, we now pass this `queryParameter` to the next view `jira-redirect.hbs`. Without passing it like this, there is no medium to pass or get the `baseUrl` into the required view.

Harminder84: User is redirected from `/session/github/redirect?baseUrl=<host>` to `/github/redirect?baseUrl=<host>`. The baseUrl is present on both views.

In `jira-redirect.js`, you can get baseURL using -
```
const params = new URLSearchParams(window.location.search.substring(1));
const gheHost = params.get("baseUrl");
```

src/routes/session/session-get.ts
krazziekay: Created a generic method `createUrlWithQueryString` with test cases.

views/jira-redirect.hbs
krazziekay: Updated

src/routes/session/session-get.ts
Harminder84: You should not pass the query parameter name. It should read all the query parameters available and passed on to the targeted URL.

krazziekay: Query parameter name removed. Updated the logic to read all the query parameters

src/util/create-url-with-query-string.test.ts
```diff
@@ -0,0 +1,19 @@
...
+	} as any;
+	const URL = "http://myrandomSite.com";
+
+	it(`Should return URL with all the query parameters`, async () => {
+		const resultUrl = createUrlWithQueryString(request, URL);
+
+		expect(decodeURIComponent(resultUrl)).toBe(`${URL}?son=goku&uzumaki=naruto&ketchum=ash`);
+	});
+});
+
```
Harminder84, line 19: Please add a test case for no query paramater.

views/session.hbs
```diff
@@ -54,7 +54,7 @@
       document.cookie = "jiraHost=" + window.jiraHost + "; path=/; max-age=" + 60 * 5 + "; samesite=none; secure";
       document.cookie = "jwt=" + window.jwt + "; path=/; max-age=" + 60 * 5 + "; samesite=none; secure";
     }
-    window.location = "{{redirectUrl}}";
+    window.location = decodeURIComponent("{{redirectUrl}}");
```
Harminder84, line 57: Can you please explain why we first encoding the URL then decoding it before using it?

src/util/create-url-with-query-string.test.ts
krazziekay: Test case added

src/util/create-url-with-query-string.ts
```diff
@@ -0,0 +1,24 @@
+import { Request } from "express";
+
+/**
+ * This method creates a URL with queryParameters that are available within the request
+ * It firstly checks if the passed Request `req` has the query parameters defined in the array `keys`
+ * And will add the query parameters as queryString within the URL
```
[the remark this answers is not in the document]
krazziekay: Comment updated

views/session.hbs
Harminder84: You reading from the URL, it is already encoded. Here you're encoding in `session-get.ts` and decoding in `session.hbs` before navigating to the URL in browser.  This cycle of encoding and decoding seems to be redundant. Please verify it.

krazziekay: The `=` gets converted, can see in the video below.

https://user-images.githubusercontent.com/13076255/182507147-685ec015-fb0e-4b7f-a56e-88071df0d47b.mov

krazziekay: Ignore, found a better solution for this

Harminder84: Tripe curly braces for HTML escaping. This is not meant for URL encoding.

krazziekay: The aim is to correctly display the special characters(`=`), and for that [triple braces](https://handlebarsjs.com/guide/expressions.html#html-escaping) would be valid, isn't it?
What are you suggesting?

src/util/create-url-with-query-string.test.ts
```diff
@@ -0,0 +1,34 @@
...
+		expect(decodeURIComponent(resultUrl)).toBe(`${URL}`);
+	});
+
+	it(`Should return just URL for null query`, async () => {
+		const request = {} as any;
+		const URL = "http://myrandomSite.com";
+		const resultUrl = createUrlWithQueryString(request, URL);
+
+		expect(decodeURIComponent(resultUrl)).toBe(`${URL}`);
+	});
```
Harminder84, line 32: `decodeURIComponent` not required now.

views/session.hbs
Harminder84: I mixed `{{` with `{{{`, it is fine.

src/util/create-url-with-query-string.test.ts
krazziekay: removed

Harminder84 (dismissed):
Looks Good 👍. Please test Github cloud flow before merging the PR.

Harminder84 (approved):

krazziekay:
> Looks Good 👍. Please test Github cloud flow before merging the PR.

Tested in staging.

https://user-images.githubusercontent.com/13076255/182524137-0a45af02-aae3-4ed9-aa0b-2f35aae30919.mov

PR #1443, opened 2022-08-02
Update alarm threshold to match autoscaling

joshkay10:
**What's in this PR?**
Increase alarm threshold for `AlarmOnTooManyMessages` to match autoscaling target * 2

**Why**
We changed autoscaling rules but not alarms to match, therefore we need to give a bit more opportunity for autoscaling to handle the messages before we panic and raise an alarm.

### DIFF
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -601,7 +601,7 @@ environmentOverrides:
             MetricName: ApproximateNumberOfMessagesVisible
             Namespace: AWS/SQS
             Description: "Message count for push queue too high! Follow the runbook: https://hello.atlassian.net/wiki/spaces/PF/pages/1550488861/HOWTO+Respond+to+Too+Many+Messages+on+the+Queue+Alert"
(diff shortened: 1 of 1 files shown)


github-for-jira.sd.yml
```diff
@@ -601,7 +601,7 @@ environmentOverrides:
             MetricName: ApproximateNumberOfMessagesVisible
             Namespace: AWS/SQS
             Description: "Message count for push queue too high! Follow the runbook: https://hello.atlassian.net/wiki/spaces/PF/pages/1550488861/HOWTO+Respond+to+Too+Many+Messages+on+the+Queue+Alert"
```
joshkay10, line 603: on line 249 is the scaling rules expression. To save time and lives here it is

            `- Expression: "MAX([cpuUsage, memoryUsage, (pushQueueMessages/250)*100, 
            (backfillQueueMessages/20)*100, (branchQueueMessages/250)*100, 
            (deploymentQueueMessages/250)*100])"`

mboudreau (approved):

PR #1446, opened 2022-08-02
Arc 1558 push app id to queue msg

gxueatlassian:
**What's in this PR?**
Add gitHubAppId to message queue payload for further processing for GHE.

**Why**
InstallationId is not unique, need to add gitHubAppId for GHE support.

**Added feature flags**
N/A

**Affected issues**  
ARC-1558

**How has this been tested?**  
N/A

**Whats Next?**
continue ARC-1558 for pushing real value into message body.


src/sqs/branch.ts
```diff
@@ -2,8 +2,9 @@ import { WebhookPayloadCreate } from "@octokit/webhooks";
 import { Context, MessageHandler } from "./sqs";
 import { processBranch } from "../github/branch";
 import { createInstallationClient } from "~/src/util/get-github-client-config";
+import { GitHubAppConfigPayload } from "./github-app-config-payload";
 
-export type BranchMessagePayload = {
+export type BranchMessagePayload = GitHubAppConfigPayload & {
```
mboudreau, line 7: Wouldn't it be simpler to just do 
```
{
   githubAppConfig: GitHubAppConfig;
   ...
}
```

gxueatlassian: Will it be risking some payload use `githubAppConfig` while some use `gitHubAppConfig` as the field name?

src/sqs/github-app-config-payload.ts
```diff
@@ -0,0 +1,19 @@
+/*
+ * Types for both Cloud and Enterprise server product
+ */
+export type GitHubAppConfigPayload = {
+	/*
+	 * undefined for cloud, real value for GHE
+	 * Otherwise, lost of duplicates values in msg body for cloud (majority traffic)
+	 */
+	gitHubAppConfig?: {
```
mboudreau, line 9: don't wrap the config into another type.  We also shouldn't create a new TS file just for a single type.  If you want to create a file for multiple types, we just the standard of using the extension `.types.ts` and adding all types related to the file name in here.

It might make sense to add all payload types in here as well.

mboudreau (changes requested):
Could do better typings

src/sqs/backfill.ts
```diff
@@ -5,8 +5,9 @@ import * as Sentry from "@sentry/node";
 import { AxiosErrorEventDecorator } from "models/axios-error-event-decorator";
 import { SentryScopeProxy } from "models/sentry-scope-proxy";
 import { TaskType } from "~/src/sync/sync.types";
+import { GitHubAppConfigPayload } from "./github-app-config-payload";
 
-export type BackfillMessagePayload = {
+export type BackfillMessagePayload = GitHubAppConfigPayload & {
```
thombergs, line 10: This syntax is new to me. Is this adding the attributes below to the type `GitHubAppConfigPayload`?

src/sqs/github-app-config-payload.ts
thombergs: I don't mind having this as its own type. If you search for all places where the github config is used in an SQS payload, this type makes it trivial to find those places, for example. I would err on the side of one type too many than one type too little.

src/sqs/github-app-config-payload.ts
```diff
@@ -0,0 +1,19 @@
...
+		appId: number,
+		clientId: string,
+		gitHubBaseUrl: string,
+		uuid: string,
+		gitHubAppId: string,
+		gitHubClientSecret: string,
+		webhookSecret: string,
+		privateKey: string
```
thombergs, line 17: Does this mean we would add the `gitHubClientSecret`, `webhookSecret`, and `privateKey` to every backfill and push SQS message? Why do we need those? Can't we just add the subscription ID to the message and then load the secrets when (and if) we need them in the message workers?

src/sqs/branch.ts
mboudreau: if you're concerned about that, you can create a `MessagePayload` parent interface, but we do need the type of the config itself as it will be used in other places.

src/sqs/backfill.ts
mboudreau: it's the same as `interface BackfillMessagePayload extends GitHubAppConfigPayload { ... }`, but done with the `type` syntax instead of `interface`

src/sqs/github-app-config-payload.ts
mboudreau: I don't think we need the webhook secret for this one specifically as that's specific to webhooks only.

The idea behind having everything in a config is that _everything_ in the code uses the github app config (API, webhooks and sync), so we reduce complexity by removing as much logic as possible from the processors.  This is standard practice for queues message as you want all information to process the job at hand with the least amount of logic.  By adding logic, you're creating another layer of complexity that also makes it much more difficult to update in the future.  

If we are to use this methodology, we can simply pass the config to any client or handlers that use it.

src/sqs/sqs.types.ts
```diff
@@ -0,0 +1,66 @@
...
+	webhookReceived?: number,
+}
+
+type GitHubAppConfig = {
+	appId: number,
+	clientId: string,
+	gitHubBaseUrl: string,
+	uuid: string,
+	gitHubAppId: string,
+	//gitHubClientSecret: string,
```
gxueatlassian, line 54: Hi @mboudreau and @thombergs , I've temporary commented this `secrets` in this PR. Let me think a bit more of it, and raise it in another PR.

----

For now, I need this PR as me and @Harminder84 is working on pushing and processing the queue. Need the interface change.

src/sqs/branch.ts
gxueatlassian: Fixed, extracted in single file.

src/sqs/github-app-config-payload.ts
gxueatlassian: Fixed, no wrapping now.

src/sqs/backfill.ts
gxueatlassian: Closing it, since not using this syntax in the PR now.

mboudreau (approved):

PR #1451, opened 2022-08-03
ARC-1560 Update github config get to pass gitHubAppId

rachellerathbone:
**What's in this PR?**
Small change to pass the gitHubAppId from gh config get to the installation client instance. 

**Why**
We are currently relying on the gitHubInstallationId to set the gh config and we can't do that because it is not a unique field.

**Affected issues**  
ARC-1560

**How has this been tested?**  
Tested that cloud path (with GHE_SERVER flag on and off) works as expected. We'll be able to test the server flow once that is complete.

**Whats Next?**
@Harminder84 and @gxueatlassian are working on updating the remaining installation client instances.


src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -69,11 +69,12 @@ const getInstallationsWithAdmin = async (
 	log: Logger,
 	login: string,
 	installations: Octokit.AppsListInstallationsForAuthenticatedUserResponseInstallationsItem[] = [],
-	jiraHost: string
+	jiraHost: string,
+	gitHubAppId: number
```
gxueatlassian, line 73: Shouldn't it be `gitHubAppId: number | undefined`?

gxueatlassian (approved):

PR #1452, opened 2022-08-03
ARC-1354 remigrate secrets Remove encryption context and apply re-migration of secrets

gxueatlassian:
**What's in this PR?**
1. Remove the clientKey in the encryptionContext
2. Update admin api to support remigration existing secrets

**Why**
1. clientKey is hashed with env.STORAGE_SECRETS, it is a big risk we might modify the value of the secrets, re-hash/update the client key and forgot the re-encrypt the secrets, If that happen, all our customer will lost secrets (from GH4j point of view) of jira. 

**Added feature flags**
Turning FF off so that it won't break production until all secrets are re-migrated without using the encryptionContext.
image

**Affected issues**  
ARC-1354

**How has this been tested?**  
Local
Stage [WIP]

**Whats Next?**
1. Run the API for all records,
2. Turn FF back on.


src/routes/api/cryptor-migrations/migration-installation.ts
```diff
@@ -29,20 +33,23 @@ export const CryptorMigrationInstallationPost = async (req: Request, res: Respon
 
 	//--------- loop and save each one of them --------------
 	let count = 0;
+	let lastInst: Installation | null = null;
```
mboudreau, line 36: I dont' see the point of this, if all you want is the ID.  Why not just store the `lastInstallationId` instead?

mboudreau (approved):

src/routes/api/cryptor-migrations/migration-installation.ts
gxueatlassian: Agree we could store either `lastInst` or `lastId`. But feels like really minor here, so not changing it.

PR #1456, opened 2022-08-04
Add gitHubAppConfig to push payload

gxueatlassian:
**What's in this PR?**
Add gitHubAppConfig to push queue

**Why**
current gitHubInstallationId is not enough.

**Added feature flags**
N/A

**Affected issues**  
ARC-1558

**How has this been tested?**  
Unit test.

**Whats Next?**
Add gitHubAppConfig data to following:
1. push ( -- this PR --)
2. backfill
3. deployment
4. branch

^
|----- it does trigger me to think shall we refactor at some time to make it more generic to add more info the payload? Right now have to add it per event type.


src/transforms/push.ts
```diff
@@ -91,7 +93,7 @@ export const processPush = async (github: GitHubInstallationClient, payload: Pus
 	const log = rootLogger.child({
 		webhookId: webhookId,
 		repoName: repo,
-		orgName: owner.name,
+		orgName: owner["name"], //Less bold, still maintain it, but guess it would be empty
```
rachellerathbone, line 96: What's the reason for the change here? Dot notation is faster to write and clearer to read, and is what we use throughout this codebase (unless we're passing a variable that changes or the property name isn't valid for dot notation).

src/github/push.unit.test.ts
```diff
@@ -0,0 +1,71 @@
+import { pushWebhookHandler } from "./push";
+import { WebhookContext } from "routes/github/webhook/webhook-context";
+import { getLogger } from "config/logger";
+import { GitHubPushData } from "../interfaces/github";
+import { enqueuePush } from "../transforms/push";
+
+jest.mock("../transforms/push");
+
+const GHES__GITHUB_INSTALLATION_ID = 1234;
```
rachellerathbone, line 9: Nit: Is there a reason for the double underscore followed by single underscore?

gxueatlassian: No, no idea why i did that 🤷 
Fixed.

src/interfaces/github.ts
```diff
@@ -56,9 +56,11 @@ interface GitHubWorkflowRunHeadCommit {
 	committer: GitHubWorkflowRunAuthorAndCommitter;
 }
 
+//refer from https://docs.github.com/en/rest/repos/repos#get-a-repository
 interface GitHubWorkflowRunRepositoryOwner {
 	login: string;
 	id: number;
+	name: string | null;
```
gxueatlassian, line 63: It has the schema as bellow from https://docs.github.com/en/rest/repos/repos#get-a-repository
```json
"owner": {
  "title": "Simple User",
  "description": "Simple User",
  "type": "object",
  "properties": {
    "name": {
      "type": [
        "string",
        "null"
      ]
    }
  }
}
```
So type it as `string | null`

src/transforms/push.ts
gxueatlassian: After adding some typings, ts complain `name` is not in the `owner` type. I check the file `src/interfaces/github.ts`, it does not have `name` under `owner`, and I check [github doc](https://docs.github.com/en/rest/repos/repos#get-a-repository), the `name` is not in the `owner` (see screenshot). So I think we have a mistake, and if there's no `name`, typescript did complain here, say you can't do `owner.name`. 

So in order to fix it at compile time, but not risk lose any information at runtime, I change it to `owner["name"]`, instead of simply remove it. 
 
image

------------

Turns out the github doc is wrong, from the "schema" tab, it actually shows that `name` is there under `owner.
```json
"owner": {
      "title": "Simple User",
      "description": "Simple User",
      "type": "object",
      "properties": {
        "name": {
          "type": [
            "string",
            "null"
          ]
        },
```

So 🤷 , I revert things back. Except type it as `string | null` instead of origin in the code `string`

Tbh, I am not 100% sure which part of the github doc is wrong, although believe the "schema" is more correct. At least our code won't break with the typing `string | null`

------------

All reverted now, should be all good.

src/transforms/push.ts
```diff
@@ -64,12 +65,13 @@ export const createJobData = (payload, jiraHost: string): PushQueueMessagePayloa
...
 		webhookId: payload.webhookId || "none",
-		webhookReceived: payload.webhookReceived || undefined
+		webhookReceived: payload.webhookReceived || undefined,
+		...(!gitHubAppConfig ? undefined : { gitHubAppConfig })
 	};
 };
 
-export const enqueuePush = async (payload: unknown, jiraHost: string) =>
-	await sqsQueues.push.sendMessage(createJobData(payload, jiraHost));
+export const enqueuePush = async (payload: GitHubPushData, jiraHost: string, gitHubAppConfig: GitHubAppConfig | undefined) =>
```
joshkay10, line 73: Can we use same signature as above for `gitHubAppConfig?: GitHubAppConfig` instead of `... | undefined`

src/transforms/push.ts
```diff
@@ -64,12 +65,13 @@ export const createJobData = (payload, jiraHost: string): PushQueueMessagePayloa
 		jiraHost,
 		installationId: payload.installation.id,
 		webhookId: payload.webhookId || "none",
-		webhookReceived: payload.webhookReceived || undefined
+		webhookReceived: payload.webhookReceived || undefined,
+		...(!gitHubAppConfig ? undefined : { gitHubAppConfig })
```
joshkay10, line 69: /nit **I would double check** but you should be able to simplify it a bit with
`...(gitHubAppConfig) && { gitHubAppConfig }`

src/transforms/push.ts
```diff
@@ -148,7 +150,7 @@ export const processPush = async (github: GitHubInstallationClient, payload: Pus
 						authorTimestamp: githubCommitAuthor.date,
 						displayId: commitSha.substring(0, 6),
 						fileCount: files.length, // Send the total count for all files
-						files: filesToSend.map((file) => mapFile(file, repo, owner.name, sha.id)),
+						files: filesToSend.map((file) => mapFile(file, repo, owner.name || "", sha.id)),
```
joshkay10, line 153: Since a **name** can be `null` do we want that to be the default?
I'm out of context a bit so non-issue if empty string it better.

src/github/push.ts
```diff
@@ -37,5 +37,5 @@ export const pushWebhookHandler = async (context: WebhookContext, jiraClient, _u
 	}
 
 	context.log.info("Enqueueing push event");
-	await enqueuePush(payload, jiraClient.baseURL);
+	await enqueuePush(payload, jiraClient.baseURL, context.gitHubAppConfig);
```
joshkay10, line 40: Does **gitHubAppConfig** need to be added to the **WebhookContext** type?

src/github/push.unit.test.ts
rachellerathbone: lol I say that to myself sometimes too 😆

src/interfaces/github.ts
mboudreau: We don't use `null` in js/ts if we can help it.  this should be `name?:string`

mboudreau: oh wait, is this a github rest API type?

src/transforms/push.ts
mboudreau: I can do you one better with just adding `githubAppConfig`.  if it's undefined, it'll be `githubAppConfig: undefined`, if not undefined, it'll show the object.  The only reason this might not work correctly is if the config returns null (which it shouldn't) and you could do `githubAppConfig: githubAppConfig || undefined` if you needed to handle that

src/transforms/push.ts
mboudreau: we should never use an empty string where undefined can be used.  Empty string just adds more complexity and a "third state" that we need to handle.

Also, was this a problem before?

mboudreau (changes requested):
few things to address

src/interfaces/github.ts
gxueatlassian: Yup.

src/transforms/push.ts
gxueatlassian: I have no idea "is there any problem at prod" right now. But from github's typing, that `name` could be `null` under the `owner`. 

And I have to do this because TS complaint about it (because I add a typing to `enqueuePush` method)

src/transforms/push.ts
gxueatlassian: Just fixed it in the simplest way. I must be over thinking on `{ }` vs `{ gitHubConfig: undefined}`

src/transforms/push.ts
gxueatlassian: fixed.

src/transforms/push.ts
gxueatlassian: Actually I think leaving it as it is is better, as the map file will produce `/some/path/null/some/file` instead of `/some/path//some/file`. Reverted.

src/github/push.ts
mboudreau: ...but you're already using it in the context here.  He's just saying to add it to the type

src/transforms/push.unit.test.ts
```diff
@@ -0,0 +1,43 @@
...
+import { sqsQueues } from "../sqs/queues";
+import { GitHubPushData } from "../interfaces/github";
+
+jest.mock("../sqs/queues", () => ({
+	sqsQueues: {
+		push: {
+			sendMessage: jest.fn()
+		}
+	}
+}));
```
mboudreau, line 11: just `jest.mock("../sqs/queues");` will do.

src/transforms/push.unit.test.ts
```diff
@@ -0,0 +1,43 @@
...
+
+describe("Enqueue push", ()=>{
+	describe("GitHub Enterprise Server", ()=>{
+		it("should push GitHubAppConfig to payload", async ()=>{
+			await enqueuePush({
+				installation: { id: 123, node_id: 456 },
+				webhookId: "wh123",
+				webhookReceived: Date.now(),
+				repository: {} as GitHubPushData["repository"],
```
mboudreau, line 20: shouldn't this be `as GitHubRepository`?

src/github/push.ts
gxueatlassian: OH, SORRY, haha, thought something else. That disturbed thing really got me. 
I did added the `gitHubAppConfig` to the `WebhookContext` types in file `src/routes/github/webhook/webhook-context.ts`. Not sure I get it right?

src/transforms/push.unit.test.ts
gxueatlassian: fixed.

src/github/push.unit.test.ts
```diff
@@ -0,0 +1,71 @@
...
+			log: getLogger("test"),
+			payload,
+			...(cloud ? undefined : {
+				gitHubAppConfig: {
+					uuid: GHES_GITHUB_UUID,
+					gitHubAppId: GHES_GITHUB_APP_ID,
+					appId: GHES_GITHUB_APP_APP_ID,
+					clientId: GHES_GITHUB_APP_CLIENT_ID,
+					gitHubBaseUrl: gheUrl
+				}
```
mboudreau, line 66: why does cloud not have the githubAppConfig?  It's not GHE specific.  Also, no need to use deconstructor, just `gitubAppConfig: cloud ? undefined : { config things }`

src/transforms/push.unit.test.ts
gxueatlassian: `GitHubRepository` is not exported. Not sure should I export it or just use `GitHubPushData["repository"]`. 
If `export`, then next time some code use some other parts, we will export again? So eventually many parts will export?

src/github/push.unit.test.ts
gxueatlassian: `why does cloud not have the githubAppConfig` <--
1. Seems many places in the code, we have this impression, `if cloud? then undefined, else if GHE, then some concrete value`. Don't want to break that yet.
2. I suspect majority traffic of the webhook event is cloud. Adding cloud appId(and others) to every single queue msg will big impact. Not like GHE we have to put at least one id in the payload, for cloud we can get away without any id. So given the impact is big, I am hesitate to put it in the payload now. 

--------------------

Can we think about it, and if decided to put cloud, do it in another PR?

-------------

For `just gitubAppConfig: cloud ? undefined : { config things }`, it is me always been thinking there's a different of an object without key `{}` vs and `object with a key, but value is undefined`. But on a second thoughts, those seems no differences from coding point of view. So I revert the typing to `gitHubAppConfig?: GitHubAppConfig` from `gitHubAppConfig: GitHubAppConfig | undefined`

src/github/push.ts
gxueatlassian: Closing this threads, since the type is already in the `WebhookContext` type.

src/transforms/push.unit.test.ts
mboudreau: personally I would, since you've hit a situation where it's needed.  Exporting types is not as much of an issue than exporting functions

src/github/push.unit.test.ts
```diff
@@ -0,0 +1,69 @@
+import { pushWebhookHandler } from "./push";
```
thombergs, line 1: Is the name `*.unit.test.ts` a new convention? I'm sure many of our existing `*.test.ts` are unit tests, too.

src/transforms/push.unit.test.ts
mboudreau: This hasn't been fixed yet.

src/github/push.unit.test.ts
```diff
@@ -0,0 +1,69 @@
...
+
+describe("PushWebhookHandler", ()=>{
+	let jiraClient: any;
+	let util: any;
+	beforeEach(() => {
+		jiraClient = { baseURL: jiraHost };
+		util = null;
+	});
+	describe("GitHub Cloud", ()=>{
+		it("should NOT push GitHubAppConfig", async ()=>{
```
thombergs, line 23: I think `should NOT push GitHubAppConfig` is a bit misleading, because no one is pushing the `GitHubAppConfig`. Maybe call it `should be called WITHOUT GitHubAppConfig` instead?

src/github/push.unit.test.ts
```diff
@@ -0,0 +1,69 @@
...
+			await pushWebhookHandler(getWebhookContext({ cloud: true }), jiraClient, util, GHES_GITHUB_INSTALLATION_ID);
+			expect(enqueuePush).toBeCalledWith(expect.anything(), expect.anything(), undefined);
+		});
+	});
+	describe("GitHub Enterprise Serer", ()=>{
```
thombergs, line 28:
```suggestion
	describe("GitHub Enterprise Server", ()=>{
```

src/interfaces/github.ts
```diff
@@ -224,10 +226,17 @@ interface GitHubInstallation {
...
+	id: string;
+	message: string;
+}
+
 export interface GitHubPushData {
 	webhookId: string;
 	webhookReceived: number;
 	repository: GitHubRepository;
-	commits: string[];
+	commits: GitHubCommit[];
```
thombergs, line 240: How did this work before when it was `string[]`? Shouldn't the code at https://github.com/atlassian/github-for-jira/blob/main/src/github/push.ts#L17 change as a result of this?

src/transforms/push.unit.test.ts
gxueatlassian: Sorry, turns out there're two places, I only fixed the other one.

Now this one is fixed as well.

src/github/push.unit.test.ts
gxueatlassian: This is a bit funny, as the current `push.test.ts` under the same folder is actually testing things somewhere else, it is NOT testing anything `push.ts` in the same directory. So I've to create another one to unit test it. Leave the current one untouched.

src/github/push.unit.test.ts
gxueatlassian: 👍 fixed.

src/github/push.unit.test.ts
gxueatlassian: fixed.

src/interfaces/github.ts
gxueatlassian: The code is doing bellow
image

At runtime the code is working regardless, at compile time typing is only `any`.

src/github/push.unit.test.ts
rachellerathbone: Which test are you talking about? There doesn't appear to be any other test for push in that directory 😕 

Screen Shot 2022-08-10 at 1 06 48 pm

gxueatlassian: Remove origin file to sqs/push.test.ts

src/github/push.unit.test.ts
gxueatlassian: 1. Put cloud in
2. Check api.github vs github

gxueatlassian: fixed.

src/github/push.unit.test.ts
gxueatlassian: fixed.

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -36,15 +37,23 @@ export const WebhookReceiverPost = async (request: Request, response: Response):
...
-			...(!gitHubServerApp ? undefined : {
-				gitHubAppConfig: {
+			gitHubAppConfig: {
+				...(!gitHubServerApp ? {
+					gitHubAppId: undefined,
+					appId: parseInt(envVars.APP_ID),
+					clientId: envVars.GITHUB_CLIENT_ID,
+					gitHubBaseUrl: GITHUB_CLOUD_HOSTNAME,
+					gitHubApiUrl: GITHUB_CLOUD_API_BASEURL,
+					uuid: undefined
```
joshkay10, line 47: i've not touched the GHE stuff so unsure, is gitHubAppId always a ghe server id?

can we just omit `gitHubAppId` and `uuid` instead of passing undefined?

gxueatlassian: `Is gitHubAppId always a ghe server id?` <-- Yes, it is always the PK in the `GitHubServerApp` table.

`can we just omit gitHubAppId and uuid instead of passing undefined?` <-- I suppose so, from code point of view, I think they are the same? Not sure what's the team's convention for `undefined`?

joshkay10 (approved):

src/routes/github/webhook/webhook-receiver-post.ts
gxueatlassian: Just talked to GHE team, we don't have an convention. so passing in `undefined` as well to maintain a similar structure.

mboudreau (approved):

PR #1458, opened 2022-08-04
ARC-1567 - Test and fix the back buttons on all the routes

krazziekay:
**What's in this PR?**
- Updated `navigation.js` to handle back navigation for all cases without passing anything, by using `AP.history`.
- Usage of `previousPagePath` and `previousPathName` removed

**Why**
- After connecting all the separate routes and with the change in the route structure, the back buttons in `navigation.js` were not working.

**How has this been tested?**  
- In local environment

### DIFF
--- a/static/js/navigation.js
+++ b/static/js/navigation.js
@@ -1,10 +1,9 @@
 /* globals $, AP */
 $(".go-back").click(function (event) {
   event.preventDefault();
-  const previousPagePath = $(event.target).data("previous-page-path");
 
-  if (previousPagePath) {
-    AP.navigator.go("addonmodule", {moduleKey: previousPagePath});
+  if (AP.history) {
+    AP.history.back();
   } else {
     history.back();
   }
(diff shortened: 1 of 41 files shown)


rachellerathbone (dismissed):

static/js/navigation.js
```diff
@@ -1,10 +1,9 @@
 /* globals $, AP */
 $(".go-back").click(function (event) {
   event.preventDefault();
-  const previousPagePath = $(event.target).data("previous-page-path");
 
-  if (previousPagePath) {
-    AP.navigator.go("addonmodule", {moduleKey: previousPagePath});
+  if (AP.history) {
```
joshkay10, line 5: is AP always defined? could add extra defense with `AP && AP.history`

krazziekay: In all of the pages where we load `navigation.js`, the `https://connect-cdn.atl-paas.net/all.js` has also been loaded. 
So, it shouldn't error out, but agreed with you. So, added an extra check.

rachellerathbone (approved):

PR #1464, opened 2022-08-04
update path to optional require... was working... now maybe isn't???

rachellerathbone:
Attempting to fix analytics yet again. Starting to feel like I'm chasing down some mythical creature.

### DIFF
--- a/src/util/analytics-client.ts
+++ b/src/util/analytics-client.ts
@@ -1,7 +1,7 @@
 import { getLogger 	} from "config/logger";
 import { omit } from "lodash";
-import { optionalRequire } from "optional-require";
 import { isNodeProd } from "utils/is-node-env";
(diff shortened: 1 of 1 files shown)


src/util/analytics-client.ts
```diff
@@ -1,7 +1,7 @@
 import { getLogger 	} from "config/logger";
 import { omit } from "lodash";
-import { optionalRequire } from "optional-require";
 import { isNodeProd } from "utils/is-node-env";
+import { optionalRequire } from "optional-require/dist";
```
krazziekay, line 4: `optional-require` is already exporting everything inside of `dist/index.d`, so not sure if we need to add another `/dist`.

krazziekay (approved):

PR #1468, opened 2022-08-07
ARC-1575 - Create error page for app creation failures

krazziekay:
**What's in this PR?**
- New error message added in ErrorRouter, when the user gets a 422 error in `github-manifest-complete`
- Updated the `error.hbs` 

**Why**
- To display the error screen when user tries to add a new GitHub server app without passing in a name.

**How has this been tested?**  
- In local 
- In staging

https://user-images.githubusercontent.com/13076255/183341544-e7deed54-d109-4338-aaba-9002a2b7a0b6.mov


src/routes/error-router.ts
```diff
@@ -63,13 +69,27 @@ ErrorRouter.use((err: Error, req: Request, res: Response, next: NextFunction) =>
 		"Not Found": 404
 	};
 
-	const messages = {
-		[Errors.MISSING_JIRA_HOST]: "Session information missing - please enable all cookies in your browser settings.",
-		[Errors.IP_ALLOWLIST_MISCONFIGURED]: `The GitHub org you are trying to connect is currently blocking our requests. To configure the GitHub IP Allow List correctly, <a href="${envVars.GITHUB_REPO_URL}/blob/main/docs/ip-allowlist.md">please follow these instructions</a>.`
-	};
+	let message: string | null = null;
```
rachellerathbone, line 72: What was the drive behind completely refactoring this? Can't we follow the same pattern and add a new error to the messages object and update the Errors enum? If we stick to this pattern then we don't need to declare ctaUrl and message as undefined variables.

src/routes/error-router.ts
```diff
@@ -63,13 +69,27 @@ ErrorRouter.use((err: Error, req: Request, res: Response, next: NextFunction) =>
...
+	let ctaUrl: Cta | null = null;
```
rachellerathbone, line 73: Couple of things here:
- The text also shouldn't be defined on the backend as the backend should have no knowledge of what is rendered in the UI. Create the url and then set the text in the UI (there's no need for it to be passed from the backend).
- I think ctaUrl should be defined directly set it in the render function: `ctaUrl: createUrlWithQueryString(req, "/session")`.

views/error.hbs
```diff
@@ -26,18 +33,24 @@
...
-      <p class="error-message">
-        We track these errors automatically, but if the problem persists feel
-        free to
-        <a
-          href="https://support.atlassian.com/contact/#/?inquiry_category=technical_issues&is_cloud=true&product_key=third-party-product"
-        >contact us</a>. In the meantime, try adding an organization again in
-        Jira.
-      </p>
+      {{#if ctaUrl}}
+          <a class="aui-button aui-button-primary m-4" href="{{ctaUrl.action}}">
```
rachellerathbone, line 37: When you update the middleware this should just be `href="{{ctaUrl}}"` and `{{ctaUrl.text}}` should just be `Retry`

rachellerathbone (changes requested):

src/routes/error-router.ts
krazziekay: The reason for changing the original pattern was to add `ctaUrl`. The call to action URL is only for the MISSING_GITHUB_APP_NAME error. We can easily add a call to action link to any of the existing errors or new errors, in this way. 
I didnt want to add another `if condition` to add the `ctaUrl` for this error.

src/routes/error-router.ts
krazziekay: The call to action URL is only for the _MISSING_GITHUB_APP_NAME_ error. So, if we go by this approach, we'll have to add another condition just to add `ctaUrl`. I wanted to avoid having to do this separate condition.

src/routes/error-router.ts
```diff
@@ -63,13 +69,27 @@ ErrorRouter.use((err: Error, req: Request, res: Response, next: NextFunction) =>
...
+	switch (err.message) {
+		case Errors.MISSING_JIRA_HOST:
+			message = "Session information missing - please enable all cookies in your browser settings.";
+			break;
+		case Errors.MISSING_GITHUB_APP_NAME:
+			message = "There was a problem creating your GitHub App. Please make sure you filled the GitHub App name and try again.";
+			ctaUrl = {
+				text: "Retry",
```
rachellerathbone, line 81: Again. It's bad practice to define the text on the backend. We shouldn't be setting this here.

src/routes/error-router.ts
rachellerathbone: You can easily add a call to action url and pass an additional value to the frontend **without** refactoring this entire file. Just add `MISSING_GITHUB_APP_NAME` to the object:

```
	const messages = {
		[Errors.MISSING_JIRA_HOST]: "Session information missing - please enable all cookies in your browser settings.",
		[Errors.IP_ALLOWLIST_MISCONFIGURED]: `The GitHub org you are trying to connect is currently blocking our requests. To configure the GitHub IP Allow List correctly, <a href="${envVars.GITHUB_REPO_URL}/blob/main/docs/ip-allowlist.md">please follow these instructions</a>.`
	};
  [Errors.MISSING_GITHUB_APP_NAME]: "message..."
```

Then ditch the ctaUrl object, define it in the render function, and hardcode the "Retry" text on the frontend you don't need to refactor everything.
- **Kill the ctaUrl object:** ctaUrl = { text: "Retry", action: createUrlWithQueryString(req, "/session")};
- **Set the url in the render function:** `ctaUrl: createUrlWithQueryString(req, "/session")`
- and change the anchor tag to: `<a class="aui-button aui-button-primary m-4" href="{{ctaUrl}}">Retry</a>`

rachellerathbone (changes requested):

src/routes/error-router.ts
krazziekay: * Reverted back to the previous logic. 
* The new error message was added as a part of the same object.
* Used query parameters for handling the action of the retry button

views/error.hbs
krazziekay: updated

src/routes/error-router.ts
krazziekay: Retry text removed from backend and hard coded in frontene

rachellerathbone (dismissed):

rachellerathbone (approved):

PR #1472, opened 2022-08-08
ARC-1563 put and delete

rachellerathbone:
**What's in this PR?**
Functionality for 'edit' and 'delete' in the GH app dropdown for GHES.

**Why**
So users can edit and delete GH apps.

**Added feature flags**
All behind GHE_SERVER

**Affected issues**  
ARC-1563

**How has this been tested?**  
Tested in stg and locally:

https://user-images.githubusercontent.com/37155488/183027316-f5880528-58c6-48d2-b8d8-960e619b4796.mov

**What's Next?**
- ARC 1572: update a GH app. In this PR I'm simply navigating to the form. Will take care of actually updating the fields in this issue.
- ARC 1577: we need to refactor the jira configuration table partial so it is an actual reusable component. We should be using this for the dropdown from cloud and for server gh apps. This will be addressed outside of the scope for this project.


src/models/github-server-app.ts
```diff
@@ -97,14 +97,10 @@ export class GitHubServerApp extends EncryptedModel {
 	 * @param installationId
 	 * @returns {GitHubServerApp[]}
 	 */
-	static async getAllForGitHubBaseUrl(
```
rachellerathbone, line 100: Renamed for clarity and modified so we're returning an empty array if no entries are found (same as Subscription and Installation).

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-router.ts
```diff
@@ -13,4 +13,4 @@ JiraConnectEnterpriseAppRouter.post("/", JiraContextJwtTokenMiddleware, JiraConn
 JiraConnectEnterpriseAppRouter.route("/:uuid")
 	.get(csrfMiddleware, JiraJwtTokenMiddleware, JiraConnectEnterpriseAppCreateOrEdit)
 	.put(csrfMiddleware, JiraContextJwtTokenMiddleware, JiraConnectEnterpriseAppPut)
-	.delete(csrfMiddleware, JiraContextJwtTokenMiddleware, JiraConnectEnterpriseAppDelete);
\ No newline at end of file
+	.delete(JiraContextJwtTokenMiddleware, JiraConnectEnterpriseAppDelete);
```
rachellerathbone, line 16: This middleware only works on GET requests. We need to investigate this separately.

static/js/jira-configuration.js
```diff
@@ -171,3 +171,43 @@ window.onclick = function(event) {
...
+});
+
+$(".jiraConfiguration__editGitHubApp").click(function(event) {
+	event.preventDefault();
+	const uuid = $(event.target).data("app-uuid");
+
+	AP.navigator.go(
+		'addonmodule',
+		{
+			moduleKey: "github-edit-app-page",
```
rachellerathbone, line 206: Navigating to edit form in this PR. Adding PUT req in 1572.

views/jira-configuration-new.hbs
```diff
@@ -214,28 +214,31 @@
...
-                              </a>
-                            </aui-section>
-                          </aui-dropdown-menu>
+                          <aui-dropdown-menu class="jiraConfiguration__table__dropdown" id="app-settings-dropdown-{{id}}">
+                          <h6 class="jiraConfiguration__table__cell__settings__header">
+                            <span class="jiraConfiguration__table__cell__settings__headerItem">Application</span>
+                          </h6>
+                          <aui-section class="jiraConfiguration__table__cell__settings__dropdownItems">
+                             <div class="jiraConfiguration__table__cell__gitHubAppItem">
+                               <button
```
rachellerathbone, line 223: Not sure why we were using links here because we're not navigating outside the app 🤷 Had to wrap each list item in a div as well otherwise clicking on one would fire the other (both at the same time).

rachellerathbone:
Got commit history all jacked up 😬

PR #1474, opened 2022-08-09
Probot sweep 090922

joshkay10:
**What's in this PR?**
Continued to pull out Probot.

**Why**
It's dead code, it exists but is not used, tests needed a little bit of tidying to use client instead, they still relied on the probot flow and no longer do.

**Added feature flags**
No dice.

**How has this been tested?**  
Soaked in stage and tested, backfill, webhook, install.

**Whats Next?**
There is still drips and drabs hanging around, will do another sweep, PR was big enough already.


bgvozdev:
What about
src/configure-robot.ts and src/github/webhooks.ts ?

gxueatlassian:
> What about src/configure-robot.ts and src/github/webhooks.ts ?

This PR I believe is for only removing the unused githubApi from the code. It is already touching 27 files, so too big for me to include removing the probot server in this PR as well.

bgvozdev:
Hey @gxueatlassian thanks for asking, it makes sense to keep it outside of this PR. But this work is planned ahead, right?

bgvozdev (approved):

PR #1477, opened 2022-08-09
ARC-1572 update gh app

rachellerathbone:
What's in this PR?
PUT request for editing form and the addition of a flag to notify users of how we are 'storing' the filename of the private key (we don't actually store the filename but instead use the app name). 

Affected issues
ARC-1572

How has this been tested?
Test added and locally.

Manual flow for a new app - confirming it still works and that the new flag renders
* first I enter a gh app name and then upload a file. You'll notice the app name from the gh app name input is included in the flag
* then, after refreshing, I start by uploading the file. Because the gh app name input is empty we default to < github-app-name >.private-key.pem (needed to add spaces in the angled brackets otherwise the markdown logic in the AUI flag component treats it as an html element)

https://user-images.githubusercontent.com/37155488/184080875-facfef8d-c824-401e-9aa2-0947639e71ff.mov

Editing app:
- we now render a file name in the format previously mentioned
- if a new file isn't uploaded, we just included the existing one
- if a new file is uploaded, we send the contents of the new file in the payload

https://user-images.githubusercontent.com/37155488/184081030-c449b329-620c-4128-a601-560385c1ed04.mov

What's Next?
n/a


src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-put.ts
```diff
@@ -8,10 +9,12 @@ export const JiraConnectEnterpriseAppPut = async (
 	try {
 		req.log.debug("Received Jira Connect Enterprise App PUT request");
 
-		// TODO: Add logic for updating apps
+		await GitHubServerApp.updateGitHubApp(req.body);
```
rachellerathbone, line 12: Interested to hear people's thoughts here: should we do a read first (get current entry in GHSA), compare the objects, and if there's a difference, then update? I figured we're only updating one entry so should be pretty fast/cheap but am open to opinions.

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-post.ts
```diff
@@ -33,13 +33,11 @@ export const JiraConnectEnterpriseAppPost = async (
 			installationId: installation.id
 		});
 
-		res.status(201).json({
-			message: `Successfully added GitHub Server App for ${installation.id}`,
-			data: githubServerApp
-		});
+		res.status(200).send({ success: true, data: githubServerApp });
```
rachellerathbone, line 36: Updated this to align with how we are sending other responses in the server flow.

views/jira-manual-app-creation.hbs
```diff
@@ -163,13 +163,15 @@
...
         <div class="jiraManualAppCreation__serverError">
-          {{> server-error-message-box
-              title="We couldn't create your GitHub app"
-              message="Please make sure all the details you entered are correct."
-          }}
+          {{> server-error-message-box }}
         </div>
 
-        <input type="submit" value="{{#if app.id}}Update{{else}}Next{{/if}}" class="jiraManualAppCreation_formSubmitBtn aui-button aui-button-primary" />
+        <input
```
krazziekay, line 169: One thing is remaining, when the user edits an app, then the form should display a file name instead of the blank _Drag and Drop_ area. .
he value for the input file `privateKeyFile` needs to set, otherwise the update button won't work as the form is not valid.

rachellerathbone: Gotcha. I'd just been uploading the file again. Will update.

src/models/github-server-app.ts
```diff
@@ -163,6 +163,53 @@ export class GitHubServerApp extends EncryptedModel {
...
+	 * 		gitHubBaseUrl: string,
+	 * 		gitHubClientSecret: string,
+	 * 		webhookSecret: string,
+	 * 		privateKey: string,
+	 * 		gitHubAppName: string,
+	 * 		installationId: number
+	 * 	}} payload
+	 */
+
+	static async updateGitHubApp(payload: GitHubServerAppPayload): Promise<void> {
```
gxueatlassian, line 182: /nit
I suggest change the name to `updateGitHubAppByUUID` so that it is explicit that `uuid` is the key to locate the record.

src/models/github-server-app.ts
```diff
@@ -163,6 +163,53 @@ export class GitHubServerApp extends EncryptedModel {
...
+			gitHubAppName,
+			gitHubBaseUrl,
+			gitHubClientId,
+			gitHubClientSecret,
+			webhookSecret,
+			privateKey,
+			installationId
+		} = payload;
+
+		await this.update(
```
gxueatlassian, line 195: Please add some tests.
It'd be great if we can test happy path and some edge cases, like the payload is only partial, will it override all data in config? Or record not found, what's the expectation?

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-put.ts
gxueatlassian: **IMPORTANT**

I think we need to rely on the `gitHubSeverAppMiddlware` to provide the `uuid/gitHubAppId` info, instead of relying on the `req.body`, as user can feed in any `uuid` they want, might accidentally override others.

src/models/github-server-app.ts
gxueatlassian: Ignore this if you think the tests are cover in another places. :)

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-put.test.ts 
```diff
@@ -0,0 +1,50 @@
...
+		});
+	});
+
+	it("should update GitHub app when uuid is found", async () => {
+		const originalApp = await GitHubServerApp.findForUuid(uuid);
+		expect(originalApp && originalApp.gitHubClientId).toEqual(clientIdWithTypo);
+		expect(originalApp && originalApp.webhookSecret).toEqual(`encrypted:${originalWebhookSecret}`);
+		expect(originalApp && originalApp.privateKey).toEqual(`encrypted:${originalPrivateKey}`);
+
+		await GitHubServerApp.updateGitHubApp({
```
gxueatlassian, line 32: This test is a bit strange? From the file name, it is for the `jira-connect-enterprise-app-put` handler, but the test itself is directly testing the db models. 

If it is testing the db models, I'd say change the file name to `models/github-server-app.test.ts`, if testing the handler, I'd say use the `supertest` to hit the api.

views/jira-manual-app-creation.hbs
gxueatlassian: So when updating, are we expecting user to upload the `privateKey` file again?
Depends on that, we may or may not want to "set the value of the input file", and also affect how we handle it in the db saving model, does it allow partial save or always full save?

gxueatlassian (dismissed):
Hi just one **IMPORTANT** comment to address. Others a minor.

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-put.ts
gxueatlassian: Okay, `gitHubSeverAppMiddlware` is not available, since this is the `jira` route. Maybe we can you please explain how is the `uud` verified that it belongs to the user?

src/models/github-server-app.ts
rachellerathbone: It would never just be partial. When we direct a user to the editing form, we are auto-populating all the fields with the data from the fb (masking secret stuff). If they try to submit the form with empty fields the frontend validation fails.

views/jira-manual-app-creation.hbs
rachellerathbone: @krazziekay seeing we don't have the name of the file, were you thinking we should instead use the name of the app? e.g. my-app.pem?

krazziekay: We can use any file name as discussed [here](https://atlassian.slack.com/archives/C039YPCGCSG/p1658795822188309). 
The only thing that the user can do is just see. If the user wants to change it, then they'll have to remove the current one and add the new one.
So, just need to check if there is existing private key, and if there is, we simply show a file already added instead of the blank file input, with any name. 
And we would also need to pass the saved private key as it is if the user hasn't upload any new file, when hitting submit.

rachellerathbone: Checking with design to see what their thoughts are from a UX standpoint.

src/models/github-server-app.test.ts
```diff
@@ -128,6 +135,92 @@ describe("GitHubServerApp", () => {
...
+					expect(updatedApp && updatedApp.uuid).not.toEqual(gitHubServerApp.uuid);
+					expect(updatedApp && updatedApp.gitHubClientId).not.toEqual("lvl.1n23j12389wnde");
+					expect(updatedApp && updatedApp.webhookSecret).not.toEqual("anewsecret");
+					expect(updatedApp && updatedApp.privateKey).not.toEqual("privatekeyversion2");
+				});
+
+				it("should only update values changed and leave other values as is", async () => {
+					// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+					// @ts-ignore
+					await GitHubServerApp.updateGitHubAppByUUID({ uuid, appId: 1, gitHubBaseUrl, webhookSecret: "updatedSecret" });
```
rachellerathbone, line 212: We shouldn't be able to submit a form with missing values as the frontend validation will fail. But added this test as proof that values in the db remain unaffected if data is missing from the payload.

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-post.ts
rachellerathbone, line 36: Modified this to align how we are sending responses elsewhere in the app.

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-put.ts
```diff
@@ -1,17 +1,27 @@
...
 export const JiraConnectEnterpriseAppPut = async (
 	req: Request,
 	res: Response,
 	next: NextFunction
 ): Promise<void> => {
+	req.log.debug("Received Jira Connect Enterprise App PUT request to update app.");
+
 	try {
-		req.log.debug("Received Jira Connect Enterprise App PUT request");
+		const verifiedApp = await GitHubServerApp.getForUuidAndInstallationId(req.params.uuid, res.locals.installation.id);
```
rachellerathbone, line 12: Added this additional check to make sure the uuid in the url matches the installationId. If it doesn't match, the request fails.

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-put.ts
```diff
@@ -1,17 +1,27 @@
-import { NextFunction, Request, Response } from "express";
+import { Request, Response, NextFunction } from "express";
+import { GitHubServerApp } from "~/src/models/github-server-app";
 
 export const JiraConnectEnterpriseAppPut = async (
 	req: Request,
 	res: Response,
 	next: NextFunction
 ): Promise<void> => {
+	req.log.debug("Received Jira Connect Enterprise App PUT request to update app.");
```
rachellerathbone, line 9: Moved this out of the try/catch. If the request fails, we don't see this log and can't be certain we hit this middleware (think this needs to be updated in other places as well).

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-router.ts
```diff
@@ -12,5 +12,5 @@ JiraConnectEnterpriseAppRouter.post("/", JiraContextJwtTokenMiddleware, JiraConn
 
 JiraConnectEnterpriseAppRouter.route("/:uuid")
 	.get(csrfMiddleware, JiraJwtTokenMiddleware, JiraConnectEnterpriseAppCreateOrEdit)
-	.put(csrfMiddleware, JiraContextJwtTokenMiddleware, JiraConnectEnterpriseAppPut)
+	.put(JiraContextJwtTokenMiddleware, JiraConnectEnterpriseAppPut)
```
rachellerathbone, line 15: Again, the csrf middleware only works in GET. We need to investigate this separately.

static/js/jira-manual-app-creation.js
```diff
@@ -1,79 +1,130 @@
...
-        'addonmodule',
-        {
-          moduleKey: "github-post-install-page"
-        }
-      );
-    }
-  }, 1000);
-
-  return child;
+	const child = window.open(url);
```
rachellerathbone, line 6: No idea why the indentation looks like this here. Fine in my IDE 🤷

static/js/jira-manual-app-creation.js
```diff
@@ -1,79 +1,130 @@
...
+			// Reading the content of the file
+			const file = $("#privateKeyFile")[0].files[0];
+			const reader = new FileReader();
+			reader.readAsDataURL(file);
+
+			reader.onload = () => {
+				data.privateKey = reader.result;
+			};
+		} else {
+			data.privateKey = existingPrivateKey;
```
rachellerathbone, line 86: For the manual form (non-edit mode) or the editing form when the private key file is changed, we'll set the privateKey to reader.result, otherwise set it to the existingPrivateKey.

static/js/jira-manual-app-creation.js
```diff
@@ -1,79 +1,148 @@
...
+					child.window.jwt = token;
+				} else {
+					handleFormErrors(isUpdate);
+				}
+			});
+		}
+	});
+});
+
+$('#jiraManualAppCreation__uploadedFile').bind('DOMSubtreeModified', function () {
```
rachellerathbone, line 106: Adding a flag when a file is uploaded. Because we aren't saving the name of the file, the best we can do to render a file name in editing mode is to use the format of <app-name>.private-key.pem. So if a user uploads a file, then goes back to edit the form, the 'incorrect' file naming will be confusing and may lead to customers thinking something has gone wrong. This notifies them of the file name formatting.

src/models/github-server-app.test.ts
```diff
@@ -128,6 +135,92 @@ describe("GitHubServerApp", () => {
...
+						gitHubClientSecret: "secret",
+						webhookSecret: "anewsecret",
+						privateKey: "privatekeyversion2",
+						installationId
+					});
+
+					const updatedApp = await GitHubServerApp.findForUuid(mismatchedUUID);
+					expect(updatedApp && updatedApp.uuid).not.toEqual(gitHubServerApp.uuid);
+					expect(updatedApp && updatedApp.gitHubClientId).not.toEqual("lvl.1n23j12389wnde");
+					expect(updatedApp && updatedApp.webhookSecret).not.toEqual("anewsecret");
```
gxueatlassian, line 205: Hmm, that's weird, shouldn't it be `updatedApp.decrypt("webhookSecret")` inorder to get the origin value?

gxueatlassian: Alright, seems like a bug in the base GitHubServerApp model. Let me have a look.

src/models/github-server-app.test.ts
```diff
@@ -128,6 +135,92 @@ describe("GitHubServerApp", () => {
...
+					expect(originalApp && originalApp.gitHubClientId).toEqual(clientIdWithTypo);
+					expect(originalApp && originalApp.webhookSecret).toEqual(`encrypted:${originalWebhookSecret}`);
+					expect(originalApp && originalApp.privateKey).toEqual(`encrypted:${originalPrivateKey}`);
+
+					await GitHubServerApp.updateGitHubAppByUUID({
+						uuid,
+						appId: 1,
+						gitHubAppName: "my awesome app",
+						gitHubBaseUrl,
+						gitHubClientId: "lvl.1n23j12389wnde",
```
gxueatlassian, line 173: /nit
Use const variable to better show the intention of the test. It took me a while to figure out where is the "typo".
So maybe either use some const var, or simpler, set the string vaule "origin_client_id" vs "new_client_id", so that the test is more readable. And "origin_webhook_secret" vs "new_webhook_secret".

src/models/github-server-app.ts
```diff
@@ -109,6 +109,25 @@ export class GitHubServerApp extends EncryptedModel {
 		});
 	}
 
+	/**
+	 * Get one GitHubServerApp for uuid with installationId
+	 *
+	 * @param uuid
+	 * @param installationId
+	 * @returns {GitHubServerApp[]}
```
mboudreau, line 117: I'd prefer we don't do JSDocs when we're using TS.  TS already has types and definition and they'll conflict with each other in the IDE

src/models/github-server-app.test.ts
gxueatlassian: `updatedApp.webhookSecret` is always `encrypted:xxx`, so never equal to anything `anewsecret` or `anothersecret`. The test itself is invalid.
I suggest change it to:
`expect(await updatedApp.decyrpt("webhookSecret")).toBe(ORIGIN_WEBHOOK_SECRET);`

src/models/github-server-app.test.ts
```diff
@@ -128,6 +135,92 @@ describe("GitHubServerApp", () => {
...
+					expect(updatedApp && updatedApp.privateKey).not.toEqual("privatekeyversion2");
+				});
+
+				it("should only update values changed and leave other values as is", async () => {
+					// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+					// @ts-ignore
+					await GitHubServerApp.updateGitHubAppByUUID({ uuid, appId: 1, gitHubBaseUrl, webhookSecret: "updatedSecret" });
+					const myApp = await GitHubServerApp.findForUuid(uuid);
+
+					expect(myApp && myApp.uuid).toEqual(uuid);
```
gxueatlassian, line 215: Btw, I think we can change all these `expect(xxx & xxx.gitHubAppName)...` to just `expect(xxx.gitHubAppName)`. 
The `myApp` here has to be not null, if null, test should fail, so it is okay to skip the `myApp &&`

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-put.test.ts
```diff
@@ -0,0 +1,186 @@
...
+		});
+		app.use(RootRouter);
+
+		jwt = encodeSymmetric({
+			qsh: "context-qsh",
+			iss: jiraHost
+		}, installation.sharedSecret);
+	});
+
+	it("should return 200 with success true when correct uuid and installation id are passed", async () => {
```
gxueatlassian, line 38: For `uuid and installation id are passed` <-- I don't think `installation id` is passed (from client browser to server)?

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-put.test.ts
```diff
@@ -0,0 +1,186 @@
...
+			gitHubClientId: "lvl.1n23j12389wndd",
+			gitHubClientSecret: "secret",
+			webhookSecret: "anothersecret",
+			privateKey: "privatekey",
+			installationId: installation.id
+		});
+
+		const payload ={
+			gitHubAppName: "my-app",
+			webhookSecret: `encrypted:${uuid}`,
```
gxueatlassian, line 124: not a big problem, but at this stage, the "webhookSecret" should be plain text. But technically, this test still works.

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-put.ts
```diff
@@ -1,17 +1,27 @@
...
-		// TODO: Add logic for updating apps
+		await GitHubServerApp.updateGitHubAppByUUID(req.body);
 
-		req.log.debug("Jira Connect Enterprise App updated successfully.", res.locals);
+		res.status(200).send({ success: true });
+		req.log.debug("Jira Connect Enterprise App updated successfully.");
 	} catch (error) {
-		return next(new Error(`Failed to render Jira Connect Enterprise App PUT request: ${error}`));
+		res.status(200).send({ success: false, message: "Failed to update GitHub App." });
+		return next(new Error(`Failed to update GitHub app: ${error}`));
```
gxueatlassian, line 25: **IMPORTANT**
I have concern on this, I am not sure how `next` is handling the `new Error`. So worst case is, the variable `error` contains some confidential info, and `next(new Error)` will be captured and render to html somewhere else by some middleware. 

From safety reason, maybe we can omit the `${error}` in the `Error constructor`? But surely we can put a `log.warn` here.

static/js/jira-manual-app-creation.js
gxueatlassian: Switch to vim!

static/js/jira-manual-app-creation.js
```diff
@@ -1,79 +1,150 @@
...
-    });
-  };
+	event.preventDefault();
+	const form = event.target;
+	const data = $(form).serializeObject();
+	const updateElement = document.getElementById("Update");
+	const isUpdate = updateElement && updateElement.innerText === "Update";
+	const uuid = $(event.target).data("app-uuid");
+	const appName = $(event.target).data("app-appname");
+	const existingPrivateKey = $(event.target).data("app-privatekey");
```
gxueatlassian, line 67: **IMPORTANT**

**NO NEED TO CHANGE IN THIS PR THOUGH**

I am concern from security point of view. I believe it is better to transmit the sensitive data as less as possible.

So in theory, we can omit sending the privateKey or webhook secret or whatever down to client browser for update. 

----

But I maybe misunderstood the code here. Let me know if I am wrong or need me to explain more.

static/js/jira-manual-app-creation.js
gxueatlassian: /nit

I think I am not 100% following the code here. Too me seems like quite a lot of code going on for the `file name`. 

I wonder is that `file name` thing a UX requirement? Or is it just a technical issue (like the file input is mandatory)? 
It is just this `$( "input[type=text][name=gitHubAppName]" ).val().replace(/\s+/g, '-').toLowerCase()` quite brain energy consuming to me.

static/js/jira-manual-app-creation.js
rachellerathbone: Yeah I had been thinking the same thing. You okay with me creating a follow-up ticket to look at an alternate solution for this?

static/js/jira-manual-app-creation.js
rachellerathbone: What do you mean by 'file name thing'? I think you and @krazziekay had discussed rendering the file name in editing mode. As we don't save the file name of a stored private key, our only option is to use the format of <app-name>.private-key.pem. Jade has confirmed, from a UX standpoint, that is better that we have the file name rendered and don't force the user to have to reupload the private key any time they want to edit another app value.

This bind call detects a change in the element that handles displaying the file name. If a change occurs, we render a flag to inform the user that the private key file name will be stored in the format of <app-name>.private-key.pem. Otherwise, if a user saves their private key, which is in the format of  <app-name>.<created-at-date>.private-key.pem, and then returns to editing mode, the mismatched naming in editing mode will lead to confusion and may cause a customer to think the file didn't save correctly.  

What about `$( "input[type=text][name=gitHubAppName]" ).val().replace(/\s+/g, '-').toLowerCase()` is causing confusion? 
- `( "input[type=text][name=gitHubAppName]" ).val()` targets the input field with name `gitHubAppName` and retrieves the value e.g. My New GitHub App
- `.replace(/\s+/g, '-').toLowerCase()` replaces the spaces with hyphens and converts all casing to lower case e.g. my-new-github-app (as this is how the file name will be rendered in editing mode)

src/models/github-server-app.test.ts
rachellerathbone: What do we set ORIGIN_WEBHOOK_SECRET to?

static/js/jira-manual-app-creation.js
gxueatlassian: Just my two cents, happy to have this thread close and move forward on this PR.

`from a UX standpoint, that is better that we have the file name rendered and` <-- what if, we don't display the `file name` at all, leave the file input box completely empty as new when editing existing app. User can choose to upload or leave it empty when click the save button. If user provides a new file, we override existing one in db. If leave it empty, we leave it as it is in db.

By doing above, do you think we can remove a whole bunch of code?

src/models/github-server-app.test.ts
gxueatlassian: `ORIGIN_WEBHOOK_SECRET` is a const var that we use above to create the existing app in `beforeEach`. 
But this is /nit, just for better readability on the test.

Only thing need to change is, to be a valid test, I think we should use `updatedApp.decrypt("webhookSecret")).toEqual` instead of `updatedApp.webhookSecret).not.toEqual`

src/models/github-server-app.test.ts
rachellerathbone: The test won't fail because in these tests the result **isn't** null. However, because the GHSA methods either return an app (if found) or null e.g.
```
	static async findForUuid(uuid: string): Promise<GitHubServerApp | null> {
		return this.findOne({
			where: {
				uuid
			}
		});
	}
```

we get a TS error if we don't confirm there is in fact an a

Screen Shot 2022-08-12 at 12 16 25 pm

So we can't simply remove this. I can tidy this up a little with optional chaining though.

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-put.test.ts
rachellerathbone: It's not. The installation is stored in res.locals.

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-put.ts
rachellerathbone: The error is handled in error-router `${error}` isn't rendered in prod. If this fails for customers, they will see our standard error page which is rendered in error-router.

gxueatlassian: Fair enough. I am too over concern. All good for this thread.

src/models/github-server-app.ts
```diff
@@ -169,6 +181,51 @@ export class GitHubServerApp extends EncryptedModel {
...
+			gitHubBaseUrl,
+			gitHubClientId,
+			gitHubClientSecret,
+			webhookSecret,
+			privateKey,
+			installationId
+		} = payload;
+
+		const existApp = await this.findForUuid(uuid);
+		if (existApp) {
```
mboudreau, line 214: Apparently, you can use the static functions to update a model while still running hooks, but you just specify that option:

```
await Model.update({ username: 'Tony' }, {
  where: { accessLevel: 0 },
  individualHooks: true // this one
});
```

mboudreau: And you don't need to check if it exists first, if it's not in the DB, it just won't update anything and continue.  Reducing your DB calls by half.

src/models/github-server-app.ts
```diff
@@ -169,6 +181,51 @@ export class GitHubServerApp extends EncryptedModel {
...
+	 * 		gitHubClientId: string,
+	 * 		uuid: string,
+	 * 		appId: number;
+	 * 		gitHubBaseUrl: string,
+	 * 		gitHubClientSecret: string,
+	 * 		webhookSecret: string,
+	 * 		privateKey: string,
+	 * 		gitHubAppName: string,
+	 * 		installationId: number
+	 * 	}} payload
```
mboudreau, line 197: more jsdoc :(

static/js/jira-manual-app-creation.js
rachellerathbone: Maybe but this was a UX request. I say we create a follow-up ticket and circle back on this one. I don't think it should be a blocker.

src/models/github-server-app.ts
rachellerathbone: This was something @gxueatlassian because he said there was a bug.

static/js/jira-manual-app-creation.js
gxueatlassian: Agree! closing this thread now.

src/models/github-server-app.ts
gxueatlassian: Yup I made that changes. I think the `individualHooks` works as well. I am okay either way, but if @mboudreau you insist, @rachellerathbone can you please help me change it? Sorry I need to away from keyboard now.

ps: performance wise, the individual hooks will query from db first. So should be the same.

static/css/jira-manual-app-creation.css
```diff
@@ -118,4 +118,14 @@ form#jiraManualAppCreation__form > div.field-group {
...
+}
+
+.jiraManualAppCreation__flag__title {
+  margin-bottom: 0.5em;
+}
+
+#aui-flag-container {
+  bottom: 30px;
+  left: 50px;
+  top: initial;
```
rachellerathbone, line 130: Setting top to initial to clear the default flag styling that is applied.

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-post.ts
```diff
@@ -33,13 +33,11 @@ export const JiraConnectEnterpriseAppPost = async (
...
 
 		req.log.debug("Jira Connect Enterprise App added successfully.");
 	} catch (error) {
+		res.status(200).send({ success: false, message: "Failed to create GitHub App." });
```
gxueatlassian, line 40: Not sure is there a way you can double check this? 
If we do `res.status(xxx).send(xxx)` here, will it conflict with whatever error router handler down the track when we `return next(new Error(xxx))`?

gxueatlassian (approved):

PR #1479, opened 2022-08-09
ARC-1562  support for GH Enterprise in /resync API route

Harminder84:
**What's in this PR?**
Update API route /resync to include support for both GH Cloud and Enterprise. Previously, it fetched subscription only for cloud. Now update the route to include `UUID` as path variable for Enterprise. Based on `UUID`, it decides whether to fetch subscriptions for Cloud or Enterprise.
API route for 
cloud: `/api/resync` 
Enterprise: `/api/<UUID>/resync`

**Why**
To provide support for GH Enterprise

**Affected issues**  
ARC-1562


src/routes/api/api-router.ts
```diff
@@ -17,18 +17,24 @@ import { EncryptionClient, EncryptionSecretKeyEnum } from "utils/encryption-clie
 import { ApiPingPost } from "routes/api/api-ping-post";
 import { CryptorMigrationRouter } from "./cryptor-migrations/migration-router";
 import { TaskType } from "~/src/sync/sync.types";
+import { GitHubServerApp } from "~/src/models/github-server-app";
+
+const UUID_REGEX = "[0-9a-fA-F]{8}\\b-[0-9a-fA-F]{4}\\b-[0-9a-fA-F]{4}\\b-[0-9a-fA-F]{4}\\b-[0-9a-fA-F]{12}";
```
joshkay10, line 22: Is it worth moving this to a shared file since github router is using it too?

src/routes/api/api-router.ts
```diff
@@ -100,8 +106,16 @@ ApiRouter.post(
...
 		}
-
-		const subscriptions = await Subscription.getAllFiltered(installationIds, statusTypes, offset, limit, inactiveForSeconds);
+		const { uuid } = req.params;
+		let gitHubServerApp;
+		if (uuid) {
+			gitHubServerApp = await GitHubServerApp.findForUuid(uuid);
+			if (!gitHubServerApp) {
+				req.log.error("No GitHub app found for provided uuid.");
+				throw new Error("No GitHub app found for provided id.");
```
joshkay10, line 115: This is most likely caused by an incorrect uuid? Can we return a 400 here?

gxueatlassian (approved):

PR #1483, opened 2022-08-10
Order by id when doing remigration

gxueatlassian:
**What's in this PR?**
Modify the migration process to do it in order. 

**Why**
The secrets is encrypted using clientKey, which is hashed with STORAGE_SECRET. This is will prevent us from decrypting the secret should that STROAGE_SECRET accidentally changed. 

**Added feature flags**
N/A

**Affected issues**  
ARC-1354

**How has this been tested?**  
Locally

**Whats Next?**
N/A in PR.
Will do prod migraiton.


mboudreau (approved):
lol

joshkay10 (approved):

PR #1490, opened 2022-08-11
ARC-1561 remove unused endpoints from jira client

rachellerathbone:
**What's in this PR?**
Removal of unused endpoints from jiraClient.

**Why**
Because we don't need them...

**Affected issues**  
ARC-1561

**How has this been tested?**  
Deployed to staging and no issues detected. Have triple checked the removed endpoints aren't being used anywhere.

**What's Next?**
Continue on ARC-1561: updating all calls the jiraClient that use gitHubInstallationId. Was going to do them all in one PR but figured that's a sure fire way to break something. Going to tackle each yellow highlighted area in this doc separately https://hello.atlassian.net/wiki/spaces/PF/pages/1856341692/Jira+Client

### DIFF
--- a/src/jira/client/jira-client.ts
+++ b/src/jira/client/jira-client.ts
@@ -36,6 +36,7 @@ export const getJiraClient = async (
 ): Promise<any> => {
 	const logger = log.child({ jiraHost, gitHubInstallationId });
 	const installation = await Installation.getForHost(jiraHost);
+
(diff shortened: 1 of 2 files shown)


gxueatlassian (approved):
I've do search in my vim to my best effort, can confirm these apis are not used. 
Not sure, but shall we wait for another approval?

PR #1491, opened 2022-08-11
ARC-1561: update jiraClient delete endpoint

rachellerathbone:
**What's in this PR?**
Updated calls to client.devinfo.installation.delete: see yellow highlighted section for this endpoint: https://hello.atlassian.net/wiki/spaces/PF/pages/1856341692/Jira+Client. These include:
- updated the way we are passing the installationId to the jiraClient endpoint for `JiraDelete`
- I had originally included work to remove `ApiInstallationDelete`, and associated tests, because it's not being used in the app but realised this endpoint is being called by the semantic polli check. We don't need the gitHubInstallationId to be unique for this polli check so doing this work right now isn't critical. Instead of tackling this now, I have created a `TECH DEBT` follow-up [ticket](https://softwareteams.atlassian.net/jira/software/projects/ARC/boards/385/backlog?selectedIssue=ARC-1585) that will need to be addressed 

**Why**
Part of the gitHubInstallationId work

**Affected issues**  
ARC-1561

**How has this been tested?**  
Locally and in staging.

**What's Next?**
Update calls to client.devinfo.repository.update


rachellerathbone:
Consolidating to one pr

PR #1492, opened 2022-08-11
ARC-1561 updating client.devinfo.repository.update

rachellerathbone:
What's in this PR?
updating client.devinfo.repository.update. See yellow highlighted section for this endpoint. https://hello.atlassian.net/wiki/spaces/PF/pages/1856341692/Jira+Client

Why
Part of the gitHubInstallationId work

Affected issues
ARC-1561

How has this been tested?
Locally and in staging.

What's Next?
Update calls to client.workflow.submit


src/jira/client/jira-client.ts
```diff
@@ -223,8 +223,14 @@ export const getJiraClient = async (
 							repositoryId
 						}
 					}),
-				update: async (data, options?: JiraSubmitOptions) => {
+				update: async (data, gitHubAppId?: number, options?: JiraSubmitOptions) => {
 					dedupIssueKeys(data);
+					const subscription = await Subscription.getSingleInstallation(
```
gxueatlassian, line 228: Am I missing something? Why moving the trunk (`await Subscription.getSingleInstallation()` and `await subscription?.update` ) out of the `if{` ?

src/jira/client/jira-client.ts
```diff
@@ -237,17 +243,12 @@ export const getJiraClient = async (
...
-							gitHubInstallationId
-						);
-						await subscription?.update({ syncWarning: issueKeyLimitWarning });
 					}
 
 					return batchedBulkUpdate(
 						data,
 						instance,
-						gitHubInstallationId,
+						subscription?.gitHubInstallationId || gitHubInstallationId,
```
gxueatlassian, line 251: Can you please explain why `subscription?.gitHubInstallationId || gitHubInstallationId,`?

src/sync/installation.ts
```diff
@@ -245,6 +246,8 @@ const doProcessInstallation = async (data: BackfillMessagePayload, sentry: Hub,
 		throw new Error(`Error processing task: installationId=${installationId}, repositoryId=${nextTask.repositoryId}, task=${task}`);
 	};
 
+	const gitHubAppConfig = await getGitHubAppConfig(subscription, logger);
```
gxueatlassian, line 249: I know the code is a bit tricky here. So it seems this `getNextTask` is called only in this file, and the caller position all contains the `BackfillMessagePayload`, which in turns contains the `gitHubAppConfig?` ( as they are msg in the queue). 

So maybe just grab it, and avoid one db call?

src/jira/client/jira-client.ts
rachellerathbone: `getSingleInstallation` either returns an entry or `null`. In the case where a subscription is found, we want to set the installationId in the payload to the subscription's gitHubInstallationId. However, because it possibly returns `null`, if we don't have a default (`|| gitHubInstallationId`) TS complains:

Screen Shot 2022-08-12 at 3 37 59 pm

src/jira/client/jira-client.ts
rachellerathbone: `await Subscription.getSingleInstallation` needs to move so that the subscription is defined outside of the conditional and therefore accessible as an arg for `batchedBulkUpdate`. But the call to subscription.update didn't need to move. Returned that to the conditional.

src/jira/client/jira-client.ts
gxueatlassian: `In the case where a subscription is found, we want to set the installationId in the payload to the subscription's gitHubInstallationId` <-- which will be the exact value of `gitHubInstallationId`? So the origin code 
```
return batchedBulkUpdate(
						data,
						instance,
						gitHubInstallationId,
```
will still work?

rachellerathbone:
Changing to identify further upstream

src/sync/sync-utils.ts
```diff
@@ -60,7 +60,7 @@ export const getCommitSinceDate = async (jiraHost: string, flagName: NumberFlags
 	return new Date(Date.now() - timeCutoffMsecs);
 };
 
-const getGitHubAppConfig = async (subscription: Subscription, logger: Logger): Promise<GitHubAppConfig> => {
+export const getGitHubAppConfig = async (subscription: Subscription, logger: Logger): Promise<GitHubAppConfig> => {
```
gxueatlassian, line 63: why export if not used?

gxueatlassian (dismissed):
Just one minor comments.

src/sync/sync-utils.ts
rachellerathbone: Forgot to remove that.

gxueatlassian (approved):

PR #1497, opened 2022-08-12
ARC-1571: Add encrypt and decrypt on cryptor check

gxueatlassian:
**What's in this PR?**
1. Add an encrypt/decrypt check for Worker group in healthcheck. 
2. Add encrypt/decrypt check for WebServer in deepcheck.

**Why**
It is complicated, please refer to the ticket ARC-1571 itself to see a more detail doc page.

**Added feature flags**
N/A

**Affected issues**  
https://softwareteams.atlassian.net/browse/ARC-1571

**How has this been tested?**  
Stage (WIP)

**Whats Next?**
N/A

### DIFF
--- a/src/routes/healthcheck/healthcheck-get.ts
+++ b/src/routes/healthcheck/healthcheck-get.ts
@@ -10,7 +10,7 @@ export const HealthcheckGet = async (_: Request, res: Response): Promise<void> =
 		try {
 			await deepCheckCryptor();
 		} catch (e) {
-			logger.warn("Cryptor not ready yet", { error: e });
+			logger.warn("Cryptor deepcheck failed.", { error: e });
 			res.status(500).send("NOT OK");
 			return;
 		}
--- a/src/routes/healthcheck/healthcheck-get.ts
+++ b/src/routes/healthcheck/healthcheck-get.ts
@@ -1,8 +1,22 @@
 import { Request, Response } from "express";
 import { getLogger } from "config/logger";
+import { envVars } from "config/env";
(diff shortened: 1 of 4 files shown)


src/routes/healthcheck/healthcheck-get.ts
```diff
@@ -1,8 +1,22 @@
...
 
 const logger = getLogger("healthcheck");
-export const HealthcheckGet = (_: Request, res: Response): void => {
+export const HealthcheckGet = async (_: Request, res: Response): Promise<void> => {
+
+	if (envVars.MICROS_GROUP === "Worker") {
+		try {
+			await deepCheckCryptor();
+		} catch (e) {
+			logger.warn("Cryptor not ready yet", { error: e });
```
joshkay10, line 13: `Cryptor not ready yet` could be confusing if its a legit Cryptor error, once it is ready.

gxueatlassian: I can't think of any way to distinguish whether it is just not ready or it is ready but has error. I can do a different text, like `Cryptor not ready yet or has error`? You reckon that's okay?

rachellerathbone: Can't we just use something more generic like `Cryptor deepcheck failed`? That plus the error should give enough info without being misleading.

gxueatlassian: thanks team, fixed per @rachellerathbone suggestion.

rachellerathbone (approved):

joshkay10 (approved):

PR #1498, opened 2022-08-12
add stg-west to sd.yml

gxueatlassian:
add new stg-east


joshkay10 (commented):
Is this environment going to run through pollinator too?

PR #1499, opened 2022-08-15
Ghe stack

thombergs:
This PR adds a CloudFormation template to the codebase that we can use to spin up GitHub Enterprise servers on-demand for testing purposes.


mboudreau (approved):
I'll approve, but I'm not certain this should go in this repo.  It's related, but it could be it's own repo really...

bgvozdev:
Here we are exposing some Atlassian-related info, like, our team's e-mail and the surname of the owner of the component... Do we really want this?

PR #1500, opened 2022-08-15
update error mapping

rachellerathbone:
**What's in this PR?**
Update error mapping when trying to make a request to a GHE server.

**Why**
Because we found a new error that we haven't yet accounted for.

**Added feature flags**
Al behind GHE_SERVER

**Affected issues**  
None, just following up on a slack thread.

**How has this been tested?**  
Locally.

**What's Next?**
Nap time? I'm tiiiiiiiired.

### DIFF
--- a/src/routes/jira/connect/enterprise/jira-connect-enterprise-post.ts
+++ b/src/routes/jira/connect/enterprise/jira-connect-enterprise-post.ts
@@ -27,6 +27,7 @@ export const gheServerUrlErrors: GheServerUrlErrors = {
 		message: "Invalid URL"
 	},
 	ENOTFOUND: GHE_ERROR_UNREACHABLE,
+	DEPTH_ZERO_SELF_SIGNED_CERT: GHE_ERROR_UNREACHABLE,
 	403: GHE_ERROR_UNREACHABLE,
 	502: {
 		errorCode: "GHE_SERVER_BAD_GATEWAY",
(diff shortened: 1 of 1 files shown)


krazziekay (approved):

src/routes/jira/connect/enterprise/jira-connect-enterprise-post.ts
```diff
@@ -27,6 +27,7 @@ export const gheServerUrlErrors: GheServerUrlErrors = {
 		message: "Invalid URL"
 	},
 	ENOTFOUND: GHE_ERROR_UNREACHABLE,
+	DEPTH_ZERO_SELF_SIGNED_CERT: GHE_ERROR_UNREACHABLE,
```
bgvozdev, line 30: Will it just say that the server is unreachable? Shouldn't we be more specific? (Or the message would contain the code itself that might help the customer?)

rachellerathbone: This isn't what we render in the UI. We're rendering a more specific message with a link to documentation. The backend isn't responsible for what we surface users in the client. Only for sending the error.

bgvozdev: Ah right, sorry, I didn't check where the change was made 🤦

bgvozdev (approved):

PR #1501, opened 2022-08-15
Adding ability to change log level per jira host

mboudreau:
**What's in this PR?**
Adding the ability to change the log level per jirahost for req.log and context.log (sync, web handler and webhooks)

**Why**
Because there's no other way of getting debug logs for a single customer

**Added feature flags**
`StringFlags.LOG_LEVEL`

**Affected issues**  
None

**How has this been tested?**  
Added unit tests and manual

**Whats Next?**
Turn it on for key customers where we need more information.


src/config/logger.ts
```diff
@@ -77,44 +76,53 @@ export const overrideProbotLoggingMethods = (probotLogger: Logger) => {
...
+	return options.fields ? logger.child({ ...options.fields }) : logger;
 };
 
 // This will log data to a restricted environment [env]-unsafe and not serialize sensitive data
-export const getUnsafeLogger = (name: string, fields?: Record<string, unknown>): Logger => {
-	const logger = createNewLogger(name, { env_suffix: "unsafe" });
-	return logger.child({ ...fields });
+export const getUnsafeLogger = (name: string, options: LoggerOptions = {}): Logger => {
+	const logger = createNewLogger(name, options);
+	return options.fields ? logger.child({ ...options.fields }) : logger;
```
joshkay10, line 125: Only thing I wonder is if when we search in the Splunk if it will be able to search on the level down.
e.g. if i splunk jiraHost="banana" is it going to search the `fields` attribute or do I need to explicitly search fields attribute for the jiraHost?🤷‍♂️

joshkay10, line 125: Only thing I wonder is if when we search in the Splunk if it will be able to search on the level down.
e.g. if i splunk jiraHost="banana" is it going to search the fields attribute or do I need to explicitly search fields attribute for the jiraHost?🤷‍♂️

src/middleware/github-webhook-middleware.ts
```diff
@@ -99,14 +99,19 @@ export const GithubWebhookMiddleware = (
...
-			...cloneAllowedLogFields(context.log.fields)
+			level: await stringFlag(StringFlags.LOG_LEVEL, defaultLogLevel, jiraHost),
+			fields: {
+				webhookId,
+				gitHubInstallationId,
+				event: webhookEvent,
+				webhookReceived,
+				repoName,
+				orgName,
+				...cloneAllowedLogFields(context.log.fields)
```
joshkay10, line 113: cloneAllowedLogFields probably isn't required anymore, since `name` is a top level restricted attribute and we are inside fields..... however name and then field.name both existing could be confusing.

src/config/logger.ts
```diff
@@ -77,44 +76,53 @@ export const overrideProbotLoggingMethods = (probotLogger: Logger) => {
 		type: "stream",
 		stream: LOG_STREAM,
 		closeOnExit: false,
-		level: globalLoggingLevel
+		level: defaultLogLevel
```
gxueatlassian, line 79: The `globalLoggingLevel` is a `number`? But the `defaultLogLevel` is string "info" by default. Will it be a problem?

joshkay10 (approved):

src/config/logger.ts
```diff
@@ -77,44 +76,53 @@ export const overrideProbotLoggingMethods = (probotLogger: Logger) => {
...
 };
 
-const createNewLogger = (name: string, options: Partial<LoggerOptions> = {}): Logger => {
-	return createLogger({
+const createNewLogger = (name: string, options: LoggerOptions = {}): Logger => {
+	return createLogger(merge<Logger.LoggerOptions, LoggerOptions>({
 		name,
 		stream: LOG_STREAM,
-		level: globalLoggingLevel,
+		level: defaultLogLevel,
```
gxueatlassian, line 87: So if I do `const logger = createNewLogger('name', { level: INFO }); logger.debug('xxx')`
What will be actual output?

src/util/match-route-with-pattern.ts
```diff
@@ -8,9 +9,12 @@ import matchstick from "matchstick";
  *
  */
 export const matchRouteWithPattern = (pattern: string, route: string): boolean => {
-	pattern = pattern.replace(/ac\./gm, ""); // Remove all the `ac.`
+	if (isEmpty(pattern) || isEmpty(route)) {
+		return false;
+	}
+	pattern = pattern.replace("ac.", ""); // Remove all the `ac.`
```
gxueatlassian, line 15: **BLOCKING**
Somehow in chrome console it says otherwise. And we do have some line with multiple "ac." `url:"/jira/connect/enterprise/{ac.serverUrl}/app?new={ac.new}",`
image



Maybe revert this change ( as it is not part of the PR)? Or try to test it more.

gxueatlassian (changes requested):

src/middleware/github-webhook-middleware.ts
mboudreau: I'll leave it as is for now :)

src/config/logger.ts
mboudreau: No.

src/config/logger.ts
mboudreau: nothing as the level gets overriden with your option.

src/util/match-route-with-pattern.ts
mboudreau: Huh, the tests did pass but I guess this scenario wasn't tested.  Added said scenario.

gxueatlassian (approved):

src/middleware/github-webhook-middleware.ts
mboudreau: okay, removed it.

src/config/logger.ts
mboudreau: yeah, sorry, that was old code I didn't removed.  Fixed.

joshkay10 (approved):

gxueatlassian (dismissed):

joshkay10 (dismissed):

gxueatlassian (approved):

gxueatlassian (approved):

PR #1505, opened 2022-08-16
double check github app id

gxueatlassian:
**What's in this PR?**
Make most of the `gitHubAppId?: number` to `gitHubAppId: number | undefined`, so that we are pretty sure that the code is ready.

**Why**

**Added feature flags**

**Affected issues**  
_Jira Issues_

**How has this been tested?**  

**Whats Next?**

### DIFF
--- a/src/github/client/key-locator.ts
+++ b/src/github/client/key-locator.ts
@@ -7,7 +7,7 @@ import isBase64 from "is-base64";
 /**
  * Look for a Github app's private key
  */
-export const keyLocator = async (gitHubAppId?: number) => {
+export const keyLocator = async (gitHubAppId: number | undefined) => {
 	if (gitHubAppId) {
 		const gitHubServerApp = await GitHubServerApp.getForGitHubServerAppId(gitHubAppId);
 		if (gitHubServerApp) {
--- a/src/models/subscription.ts
+++ b/src/models/subscription.ts
@@ -39,7 +39,7 @@ export class Subscription extends Model {
 	numberOfSyncedRepos?: number;
 	repositoryCursor?: string;
 	repositoryStatus?: TaskStatus;
-	gitHubAppId?: number;
+	gitHubAppId: number | undefined;
 
 	static async getAllForHost(jiraHost: string): Promise<Subscription[]> {
 		return this.findAll({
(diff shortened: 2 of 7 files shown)


src/github/client/key-locator.ts
```diff
@@ -7,7 +7,7 @@ import isBase64 from "is-base64";
 /**
  * Look for a Github app's private key
  */
-export const keyLocator = async (gitHubAppId?: number) => {
+export const keyLocator = async (gitHubAppId: number | undefined) => {
```
mboudreau, line 10: in TS, `gitHubAppId?: number` is the same type as `gitHubAppId: number | undefined`, the `?` is just syntactic sugar and makes it more concise IMO.  The former is the standard we want to use to reduce boilerplate.

gxueatlassian: So for example, in bellow screenshot, if I use `gitHubAppId: number | undefined`, it will prompt me a TS error
image

But `gitHubAppId?: number` does not.
image


So my goal is to ensure, our code doesn't forget to provide the `gitHubAppId`, even it is undefined. Otherwise, we maybe forgot to modify somewhere in our code to supply to `gitHubAppId` and still pass the TS checking, which will mix cloud data with GHE data.

gxueatlassian: It is related to recent series of PR that try to fix the `gitHubInstallationId` is not unique in GHE, I am a bit concern somewhere in the code we forgot to add the `gitHubAppId`.

rachellerathbone: Are you saying we should make gitHubAppId a required arg instead of optional? If that's the case, we would need to update every method in Subscription that has the gitHubAppId as a param and also update every call to each of those methods. From what I can see, getAllFiltered, getSingleInstallation, findOneForGitHubInstallationId, and getAllForInstallation would need to be updated and it would be best to break this out into 3 small prs to also address updating where these methods are being called:

- Subscription.getAllForInstallation is called in 7 places: should be it's own PR
- Subscription.getSingleInstallation is called in close to 20 places: should be it's own PR
- Subscription.getAllFiltered and Subscription.findOneForGitHubInstallationId are both only called in 3 places so could be tackled together.

mboudreau (approved):

gxueatlassian:
Close this one, prefer another PR https://github.com/atlassian/github-for-jira/pull/1530/files

PR #1511, opened 2022-08-17
Arc 1588 make GitHub app config to discovery

gxueatlassian:
**What's in this PR?**
Add gitHubAppId to discovery task

**Why**
gitHubInstallationId is not unique

**Added feature flags**
N/A

**Affected issues**  
ARC-1588

**How has this been tested?**  
unit test

**Whats Next?**
N/A


src/sync/installation.test.ts
```diff
@@ -90,9 +105,16 @@ describe("sync/installation", () => {
 
 	describe("processInstallation", () => {
 
-		it("should process the installation with deduplication", async () => {
+		it("should process the installation with deduplication for cloud", async () => {
 			await processInstallation()(JOB_DATA, sentry, TEST_LOGGER);
 			expect(mockedExecuteWithDeduplication.mock.calls.length).toBe(1);
+			expect(mockedExecuteWithDeduplication).toBeCalledWith(`i-1-http://foo-ghaid-undefined`, expect.anything());
```
krazziekay, line 111: Do we want this to be showing `undefined` like this?

src/sync/installation.ts
```diff
@@ -429,7 +431,7 @@ export const processInstallation = () => {
 			const nextTaskDelaysMs: Array<number> = [];
 
 			const result = await deduplicator.executeWithDeduplication(
-				"i-" + installationId + "-" + jiraHost,
+				`i-${installationId}-${jiraHost}-ghaid-${gitHubAppId}`,
```
krazziekay, line 434: Instead of showing up `undefined`, can we add a condition for cloud and server, where the previous value is set for cloud and this new value is set for server?

rachellerathbone: I agree. I don't think we should just have undefined here.

gxueatlassian: I am fine as set for `cloud`, but it won't work to set for `server`, because this id is used to do the deduplication I believe. If two servers happen to have same `installationId`, then it will show `i-${installationId}-${jiraHost}-ghaid-server` and conflict.

Tbh as the id is only used in the key inside redis, so `undefined` technically works, but agree it doesn't looks nice. I can change that.

gxueatlassian: fixed.

krazziekay (approved):

PR #1516, opened 2022-08-17
ARC-1592 fix webhooks

rachellerathbone:
**What's in this PR?**
Change to github server app middleware so webhooks will work.

**Why**
The jiraHost doesn't appear to be available in res.locals here but we don't need it. Can confirm we have the correct app/installation without it.

**Added feature flags**
GHE_SERVER

**Affected issues**  
ARC-1592

**How has this been tested?**  
Locally and in staging.

**What's Next?**
More bug fixes for GHES

### DIFF
--- a/src/middleware/github-server-app-middleware.test.ts
+++ b/src/middleware/github-server-app-middleware.test.ts
@@ -24,9 +24,7 @@ describe("github-server-app-middleware", () => {
 	beforeEach(async () => {
 		next = jest.fn();
 		res = {
-			locals: {
-				jiraHost: "https://testatlassian.net"
-			}
+			locals: {}
 		};
 
 		req = {
@@ -48,7 +46,7 @@ describe("github-server-app-middleware", () => {
 			.toThrow("No GitHub app found for provided id.");
 	});
 
-	it("should throw an error if an uuid is provided and a GitHub server app is found but the installation id doesn't match", async () => {
+	it("should throw an error if no installation is found for GitHub Server App installation ID", async () => {
 		req.params.uuid = UUID;
 
 		payload = {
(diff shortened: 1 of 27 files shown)


src/middleware/github-server-app-middleware.test.ts
```diff
@@ -59,24 +57,25 @@ describe("github-server-app-middleware", () => {
...
 			webhookSecret: "mywebhooksecret",
 			privateKey: "myprivatekey",
-			installationId: JIRA_INSTALLATION_ID
+			installationId: JIRA_INSTALLATION_ID,
+			decrypt: async (s: any) => s
 		};
 
 		installation = {
-			jiraHost: "https://testatlassian.com",
+			jiraHost,
```
rachellerathbone, line 65: We don't need to manually set the jiraHost. It's available in the global variables defined in setup.

src/middleware/github-server-app-middleware.ts
```diff
@@ -39,10 +42,10 @@ export const GithubServerAppMiddleware = async (req: Request, res: Response, nex
 		};
 
 	} else {
-		// is cloud app
+		req.log.debug("Received request for GitHub Cloud");
```
rachellerathbone, line 45: Replaced the comment with a log as this helps more with local development.

src/middleware/github-server-app-middleware.ts
```diff
@@ -12,20 +12,24 @@ export const GithubServerAppMiddleware = async (req: Request, res: Response, nex
...
+
 		const installation = await Installation.findByPk(gitHubServerApp.installationId);
+
 		if (installation?.jiraHost !== jiraHost) {
 			req.log.error({ uuid, jiraHost }, "Jira hosts do not match");
 			throw new Error("Jira hosts do not match.");
 		}
-		req.log.info("Found GitHub server app for installation");
+
+		req.log.info("Found server app for installation. Defining GitHub app config for GitHub Enterprise Server.");
```
rachellerathbone, line 31: Made this more explicit.

src/middleware/github-server-app-middleware.ts
```diff
@@ -37,9 +41,8 @@ export const GithubServerAppMiddleware = async (req: Request, res: Response, nex
 			webhookSecret: await gitHubServerApp.decrypt("webhookSecret"),
 			privateKey: await gitHubServerApp.decrypt("privateKey")
 		};
-
 	} else {
-		// is cloud app
+		req.log.info("Defining GitHub app config for GitHub Cloud.");
```
rachellerathbone, line 45: Replaced comment with more explicit debug log.

src/routes/github/github-router.ts
```diff
@@ -15,19 +15,19 @@ export const GithubRouter = Router();
 const subRouter = Router({ mergeParams: true });
 GithubRouter.use(`/:uuid(${UUID_REGEX})?`, subRouter);
 
+// Webhook Route
```
rachellerathbone, line 18: Just needed to move this bad boy up so webhook events don't hit the ghes app middleware.

krazziekay (approved):

PR #1518, opened 2022-08-18
ARC-1354 Supress error logging during healthcheck for Worker

gxueatlassian:
**What's in this PR?**
Move the error logging from within EncryptionClient to EncryptedModel

**Why**
Don't want to get too much noise during worker starting up and when cryptor is not ready. It is okay because:
1. If it stays error, we will found out, the cicd pipeline will be red.
2. If it is only transient, it will turns green when the cryptor sidecar catchup and started, then we shouldn't care about this error. So warn instead.

**Added feature flags**
N/A

**Affected issues**  
ARC-1354

**How has this been tested?**  
N/A

**Whats Next?**
N/A

### DIFF
--- a/src/models/encrypted-model.ts
+++ b/src/models/encrypted-model.ts
@@ -1,5 +1,6 @@
 import { Model } from "sequelize";
 import { EncryptionClient, EncryptionSecretKeyEnum } from "utils/encryption-client";
+import { getLogger } from "config/logger";
 
 type StringValues<Obj> = {
 	[Prop in keyof Obj]: Obj[Prop] extends string ? Prop : never
@@ -18,7 +19,12 @@ export abstract class EncryptedModel extends Model {
 		if (typeof value !== "string") {
 			throw new Error(`Cannot decrypt '${field}', it is not a string.`);
 		}
-		return await EncryptionClient.decrypt(value, await this.getEncryptContext(field));
+		try {
+			return await EncryptionClient.decrypt(value, await this.getEncryptContext(field));
+		} catch (e) {
+			getLogger("cryptor").error(`Fail to decrypt field ${field}`, { error: e });
+			throw e;
+		}
 	}
 
 	protected async encrypt(field: (keyof StringValues<this>)): Promise<string> {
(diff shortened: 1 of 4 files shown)


src/routes/healthcheck/healthcheck-get.ts
```diff
@@ -10,7 +10,7 @@ export const HealthcheckGet = async (_: Request, res: Response): Promise<void> =
 		try {
 			await deepCheckCryptor();
 		} catch (e) {
-			logger.warn("Cryptor deepcheck failed.", { error: e });
+			logger.warn("Cryptor deepcheck failed in work. This happen during startup when Worker started by Cryptor sidecar is not ready.", { error: e });
```
mboudreau, line 13: I'd revert back this message, I'm not sure it's helpful and there is a possibility for the sidecar to fail in production.  In which case, we don't want to give the illusion that it's just a temporary issue

src/models/encrypted-model.ts
```diff
@@ -18,15 +19,25 @@ export abstract class EncryptedModel extends Model {
...
 	protected async encrypt(field: (keyof StringValues<this>)): Promise<string> {
 		const value = this[field];
 		if (typeof value !== "string") {
 			throw new Error(`Cannot encrypt '${field}', it is not a string.`);
 		}
-		return await EncryptionClient.encrypt(this.getEncryptionSecretKey(), value, await this.getEncryptContext(field));
+		try {
+			return await EncryptionClient.encrypt(this.getEncryptionSecretKey(), value, await this.getEncryptContext(field));
+		} catch (e) {
+			getLogger("cryptor").error(`Fail to encrypt field ${field}`, { error: e });
```
mboudreau, line 38: set the logger as a constant, no need to keep recreating it...

src/routes/healthcheck/healthcheck-get.ts
gxueatlassian: I get your concerning, but here's the reason:
1. This is line is within the `envVars.MICROS_GROUP === "Worker"`, so only `Worker` hit this line.
2. And according the `micros` settings (let me know I can further point you to the doc), `healtCheck` within `Worker` is only hit once during startup, unless we config to periodically hit worker healthcheck (which we did not).

So in short, this does only happen during worker startup.

src/models/encrypted-model.ts
gxueatlassian: fixed.

src/routes/healthcheck/healthcheck-get.ts
mboudreau: Then what's the benefit of updating the message?  If you don't care about it, just remove the warning?

gxueatlassian: The benefit is I think the updated message is more clear than existing one. If anyone stumble upon this warning message, it will make it clear to them.

Not sure about removing the warning. Was thinking leave it for now since it is not `error`, not a super noise to me yet (and still maybe helpful to diagnose startup)

-------------

So I am like 6 out of 10 prefer to leave the updated message as it is. But if you are feeling stronger than a `6`, let me know, I am happy to either remove the warning message or revert the message.

mboudreau (approved):

PR #1521, opened 2022-08-18
ARC 1605 - Endpoint GitHubConfigurationGet removes the previous GH cloudconnection 

krazziekay:
**What's in this PR?**
- Added `gitHubAppId` to fetch the correct subscriptions in `github-configuration-get`
- Removed multiple DB calls.

**Why**
- The previous usage of `Subscription.getAllForHost` was pulling both cloud and server apps. This caused the cloud apps to be failed installations because they don't exist in the GHES. So, it was being removed from the DB as failed installations are all being removed. 

**How has this been tested?**  
- In local and ddev environment

### DIFF
--- a/src/models/subscription.ts
+++ b/src/models/subscription.ts
@@ -41,9 +41,10 @@ export class Subscription extends Model {
 	repositoryStatus?: TaskStatus;
 	gitHubAppId?: number;
 
-	static async getAllForHost(jiraHost: string): Promise<Subscription[]> {
+	static async getAllForHost(jiraHost: string, gitHubAppId?: number): Promise<Subscription[]> {
 		return this.findAll({
 			where: {
+				...(gitHubAppId && { gitHubAppId }), // Add gitHubAppId only if passed
 				jiraHost
 			}
 		});
--- a/src/routes/github/configuration/github-configuration-get.ts
+++ b/src/routes/github/configuration/github-configuration-get.ts
@@ -56,7 +56,7 @@ const installationConnectedStatus = async (
 	reqLog: Logger,
 	gitHubAppId?: number
 ): Promise<MergedInstallation[]> => {
-	const subscriptions = await Subscription.getAllForHost(jiraHost);
+	const subscriptions = await Subscription.getAllForHost(jiraHost, gitHubAppId);
 	const installationsWithSubscriptions = await getInstallations(subscriptions, reqLog, gitHubAppId);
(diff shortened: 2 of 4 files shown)


src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -56,8 +56,11 @@ const installationConnectedStatus = async (
 	reqLog: Logger,
 	gitHubAppId?: number
 ): Promise<MergedInstallation[]> => {
-	const subscriptions = await Subscription.getAllForHost(jiraHost);
+	const subscriptions = await Subscription.getAllForHost(jiraHost, gitHubAppId);
 	const installationsWithSubscriptions = await getInstallations(subscriptions, reqLog, gitHubAppId);
+	await removeFailedConnectionsFromDb(reqLog, installationsWithSubscriptions, jiraHost);
+	reqLog.debug(`Removed failed installations`);
```
rachellerathbone, line 62: Should probably just call this logger as this is how it's typically approached in the code when passing req.log

rachellerathbone: Nit: don't need backticks here as we're not passing any values in.

src/routes/github/configuration/github-configuration-router.test.ts
```diff
@@ -125,22 +125,6 @@ describe("Github Configuration", () => {
 
 			githubUserTokenNock(sub.gitHubInstallationId);
 
-			githubNock
```
rachellerathbone, line 128: Why have we simply removed these tests?

krazziekay: Reverted back and just removed `twice`. These calls are not made twice anymore, cause of the removal of the duplicated lines in `github-configuration-get`

src/models/subscription.ts
```diff
@@ -41,9 +41,10 @@ export class Subscription extends Model {
 	repositoryStatus?: TaskStatus;
 	gitHubAppId?: number;
 
-	static async getAllForHost(jiraHost: string): Promise<Subscription[]> {
+	static async getAllForHost(jiraHost: string, gitHubAppId?: number): Promise<Subscription[]> {
```
gxueatlassian, line 44: Hey @krazziekay , can you please change it to `gitHubAppId: number | undefined`?

krazziekay: `getAllForHost` is already being used in a lot of places and if we define it like this, then the second parameter will be required. Thats why I made the second params as optional...

gxueatlassian (approved):

PR #1522, opened 2022-08-18
Bypassing the jiraHost check if provided value is same as session

gxueatlassian:
**What's in this PR?**
Bypassing the jiraHost check if provided value is same as session

**Why**
Currently in prod, we can't delete a subscription as the route is `/github/subscription` and contains `jiraHost` in body. Therefore code try to verify it with the jwt token, but jwt token is missing on that subscription managing page.

So instead, I decided to bypass that check if the jiraHost value in request body is same as the one in the session cookie, coz we ALWAYS TRUST session cookie.

**Added feature flags**
N/A

**Affected issues**  
ARC-1614

**How has this been tested?**  
Unit test.
Stage [WIP]

**Whats Next?**
N/A

### DIFF
--- a/src/middleware/jirahost-middleware.ts
+++ b/src/middleware/jirahost-middleware.ts
@@ -50,7 +50,7 @@ export const jirahostMiddleware = async (req: Request, res: Response, next: Next
 	const takenFromCookies = unsafeJiraHost === req.cookies.jiraHost;
 	res.clearCookie("jiraHost");
 
-	if (unsafeJiraHost && isDifferentFromSession(unsafeJiraHost, req.session.jiraHost)) {
+	if (unsafeJiraHost && unsafeJiraHost !== req.session.jiraHost) {
 		// Even though it is unsafe, we are verifying it straight away below (in "verifyJwtBlahBlah" call)
 		res.locals.jiraHost = unsafeJiraHost;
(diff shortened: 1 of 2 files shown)


src/middleware/jirahost-middleware.ts
```diff
@@ -43,15 +43,14 @@ const detectJwtTokenType = (req: Request): TokenType => {
...
-
 	req.addLogFields({ jiraHost: unsafeJiraHost });
 
 	// JWT validation makes sure "res.locals.jiraHost" is legit, not the cookie value. To avoid
 	// any temptation to use it later, let's remove it straight away!
 	const takenFromCookies = unsafeJiraHost === req.cookies.jiraHost;
 	res.clearCookie("jiraHost");
 
-	if (unsafeJiraHost) {
+	if (unsafeJiraHost && isDifferentFromSession(unsafeJiraHost, req.session.jiraHost)) {
```
mboudreau, line 53: you don't need a separate function just do do `unsafeJiraHost !== res.session.jiraHost`

gxueatlassian: true

mboudreau: Also, I'm fairly certain we want to verify the JWT token on every request, even if the jirahost in the session hasn't changed because the token might have

rachellerathbone (approved):

PR #1524, opened 2022-08-18
Arc-1596 fix manage subscriptions for GHES

rachellerathbone:
**What's in this PR?**
Fixing navigation to gitub subscriptions for GHES

**Why**
Because we previously saw this 🙃 
Screen Shot 2022-08-17 at 10 00 54 am

**Added feature flags**
GHE_SERVER

**Affected issues**  
ARC-1596

**How has this been tested?**  
Locally and in dev with flag off, and with flag on for cloud & server.

**What's Next?**
Bug hunting

### DIFF
--- a/src/models/github-server-app.ts
+++ b/src/models/github-server-app.ts
@@ -175,7 +175,7 @@ export class GitHubServerApp extends EncryptedModel {
 
 	}
 
-	static async findForUuid(uuid: string): Promise<GitHubServerApp | null> {
+	static async findForUuid(uuid: string | undefined): Promise<GitHubServerApp | null> {
 		if (!uuid) return null;
 
 		return this.findOne({
--- a/src/models/github-server-app.ts
+++ b/src/models/github-server-app.ts
@@ -175,9 +175,7 @@ export class GitHubServerApp extends EncryptedModel {
 
 	}
 
-	static async findForUuid(uuid: string | undefined): Promise<GitHubServerApp | null> {
-		if (!uuid) return null;
-
+	static async findForUuid(uuid: string): Promise<GitHubServerApp | null> {
 		return this.findOne({
 			where: {
 				uuid
--- a/src/models/github-server-app.ts
+++ b/src/models/github-server-app.ts
@@ -175,7 +175,7 @@ export class GitHubServerApp extends EncryptedModel {
 
 	}
 
-	static async findForUuid(uuid: string | undefined): Promise<GitHubServerApp | null> {
+	static async findForUuid(uuid: string): Promise<GitHubServerApp | null> {
 		if (!uuid) return null;
 
 		return this.findOne({
--- a/src/models/github-server-app.ts
+++ b/src/models/github-server-app.ts
@@ -175,7 +175,9 @@ export class GitHubServerApp extends EncryptedModel {
 
 	}
 
-	static async findForUuid(uuid: string): Promise<GitHubServerApp | null> {
+	static async findForUuid(uuid: string | undefined): Promise<GitHubServerApp | null> {
+		if (!uuid) return null;
+
 		return this.findOne({
 			where: {
 				uuid
--- a/src/routes/github/github-oauth-router.ts
+++ b/src/routes/github/github-oauth-router.ts
@@ -128,7 +128,7 @@ export const GithubAuthMiddleware = async (req: Request, res: Response, next: Ne
 	try {
 		const { githubToken, gitHubUuid } = req.session;
 		const { jiraHost, gitHubAppConfig } = res.locals;
-		const gitHubAppId = res.locals.gitHubAppId || gitHubAppConfig.id;
+		const gitHubAppId = res.locals.gitHubAppId || gitHubAppConfig.gitHubAppId;
 
 		/**
 		 * Comparing the `UUID` saved in the session with the `UUID` inside `gitHubAppConfig`,
--- a/src/routes/github/github-oauth-router.ts
+++ b/src/routes/github/github-oauth-router.ts
@@ -8,6 +8,7 @@ import { GithubAPI } from "config/github-api";
 import { Errors } from "config/errors";
 import { getGitHubApiUrl } from "~/src/util/get-github-client-config";
 import { createHashWithSharedSecret } from "utils/encryption";
+import { GitHubServerApp } from "~/src/models/github-server-app";
 
 const logger = getLogger("github-oauth");
 const appUrl = envVars.APP_URL;
@@ -79,7 +80,7 @@ const GithubOAuthCallbackGet = async (req: Request, res: Response, next: NextFun
 	if (!code) return next("Missing OAuth Code");
 
 	const { jiraHost, gitHubAppConfig } = res.locals;
-	const { hostname, clientId, gitHubClientSecret, uuid } = gitHubAppConfig;
+	const { hostname, clientId, gitHubClientSecret } = gitHubAppConfig;
 	req.log.info({ jiraHost }, "Jira Host attempting to auth with GitHub");
 	req.log.debug(`extracted jiraHost from redirect url: ${jiraHost}`);
 
@@ -106,9 +107,6 @@ const GithubOAuthCallbackGet = async (req: Request, res: Response, next: NextFun
 		// Saving it to session be used later
 		req.session.githubToken = response.data.access_token;
 
-		// Saving UUID for each GitHubServerApp
-		req.session.gitHubUuid = uuid;
-
 		if (!req.session.githubToken) {
 			req.log.debug(`didn't get access token from GitHub`);
 			return next(new Error("Missing Access Token from Github OAuth Flow."));
@@ -127,7 +125,9 @@ const GithubOAuthCallbackGet = async (req: Request, res: Response, next: NextFun
 export const GithubAuthMiddleware = async (req: Request, res: Response, next: NextFunction) => {
 	try {
 		const { githubToken, gitHubUuid } = req.session;
-		const { jiraHost, gitHubAppId, gitHubAppConfig } = res.locals;
+		const { jiraHost, gitHubAppConfig } = res.locals;
+		const gitHubAppId = res.locals.gitHubAppId || gitHubAppConfig.id;
+
 		/**
 		 * Comparing the `UUID` saved in the session with the `UUID` inside `gitHubAppConfig`,
 		 * to trigger another GitHub Login and fetch new githubToken.
@@ -138,6 +138,7 @@ export const GithubAuthMiddleware = async (req: Request, res: Response, next: Ne
 			req.log.info("github token missing, calling login()");
 			throw "Missing github token";
 		}
+
 		req.log.debug("found github token in session. validating token with API.");
 
 		const url = await getGitHubApiUrl(jiraHost, gitHubAppId);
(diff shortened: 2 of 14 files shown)


src/routes/github/github-oauth-router.ts
```diff
@@ -122,9 +122,13 @@ const GithubOAuthCallbackGet = async (req: Request, res: Response, next: NextFun
 };
 
 export const GithubAuthMiddleware = async (req: Request, res: Response, next: NextFunction) => {
+	req.log.debug("GitHub Auth request received.");
```
rachellerathbone, line 125: Added some new debugs logs to make it easier to see request pre debug mode.

src/util/handlebars/handlebar-helpers.ts
```diff
@@ -25,6 +25,11 @@ export const registerHandlebarsHelpers = () => {
 
 	hbs.registerHelper("isNotConnected", (syncStatus) => syncStatus == null);
 
+	hbs.registerHelper("setSubscriptionUrl", (appId, uuid, installationId) => appId
+		? `/github/subscriptions/${installationId}/ghes?appId=${appId}&uuid=${uuid}`
```
krazziekay, line 29: Do we need this `ghes`? I don't see any use of this additional parameter.

rachellerathbone: It's in the router `GithubSubscriptionRouter.route("/:installationId/:ghes?")`

krazziekay: But its not used inside `GithubSubscriptionGet` nor `GithubSubscriptionDelete`. Won't it still work if we remove this from the route?

rachellerathbone: It's needed to define the query params.

krazziekay: Can't we do `/github/subscriptions/${installationId}?appId=${appId}&uuid=${uuid}` ? The query parameters are pulled in the exact same manner. 
Its just that word `ghes` brings up a bit of confusion cause that parameter itself is not used inside any method.

krazziekay (approved):

src/middleware/github-server-app-middleware.ts
```diff
@@ -7,7 +7,7 @@ import { GITHUB_CLOUD_HOSTNAME } from "utils/get-github-client-config";
 
 export const GithubServerAppMiddleware = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
 	const { jiraHost } = res.locals;
-	const { uuid } = req.params;
+	const uuid = req.params.uuid || req.query.uuid?.toString();
```
gxueatlassian, line 10: Which url require this `uuid` to be in the query but not in the path? 
If any, is it better to update that url to include uuid in the path? Or is it not feasible?

src/routes/github/subscription/github-subscription-delete.ts
```diff
@@ -4,24 +4,28 @@ import { isUserAdminOfOrganization } from "~/src/util/github-utils";
 import { createAppClient, createUserClient } from "~/src/util/get-github-client-config";
 
 export const GithubSubscriptionDelete = async (req: Request, res: Response): Promise<void> => {
-	const { githubToken, jiraHost, gitHubAppId } = res.locals;
+	const { githubToken, jiraHost } = res.locals;
+	const gitHubAppId = res.locals.gitHubAppId || req.body.gitHubAppId;
```
gxueatlassian, line 8: Same here, can we trust the `req.body.gitHubAppId`? Should we instead put the `uuid` in the url path, and let the middlware to verify it?

src/routes/github/subscription/github-subscription-get.ts
```diff
@@ -4,18 +4,23 @@ import { isUserAdminOfOrganization } from "utils/github-utils";
 import { createAppClient, createUserClient } from "~/src/util/get-github-client-config";
 
 export const GithubSubscriptionGet = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
-	const { githubToken, jiraHost, gitHubAppId } = res.locals;
-
+	const { githubToken, jiraHost } = res.locals;
 	const gitHubInstallationId = Number(req.params.installationId);
+	const gitHubAppId = Number(req.query.gitHubAppId);
```
gxueatlassian, line 9: Same security concern.

gxueatlassian (changes requested):
Some security concern comments.

src/middleware/github-server-app-middleware.ts
rachellerathbone: For the GH manage subscriptions page for GHES we navigate to /github/subscriptions/${installationId}/?gitHubAppId=${gitHubAppId}&uuid=${uuid} and for cloud we navigate to /github/subscriptions/${installationId}. I chose to go with a query string instead of params otherwise I would need to add 2 optional params here: GithubSubscriptionRouter.route("/:installationId") and figured it was better to leave that as is as opposed to the alternative of GithubSubscriptionRouter.route("/:installationId/:gitHubAppId?/:uuid?")

src/routes/github/subscription/github-subscription-delete.ts
rachellerathbone: Uuid is already part of the query string. I don't think it would be necessary to make this additional check as passing a bogus gitHubAppId would fail on lines 14/15 unless someone also happened to obtain another users jiraHost and githubToken. Unless these 3 things are valid, I don't see an issue with using the gitHubAppId in the body.

src/routes/github/subscription/github-subscription-get.ts
rachellerathbone: Same thing here. Unless someone knows what my GH token is, as well as my jiraHost, I don't see any issues with passing the gitHubAppId from the query string.

src/middleware/github-server-app-middleware.ts
gxueatlassian: I'd suggest change the frontend to navigate to `/github/:uuid/subscriptions/${installationId}` for managing ghe use cases. And then the current `gitHubServerMiddleware` will extract `gitHubAppId` from the uuid.

src/routes/github/subscription/github-subscription-delete.ts
gxueatlassian: If the `uuid` is passed in query path as `/github/:uuid/subscription/${installationId}` etc, `gitHubAppId` should be already available from the `res.locals.gitHubAppConfig`, no need to get it from body.

src/models/github-server-app.ts
```diff
@@ -175,7 +175,9 @@ export class GitHubServerApp extends EncryptedModel {
 
 	}
 
-	static async findForUuid(uuid: string): Promise<GitHubServerApp | null> {
+	static async findForUuid(uuid: string | null): Promise<GitHubServerApp | undefined> {
```
rachellerathbone, line 178: We are now calling this method and passing req.params.uuid from middleware where that param may not exist (cloud path) and the where clause fails when an undefined value is passed.

src/routes/github/github-oauth-router.ts
```diff
@@ -124,7 +125,10 @@ const GithubOAuthCallbackGet = async (req: Request, res: Response, next: NextFun
 export const GithubAuthMiddleware = async (req: Request, res: Response, next: NextFunction) => {
 	try {
 		const { githubToken } = req.session;
-		const { jiraHost, gitHubAppId } = res.locals;
+		const { jiraHost } = res.locals;
+		const gitHubApp = await GitHubServerApp.findForUuid(req.params.uuid);
+		const gitHubAppId = res.locals.gitHubAppId || gitHubApp?.id;
```
rachellerathbone, line 130: If gitHubAppId isn't available in res.locals, we try to get it from the app ID.

src/routes/github/github-oauth-router.ts
```diff
@@ -161,6 +165,6 @@ export const GithubAuthMiddleware = async (req: Request, res: Response, next: Ne
 
 // IMPORTANT: We need to keep the login/callback/middleware functions
 // in the same file as they reference each other
-export const GithubOAuthRouter = Router();
+export const GithubOAuthRouter = Router({ mergeParams: true });
```
rachellerathbone, line 168: Turns out uuid was an undefined value for any gh paths being set by a nested route. If we want the uuid to be available as a param, we need to pass this property to individual routers, as well as to the subrouter.

src/models/github-server-app.ts
```diff
@@ -175,7 +175,9 @@ export class GitHubServerApp extends EncryptedModel {
 
 	}
 
-	static async findForUuid(uuid: string): Promise<GitHubServerApp | null> {
+	static async findForUuid(uuid: string | undefined): Promise<GitHubServerApp | null> {
+		if (!uuid) return null;
```
rachellerathbone, line 179: Added this because the where clause fails if you pass an undefined value (as may happen for cloud path)

src/routes/github/github-oauth-router.ts
```diff
@@ -127,7 +127,9 @@ const GithubOAuthCallbackGet = async (req: Request, res: Response, next: NextFun
 export const GithubAuthMiddleware = async (req: Request, res: Response, next: NextFunction) => {
 	try {
 		const { githubToken, gitHubUuid } = req.session;
-		const { jiraHost, gitHubAppId, gitHubAppConfig } = res.locals;
+		const { jiraHost, gitHubAppConfig } = res.locals;
+		const gitHubAppId = res.locals.gitHubAppId || gitHubAppConfig.id;
```
rachellerathbone, line 131: gitHubAppId isn't available from res.locals in all paths. For the subscription portion of the flow gitHubAppConfig, but not gitHubAppId

src/routes/github/subscription/github-subscription-router.ts
```diff
@@ -4,7 +4,7 @@ import { GithubSubscriptionDelete } from "./github-subscription-delete";
 import { param } from "express-validator";
 import { returnOnValidationError } from "../../api/api-utils";
 
-export const GithubSubscriptionRouter = Router();
+export const GithubSubscriptionRouter = Router({ mergeParams: true });
```
rachellerathbone, line 7: Nested routers need this property to be set otherwise the uuid in the params is undefined.

src/routes/github/subscription/github-subscription-delete.ts
```diff
@@ -2,26 +2,38 @@ import { Request, Response } from "express";
...
 	const { installationId: gitHubInstallationId } = req.body;
 	const logger = req.log.child({ jiraHost, gitHubInstallationId });
+
+	logger.debug("Received DELETE subscription request");
+
+	const gitHubApp = await GitHubServerApp.findForUuid(req.params.uuid);
+
+	if (gitHubApp?.id !== gitHubAppConfig?.gitHubAppId) {
+		logger.debug("GitHub app IDs do not match. Cannot DELETE subscription.");
+		throw new Error("Cannot DELETE subscription.");
```
rachellerathbone, line 18: Extra precaution to make sure we're deleting the correct gh app and that the user has permissions to do so.

src/routes/github/github-oauth-router.ts
gxueatlassian: In the `github-server-app-middlware` doesn't generate `gitHubAppConfig.id`, it generates `gitHubAppConfig.gitHubAppId`. That's the same one as `res.locals.gitHubAppId`.

gxueatlassian: Can use whichever one.

gxueatlassian (approved):

gxueatlassian (approved):

PR #1525, opened 2022-08-18
ARC-1594 - Only render orgs connected to an app on the connect an org page

krazziekay:
**What's in this PR?**
- Added github UUID in the session.
- Checking this saved UUID with the UUID from githubAppConfig.
  - If same do nothing
  - If different, do another login to fetch new gitHubToken. 

**Why**
- The `githubToken` for the endpoint `user/installations`, which fetches the list of installations for a user, is based on the app. So, for every new app, we need a new token. Thats why we are re-authenticating every time when the UUIDs do not match.

**How has this been tested?**  
- In local
- In ddev

**Video**

https://user-images.githubusercontent.com/13076255/185519062-a4b7bc43-e5b1-498a-af7f-e789f9790085.mov

### DIFF
--- a/src/interfaces/common.ts
+++ b/src/interfaces/common.ts
@@ -37,7 +37,7 @@ declare global {
 			session: {
 				jiraHost?: string;
 				githubToken?: string;
-				githubUuid?: string;
+				gitHubUuid?: string;
 				temp?:  {
 					[key: string]: any;
 				}
--- a/src/interfaces/common.ts
+++ b/src/interfaces/common.ts
@@ -37,6 +37,7 @@ declare global {
 			session: {
 				jiraHost?: string;
 				githubToken?: string;
+				githubUuid?: string;
 				temp?:  {
 					[key: string]: any;
 				}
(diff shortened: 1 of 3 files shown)


src/routes/github/github-oauth-router.ts
```diff
@@ -106,6 +106,9 @@ const GithubOAuthCallbackGet = async (req: Request, res: Response, next: NextFun
 		// Saving it to session be used later
 		req.session.githubToken = response.data.access_token;
 
+		// Saving UUID for each GitHubServerApp
+		req.session.githubUuid = uuid;
```
rachellerathbone, line 110: Nit: casing should be gitHubUuid

krazziekay: Updated to `gitHubUuid`

src/interfaces/common.ts
```diff
@@ -37,6 +37,7 @@ declare global {
 			session: {
 				jiraHost?: string;
 				githubToken?: string;
+				gitHubUuid?: string;
```
gxueatlassian, line 40: /nit or JFYI,

Personally I'd prefer to carry the `uuid` to different pages via url (`/github/uuid`), instead put it in the session token.

But can leave it as it is.

gxueatlassian (changes requested):
Seems there's better way to get the uuid?

gxueatlassian (approved):

PR #1526, opened 2022-08-18
ARC-1617 - Show the correct repository number in the list of connections for GHES

krazziekay:
**What's in this PR?**
- Updated the logic to fetch the `privateKey` using the `ghsaId` directly instead of using the extra `Subscription` call.

**Why**
- The number of repositories are being fetched from `GithubInstallationClient`, which requires a JWT token, which is created using the private key of the installation/organization/subscription.
- But any installation/organization/subscription that has not been connected yet is not saved in our DB. So, they don't have any Subscription yet, and that was causing the `if-else` check to always go to the `else` part. This causes the privateKey of the cloud to be used and triggers a 401 error.

**How has this been tested?**  
- In local
- In ddev

**Screenshot**
- GHE server
image
- GH Cloud
image


src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -67,18 +66,17 @@ const getInstallationsWithAdmin = async (
...
 	return await Promise.all(installations.map(async (installation) => {
 		const errors: Error[] = [];
-		const gitHubClient = await createInstallationClient(installation.id, jiraHost, log, gitHubAppId);
 
-		const numberOfReposPromise = await gitHubClient.getNumberOfReposForInstallation().catch((err) => {
-			errors.push(err);
-			return 0;
-		});
+		const numberOfReposPromise = await gitHubUserClient.getNumberOfReposForInstallation(installation.id)
+			.then((response) => response.data.total_count).
```
rachellerathbone, line 75: So we're not using getNumberOfReposForInstallation that's defined in the installation client anymore? If not, we should probably delete later on but for now, this is prod code, we should probably feature flag this. Leave the original implementation if the flag is off and only call this new endpoint if it's on.

src/github/client/github-user-client.ts
```diff
@@ -58,6 +58,12 @@ export class GitHubUserClient extends GitHubClient {
 		return await this.get<Octokit.AppsListInstallationsForAuthenticatedUserResponse>("/user/installations");
 	}
 
+	public async getNumberOfReposForInstallation(installationId: number): Promise<AxiosResponse<Octokit.AppsListInstallationsForAuthenticatedUserResponse>> {
+		return await this.get<Octokit.AppsListInstallationsForAuthenticatedUserResponse>(`/user/installations/{installationId}/repositories`, {
```
rachellerathbone, line 62: Why the shift from graphql? If we're using graphql on the installation method to get the number of repos, shouldn't we do the same here?

src/routes/github/configuration/github-configuration-get.ts
krazziekay: We can add FF to toggle this, but I feel its unnecessary, cause both endpoints give the same result. 
The previous one used JWT token through graphql request, and this one uses githubToken through a [GET request](https://docs.github.com/en/rest/apps/installations#list-repositories-accessible-to-the-user-access-token).

src/github/client/github-user-client.ts
krazziekay: The previous graphql implementation used `GithubInstallationClient`, which requires a JWT token, which is created using the private key of the installation/organization/subscription.
But in our case, the installation/organization/subscription has not been connected and yet is not saved in our DB. So, they don't have a privateKey. So, we can't use grahpql request cause we have no privateKey to create the JWT.
As a result we are getting a 401 error, cause theres a mismatch of the privateKeys.
Thats why used a different endpoint in `GithubUserClient`, cause the token here is generated during the oauth authentication happening in middleware and gives us the same result.

rachellerathbone: Yeah yeah I get that. Using the user client isn't what I was calling out. I was asking why we switched from making a request via graphql to REST. GHES has a graphql api and the gh client supports it `this.graphqlUrl = `${baseUrl}/api/graphql`

krazziekay: I wasn't sure if the `graphql` works with the `githubToken` here in `GithubUserClient`. And as I was looking into the Github docs, found this endpoint:[ user/installations/{id}/repositories](https://docs.github.com/en/rest/apps/installations#list-repositories-accessible-to-the-user-access-token). So just used this.

src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -67,18 +66,17 @@ const getInstallationsWithAdmin = async (
 	gitHubUserClient: GitHubUserClient,
 	log: Logger,
 	login: string,
-	installations: Octokit.AppsListInstallationsForAuthenticatedUserResponseInstallationsItem[] = [],
-	jiraHost: string,
-	gitHubAppId?: number
```
gxueatlassian, line 72: while you are here, change it to `gitHubAppId: number | undefined`?

src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -67,18 +66,17 @@ const getInstallationsWithAdmin = async (
...
 ): Promise<InstallationWithAdmin[]> => {
 	return await Promise.all(installations.map(async (installation) => {
 		const errors: Error[] = [];
-		const gitHubClient = await createInstallationClient(installation.id, jiraHost, log, gitHubAppId);
 
-		const numberOfReposPromise = await gitHubClient.getNumberOfReposForInstallation().catch((err) => {
-			errors.push(err);
-			return 0;
-		});
+		const numberOfReposPromise = await gitHubUserClient.getNumberOfReposForInstallation(installation.id)
```
gxueatlassian, line 74: /nit
Is the return a `promise` or `number`?

krazziekay: The return is a number. Cleaned up the `Promise.all` by using `await`.

src/github/client/github-user-client.ts
krazziekay: This has been reverted.

src/routes/github/configuration/github-configuration-get.ts
krazziekay: This has been reverted

gxueatlassian (dismissed):

gxueatlassian (approved):

PR #1530, opened 2022-08-19
Arc 1619 make GitHub app id mandatory

gxueatlassian:
**What's in this PR?**
Change many method signature from 

```javascript
someMethod(..., gitHubAppId?: number, ...)
```
to 
```javascript
someMethod(..., gitHubAppId: number | undefined, ...)
```

**Why**
To have more confident in the code that it is doing right thing (passing the gitHubAppId for GHES), I believe making the argument mandatory is better. In deed, I believe some bugs are found in this refactoring. 

In current code base, if the code that intends to support both `cloud` and `GHES`, but forgot the passing the `gitHubAppId`, can still pass the TS checking. I want to ensure that doesn't happen.

So calling bellow will fail
```javascript
Subscriptions.getAllForInstallation(gitHubIntallationId)
```

**Notes**
To ensure catching the bug by refactoring the code without making huge PR or introduce new bugs, I intentionally impl code with following guidelines:
1. First change `gitHubAppId?: number` to `gitHubAppId: number | undefined` by global find&replace
2. Do `tsc --watch` to compile and fix whatever failing the ts check
3. Try NOT to change code behaviour as best as I can (unless it is obvious bug, which is one of the purpose of this PR).
4. DID NOT refactoring any code structure, like merging options, putting the gitHubAppId in a better place, change payload structure to better accommodate the id, fixing any tech debt in the `gitHubServerMiddlware`, etc / blah / blah
5. There're some api routes missing the `gitHubAppId`. I think better way is to introduce it in the `path`, but DID NOT do that because I am afraid it will break the pollinator. Instead make it optional as query string.

**Added feature flags**
N/A

**Affected issues**  
ARC-1619

**How has this been tested?**  
unit test
stage [pollinator and manual: https://bitbucket.org/atlassian/github-for-jira-deployment/pipelines/results/5674]

**Whats Next?**
N/A

### DIFF
--- a/src/jira/client/jira-client.ts
+++ b/src/jira/client/jira-client.ts
@@ -32,8 +32,8 @@ export interface DeploymentsResult {
 export const getJiraClient = async (
 	jiraHost: string,
 	gitHubInstallationId: number,
-	log: Logger = getLogger("jira-client"),
-	gitHubAppId: number | undefined
+	gitHubAppId: number | undefined,
+	log: Logger = getLogger("jira-client")
 ): Promise<any> => {
 	const logger = log.child({ jiraHost, gitHubInstallationId });
 	const installation = await Installation.getForHost(jiraHost);
--- a/src/jira/client/jira-client.ts
+++ b/src/jira/client/jira-client.ts
@@ -33,7 +33,7 @@ export const getJiraClient = async (
 	jiraHost: string,
 	gitHubInstallationId: number,
 	log: Logger = getLogger("jira-client"),
-	gitHubAppId?: number
+	gitHubAppId: number | undefined
 ): Promise<any> => {
 	const logger = log.child({ jiraHost, gitHubInstallationId });
 	const installation = await Installation.getForHost(jiraHost);
--- a/src/routes/api/installation/api-installation-delete.ts
+++ b/src/routes/api/installation/api-installation-delete.ts
@@ -9,7 +9,7 @@ export const ApiInstallationDelete = async (req: Request, res: Response): Promis
 	//TODO: ARC-1619 Maybe need to fix this and put it into the path
 	//Not doing it now as it might break pollinator if it use this api
 	const { gitHubAppIdStr } = req.query;
-	const gitHubAppId = parseInt(gitHubAppIdStr as string) || undefined
+	const gitHubAppId = parseInt(gitHubAppIdStr as string) || undefined;
 
 	if (!jiraHost || !githubInstallationId) {
 		const msg = "Missing Jira Host or Installation ID";
--- a/src/routes/api/installation/api-installation-delete.ts
+++ b/src/routes/api/installation/api-installation-delete.ts
@@ -6,11 +6,6 @@ export const ApiInstallationDelete = async (req: Request, res: Response): Promis
 	const githubInstallationId = req.params.installationId;
 	const jiraHost = req.params.jiraHost;
 
-	//TODO: ARC-1619 Maybe need to fix this and put it into the path
-	//Not doing it now as it might break pollinator if it use this api
-	const { gitHubAppIdStr } = req.query;
-	const gitHubAppId = parseInt(gitHubAppIdStr as string) || undefined;
-
 	if (!jiraHost || !githubInstallationId) {
 		const msg = "Missing Jira Host or Installation ID";
 		req.log.warn({ req, res }, msg);
@@ -21,7 +16,7 @@ export const ApiInstallationDelete = async (req: Request, res: Response): Promis
 	const subscription = await Subscription.getSingleInstallation(
 		jiraHost,
 		Number(githubInstallationId),
-		gitHubAppId
+		undefined
 	);
 
 	if (!subscription) {
@@ -31,7 +26,7 @@ export const ApiInstallationDelete = async (req: Request, res: Response): Promis
 	}
 
 	try {
-		const jiraClient = await getJiraClient(jiraHost, Number(githubInstallationId), req.log, gitHubAppId);
+		const jiraClient = await getJiraClient(jiraHost, Number(githubInstallationId), undefined, req.log);
 		req.log.info({ jiraHost, githubInstallationId }, `Deleting DevInfo`);
 		await jiraClient.devinfo.installation.delete(githubInstallationId);
 		res.status(200).send(`DevInfo deleted for jiraHost: ${jiraHost} githubInstallationId: ${githubInstallationId}`);
--- a/src/routes/api/installation/api-installation-delete.ts
+++ b/src/routes/api/installation/api-installation-delete.ts
@@ -6,6 +6,11 @@ export const ApiInstallationDelete = async (req: Request, res: Response): Promis
 	const githubInstallationId = req.params.installationId;
 	const jiraHost = req.params.jiraHost;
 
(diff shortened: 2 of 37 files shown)


src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -103,15 +103,16 @@ const getInstallationsWithAdmin = async (
...
 };
 
-const removeFailedConnectionsFromDb = async (req: Request, installations: InstallationResults, jiraHost: string): Promise<void> => {
+const removeFailedConnectionsFromDb = async (req: Request, installations: InstallationResults, jiraHost: string, gitHubAppId: number | undefined): Promise<void> => {
 	await Promise.all(installations.rejected
 		// Only uninstall deleted installations
 		.filter(failedInstallation => failedInstallation.deleted)
 		.map(async (failedInstallation) => {
 			try {
 				await Subscription.uninstall({
```
gxueatlassian, line 112: **NOTE**
Potential bug

src/routes/github/subscription/github-subscription-delete.ts
```diff
@@ -38,7 +38,7 @@ export const GithubSubscriptionDelete = async (req: Request, res: Response): Pro
 			return;
 		}
 		try {
-			const subscription = await Subscription.getSingleInstallation(jiraHost, gitHubInstallationId);
```
gxueatlassian, line 41: **NOTE**
Potential bug

src/routes/github/subscription/github-subscription-get.ts
```diff
@@ -26,7 +26,7 @@ export const GithubSubscriptionGet = async (req: Request, res: Response, next: N
 		const { data: installation } = await gitHubAppClient.getInstallation(gitHubInstallationId);
 
 		// get all subscriptions from the database for this installation ID
-		const subscriptions = await Subscription.getAllForInstallation(gitHubInstallationId);
```
gxueatlassian, line 29: **NOTE**
Potential bug

src/transforms/transform-deployment.ts
```diff
@@ -234,7 +234,7 @@ export const transformDeployment = async (githubInstallationClient: GitHubInstal
 	let config: Config | undefined;
 
 	if (await booleanFlag(BooleanFlags.CONFIG_AS_CODE, false, jiraHost)) {
-		const subscription = await Subscription.getSingleInstallation(jiraHost, githubInstallationClient.githubInstallationId.installationId);
```
gxueatlassian, line 237: **NOTE**
Potential bug

src/jira/client/jira-client.ts
```diff
@@ -33,7 +33,7 @@ export const getJiraClient = async (
 	jiraHost: string,
 	gitHubInstallationId: number,
 	log: Logger = getLogger("jira-client"),
-	gitHubAppId?: number
+	gitHubAppId: number | undefined
```
mboudreau, line 36: if you're going to do this change, can you please put the log as being the last parameter?  The log isn't important and you should never have a require parameter _after_ an optional one.

src/routes/api/installation/api-installation-delete.ts
```diff
@@ -6,6 +6,11 @@ export const ApiInstallationDelete = async (req: Request, res: Response): Promis
 	const githubInstallationId = req.params.installationId;
 	const jiraHost = req.params.jiraHost;
 
+	//TODO: ARC-1619 Maybe need to fix this and put it into the path
+	//Not doing it now as it might break pollinator if it use this api
+	const { gitHubAppIdStr } = req.query;
+	const gitHubAppId = parseInt(gitHubAppIdStr as string) || undefined;
```
mboudreau, line 12: isn't gitHubAppIdStr already a string?

src/routes/api/installation/api-installation-delete.ts
```diff
@@ -6,6 +6,11 @@ export const ApiInstallationDelete = async (req: Request, res: Response): Promis
 	const githubInstallationId = req.params.installationId;
 	const jiraHost = req.params.jiraHost;
 
+	//TODO: ARC-1619 Maybe need to fix this and put it into the path
+	//Not doing it now as it might break pollinator if it use this api
+	const { gitHubAppIdStr } = req.query;
```
mboudreau, line 11: is the value set in the query actually `gitHubAppIdStr`?  We shouldn't be using query for entity information as it breaks the REST conventions - it should be part of the url params.

mboudreau: also, why is this needed in the API?  It's kinda outside the scope of this PR

src/routes/jira/jira-get.ts
```diff
@@ -49,7 +49,7 @@ const mapSyncStatus = (syncStatus: SyncStatus = SyncStatus.PENDING): string => {
 	}
 };
 
-export const getInstallations = async (subscriptions: Subscription[], log: Logger, gitHubAppId?: number): Promise<InstallationResults> => {
+export const getInstallations = async (subscriptions: Subscription[], log: Logger, gitHubAppId: number | undefined): Promise<InstallationResults> => {
```
mboudreau, line 52: `gitHubAppId` should be before logger

src/routes/api/installation/api-installation-delete.ts
gxueatlassian: Got this 
image

src/routes/jira/jira-get.ts
gxueatlassian: The refactor on this arg position, in this case, is not purpose of this PR, in this line, it merely change the `?`

src/jira/client/jira-client.ts
gxueatlassian: fixed.

src/routes/api/installation/api-installation-delete.ts
gxueatlassian: Actually I've remove/revert the api changes here. Created new ticket to tackle that ARC-1622

src/routes/jira/jira-get.ts
mboudreau: The reason for this PR is to make something mandatory, as such, I'd expect you'd use the proper coding conventions when doing so.

gxueatlassian: fixed. please review.

src/routes/github/subscription/github-subscription-delete.ts
mboudreau: why?

gxueatlassian: I mean before change, it is a bug. After change, it should be fine.

test/setup/setup.ts
```diff
@@ -130,6 +130,7 @@ beforeAll(async () => {
 });
 
 const globalGitHubAppConfig = {
+	gitHubAppId: -1, //TODO: ARC-1619, this value is not known yet.
```
mboudreau, line 133: shouldn't this be `undefined`?

gxueatlassian: I've removed `globalGitHubAppConfig` completely, it is not being used at all. And from the value itself, I can't tell whether it is a cloud config or a ghe config. So remove it.

mboudreau (approved):

PR #1534, opened 2022-08-22
ARC-1618 - Fix navigation from manage subscriptions page

krazziekay:
**What's in this PR?**
- Passed the `moduleKey` as a message from the child window to the parent window.
- Added an event listener to the parent, which checks the origin and the `moduleKey` and will navigate to that `moduleKey`.

**Why**
- Fixing routes.

**How has this been tested?**  
- Locally


static/js/github-configuration.js
```diff
@@ -26,6 +26,7 @@ $('.delete-link').click(function (event) {
     if (data.err) {
       return console.log(data.err)
     }
+		window.opener.postMessage({moduleKey: "github-post-install-page"}, window.location.origin);
```
gxueatlassian, line 29: /nit
maybe add a messageType along with the moduleKey, with 'goto_module_page'.
purely for code readability

gxueatlassian (approved):

static/js/github-configuration.js
krazziekay: I thought about adding it, but felt like its of not much value at this point. Will leave it for now...

rachellerathbone:
Converted to draft because this needs to wait until Arc-1596 is merged.

PR #1535, opened 2022-08-22
ARC-1621: increase max_client_conn for pgbouncer

bgvozdev:
**What's in this PR?**

Increasing the number of allowed connections to pgbouncer

**Why**

Because it refuses connections in peak hours, which causes broken syncs: SequelizeConnectionError strongly correlates with "no more connections allowed (max_client_conn)" from pgbouncer (check the ticket)

**Added feature flags**

**Affected issues**  
ARC-1621

**How has this been tested?**  
We will check the effect in prod :cowboy:

**Whats Next?**
:susp:


yvonnevictoria:
```
How has this been tested?
We will check the effect in prod :cowboy:
```

I don't know enough about this ticket to advise against this, but be sure before you merge 😄

yvonnevictoria (approved):

PR #1545, opened 2022-08-24
Arc 1622 api installtion

gxueatlassian:
**What's in this PR?**
Add gitHubAppId to the route and each handler.

**Why**
Just installationId is not enough.

**Added feature flags**
N/A

**Affected issues**  
ARC-1622

**How has this been tested?**  
Unit test.

**Whats Next?**
N/A

### DIFF
--- a/src/routes/api/installation/api-installation-delete.ts
+++ b/src/routes/api/installation/api-installation-delete.ts
@@ -3,11 +3,11 @@ import { Subscription } from "models/subscription";
 import { getJiraClient } from "~/src/jira/client/jira-client";
 
 export const ApiInstallationDelete = async (req: Request, res: Response): Promise<void> => {
-	const githubInstallationId = req.params.installationId;
+	const gitHubInstallationId = req.params.installationId;
 	const gitHubAppId = Number(req.params.gitHubAppId) || undefined;
 	const jiraHost = req.params.jiraHost;
 
-	if (!jiraHost || !githubInstallationId) {
+	if (!jiraHost || !gitHubInstallationId) {
 		const msg = "Missing Jira Host or Installation ID";
 		req.log.warn({ req, res }, msg);
 		res.status(400).send(msg);
@@ -16,7 +16,7 @@ export const ApiInstallationDelete = async (req: Request, res: Response): Promis
 
 	const subscription = await Subscription.getSingleInstallation(
 		jiraHost,
-		Number(githubInstallationId),
+		Number(gitHubInstallationId),
 		gitHubAppId
 	);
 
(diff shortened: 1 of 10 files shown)


src/routes/api/installation/api-installation-delete.ts
```diff
@@ -4,6 +4,7 @@ import { getJiraClient } from "~/src/jira/client/jira-client";
 
 export const ApiInstallationDelete = async (req: Request, res: Response): Promise<void> => {
 	const githubInstallationId = req.params.installationId;
```
rachellerathbone, line 6: Nit: all casing for GH should be gitHub.

src/routes/api/installation/api-installation-router.ts
```diff
@@ -7,25 +7,27 @@ import { ApiInstallationSyncPost } from "./api-installation-sync-post";
 import { ApiInstallationGet } from "./api-installation-get";
 
 export const ApiInstallationRouter = Router({ mergeParams: true });
+const subRouter = Router({ mergeParams: true });
+ApiInstallationRouter.use(`(/githubapp/:gitHubAppId(\\d+))?`, subRouter);
 
-ApiInstallationRouter.post(
+subRouter.post(
```
bgvozdev, line 13: Curious what would be the path to trigger resync for a cloud customer?

src/routes/api/installation/api-installation-router.ts
```diff
@@ -7,25 +7,27 @@ import { ApiInstallationSyncPost } from "./api-installation-sync-post";
 import { ApiInstallationGet } from "./api-installation-get";
```
bgvozdev, line 7: I believe we also need to update all our runbooks?

src/routes/api/installation/api-installation-router.test.ts
```diff
@@ -0,0 +1,104 @@
+import { Application } from "express";
```
bgvozdev, line 1: 🥇

src/routes/api/installation/api-installation-router.ts
gxueatlassian: Ticket created ARC-1644.
For cloud, run book is same, so no changes.
For GHES, will do it in above ticket.

src/routes/api/installation/api-installation-router.ts
gxueatlassian: Seems the resync is routed in the `ApiRouter` as
```javascript
ApiRouter.post(`/:uuid(${UUID_REGEX})?/resync`,..., async (req, res) => {});
```

So it is not included inside this ApiInstallationRouter.

src/routes/api/installation/api-installation-delete.ts
gxueatlassian: fixed.

src/routes/api/installation/api-installation-router.ts
bgvozdev: Ok,sorry, I'm confused now :) Let me summarise my questions/concerns:

1. Could you please clarify who calls these endpoints?
2. Since we are prefixing it essentially with `githubapp` now, everything that was calling these endpoints will be broken now (no matter if it is in the code or in the runbooks), which is the cloud installations essentially (because GHES is not in prod yet)... What's the plan about it? Or noone is calling them now?

gxueatlassian: No worries, let me try to explain:
1. I believe only human or machine call these apis. (LOL, I mean pollinator or anyone does some ops in prod)
2. There's a `?` in the line `ApiInstallationRouter.use(`(/githubapp/:gitHubAppId(\\d+))?`, subRouter);`. I think that does the trick, that it can accept with or without the `/githubapp/:gitHubAppId` in the route. If you look closer to the test, it should test the scenarios that api is hitting both with and without the new route.

So in short, for current runbook (pollinator, human), cloud supporting is unchanged. Only added support is GHES, therefore I think runbook for GHES can be done in another ticket. 

Let me know if you spot anything wrong in the routing still. Thanks.

bgvozdev: Gotcha @gxueatlassian , thanks for the explanation! It makes much more sense now

bgvozdev (approved):

PR #1547, opened 2022-08-24
ARC-1641 - Block uploading files that don't have the .pem extension

krazziekay:
**What's in this PR?**
- Condition updated to check if the file is valid before displaying the success flag.

**Why**
- To avoid confusion when uploading an invalid file.

**How has this been tested?**  
- Tested in local and staging.

**Video**

https://user-images.githubusercontent.com/13076255/186343546-72034b2b-a4b7-4234-9aaf-b0375078139f.mov

### DIFF
--- a/static/js/jira-manual-app-creation.js
+++ b/static/js/jira-manual-app-creation.js
@@ -109,7 +109,7 @@ AJS.$("#jiraManualAppCreation__form").on("aui-valid-submit", (event) => {
 const replaceSpacesAndChangeCasing = (str) => str.replace(/\s+/g, '-').toLowerCase();
 
 $('#jiraManualAppCreation__uploadedFile').bind('DOMSubtreeModified', function () {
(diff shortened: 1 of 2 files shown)


static/js/jira-manual-app-creation.js
```diff
@@ -125,9 +125,10 @@ $('#jiraManualAppCreation__uploadedFile').bind('DOMSubtreeModified', function ()
 			<p>Your file has been uploaded and will be </br>
 			stored with the filename in the format of </br>
 			${appName}.private-key.pem</p>
-	`
+	`;
+	const pemFilePattern = "^\s*$|^.*\.(pem|PEM)$";
```
gxueatlassian, line 129: what does `^\s*$` means?

krazziekay: Regex updated

gxueatlassian (approved):

PR #1548, opened 2022-08-24
ARC-1507 update logs

rachellerathbone:
**What's in this PR?**
- Updated some logs to include `gitHubProduct` (cloud/server). 
- Changed a handful of info logs to debug. 

**Why**
- To make it easier to split out customer groups in our logs.
- These logs add value locally but just add noise to Splunk.

**Added feature flags**

**Affected issues**  
ARC-1507

### DIFF
--- a/src/github/branch.ts
+++ b/src/github/branch.ts
@@ -9,6 +9,7 @@ import { GitHubInstallationClient } from "./client/github-installation-client";
 import { JiraBranchData } from "../interfaces/jira";
 import { jiraIssueKeyParser } from "utils/jira-utils";
 import { WebhookContext } from "../routes/github/webhook/webhook-context";
+import { getCloudOrServerFromGitHubAppId } from "utils/get-cloud-or-server";
 
 export const createBranchWebhookHandler = async (context: WebhookContext, jiraClient, _util, gitHubInstallationId: number): Promise<void> => {
 
@@ -48,7 +49,8 @@ export const processBranch = async (
 		return;
 	}
 
-	logger.info(`Sending jira update for create branch event`);
+	const gitHubProduct = getCloudOrServerFromGitHubAppId(gitHubAppId);
+	logger.info({ gitHubProduct },"Sending update for create branch event to Jira.");
 
 	const jiraClient = await getJiraClient(
 		jiraHost,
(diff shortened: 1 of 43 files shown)


src/github/branch.ts
```diff
@@ -77,14 +79,23 @@ export const deleteBranchWebhookHandler = async (context: WebhookContext, jiraCl
 		return;
 	}
 
-	context.log.info(`Deleting branch for repo ${context.payload.repository?.id} with ref ${context.payload.ref}`);
+	const gitHubAppId = context.gitHubAppConfig?.gitHubAppId;
+	const gitHubProduct = getCloudOrServerFromGitHubAppId(gitHubAppId);
+
+	context.log.info(
```
gxueatlassian, line 85: Hi, there're quite a lot of changes in the webhook related handlers, I assume they are all for the main purpose of adding a product field in the log?

If so, we can do it in the GithubWebhookMiddlware at [this line](https://github.com/atlassian/github-for-jira/blob/7b3e3f5de2af73f01a331595acc55724e3a8089a/src/middleware/github-webhook-middleware.ts#L106), like `context.log.addFields({...product})`, so that the change is just in one place. 

Do you think it is okay to do that?

src/jira/client/jira-client.ts
```diff
@@ -265,7 +267,7 @@ export const getJiraClient = async (
 					};
 				}
 				logger?.debug(`Sending builds payload to jira. Payload: ${payload}`);
-				logger?.info("Sending builds payload to jira.");
+				logger?.info({ gitHubProduct }, "Sending builds payload to jira.");
```
gxueatlassian, line 270: Should we do this in the first line of this method when `logger` is created? 
`const logger = log.child({ jiraHost, gitHubInstallationId, gitHubProduct });`

So that we can save the changes in all other places?

src/github/branch.ts
gxueatlassian: Wait, haha, I see you already added that in the middleware, so I guess if just for adding the `product` in the logs fields, we don't have to do it in all these handlers? Or is there a reason that the `product` set in the middleware doesn't work here?

gxueatlassian (approved):

gxueatlassian (approved):

PR #1549, opened 2022-08-24
Load user config during discovery

thombergs:
This PR adds some missing pieces to the config-as-code feature.

A [previous PR ](https://github.com/atlassian/github-for-jira/pulls?q=is%3Apr+is%3Aclosed+config-as-code) added the feature that when someone changes the contents of the file `.jira/config.yml` in a repo, this config is saved in our database.

This PR adds the same thing during discovery (i.e. at the start of a backfilling), so that the config is available shortly after the backfilling has been started.

### DIFF
--- a/src/services/user-config-service.ts
+++ b/src/services/user-config-service.ts
@@ -56,12 +56,26 @@ export const updateRepoConfig = async (
 /**
  * Returns the config for a given repo.
  */
-export const getRepoConfig = async (subscription: Subscription, repositoryId: number): Promise<Config | undefined> => {
+export const getRepoConfig = async (
+	subscription: Subscription,
+	installationId: InstallationId,
+	repositoryId: number,
+	repoOwner: string,
+	repoName: string
+): Promise<Config | undefined> => {
 	// In the future, we may look in other places for a config than just in the RepoSyncState (for example,
 	// we might fall back to default configs on the level of a subscription or an installation).
 	const repoSyncState = await RepoSyncState.findByRepoId(subscription, repositoryId);
-	// TODO: if repoSyncState is not available, try to get the config file from GitHub!
-	return repoSyncState?.config;
+
+	// Edge case: we don't have a record of the repository in our DB, yet, so we're loading the
+	// config directly from the config file in the GitHub repo.
+	if (!repoSyncState) {
+		const yamlConfig = await getRepoConfigFromGitHub(installationId, repoOwner, repoName);
+		return convertYamlToUserConfig(yamlConfig);
+	}
+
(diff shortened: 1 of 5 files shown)


src/services/user-config-service.ts
```diff
@@ -56,12 +56,26 @@ export const updateRepoConfig = async (
...
 	const repoSyncState = await RepoSyncState.findByRepoId(subscription, repositoryId);
-	// TODO: if repoSyncState is not available, try to get the config file from GitHub!
-	return repoSyncState?.config;
+
+	// Edge case: we don't have a record of the repository in our DB, yet, so we're loading the
+	// config directly from the config file in the GitHub repo.
+	if (!repoSyncState) {
+		const yamlConfig = await getRepoConfigFromGitHub(installationId, repoOwner, repoName);
+		return convertYamlToUserConfig(yamlConfig);
+	}
```
thombergs, line 75: Added handling of this edge case as it might be that we don't have a `RepoSyncState` record in the database, yet, for example, when a customer has many repos and the repo in question is last in line to be "discovered" at the start of a backfill. 

In this case, we fall back to loading the config directly from GitHub.

src/services/user-config-service.ts
```diff
@@ -128,3 +142,21 @@ const updateRepoConfigFromGitHub = async (repoSyncState: RepoSyncState, githubIn
...
+export const updateRepoConfigsFromGitHub = async (repoSyncStates: RepoSyncState[], githubInstallationId: InstallationId): Promise<void> => {
+	await Promise.all(repoSyncStates.map(async (repoSyncState) => {
+		await updateRepoConfigFromGitHub(repoSyncState, githubInstallationId)
+			.catch(err => logger.error({
+				err,
+				githubInstallationId,
+				repositoryId: repoSyncState.repoId
+			}, "error while updating a single repo config"));
+	}));
+};
```
thombergs, line 162: I added this new function to handle multiple repo config updates in parallel.

src/services/user-config-service.ts
```diff
@@ -34,15 +34,15 @@ export const updateRepoConfig = async (
...
+			if (!repoSyncState) {
 				logger.error({
 					githubInstallationId,
 					repositoryId
 				}, "could not find RepoSyncState for repo");
 				return;
 			}
 
-			await updateRepoConfigFromGitHub(repoSyncState, githubInstallationId);
+			await updateRepoConfigsFromGitHub([repoSyncState], githubInstallationId);
```
thombergs, line 45: Switched to the new function so that the new function is covered by the existing tests.

mboudreau (approved):

PR #1551, opened 2022-08-25
Unsafe log stream

joshkay10:
**What's in this PR?**
Data serialization is moved to the stream level, so it's not hashed until it is written, and captures everything!!

There is two steams, **safe** - which hashes all sensitive data no matter what!!
**unsafe** - doesn't hash anything, but it only writes out debug and lower logs(bunyans logging levels). It also send the logs to a seperate logging environment which is locked down with a ACL.

**Why**
We need to hash UGC/PII for privacy reasons.
This is also the second attempt at this solution, this way adds less risk by not making the unsafe stream esoteric.

**How has this been tested?**  
Tested in staging


src/config/logger.test.ts
```diff
@@ -10,19 +10,6 @@ describe("logger behaviour", () => {
 			ringBuffer = new RingBuffer({ limit: 5 });
 		});
 
-		it("should serialize sensitive data as part of getlogger", () => {
```
joshkay10, line 13: Serialization is done on the stream level now and tests have been moved there.

src/config/logger.test.ts
```diff
@@ -55,13 +42,6 @@ describe("logger behaviour", () => {
 		});
 	});
 
-	describe("unsafe logger", () => {
```
joshkay10, line 58: Serialization is done on the stream level now and tests have been moved there.

src/config/logger.ts
```diff
@@ -1,26 +1,8 @@
...
-import { filteringHttpLogsStream } from "utils/filtering-http-logs-stream";
-import { createHashWithSharedSecret } from "utils/encryption";
-import bformat from "bunyan-format";
 import { isArray, isString, merge, omit } from "lodash";
+import { SafeRawLogStream, UnsafeRawLogStream } from "utils/logger-utils";
 
-// For any Micros env we want the logs to be in JSON format.
-// Otherwise, if local development, we want human readable logs.
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-const outputMode: any = process.env.MICROS_ENV ? "json" : "short";
```
joshkay10, line 10: All this stuff has been moved to logger-utils

src/util/logger-utils.ts
```diff
@@ -0,0 +1,82 @@
+import { Writable } from "stream";
+import safeJsonStringify from "safe-json-stringify";
+import bformat from "bunyan-format";
+import { DEBUG } from "bunyan";
+import { createHashWithSharedSecret } from "utils/encryption";
+
+const SENSITIVE_DATA_FIELDS = ["jiraHost", "orgName", "repoName", "userGroup", "userGroup", "aaid", "username"];
```
joshkay10, line 7: As we find more fields that need to be hashed we can just add/remove from this array

src/util/logger-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+		Object.keys(recordClone).forEach(key => {
+			if (SENSITIVE_DATA_FIELDS.includes(key)) {
+				recordClone[key] = createHashWithSharedSecret(recordClone[key]);
+			}
+		});
+		return recordClone;
+	}
+}
+
+export class UnsafeRawLogStream extends RawLogStream {
```
joshkay10, line 66: Unsafe stream... doesn't hash anything and logs all debug level and lower.

src/util/logger-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+			return next();
+		}
+
+		const chunk = safeJsonStringify(record) + "\n";
+		this.writeStream.write(chunk, encoding);
+		next();
+	}
+}
+
+export class SafeRawLogStream extends RawLogStream {
```
joshkay10, line 45: Safe stream: hashes all sensitive datas.

src/util/logger-utils.ts
```diff
@@ -0,0 +1,82 @@
...
+export const filterHttpRequests = (record: Record<string, any>, filteredLoggerName: string) => {
+	const { msg, name } = record;
+	if (name !== filteredLoggerName) {
+		return false;
+	}
+	return !!msg.match(/(GET|POST|DELETE|PUT|PATCH)/);
+};
+
+class RawLogStream extends Writable {
+	private readonly  filteredHttpLoggerName: string;
```
joshkay10, line 23: Unsafe stream: hashes nothing, but only writes out debug level or lower to a locked down logging environment

src/util/logger-utils.test.ts
```diff
@@ -0,0 +1,150 @@
...
+describe("Logger Utils", () => {
+
+	describe("filterHttpRequests", () => {
+
+		it("should return false for msg with http request keyword and not targeted logger", () => {
+			const record = {
+				msg: "GET - Im a filterable msg",
+				name: "NO-FILTEREDNAME"
+			};
+			expect(filterHttpRequests(record, "FILTEREDNAME")).toEqual(false);
```
gxueatlassian, line 13: /nit
I understand `filterHttpRequests` is only exported for testing purpose. I suggest it could be better if the test is not against the implementation detail (`filterHttpRequests`), but instead testing on the `stdoutSpy.mock` not being called if filtered out.

src/config/logger.ts
```diff
@@ -56,34 +38,38 @@ const errorSerializer = (err) => {
...
-	return createHashWithSharedSecret(data);
-};
-
 export const defaultLogLevel: LogLevel = process.env.LOG_LEVEL as LogLevel || "info";
 
+const loggerStream = (safe = true): Logger.Stream => ({
+	type: "raw",
+	stream: safe ? new SafeRawLogStream(FILTERING_FRONTEND_HTTP_LOGS_MIDDLEWARE_NAME) : new UnsafeRawLogStream(FILTERING_FRONTEND_HTTP_LOGS_MIDDLEWARE_NAME),
+	closeOnExit: false
+});
```
mboudreau, line 47: Not sure I'm a fan of this.  I'd probably just create 2 constants, one for a safe log stream and one unsafe, then add those both to the logger.  Unless there's a reason for this?

src/util/logger-utils.ts
```diff
@@ -0,0 +1,81 @@
...
+// Otherwise, if local development, we want human readable logs.
+const outputMode = process.env.MICROS_ENV ? "json" : "short";
+
+//  See https://github.com/probot/probot/issues/1577
+export const filterHttpRequests = (record: Record<string, any>, filteredLoggerName: string) => {
+	const { msg, name } = record;
+	if (name !== filteredLoggerName) {
+		return false;
+	}
+	return !!msg.match(/(GET|POST|DELETE|PUT|PATCH)/);
```
mboudreau, line 18: could this be simplified as `return name === filteredLoggerName && /(GET|POST|DELETE|PUT|PATCH)/.test(msg);`?

src/util/logger-utils.ts
```diff
@@ -0,0 +1,81 @@
...
+};
+
+class RawLogStream extends Writable {
+	private readonly  filteredHttpLoggerName: string;
+	private writeStream: NodeJS.WritableStream;
```
mboudreau, line 23: should the write stream be readonly also?

src/util/logger-utils.ts
```diff
@@ -0,0 +1,81 @@
...
+export class UnsafeRawLogStream extends RawLogStream {
+
+	public constructor(filteredHttpLoggerName: string) {
+		super(filteredHttpLoggerName);
+	}
+
+	public async _write(record: any, encoding: BufferEncoding, next): Promise<void> {
+
+		// Skip any log above DEBUG level
+		if (!record.level || record.level > DEBUG) {
```
gxueatlassian, line 74: Since the `record` is `any`, I am not sure whether the `record.level` is always a number.
Maybe if we really want exact match, we can do 

```javascript
const { level } = record;
if(level === 'DEBUG' || level === 'TRACE' || ( isNaN(level) === false && level > DEBUG ) ){}
```

src/config/logger.ts
```diff
@@ -56,34 +38,38 @@ const errorSerializer = (err) => {
...
+	stream?: NodeJS.WritableStream;
+	serializers?: Serializers;
+	src?: boolean;
+}
+
+export const getLogger = (name: string, options: LoggerOptions = {}): Logger => {
 	return createLogger(merge<Logger.LoggerOptions, LoggerOptions>({
 		name,
-		stream: LOG_STREAM,
+		streams: [ loggerStream(), loggerStream(false) ],
```
gxueatlassian, line 72: Isn't this `getLogger` also used by non frontend middleware code? Then will the `FILTERING_FRONTEND_HTTP_LOGS_MIDDLEWARE_NAME` still make sense?

src/util/logger-utils.ts
```diff
@@ -0,0 +1,81 @@
...
+	}
+
+	private hashSensitiveData(record: Record<string, any>): Record<string, string | undefined> {
+		const recordClone = { ...record };
+		Object.keys(recordClone).forEach(key => {
+			if (SENSITIVE_DATA_FIELDS.includes(key)) {
+				recordClone[key] = createHashWithSharedSecret(recordClone[key]);
+			}
+		});
+		return recordClone;
```
mboudreau, line 61: Couldn't you use something like `omit` to do this for you?

src/util/logger-utils.ts
```diff
@@ -0,0 +1,81 @@
...
+	}
+}
+
+export class UnsafeRawLogStream extends RawLogStream {
+
+	public constructor(filteredHttpLoggerName: string) {
+		super(filteredHttpLoggerName);
+	}
+
+	public async _write(record: any, encoding: BufferEncoding, next): Promise<void> {
```
mboudreau, line 71: is `record:any` correct?

src/util/logger-utils.ts
```diff
@@ -0,0 +1,81 @@
...
+
+		// Skip any log above DEBUG level
+		if (!record.level || record.level > DEBUG) {
+			return next();
+		}
+		// Tag the record do it gets indexed to the _unsafe logging environment
+		record.env_suffix = "unsafe";
```
mboudreau, line 78: if record can be `any` it means that `record` could be a string right?

mboudreau (commented):
Maybe we should pair quickly to make sure my understanding is correct :+1:

src/util/logger-utils.ts
joshkay10: fixed, and the double spacing above it.

src/util/logger-utils.ts
joshkay10: Maybe something like i, i'll have a looksy, basically trying to do an Array.map equivalent on an object here.
Does look a bit long in the tooth though.

src/util/logger-utils.ts
joshkay10: good catch! thats left over from when it was much more wild. The `super({ objectMode: true });` in rawlogstream should force it to always be some type of object, I will put in something more useful 😆

src/util/logger-utils.ts
joshkay10: Yea, am updating the type for record, but still can't guarantee that level will be a number since it intercepts all logging, e.g. bunyan, console.log 

But its still a valid point, I'll do a NaN check

src/util/logger-utils.ts
joshkay10: Luckily a consequence of your other comment will fix this one too, will enforce a Record type type of some type.

src/util/logger-utils.ts
joshkay10: I've gone for `Record<string, any>` its not perfect or tight, but it can have any logging data come through here so can't be too restrictive. Open to alternatives if you can think of anything though

src/config/logger.ts
joshkay10: yes :( you pick on my most annoying thing lol, It was inherited from the previous implementation so was originally trying not to change too many things, but that ship sailed so I will have another look if I can do it cleaner!

src/util/logger-utils.ts
joshkay10: Why yes, it can. Have updated it and confirmed with tests 👍

src/util/logger-utils.test.ts
joshkay10: !!!!! I am having feelings here, I accept what you are saying, I would like to discuss this in a more more general standards kind of discussion though. I think we have a few areas that we could improve how we test and expose things etc.

src/config/logger.ts
joshkay10: I agree it is a bit messy. Have split it out.

src/config/logger.ts
joshkay10: @gxueatlassian I have completely removed the `FILTERING_FRONTEND_HTTP_LOGS_MIDDLEWARE_NAME` const, it has been replaced with a config value. so its much cleaner now its not passing that around everywhere!

gxueatlassian (approved):

mboudreau (approved):

PR #1555, opened 2022-08-30
ARC-1643: remove direct instantiation of GitHubInstallationClient

bgvozdev:
**What's in this PR?**

A small refactoring to get rid of direct instantiation of GitHubInstallationClient

**Why**

We need to introduce proxy config, and except of this place the rest is encapsulated in a single place, `get-github-client-config.ts`

**Added feature flags**

Already feature-flagged by "config-as-code" FF

**Affected issues**  
ARC-1643

**How has this been tested?**  
I'll pair up with Tom some time later.

**Whats Next?**
Introducing proxy config to `get-github-client-config.ts`


src/services/user-config-service.ts
```diff
@@ -137,8 +137,8 @@ const convertYamlToUserConfig = (input?: string): Config => {
 	return output;
 };
 
-const updateRepoConfigFromGitHub = async (repoSyncState: RepoSyncState, githubInstallationId: InstallationId): Promise<void> => {
-	const yamlConfig = await getRepoConfigFromGitHub(githubInstallationId, repoSyncState.repoOwner, repoSyncState.repoName);
+const updateRepoConfigFromGitHub = async (repoSyncState: RepoSyncState, githubInstallationId: InstallationId, jiraHost: string, gitHubAppId?: number): Promise<void> => {
```
gxueatlassian, line 140: Hey, just FYI I previously created a PR regarding the `?` for all these `gitHubAppId`. 
https://github.com/atlassian/github-for-jira/pull/1530

I think it should be fine here, as this is internal method, but only FYI, on external method I think using `gitHubAppId: number | undefined` is better.

gxueatlassian (approved):

src/services/user-config-service.ts
bgvozdev: Thanks @gxueatlassian , changed that to `number | undefined`

gxueatlassian (approved):

PR #1558, opened 2022-08-30
ARC-1590 ARC-1589 connect app create branch action

joshkay10:
**What's in this PR?**
Update the connect manifest to include the create branch action.

This has made the assumption that the path will be **/github/create-branch**

Also put the config behind a feature flag create-branch.

**Why**
So we can create branches!

**Added feature flags**
[create-branch](https://switcheroo.atlassian.com/changes/github-for-jira/create-branch)

**Affected issues**  
ARC-1590 ARC-1589

**How has this been tested?**  
Tested locally and on stage.
**Test A** - Does the Create Branch button for Github appear(when FF serves true)
**Test B** - Does the Create Branch button for Github not appear(when FF serves false)

**Whats Next?**
Base route for create-branch

Screen Shot 2022-08-31 at 9 20 32 AM

### DIFF
--- a/src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
+++ b/src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
@@ -2,8 +2,10 @@ import { Request, Response } from "express";
 import { envVars } from "config/env";
 import { EnvironmentEnum } from "interfaces/common";
 import { compact, map } from "lodash";
+import { booleanFlag, BooleanFlags } from "config/feature-flags";
 
 const instance = envVars.INSTANCE_NAME;
+const jiraHost = `https://${instance}.atlassian.net`;
 const isProd = (instance === EnvironmentEnum.production);
 // TODO: implement named routes (https://www.npmjs.com/package/named-routes) to facilitate rerouting between files
 export const postInstallUrl = "/jira";
@@ -32,7 +34,7 @@ const modules = {
 		url: "https://github.com",
 		actions: {
 			createBranch: {
-				templateUrl: `${envVars.APP_URL}/github/create-branch?issue_key={issue.key}&issue_summary={issue.summary}`
+				templateUrl: `${envVars.APP_URL}/github/create-branch?issue_key={issue.key}&issue_summary={issue.summary}&instance=${instance}`
 			}
 		}
 	},
@@ -165,6 +167,21 @@ const modules = {
 
 export const moduleUrls = compact(map([...modules.adminPages, ...modules.generalPages], "url"));
 
+// Remove this function when CREATE_BRANCH flag is complete
+const moduleModifier = async (modules) => {
+	console.log("MODIFY MEEEE");
+	console.log(await booleanFlag(BooleanFlags.CREATE_BRANCH, false, jiraHost));
+	if (await booleanFlag(BooleanFlags.CREATE_BRANCH, false, jiraHost)) {
+		console.log("BRANCH WAS TRUE");
+		console.log(modules);
+		return modules;
+	}
+	console.log("BRANCH WAS FALSE");
+	const noCreateBranchModule = { ...modules };
+	delete noCreateBranchModule.jiraDevelopmentTool.actions;
+	return noCreateBranchModule;
+};
+
 export const JiraAtlassianConnectGet = async (_: Request, res: Response): Promise<void> => {
 	res.status(200).json({
 		// Will need to be set to `true` once we verify the app will work with
@@ -195,6 +212,6 @@ export const JiraAtlassianConnectGet = async (_: Request, res: Response): Promis
 			"DELETE"
 		],
 		apiVersion: 1,
-		modules
+		modules: await moduleModifier(modules)
 	});
 };
--- a/src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
+++ b/src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
@@ -31,12 +31,7 @@ const modules = {
 		name: {
 			value: "GitHub"
 		},
-		url: "https://github.com",
-		actions: {
-			createBranch: {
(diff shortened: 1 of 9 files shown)


src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
```diff
@@ -29,7 +31,12 @@ const modules = {
 		name: {
 			value: "GitHub"
 		},
-		url: "https://github.com"
+		url: "https://github.com",
+		actions: {
+			createBranch: {
+				templateUrl: `${envVars.APP_URL}/github/create-branch?issue_key={issue.key}&issue_summary={issue.summary}`
```
joshkay10, line 37: If required we can chuck `instance=${instance}` or `jira_host=${jiraHost}` here

test/snapshots/routes/jira/atlassian-connect/jira-atlassian-connect.test.ts.snap
```diff
@@ -153,6 +153,222 @@ Object {
         "value": "GitHub Actions",
       },
     },
+    "jiraDevelopmentTool": Object {
```
joshkay10, line 156: New snapshot since we test true and false on the feature flag

src/routes/jira/atlassian-connect/jira-atlassian-connect.test.ts
```diff
@@ -23,7 +27,16 @@ describe("Atlassian Connect", () => {
 			}));
 		});
 
-		describe("Atlassian Connect", () => {
+		describe.each([true, false])("Atlassian Connect", (createBranch) => {
```
joshkay10, line 30: Testing both variations of the flag

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
krazziekay, line 37: Just something to avoid conflict later on. 
There already is a PR that removes the usage of all the `envVars` -> [link](https://github.com/atlassian/github-for-jira/pull/1557). So might suggest to just use `process.env.APP_URL`

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
```diff
@@ -160,6 +167,16 @@ const modules = {
 
 export const moduleUrls = compact(map([...modules.adminPages, ...modules.generalPages], "url"));
 
+// Remove this function when CREATE_BRANCH flag is complete
+const moduleModifier = async (modules) => {
```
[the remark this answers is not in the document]
krazziekay: Instead of adding and then deleting depending upon the condition, can't we just add conditionally? 
What if we just do the FF check before, if true then we add the `actions` property in the `modules`, and if not keep the `modules` as it is.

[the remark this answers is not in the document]
bgvozdev: Yeah, I agree with @krazziekay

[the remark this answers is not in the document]
krazziekay: Instead of adding and then deleting depending upon the condition, can't we just add conditionally?
What if we just do the FF check before, if true then we add the actions property in the modules, and if not keep the modules as it is.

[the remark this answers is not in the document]
joshkay10: Yeah, I also agree

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
mboudreau: nah, leave it as is until mine is merged.  Either mine is merged first and he has to change his, or his is merged first and I need to fix it in mine.

mboudreau: can we though?

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
```diff
@@ -2,8 +2,10 @@ import { Request, Response } from "express";
 import { envVars } from "config/env";
 import { EnvironmentEnum } from "interfaces/common";
 import { compact, map } from "lodash";
+import { booleanFlag, BooleanFlags } from "config/feature-flags";
 
 const instance = envVars.INSTANCE_NAME;
+const jiraHost = `https://${instance}.atlassian.net`;
```
mboudreau, line 8: instance name is the jira app instance name, not the jira host name.

joshkay10: gahhhhhh

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
joshkay10: no :(

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
```diff
@@ -160,6 +162,18 @@ const modules = {
 
 export const moduleUrls = compact(map([...modules.adminPages, ...modules.generalPages], "url"));
 
+// Remove this function when CREATE_BRANCH flag is complete
+const addCreateBranchAction = async (modules) => {
+	if (await booleanFlag(BooleanFlags.CREATE_BRANCH, false)) {
```
joshkay10, line 167: We don't have access to the Jira host at this point :(

krazziekay (approved):

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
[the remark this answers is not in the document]
joshkay10: Updated to be additive

atrigueiro (approved):

mboudreau (approved):

Harminder84 (approved):

PR #1564, opened 2022-09-01
ARC-1643 refactor config class a bit

bgvozdev:
**What's in this PR?**

Currently we have 2 places where we configure the client with duplicating code: 
 - `src/github/client/github-client.ts` and
 - `src/util/get-github-client-config.ts`

This refactoring will leave in only one place in `get-github-client-config` (after FF is removed). In that place, there's also access to `jiraHost` , which I need in order to configure the proxy (next step).

**Why**

Currently there's no way to configure outbound proxy based on `jiraHost`, because 
1. we don't have it in the place where we do configuration now (client)
2. even if we bring it over, we cannot use it because that's in a constructor and constructors cannot be async (the list of hosts that should be skipped are to be fetched from a FF)
3. this needs to be done in a common place so we could reuse it in GHE config screen, that doesn't use the client

**Added feature flags**
https://switcheroo.atlassian.com/changes/github-for-jira/use-githubconfig-in-base-client

**Affected issues**  
ARC-1643

**How has this been tested?**  
Staging

**Whats Next?**
Remove the FF
Add outbound proxy config based on FF value (jiraHost)

### DIFF
--- a/src/github/client/github-app-client.ts
+++ b/src/github/client/github-app-client.ts
@@ -22,9 +22,9 @@ export class GitHubAppClient extends GitHubClient {
 	private readonly appToken: AuthToken;
 
 	constructor(
-		gitHubConfig?: GitHubConfig,
+		gitHubConfig?: GitHubConfig, // becomes mandatory once we remove the FF
 		logger?: Logger,
-		baseUrl?: string,
+		baseUrl?: string, // will go away once gitHubConfig is mandatory
 		appId = envVars.APP_ID,
 		privateKey = PrivateKey.findPrivateKey() || ""
 	) {
--- a/src/github/client/github-app-client.ts
+++ b/src/github/client/github-app-client.ts
@@ -22,9 +22,9 @@ export class GitHubAppClient extends GitHubClient {
 	private readonly appToken: AuthToken;
 
 	constructor(
-		gitHubConfig?: GitHubConfig, // becomes mandatory once we remove the FF
+		gitHubConfig: GitHubConfig,
 		logger?: Logger,
-		baseUrl?: string, // will go away once gitHubConfig is mandatory
+		baseUrl?: string, // goes away once it goes away in the parent class
 		appId = envVars.APP_ID,
 		privateKey = PrivateKey.findPrivateKey() || ""
 	) {
--- a/src/github/client/github-app-client.ts
+++ b/src/github/client/github-app-client.ts
@@ -9,7 +9,7 @@ import * as PrivateKey from "probot/lib/private-key";
 import { envVars } from "config/env";
 import { AuthToken } from "~/src/github/client/auth-token";
 import { GITHUB_ACCEPT_HEADER } from "~/src/util/get-github-client-config";
-import { GitHubClient } from "./github-client";
+import { GitHubClient, GitHubConfig } from "./github-client";
 
 
 /**
@@ -22,12 +22,13 @@ export class GitHubAppClient extends GitHubClient {
 	private readonly appToken: AuthToken;
 
 	constructor(
+		gitHubConfig?: GitHubConfig,
 		logger?: Logger,
 		baseUrl?: string,
 		appId = envVars.APP_ID,
 		privateKey = PrivateKey.findPrivateKey() || ""
 	) {
-		super(logger, baseUrl);
+		super(gitHubConfig, logger, baseUrl);
 		this.appToken = AppTokenHolder.createAppJwt(privateKey, appId);
 
 		this.axios.interceptors.request.use(setRequestStartTime);
--- a/src/github/client/github-client.test.ts
+++ b/src/github/client/github-client.test.ts
@@ -4,13 +4,14 @@ import "config/env";
 
 import * as axios from "axios";
 import { GitHubClient, GitHubConfig } from "~/src/github/client/github-client";
+import { booleanFlag } from "config/feature-flags";
 
 jest.mock("axios");
 
 jest.mock("config/feature-flags");
 
 class TestGitHubClient extends GitHubClient {
-	constructor(config?: GitHubConfig) {
+	constructor(config: GitHubConfig) {
 		super(config);
 	}
 	public doTestGraphqlCall() {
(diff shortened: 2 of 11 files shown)


src/github/client/github-client.test.ts
```diff
@@ -1,27 +1,86 @@
...
+	}
+	public doTestGraphqlCall() {
+		return this.graphql("foo", {});
+	}
+}
+
+const TEST_API_URL = 'http://api.myBaseUrl.com';
+const TEST_GRAPHQL_URL = 'http://graphql.myBaseUrl.com';
+
+const TEST_GITHUB_CONFIG = {
```
rachellerathbone, line 25: Should we define this object in setup? Then it's globally accessible and doesn't need to be defined in individual tests.

src/github/client/github-installation-client.ts
```diff
@@ -39,19 +38,18 @@ import { GitHubClient } from "./github-client";
...
  */
 export class GitHubInstallationClient extends GitHubClient {
-	private readonly appTokenHolder: AppTokenHolder;
 	private readonly installationTokenCache: InstallationTokenCache;
 	public readonly githubInstallationId: InstallationId;
 	public readonly gitHubServerAppId?: number;
 
 	constructor(
 		githubInstallationId: InstallationId,
+		githubConfig: GitHubConfig,
```
rachellerathbone, line 47: Nit: should have uppercase H in github. I know there's a lot of places in the code where that's not the case (need to clean that up) but should strive to avoid in new code.

src/github/client/github-client.test.ts
bgvozdev: Nah, people will forget anyway and will create their own constants. I don't think we have a single place for such thing, do we?

src/github/client/github-installation-client.ts
bgvozdev: Well spotted! Thanks. Fixed :)

src/github/client/github-app-client.ts
```diff
@@ -22,12 +22,13 @@ export class GitHubAppClient extends GitHubClient {
 	private readonly appToken: AuthToken;
 
 	constructor(
+		gitHubConfig: GitHubConfig,
 		logger?: Logger,
-		baseUrl?: string,
+		baseUrl?: string, // goes away once it goes away in the parent class
```
mboudreau, line 27: goes away because we're using the baseUrl in the config instead right?

src/github/client/github-client.ts
```diff
@@ -15,20 +28,37 @@ export class GitHubClient {
...
 	constructor(
+		gitHubConfig: GitHubConfig,
 		logger: Logger = getLogger("gitHub-client"),
-		baseUrl?: string
+		baseUrl?: string // goes away when we remove FF
 	) {
 		this.logger = logger;
 
+		// constructor cannot be async, therefore cannot use await
+		booleanFlag(BooleanFlags.USE_GITHUB_CONFIG_IN_BASE_CLIENT, false)
```
mboudreau, line 38: I'm not sure an FF is really needed for this change as it would be thoroughly tested in unit test and pollinator.  Any problem here would fail spectacularly.  There's also the issue of potential race conditions with this...

src/github/client/github-client.ts
```diff
@@ -65,4 +95,26 @@ export class GitHubClient {
 			proxy: false
 		};
 	};
+
+	protected async graphql<T>(query: string, config: AxiosRequestConfig, variables?: Record<string, string | number | undefined>): Promise<AxiosResponse<GraphQlQueryResponse<T>>> {
```
mboudreau, line 99: love that you moved this to the superclass :+1: 

We need to do the same with all other base function like get, post, put, delete, etc.  For a later PR :)

src/github/client/github-client.test.ts
mboudreau: we do, in `test/setup.ts`.  It's where we set globals for tests like `jiraHost`.  I could see this being useful to have a `githubConfig` and `gheConfig` object premade to be easily used in tests :)

mboudreau (approved):
Looks good, few things that could be improved on, but no blockers.

src/github/client/github-app-client.ts
bgvozdev: yep

src/github/client/github-installation-client.ts
```diff
@@ -150,7 +147,7 @@ export class GitHubInstallationClient extends GitHubClient {
 	 */
 	public getRepositoriesPage = async (per_page = 1, cursor?: string): Promise<GetRepositoriesResponse> => {
 		try {
-			const response = await this.graphql<GetRepositoriesResponse>(GetRepositoriesQuery, {
+			const response = await this.graphql<GetRepositoriesResponse>(GetRepositoriesQuery, await this.installationAuthenticationHeaders(), {
```
gxueatlassian, line 150: I'd think making `appAuthenticationHeaders()` or `getAxiosConfig()` as abstract in the parent class, (but impl it in child class) will be cleaner, so that in the actual call of `this.graphql`, we don't need to remember to pass in the config.

src/util/get-github-client-config.ts
```diff
@@ -75,21 +75,20 @@ export async function getGitHubHostname(jiraHost: string, gitHubAppId: number) {
...
+		: new GitHubAppClient(gitHubClientConfig, logger);
 }
 
 /**
  * Factory function to create a GitHub client that authenticates as the installation of our GitHub app to get
  * information specific to an organization.
  */
 export async function createInstallationClient(gitHubInstallationId: number, jiraHost: string, logger: Logger, gitHubAppId: number | undefined): Promise<GitHubInstallationClient> {
-
+	const gitHubClientConfig = await getGitHubClientConfigFromAppId(gitHubAppId, jiraHost);
```
gxueatlassian, line 87: Maybe we should put the FF here, and pass `null` if FF is off.
Async in side the gitHubClient constructor is kind of weird.

src/github/client/github-client.test.ts
bgvozdev: Gotcha, will change then. Thanks, @mboudreau !

src/github/client/github-client.ts
bgvozdev: According to https://hello.atlassian.net/wiki/spaces/BPP/pages/128871659/Feature+Delivery+FAQ , we should use feature flags unless there's a good excuse not to ¯\_(ツ)_/¯

src/github/client/github-installation-client.ts
bgvozdev: Yeah, that would be nice, but I think it is better to do that once we have more methods that will use it. Currently, that's only one and I prefer to be explicit here, therefore calling this YAGNI for now 😋

src/util/get-github-client-config.ts
bgvozdev: Sorry, I'm too lazy to do that. I'd need to do that in 3 places for 3 clients, which means ideally I should abstract this as some helper... given that this code will live a few hours... nah

bgvozdev:
Merging because I can :) fixing nits in the follow-up PR.

PR #1565, opened 2022-09-01
ARC-1611 create branch endpoint

joshkay10:
**What's in this PR?**
New /github/create-branch POST endpoint to create a new branch!!!

**Why**
Part of the the create branch flow, but also unblock the submit from UI work

**Affected issues**  
ARC-1611

**How has this been tested?**  
New test script written.
tested locally and on stage by manually calling.

**Whats Next?**
Submit create branch form


src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -0,0 +1,34 @@
...
+	}
+
+	if (!owner || !repo || !sourceBranchName || !newBranchName) {
+		res.status(400).json({ err: "Missing required data." });
+		return;
+	}
+
+	try {
+		const gitHubUserClient = await createUserClient(githubToken, jiraHost, req.log, undefined);
+		const { data: baseBranchRef } = await gitHubUserClient.getReference(owner, repo, sourceBranchName);
```
joshkay10, line 20: If we send the SHA for the base branch from the UI we can skip this call entirely

src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -0,0 +1,34 @@
...
+		return;
+	}
+
+	if (!owner || !repo || !sourceBranchName || !newBranchName) {
+		res.status(400).json({ err: "Missing required data." });
+		return;
+	}
+
+	try {
+		const gitHubUserClient = await createUserClient(githubToken, jiraHost, req.log, undefined);
```
krazziekay, line 19: Passing `undefined` would only work for GitHub cloud, what about GitHub servers? 
For GHES, we will be passing a `gitHubServerAppId`(primary key), which can be passed here.

src/routes/github/create-branch/github-create-branch-post.ts
krazziekay: The list of source branch in the frontend will need to be pulled from another API request. So, I think we will have `SHA` present for the branch and in that case we can directly pass the `SHA` value for the source branch and can skip this extra call.

src/routes/github/create-branch/github-create-branch-post.ts
joshkay10: yeah, have left out the GHES stuff intentionally until we plug it in proper.

krazziekay: /nit Please add a TODO so that its not lost or forgotten...

src/routes/github/create-branch/github-create-branch-post.ts
Harminder84: `SHA` value for source branch keeps on changing if branch updated with commit, PR and so on. It may be possible that new branch created with old SHA value if source branch updated from the time source branches pulled and hit 'create' button. I think, it is better to fetch `SHA` value just before creating branch.

src/routes/github/create-branch/github-create-branch-post.ts
joshkay10: word

src/routes/github/create-branch/github-create-branch-post.ts
joshkay10: Good shout

src/github/client/github-user-client.ts
```diff
@@ -58,7 +70,22 @@ export class GitHubUserClient extends GitHubClient {
...
+		return await this.get<Octokit.GitGetRefResponse>(`/repos/{owner}/{repo}/git/refs/heads/{branch}`, {
+			urlParams: {
+				owner,
+				repo,
+				branch
+			}
+		});
 	}
+
+	public async createBranch(owner: string, repo: string, body: CreateBranchBody): Promise<AxiosResponse<Octokit.GitCreateRefResponse>> {
```
Harminder84, line 83: Can we call this method `createReference` so that it should be in sync with get branch method name which is `getReference`?

joshkay10: Fair, have updated.

Harminder84 (approved):

PR #1570, opened 2022-09-02
ARC-1601 - Get initial repo list

krazziekay:
**What's in this PR?**
- Added new graphql endpoint to fetch the list of owned repositories
- Pulled these list in `create-branch-get`

**Why**
- For displaying the first initial list of repositories

**How has this been tested?**  
- Local
- Staging

**Screenshot**
image


src/github/client/github-user-client.ts
```diff
@@ -42,10 +44,30 @@ export class GitHubUserClient extends GitHubClient {
...
 
+	public async getUserRepositories(per_page = 20, cursor?: string): Promise<GetRepositoriesResponse> {
+		try {
+			const response = await this.graphql<GetRepositoriesResponse>(GetRepositoriesQuery, {
+				per_page,
+				cursor
+			});
+			return response.data.data;
+		} catch (err) {
+			err.isRetryable = true;
```
joshkay10, line 66: You can drop the isRetryable, its used in the installation/backfill code.

src/routes/github/create-branch/github-create-branch-get.test.ts
```diff
@@ -3,6 +3,7 @@ import supertest from "supertest";
 import { getLogger } from "config/logger";
 import { getFrontendApp } from "~/src/app";
 import { getSignedCookieHeader } from "test/utils/cookies";
+import { GetRepositoriesQuery } from "~/src/github/client/github-queries";
```
joshkay10, line 6: if you slap 
```
defaultBranchRef {
            name
          }
```

into the `GetRepositoriesQuery` we can get the default branch name and pass that through as the first selection in the branch list.

Can do this in seperate PR if you wanna.

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -1,31 +1,34 @@
...
-		githubToken
+		githubToken,
+		hostname
 	} = res.locals;
 
 	if (!githubToken) {
 		return next(new Error(Errors.MISSING_GITHUB_TOKEN));
 	}
 
+	const gitHubAppClient = new GitHubUserClient(githubToken, req.log, hostname);
```
joshkay10, line 21:
```suggestion
	const gitHubUserClient = new GitHubUserClient(githubToken, req.log, hostname);
```

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -1,31 +1,34 @@
...
+	const response = await gitHubAppClient.getUserRepositories();
```
joshkay10, line 22:
```suggestion
	const response = await gitHubUserClient.getUserRepositories();
```

src/routes/github/create-branch/github-create-branch-get.ts
joshkay10, line 21: Theres a definite possibility that we change this user client to the installation client. But for now it does what we need so we can revisit that later on. More of a by the way than anything to action.

src/routes/github/create-branch/github-create-branch-get.test.ts
joshkay10: in fact, there is a story [ARC-1608](https://softwareteams.atlassian.net/browse/ARC-1608) which we can cover it in

krazziekay: I tried using `defaultBranchRef`, it just returns null and a Forbidden error. Lets just do it separate in Arc-1608

src/github/client/github-user-client.ts
krazziekay: Removed

src/routes/github/create-branch/github-create-branch-get.ts
krazziekay: Updated

src/routes/github/create-branch/github-create-branch-get.ts
krazziekay: Updated

src/routes/github/create-branch/github-create-branch-get.ts
krazziekay: We need a specific `githubInstallationId` for using `GitHubInstallationClient`. And I don't think we will have just one `githubInstallationId`. Can discuss and revisit later.

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -1,31 +1,34 @@
...
 const branches = [{ id: 1, name: "first-branch" }, { id: 2, name: "second-branch" }, { id: 3, name: "third-branch" }];
 
 
 export const GithubCreateBranchGet = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
 	const {
 		jiraHost,
-		githubToken
+		githubToken,
+		gitHubAppId
 	} = res.locals;
```
Harminder84, line 15: gitHubAppId should be fetched from `res.locals.gitHubAppConfig`

views/github-create-branch.hbs
```diff
@@ -48,7 +48,7 @@
       <aui-label for="repo">Repository</aui-label>
       <select id="ghRepo" name="repo" class="full-width-field" placeholder="Select a repository">
         {{#each repos}}
-          <option value="{{id}}">{{org}} / {{name}}</option>
+          <option value="{{node.id}}">{{node.owner.login}} / {{node.name}}</option>
```
Harminder84, line 51: We need `orgName` and `repo` for creating branch. will it make sense to put this information in value as well?  We're not using id anywhere.

src/github/client/github-user-client.ts
```diff
@@ -58,6 +59,19 @@ export class GitHubUserClient extends GitHubClient {
...
+				cursor
+			});
+			return response.data.data;
+		} catch (err) {
+			this.logger.error({ err }, "Could not fetch repositories");
+			throw err;
+		}
+	}
+
 	public getMembershipForOrg = async (org: string): Promise<AxiosResponse<Octokit.OrgsGetMembershipResponse>> => {
```
Harminder84, line 75: Unit test case missing!

src/routes/github/create-branch/github-create-branch-get.ts
krazziekay: Updated.

views/github-create-branch.hbs
krazziekay: Replaced the `id` by `full_name`(which is the same value as `owner.login / name`)

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -1,31 +1,34 @@
...
-const repos = [{ id: 1, name: "first-repo", org: "org-1" }, { id: 2, name: "second-repo", org: "org-1"  }, { id: 3, name: "third-repo", org: "org-2" }];
 const branches = [{ id: 1, name: "first-branch" }, { id: 2, name: "second-branch" }, { id: 3, name: "third-branch" }];
 
 
 export const GithubCreateBranchGet = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
 	const {
 		jiraHost,
-		githubToken
+		githubToken,
+		gitHubAppConfig
```
gxueatlassian, line 14: If we assume `gitHubAppConfig` is available, that means we have the `uuid` present in the url /github/uuid/xxx, and that means we will have the gitHubInstallationId, that means we don't have to use the user token?

krazziekay: That would work for Github enterprise, but not for cloud. `uuid` will be undefined for GH cloud users and we won't be able to recognize the exact user.

src/github/client/github-user-client.ts
krazziekay: Ignoring for now, since there is no test case for this file yet. Can tackle this problem in the future...

joshkay10 (approved):
Noice, Maybe chuck a card on Jira to make sure we don't forget those tests!?

Harminder84 (approved):

PR #1572, opened 2022-09-05
Arc-1603 - Autofill branch name with issue key and summary

krazziekay:
**What's in this PR?**
- Used the `issue_key` and `issue_summary` passed in the query parameters to display in the view.

**How has this been tested?**  
- Local

**Screenshot**
image


Harminder84:
I'm assuming this PR will auto populate branch name as PR title suggests. Branch name text field is still shows as empty field initially. What is the motive of this PR?

krazziekay:
> I'm assuming this PR will auto populate branch name as PR title suggests. Branch name text field is still shows as empty field initially. What is the motive of this PR?

The motive is to display the issue key in the view. 
Good catch on the input field. I have added the same issue key as the default name for the branch.

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -17,10 +17,16 @@ export const GithubCreateBranchGet = async (req: Request, res: Response, next: N
 		return next(new Error(Errors.MISSING_GITHUB_TOKEN));
 	}
 
+	const { issue_key: key, issue_summary: summary } = req.query;
+	if (!key ||!summary) {
```
gxueatlassian, line 21: /nit
From UX's point of view, if somehow key || summary is missing, it wouldn't be too bad to still show a page with blank prefill, instead of throwing error to user?

gxueatlassian (dismissed):

src/routes/github/create-branch/github-create-branch-get.ts
joshkay10, line 21: /nit 
```suggestion
	if (!key || !summary) {
```

joshkay10, line 21: It's not strictly necessary that the issue summary exists, the key is a must, the summary is nice to have. None of this is on the story though lolol.

As a side note, there is also work thats happening in a different team that can manipulate these query params, so it can pass different issue values.

Even if the above work gets done, the only required field that remains is **Issue key**.

```suggestion
	if (!key) {
```

krazziekay: We can ignore summary since we are not using it, but I think we need the key.  Will remove the `summary` part.

views/github-create-branch.hbs
```diff
@@ -63,7 +63,7 @@
     <div class="field-group">
       <aui-label for="branch-name">Branch name</aui-label>
       <input class="text full-width-field" type="text" data-aui-validation-field required name="branch-name"
-             placeholder="Enter branch name"/>
+             value="{{issue.key}}" placeholder="Enter branch name"/>
```
joshkay10, line 66: Add the issue summary in the default name... if it exists.
kebab cased and all lower case to incase it comes through strange.
e.g. `key-99-summary-of-goodies`

strip out all none alphanumeric characters toooooooo

gxueatlassian (approved):

views/github-create-branch.hbs
krazziekay: Combined summary and key for the branch name.
Added a new helper for removing non alphanumeric chars.

Harminder84 (approved):

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -17,10 +21,21 @@ export const GithubCreateBranchGet = async (req: Request, res: Response, next: N
 		return next(new Error(Errors.MISSING_GITHUB_TOKEN));
 	}
 
+	const { issue_key: key, issue_summary: summary } = req.query;
+	if (!key) {
+		return next(new Error(Errors.MISSING_ISSUE_KEY));
+	}
+
+	const branchSuffix = summary ? replaceSpaceWithHyphenHelper(replaceNonAlphaNumericWithHyphenHelper(summary as string)) : "";
```
Harminder84, line 29: Should we put limit on branch name length?

joshkay10: we had a google and github has no limit so its a free for all

Harminder84: @joshkay10 , we'll end up with a very long branch name in some cases. I think, it is better to truncate description to have some relevant branch name. your thoughts?

PR #1583, opened 2022-09-12
Remove the config-as-code feature flag

thombergs:


src/github/push.test.ts
```diff
@@ -65,7 +65,10 @@ describe("PushWebhookHandler", ()=>{
 			repository: {} as GitHubRepository, //force it as not required in test
 			commits: [{
 				id: "commit-1",
-				message: "ARC-0001 some commit message"
+				message: "ARC-0001 some commit message",
+				added: [],
+				modified: [],
+				removed: []
```
thombergs, line 71: Had to add these three fields from the official GitHub push webhook type to make the test `push.test.ts` pass. It seems that the feature flag wasn't covered in that particular test 😬.

src/github/push.ts
```diff
@@ -33,15 +32,13 @@ export const pushWebhookHandler = async (context: WebhookContext, jiraClient, _u
...
-			const modifiedFiles = context.payload?.commits?.reduce((acc, commit) =>
-				([...acc, ...commit.added, ...commit.modified, ...commit.removed]), []);
-			// TODO: this call must be updated to support GitHub Server events
-			await updateRepoConfig(subscription, payload.repository.id, getInstallationId(gitHubInstallationId), modifiedFiles);
-		} else {
-			context.log.warn("could not load user config because subscription does not exist");
-		}
+	if (subscription) {
+		const modifiedFiles = context.payload?.commits?.reduce((acc, commit) =>
+			([...acc, ...commit.added, ...commit.modified, ...commit.removed]), []);
```
thombergs, line 37: This is the spot that crashed when removing the FF, because `commit.added` wasn't set

src/interfaces/github.ts
```diff
@@ -231,6 +231,9 @@ interface GitHubInstallation {
 interface GitHubCommit {
 	id: string;
 	message: string;
+	added: string[],
+	modified: string[],
+	removed: string[]
```
thombergs, line 236: Had to add these to the type for the test to pass

gxueatlassian (dismissed):

gxueatlassian (approved):

PR #1585, opened 2022-09-12
Arc 1354 add ff to write to cryptor

gxueatlassian:
**What's in this PR?**
Add logic to use FF to write to new secret column only.

**Why**
To deprecate the old col

**Added feature flags**
"write-only-to-new-cryptor-shared-secret"

**Affected issues**  
ARC-1354

**How has this been tested?**  
Unit test.
Stage (WIP)

**Whats Next?**
Rollout the FF, and then tidy up.

### DIFF
--- a/src/github/client/github-client.test.ts
+++ b/src/github/client/github-client.test.ts
@@ -50,13 +50,13 @@ describe("GitHub Client", () => {
 		expect(outboundProxyConfig.httpsAgent).toBeUndefined();
 	});
 
-	describe('config object', () => {
-		const TEST_API_URL = 'http://api.myBaseUrl.com';
-		const TEST_GRAPHQL_URL = 'http://graphql.myBaseUrl.com';
+	describe("config object", () => {
+		const TEST_API_URL = "http://api.myBaseUrl.com";
+		const TEST_GRAPHQL_URL = "http://graphql.myBaseUrl.com";
 
 		const TEST_GITHUB_CONFIG = {
-			hostname: 'myHostname',
-			baseUrl: 'http://myBaseUrl.com',
+			hostname: "myHostname",
+			baseUrl: "http://myBaseUrl.com",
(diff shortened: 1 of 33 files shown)


src/github/client/github-client.test.ts
```diff
@@ -50,13 +50,13 @@ describe("GitHub Client", () => {
 		expect(outboundProxyConfig.httpsAgent).toBeUndefined();
 	});
 
-	describe('config object', () => {
-		const TEST_API_URL = 'http://api.myBaseUrl.com';
-		const TEST_GRAPHQL_URL = 'http://graphql.myBaseUrl.com';
+	describe("config object", () => {
```
gxueatlassian, line 53: I do this before commit `yarn lint:ts --fix`, not sure why there was `'`, but seems the linter always change it to `"`

src/github/issue-comment.test.ts
```diff
@@ -21,7 +21,7 @@ describe("Issue Comment Webhook", () => {
 		await Installation.create({
 			jiraHost,
 			clientKey: "client-key",
-			sharedSecret: "shared-secret"
+			encryptedSharedSecret: "shared-secret"
```
gxueatlassian, line 24: So these tests are not directly related to the change.
But instead of mocking FF to true and/or false for all these test, I simply convert it to using the new col only, so that it is easier and closer to the final state we want.

src/models/github-server-app.ts
```diff
@@ -38,11 +38,6 @@ export class GitHubServerApp extends EncryptedModel {
 	}
 
 	async getEncryptContext() {
-		//For example: we can call database to fetch sharedSecret for use as EncryptionContext
```
gxueatlassian, line 41: No idea that the he** is this block of comments? lol

mboudreau (approved):

PR #1586, opened 2022-09-12
ARC-1607 -  UI - Search for repos dropdown

krazziekay:
**What's in this PR?**
- Removed the link in the sub-header.
- Added a lil gap at the top of the page.
- Debounce method added.
- Key up event for the input added

**Why**
- Adding frontend event handlers for select2 input.

**How has this been tested?**  
- Locally

**Screenshot**
Screen Shot 2022-09-12 at 2 14 58 pm


static/js/github-create-branch.js
```diff
@@ -31,17 +31,37 @@ const toggleSubmitDisabled = (bool) => {
...
 		event.preventDefault();
 		createBranchPost();
 	});
-});
+}).on('keyup', '.select2-input', debounce(event => {
+  const parentContainer = $(event.target).parent().parent();
+  const userInput = event.target.value;
+
+  if (userInput) {
+    if (parentContainer.hasClass("aui-select2-dropdown-server")) {
```
Harminder84, line 53: Can't we use select's component`query` event to handle this instead of relying on CSS classes?

krazziekay: To use `query` we would have to use `input type="hidden"`, which would just add complexity when trying to add more options based off from searching. 

image

Harminder84: What complexities it will add if we use `input type="hidden"`? This is the recommended way to use.

krazziekay: I misunderstood its usage before.
Updated to using `query`.

static/js/github-create-branch.js
```diff
@@ -31,17 +31,35 @@ const toggleSubmitDisabled = (bool) => {
 }
 
 $(document).ready(() => {
+	// Fetching the list of default repos
+	const defaultRepos = $(".default-repos").map((_, option) => ({
+		id: $(option).html(),
+		text: $(option).html()
+	})).toArray();
+
```
Harminder84, line 39: Is it worth to create a seperate endpoint that will return default repo's?  Then we don't need this workaround.

krazziekay: I don't like this idea of separate endpoint just to get the default repos, as it adds loading time.
Thoughts @joshkay10

Harminder84: This will improve the inital rendering time of `create-branch` page. Moreover, we can leverage repo's search endpoint. That will return default repo's if search string is empty.

joshkay10: This ones worth a chat at the catchup.

I like the repos being available and not doing another request from here.

Initial rendering speed is a good call, we might have a couple more calls on the backend that we could batch anyway so could get some time back there.

As for the above work around.... dunno, could pass the data in, in a state that it doesn't need to be transformed but do we gain much?

joshkay10 (approved):
I'm happy to do it this way, maybe once the search endpoint is done we can re-evaluate again.

PR #1588, opened 2022-09-12
ARC-1608 Get branches endpoint

Harminder84:
**What's in this PR?**
Implented an endpint that return all the branches for a given `org` and `repo`

**Why**
This endpoint will be used to display branches in drop down for create-branch UI.

**Affected issues**  
[ARC-1608]

**How has this been tested?**  
Unit test cases added

[ARC-1608]: https://softwareteams.atlassian.net/browse/ARC-1608

### DIFF
--- a/src/github/client/github-queries.ts
+++ b/src/github/client/github-queries.ts
@@ -218,6 +218,20 @@ export type getBranchesResponse = {
 	}
 };
 
+export type getBranchesNameResponse = {
+	repository: {
+		refs: {
+      totalCount: number,
+			edges: {
+				cursor: string;
+				node: {
+					name: string;
+				}
+			}[]
+		}
+	}
+};
+
 export const getBranchesQueryWithChangedFiles = `query ($owner: String!, $repo: String!, $per_page: Int!, $commitSince: GitTimestamp, $cursor: String) {
     repository(owner: $owner, name: $repo) {
       refs(first: $per_page, refPrefix: "refs/heads/", after: $cursor) {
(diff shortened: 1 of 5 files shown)


src/routes/github/create-branch/github-branches-get.ts
```diff
@@ -0,0 +1,40 @@
+import { Request, Response } from "express";
+import { createUserClient } from "~/src/util/get-github-client-config";
+
+export const GithubBranchesGet = async (req: Request, res: Response): Promise<void> => {
+	const { githubToken, jiraHost, gitHubAppConfig } = res.locals;
+
+	if (!githubToken || !jiraHost || !gitHubAppConfig) {
```
joshkay10, line 7: We can pull this out `|| !jiraHost` we can't guarantee we will have it :( so we've set it optional everywhere

src/routes/github/create-branch/github-branches-get.ts
```diff
@@ -0,0 +1,40 @@
...
+		res.status(400).json({ err: "Missing required data." });
+		return;
+	}
+
+	try {
+		const gitHubUserClient = await createUserClient(githubToken, jiraHost, req.log, gitHubAppConfig.gitHubAppId);
+		let allBranchesFetched = false;
+		let allBranches;
+		let cursor;
+		while (!allBranchesFetched) {
```
joshkay10, line 23: Unsure here, probably worth a chat at tomorrows catchup!!

Just concerned if we have super large amount of branches, even 1000+ we have to wait a long time for it.

We may still land here....

Other options we can add to and discuss tomorrow:
- Get the 100 most recent branches
- Have the data streaming in behind the scenes so its non-blocking
- Get the 100 most recent branches plus a manual entry if can't find it??

joshkay10 (changes requested):
Like everything I see, just want to make sure we chat about the best way to handle massive amounts of branches in src/routes/github/create-branch/github-branches-get.ts

src/routes/github/create-branch/github-branches-get.ts
joshkay10: Sweet, so as I see it. we can drop the while loop and just return the 100.

We can create a seperate story to handle the UI and the branch validation if its manual.

src/github/client/github-queries.ts
```diff
@@ -218,6 +218,20 @@ export type getBranchesResponse = {
 	}
 };
 
+export type getBranchesNameResponse = {
+	repository: {
+		refs: {
+      totalCount: number,
```
joshkay10, line 224: Can drop total count now

Harminder84: It may be helpful if we want to tweak UI (display message or make drop down editable) if branches are more than 100

joshkay10 (approved):

PR #1589, opened 2022-09-12
ARC-1598 Show issue icon and convert key into link

Harminder84:
**What's in this PR?**
Show issue icon and covert issue key into hyperlink in sub heading

**Why**
Updated UI as per design

**Affected issues**  
[ARC-1598]

**How has this been tested?**  
Local

[ARC-1598]: https://softwareteams.atlassian.net/browse/ARC-1598


krazziekay (commented):
This PR Is not needed.
We agreed to not add the link([discussed here](https://atlassian.slack.com/archives/C04079L8Y10/p1662942465810629)) and the icon for the issue in the subheader because we cant get the value of 'jiraHost'.

PR #1593, opened 2022-09-12
Pr resync script

joshkay10:
**What's in this PR?**
A Python script that calls the /resync endpoint.
Currently it is targeting Pull requests only - as per the requirement from Saiyans

**Why**
This gives us the ability to run the endpoint from a CSV.
Additionally it outputs results to output.csv so we stop the process we can pickup where we left off. (much like the DSS version)

**How has this been tested?**  
I've run it against stage with various datasets


resync-from-csv.py
```diff
@@ -0,0 +1,210 @@
...
+    for row in output_reader:
+        processed.add(Installation.from_dict(row))
+    output_writer = csv.DictWriter(output_file, fieldnames=['installation_id', 'status'])
+    input_reader = csv.DictReader(input_file, fieldnames=['installation_id'])
+
+    batch = []
+    batchRow = ''
+    batchCount = 0
+
+    for row in input_reader:
```
joshkay10, line 172: The batching code in this section is pretty cumbersome, could do with some proper refactoring but it works, and can improve the script over time.... was my thinking. But aware of the inefficiencies.

resync-from-csv.py
```diff
@@ -0,0 +1,210 @@
...
+    installationId: str
+
+    @staticmethod
+    def from_dict(d):
+        return Installation(d['installation_id'])
+
+def process_installation(env: Environment, installationsString) -> bool:
+    url = '{}/api/resync'.format(env.github_for_jira_url)
+
+#   Convert the string of ids to a actual integer list
```
bgvozdev, line 120: nit: I don't think we need clarifying comments. If the code is not clear, we should refactor ;)

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -7,6 +7,18 @@ export const GithubCreateBranchGet = async (req: Request, res: Response, next: N
 		githubToken
 	} = res.locals;
 
+	console.log('res.locals');
```
bgvozdev, line 10: :wat:

etc/scripts/resync-from-csv.py
```diff
@@ -0,0 +1,203 @@
...
+            continue
+
+        batchRow.append(int(installation.installationId))
+
+        if len(batchRow) > args.batchsize:
+          batch.append(batchRow)
+          batchRow = []
+
+    if len(batchRow) > 0:
+        batch.append(batchRow)
```
bgvozdev, line 186: /nit probably could've been simplified by first reading all data in one array and then breaking it to chunks with a function like this one: https://stackoverflow.com/a/312464/1432640

bgvozdev (dismissed):

bgvozdev (approved):

PR #1594, opened 2022-09-13
ARC-1643: GHES connect page to call the server via outbound proxy

bgvozdev:
**What's in this PR?**
Adding dummy anonymous client with outbound proxy properly configured that is used on GHES connect page

**Why**
Because currently GHES connect page using axios directly, which doesn't provide outbound proxy configuration

**Added feature flags**
`use-refactored-config-builder`

**Affected issues**  
ARC-1643

**How has this been tested?**  
locally, staging (when deployed)

**Whats Next?**
FF cleanup


src/util/get-github-client-config.ts
```diff
@@ -65,7 +67,64 @@ async function calculateProxyBaseUrl(jiraHost: string, gitHubBaseUrl: string | u
 	return envVars.PROXY;
 }
 
-export const getGitHubClientConfigFromAppId = async (gitHubAppId: number | undefined, logger: Logger, jiraHost: string): Promise<GitHubClientConfig> => {
+async function buildGitHubServerConfig(githubServerBaseUrl: string, jiraHost: string, logger: Logger): Promise<GitHubConfig> {
```
bgvozdev, line 70: The diff looks big, however this is just splitting up the old super-method smaller ones.

krazziekay (approved):

PR #1599, opened 2022-09-14
SALTBAE-337 Add service association to deployments

kaganatlassian2:
**What's in this PR?**
Service association for deployments has been added. 

**Why**
We need to add a support service association for Deployments that has been created by GitHub. In order to use those at multiple features at JSM. 

**Added feature flags**
-

**Affected issues**  
[SALTBAE-337](https://jdog.jira-dev.com/browse/SALTBAE-337)

**How has this been tested?**  
Locally
Unit Test cases also added. 

**Whats Next?**
We're going to add Deployment Gating capability for GitHub Deployments.

### DIFF
--- a/src/config/feature-flags.ts
+++ b/src/config/feature-flags.ts
@@ -24,7 +24,8 @@ export enum BooleanFlags {
 	CREATE_BRANCH = "create-branch",
 	SEND_PR_COMMENTS_TO_JIRA = "send-pr-comments-to-jira_zy5ib",
 	USE_REPO_ID_TRANSFORMER = "use-repo-id-transformer",
-	USE_OUTBOUND_PROXY_FOR_OUATH_ROUTER = "use-outbound-proxy-for-oauth-router"
+	USE_OUTBOUND_PROXY_FOR_OUATH_ROUTER = "use-outbound-proxy-for-oauth-router",
+	SERVICE_ASSOCIATIONS_FOR_DEPLOYMENTS = "service-associations-for-deployments"
 }
 
 export enum StringFlags {
--- a/src/services/user-config-service.test.ts
+++ b/src/services/user-config-service.test.ts
@@ -3,6 +3,16 @@ import { RepoSyncState } from "models/reposyncstate";
 import { Subscription } from "models/subscription";
 import { getRepoConfig, updateRepoConfig } from "services/user-config-service";
 import { getInstallationId } from "~/src/github/client/installation-id";
+import { when } from "jest-when";
+import { booleanFlag, BooleanFlags } from "config/feature-flags";
+
+jest.mock("config/feature-flags");
+
+const turnFF_OnOff_service = (newStatus: boolean) => {
+	when(jest.mocked(booleanFlag))
+		.calledWith(BooleanFlags.SERVICE_ASSOCIATIONS_FOR_DEPLOYMENTS, expect.anything())
+		.mockResolvedValue(newStatus);
+};
 
 describe("User Config Service", () => {
 	const gitHubInstallationId = 1234;
@@ -24,12 +34,19 @@ describe("User Config Service", () => {
 		"      - \"Produktion\"\n" +
 		"      - \"produção\"\n" +
 		"      - \"продакшн\"\n" +
-		"      - \"PROD-*\"";
+		"      - \"PROD-*\"\n" +
+		"  services:\n" +
+		"    ids:\n" +
+		"      - \"test-id-1\"\n" +
+		"      - \"test-id-2\"\n" +
+		"      - \"test-id-3\"\n" +
+		"      - \"test-id-4\"";
 
 	const configFileContentBase64 = Buffer.from(configFileContent).toString("base64");
 
 
 	beforeEach(async () => {
+		turnFF_OnOff_service(false);
 		subscription = await Subscription.create({
 			gitHubInstallationId,
 			jiraHost,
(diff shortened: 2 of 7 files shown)


mboudreau:
@kaganatlassian2 if you want this merged, I'd recommend you fill in the PR description :)

kaganatlassian2:
@mboudreau done. Sorry I've missed that.

src/services/user-config-service.ts
```diff
@@ -115,20 +116,31 @@ const convertYamlToUserConfig = (input?: string): Config => {
...
+	if (configDeployments != null) {
+		if (configDeployments.environmentMapping) {
+			deployments["environmentMapping"] =  {
+				development: configDeployments.environmentMapping.development,
+				testing: configDeployments.environmentMapping.testing,
+				staging: configDeployments.environmentMapping.staging,
+				production: configDeployments.environmentMapping.production
+			};
+		}
+		if (configDeployments.services) {
```
bgvozdev, line 130: Should we also check for ids?

src/services/user-config-service.ts
```diff
@@ -115,20 +116,31 @@ const convertYamlToUserConfig = (input?: string): Config => {
...
+			deployments["services"] = {
+				ids: configDeployments.services.ids?.slice(0, MAX_SERVICE_ID_COUNT)
```
bgvozdev, line 132: Should we throw an exception if ids is not an array?

src/services/user-config-service.ts
```diff
@@ -115,20 +116,31 @@ const convertYamlToUserConfig = (input?: string): Config => {
...
 	const output = {
-		deployments: {
-			environmentMapping: {
-				development: config.deployments.environmentMapping.development,
-				testing: config.deployments.environmentMapping.testing,
-				staging: config.deployments.environmentMapping.staging,
-				production: config.deployments.environmentMapping.production
-			}
-		}
+		deployments: deployments
```
bgvozdev, line 143: /nit could be just `const output = { deployments }`

src/transforms/transform-deployment.ts
```diff
@@ -166,27 +166,40 @@ export const mapEnvironment = (environment: string, config?: Config): string =>
...
-		}
-	];
+	if (config?.deployments?.services?.ids) {
+		associations.push(
+			{
+				associationType: "serviceIdOrKeys",
+				values: config.deployments.services.ids
+			}
+		);
+		totalAssociationCount += config.deployments.services.ids.length;
```
bgvozdev, line 198: Can we exceed `MAX_ASSOCIATIONS_PER_ENTITY` here?

src/services/user-config-service.ts
kaganatlassian2: at common.ts Config we define ids as string[] should we need to check the typeof it here again?

src/config/feature-flags.ts
```diff
@@ -89,4 +90,9 @@ export const shouldTagBackfillRequests = async (): Promise<boolean> => {
 	return booleanFlag(BooleanFlags.TAG_BACKFILL_REQUESTS, false);
 };
 
+export const shouldEnableServiceAssociationForDeployments = async (): Promise<boolean> => {
+	logger.error(booleanFlag(BooleanFlags.SERVICE_ASSOCIATIONS_FOR_DEPLOYMENTS, false));
```
bgvozdev, line 94: Curious why do we have this log line?

src/services/user-config-service.ts
```diff
@@ -139,7 +153,7 @@ const convertYamlToUserConfig = (input?: string): Config => {
 
 const updateRepoConfigFromGitHub = async (repoSyncState: RepoSyncState, githubInstallationId: InstallationId, jiraHost: string, gitHubAppId: number | undefined): Promise<void> => {
 	const yamlConfig = await getRepoConfigFromGitHub(githubInstallationId, repoSyncState.repoOwner, repoSyncState.repoName, jiraHost, gitHubAppId);
-	const config = convertYamlToUserConfig(yamlConfig);
+	const config = await convertYamlToUserConfig(yamlConfig);
```
bgvozdev, line 156: Hmmm... was it broken before?

src/transforms/transform-deployment.test.ts
```diff
@@ -122,6 +135,116 @@ describe("transform GitHub webhook payload to Jira payload", () => {
...
+						commit: {
+							message: "ABC-2"
+						},
+						sha: "6e87a40179eb7ecf5094b9c8d690db727472d5bc2"
+					}
+				]
+			}
+			);
+
+		when(getRepoConfig).calledWith(
```
bgvozdev, line 211: /nit Would it make sense to have it as a nock, too?

src/services/user-config-service.test.ts
```diff
@@ -24,7 +24,13 @@ describe("User Config Service", () => {
 		"      - \"Produktion\"\n" +
 		"      - \"produção\"\n" +
 		"      - \"продакшн\"\n" +
-		"      - \"PROD-*\"";
+		"      - \"PROD-*\"\n" +
+		"  services:\n" +
```
bgvozdev, line 28: So, that's content. Where do we have assertions that these services are actually available via the service?

src/config/feature-flags.ts
kaganatlassian2: It was for debugging purposes but already removed it

src/services/user-config-service.ts
kaganatlassian2: No because of the FF I needed to make the function async that's why I needed to make it await here.

src/services/user-config-service.ts
```diff
@@ -107,29 +109,41 @@ const hasTooManyPatternsPerEnvironment = (config: Config): boolean => {
...
+			}
+		}
+	}
+
+
+	if (!deployments) {
 		throw new Error(`Invalid .jira/config.yml structure`);
 	}
 
 	// Trim the input data to only include the required attributes
```
bgvozdev, line 145: /nit I think this line should be also removed

bgvozdev (approved):

PR #1606, opened 2022-09-19
prTitle and prRef are now hashed - removed payload dump on log

joshkay10:
**What's in this PR?**
Adding prTitle and prRef to the sensitive data list

Removing a complete payload object in a log

**Why**
prTitle and prRef is UGC and needs to be hidden

an entire payload it too much logging! we need to be targeted, payload is ok during debug but not as a default, we cannot always be sure what is in it and could potentially leak PII/UGC

**Added feature flags**
nope

**Affected issues**  
nope

**How has this been tested?**  
locally only, changes are minor enough that I feel this is sufficient.

**Whats Next?**
Keep hunting for UGC!

### DIFF
--- a/src/util/logger-utils.ts
+++ b/src/util/logger-utils.ts
@@ -4,7 +4,7 @@ import bformat from "bunyan-format";
 import { DEBUG } from "bunyan";
 import { createHashWithSharedSecret } from "utils/encryption";
 
-const SENSITIVE_DATA_FIELDS = ["jiraHost", "orgName", "repoName", "userGroup", "userGroup", "aaid", "username"];
+const SENSITIVE_DATA_FIELDS = ["jiraHost", "orgName", "repoName", "userGroup", "userGroup", "aaid", "username", "prTitle", "prRef"];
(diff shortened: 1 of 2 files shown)


src/util/logger-utils.ts
```diff
@@ -4,7 +4,7 @@ import bformat from "bunyan-format";
 import { DEBUG } from "bunyan";
 import { createHashWithSharedSecret } from "utils/encryption";
 
-const SENSITIVE_DATA_FIELDS = ["jiraHost", "orgName", "repoName", "userGroup", "userGroup", "aaid", "username"];
+const SENSITIVE_DATA_FIELDS = ["jiraHost", "orgName", "repoName", "userGroup", "userGroup", "aaid", "username", "prTitle", "prRef"];
```
joshkay10, line 7: As we find more UGC/PII we continue to add to this list.
Its manageable at the moment but may need to consider moving it to a config file of some sort if it gets too massive!!

bgvozdev (approved):

PR #1607, opened 2022-09-19
More ugc

joshkay10:
**What's in this PR?**
Hash owner and description values as they contain UGC

### DIFF
--- a/src/github/client/github-client-interceptors.ts
+++ b/src/github/client/github-client-interceptors.ts
@@ -94,9 +94,9 @@ export const instrumentFailedRequest = (metricName: string, host: string) =>
 
 export const handleFailedRequest = (logger: Logger) =>
 	(error: AxiosError) => {
-		const { response, config, request } = error;
+		const { response, config  } = error;
 		const requestId = response?.headers?.["x-github-request-id"];
-		logger = logger.child({ res: response, config, req: request, err: error, requestId });
+		logger = logger.child({ requestId });
 
 		if (response?.status === 408 || error.code === "ETIMEDOUT") {
 			logger.warn("Request timed out");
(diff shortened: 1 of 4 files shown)


mboudreau (dismissed):

src/github/client/github-client-interceptors.ts
```diff
@@ -94,9 +94,9 @@ export const instrumentFailedRequest = (metricName: string, host: string) =>
 
 export const handleFailedRequest = (logger: Logger) =>
 	(error: AxiosError) => {
-		const { response, config, request } = error;
+		const { response, config  } = error;
 		const requestId = response?.headers?.["x-github-request-id"];
-		logger = logger.child({ res: response, config, req: request, err: error, requestId });
+		logger = logger.child({ requestId });
```
bgvozdev, line 99: Would it make sense to log request's URL + method and error message?

src/sync/installation.ts
```diff
@@ -224,7 +224,6 @@ const doProcessInstallation = async (data: BackfillMessagePayload, sentry: Hub,
 			} catch (err) {
 				const log = logger.child({
 					err,
-					repository,
```
bgvozdev, line 227: Shall we log repository ID maybe?

src/github/client/github-client-interceptors.ts
joshkay10: the log.error is **returning, status, path, method**.

The err obj doesn't have a additional msg in this case :shakes-fist:
example error:
`Error executing Axios Request (POST https://api.github.com/app/installations/26528447/access_tokens): Request failed with status code 404
`

Ahh, but theres a couple after that missing it, will add as childs

bgvozdev (approved):

PR #1608, opened 2022-09-19
ARC 1686 - Handling un-synced repositories

krazziekay:
**What's in this PR?**
- Replaced the Graphql query to fetch branch by REST API.
- Error messages added for 2 cases.
  - For Repository that is not connected/synced to GH4J yet.
  - For branch name that already exists. 
  - For a non-existing source branch.

**Why**
- When fetching the branch of the repositories that are not connected/synced to GH4J, the Graphql query throws a `FORBIDDEN` error.

**How has this been tested?**  
- Local
- Staging

**Screenshots**

Screen Shot 2022-09-21 at 4 07 44 pm


static/js/github-create-branch.js
```diff
@@ -86,16 +86,16 @@ const loadBranches = () => {
       const allBranchesfetched = data?.repository?.refs?.totalCount === data?.repository?.refs?.edges.length;
```
Harminder84, line 86: Need to refactor this as well? can we return branches count?

static/js/github-create-branch.js
```diff
@@ -106,7 +106,7 @@ const loadBranches = () => {
           }
         }
       });
-      $("#ghParentBranch").select2("val", data.repository.defaultBranchRef.name);
+      $("#ghParentBranch").select2("val", data[0].name);
```
Harminder84, line 109: We need to select default branch.

static/js/github-create-branch.js
krazziekay: Getting the total number of branches is not possible with REST. There is no such property.

static/js/github-create-branch.js
krazziekay: Unfortunately this REST API simply returns the list of the branch and there is no attribute to recognize the default branch.

static/js/github-create-branch.js
```diff
@@ -136,13 +136,23 @@ const createBranchPost = () => {
...
+          // TODO: Fix the url later, once you figure out how to get the `installationId`
+          showValidationErrorMessage("s2id_ghRepo", "This GitHub repository hasn't been configured to your Jira site. <a href='#'>Allow access to this repository.</a>");
+          break;
+        case 422:
+          showValidationErrorMessage("branchNameText", "This GitHub branch already exists. Please use a different branch name.");
+          break;
+        case 404:
+          showValidationErrorMessage("s2id_ghParentBranch", "This GitHub source branch does not exist. Please use a different branch.");
+          break;
+        default:
```
Harminder84, line 150: All above are server error messages which are returned by GitHub Api's.  It can't be  guaranteed that `404` only be thrown only in one case.  We should not display them as validation error messages.

static/js/github-create-branch.js
krazziekay: Added another Request to fetch the default branch for the selected Repo

static/js/github-create-branch.js
krazziekay: For now, I've added another Request to fetch the default branch for the selected Repo

static/js/github-create-branch.js
krazziekay: Removed the specific error message and listed as a list of possible causes.

src/routes/github/create-branch/github-branches-get.ts
```diff
@@ -17,8 +17,13 @@ export const GithubBranchesGet = async (req: Request, res: Response): Promise<vo
 
 	try {
 		const gitHubUserClient = await createUserClient(githubToken, jiraHost, req.log, gitHubAppConfig.gitHubAppId);
-		const branches = await gitHubUserClient.getReferences(owner, repo, 100);
-		res.send(branches);
+		const branches = await gitHubUserClient.getReferences(owner, repo);
+		const repository = await gitHubUserClient.getRepository(owner, repo);
```
joshkay10, line 21: Can we bundle these in a promise.all?
```
const branches = await gitHubUserClient.getReferences(owner, repo);
const repository = await gitHubUserClient.getRepository(owner, repo);
```

src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -40,6 +40,20 @@ export const GithubCreateBranchPost = async (req: Request, res: Response): Promi
...
-		res.sendStatus(500);
+		res.status(500).json(verifyError(err));
 	}
 };
+
+const verifyError = (error) => {
+	switch (error.status) {
+		case 403:
+			// TODO: Fix the url later, once you figure out how to get the `installationId`
+			return ["This GitHub repository hasn't been configured to your Jira site. <a href='#'>Allow access to this repository.</a>"];
```
joshkay10, line 51: Are these returning as an array because the error hbs file expect this?

src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -40,6 +40,20 @@ export const GithubCreateBranchPost = async (req: Request, res: Response): Promi
 		res.sendStatus(200);
 	} catch (err) {
 		req.log.error({ err }, "Error creating branch");
-		res.sendStatus(500);
+		res.status(500).json(verifyError(err));
 	}
 };
+
+const verifyError = (error) => {
+	switch (error.status) {
```
joshkay10, line 48: /nit
/Opinion piece

Thoughts on Object map instead of switch? Could replace the `verifyError` function.
i've ignored the fact the messaged are an array, but as an example......

```
const errorMessages = {
    401: "This GitHub repo.."
    403: "This GitHub repo..",
    default: "default message..."
}

```
`res.status(500).json(verifyError(err));` can be 
`res.status(500).json(errorMessages[err?.status] || errorMessages.default);`

src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -40,6 +40,20 @@ export const GithubCreateBranchPost = async (req: Request, res: Response): Promi
 		res.sendStatus(200);
 	} catch (err) {
 		req.log.error({ err }, "Error creating branch");
-		res.sendStatus(500);
+		res.status(500).json(verifyError(err));
```
joshkay10, line 43: If we have the correct code can we want it pass it in, in place of the 500?

src/routes/github/create-branch/github-branches-get.ts
krazziekay: Updated

src/routes/github/create-branch/github-create-branch-post.ts
krazziekay: Yep. The error messages from GitHub is super vague and we can't be fully sure that the error messages that we have are the only reasons for causing that error. For example, 403 may be caused by other reasons too(which I haven't seen yet) and not just the one we have, thats why instead of using a normal string, changed that to an array to show the list of possible reasons for the failure.

src/routes/github/create-branch/github-create-branch-post.ts
krazziekay: Updated

src/routes/github/create-branch/github-create-branch-post.ts
krazziekay: Updated

joshkay10 (approved):

src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -40,6 +48,6 @@ export const GithubCreateBranchPost = async (req: Request, res: Response): Promi
 		res.sendStatus(200);
 	} catch (err) {
 		req.log.error({ err }, "Error creating branch");
-		res.sendStatus(500);
+		res.status(err.status).json(errorMessages[err?.status || 500]);
```
joshkay10, line 51: `|| 500` was a good touch!!

src/routes/github/create-branch/github-branches-get.ts
```diff
@@ -17,8 +17,15 @@ export const GithubBranchesGet = async (req: Request, res: Response): Promise<vo
...
 	try {
 		const gitHubUserClient = await createUserClient(githubToken, jiraHost, req.log, gitHubAppConfig.gitHubAppId);
-		const branches = await gitHubUserClient.getReferences(owner, repo, 100);
-		res.send(branches);
+		const [ branches, repository ] = await Promise.all([
+			gitHubUserClient.getReferences(owner, repo),
+			gitHubUserClient.getRepository(owner, repo)
+		]);
+
+		res.send(branches.data.map(branch => ({
```
joshkay10, line 25: We can determine the default as part of getReferences, I reckon it would be a bit cleaner.

Then we can drop the `getRepository` call maybe??

The getReferences graphQL is returning defaultBranchRef name, so we could just mutate the response in `getReferences` with default boolean like your map is doing.
```
 defaultBranchRef {
        name
      } 
```

If thats annoying to reference, we can always slap the default ref in each branch node with:
**In getBranchesNameQuery**
```

edges {
        node {
          name
          repository {
                      defaultBranchRef {
                        name
                      }
                    }
```

krazziekay: Using `defaultBranchRef` in the Graphql throws 403 error for any repository that is not connected/synced to GH4J yet. That is why an added REST request is done.

static/js/github-create-branch.js
```diff
@@ -90,34 +90,24 @@ const loadBranches = () => {
...
-            text: item => item.node.name,
-            results: data.repository.refs.edges
+            text: item => item.name,
+            results: data
           }
         },
-        formatSelection: item => item.node.name,
-        formatResult: item => item.node.name,
-        createSearchChoice: (term) => {
-          if (allBranchesfetched) {
```
Harminder84, line 107: why you remove `createSearchChoice`?

src/routes/github/create-branch/github-branches-get.ts
```diff
@@ -17,8 +17,15 @@ export const GithubBranchesGet = async (req: Request, res: Response): Promise<vo
...
+			...branch,
+			default: branch.name === repository.data.default_branch
+		})));
```
Harminder84, line 28: Why to put default propery in each object? It would be more simple if we use structure like
{branches: [ { name: "branch-name", "other-prop" : "other-value"} ....], "default" : "branch-name"}

src/routes/github/create-branch/github-branches-get.ts
```diff
@@ -17,8 +17,15 @@ export const GithubBranchesGet = async (req: Request, res: Response): Promise<vo
 
 	try {
 		const gitHubUserClient = await createUserClient(githubToken, jiraHost, req.log, gitHubAppConfig.gitHubAppId);
-		const branches = await gitHubUserClient.getReferences(owner, repo, 100);
-		res.send(branches);
+		const [ branches, repository ] = await Promise.all([
+			gitHubUserClient.getReferences(owner, repo),
```
Harminder84, line 21: Is it returning all branches? if not, then make sure default branch is part of the response.

src/routes/github/create-branch/github-branches-get.ts
joshkay10: ahhh sorry, you are most correct. forgot about that!

static/js/github-create-branch.js
krazziekay: This was removed because we do not have `allBranchFetched`. The Graphql doesn't return the total number of repos.

src/routes/github/create-branch/github-branches-get.ts
krazziekay: Thats actually a good point.This request only returns the first 100 branches. 
So updated the response of this endpoint to return:
``````
{  
    branches: [] , // all the branches filtering out the default branch 
    defaultBranch: XXXX
}
``````

src/routes/github/create-branch/github-branches-get.ts
krazziekay: Updated the response to return response in the mentioned format: 
````
{  
    branches: [] , // all the branches filtering out the default branch 
    defaultBranch: XXXX
}
````

joshkay10 (approved):

static/js/github-create-branch.js
Harminder84: Earlier it is allowing to create new choice only in case if all branches are not fetched.  If we're unable to identify total number of repos, then it make sense to allow it in every case.

krazziekay: Reverted this back excluding the `if` case.

Harminder84 (approved):

PR #1609, opened 2022-09-19
Arc 1443 api for extract db migration from deployment.

gxueatlassian:
**What's in this PR?**
1. Remove auto db migration during startup
2. Add two apis to support db migrate up and down.

**Why**
1. http://go/fast-five and decouple the db migration from code deployment.

**Added feature flags**
Nah, I dont' think it is possible. 

**Affected issues**  
ARC-1443

**How has this been tested?**  

Locally manual testing. 

Dev
image

Stage [WIP]

Not sure how to test on prod though.

**Whats Next?**
Some run book to show how to run this code.
Already added here. https://hello.atlassian.net/wiki/spaces/OTFS/pages/1958321242/HOWTO+Run+db+migration+manually+on+ddev+stg+prod


db/migrations/20220919123300-test-db-migration.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
```
gxueatlassian, line 1: Not 100% sure, but having this just for testing on dev/stage (and maybe PROD?)

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,90 @@
...
+			statusCode: 400,
+			message: `"targetScript" is mandatory in the request body, but found none.`
+		};
+	}
+	//just make sure the script name matching the ones in db down the track.
+	if (!targetScript.endsWith(".js")) targetScript = targetScript + ".js";
+	return targetScript;
+};
+
+/**
```
gxueatlassian, line 29: Thoughts please.

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,90 @@
...
+ * So following scenarios will failed.
+ * Person A merge script 1, person B merge scripts 2. Now we CANNOT migrate anymore. Need to revert one PR first inorder to migrate.
+ *
+ * In short, encourage ppl to over-communicate when it comes to db migrations.
+ */
+export const validateScriptLocally = async (targetScript: string) => {
+
+	const scripts = await fs.promises.readdir(path.resolve(process.cwd(), "db/migrations"));
+	//Sort by name, asc, so filename order has to be in order now.
+	//So this now will be mandatory that, db migrtions scripts has to be alphabetically ordered.
```
gxueatlassian, line 40: `db migrtions scripts has to be alphabetically ordered` <-- I prefer this and think this requirement should be fine. But thoughts welcome .

src/routes/api/db-migrations/db-migration-down.ts
```diff
@@ -0,0 +1,59 @@
...
+	try {
+
+		const targetScript = getTargetScript(req);
+		logger.info(`Received DB target script to mgiration DOWN: ${targetScript}`);
+
+		await validateScriptLocally(targetScript);
+		await validateScriptAgainstDB(targetScript);
+		const { isSuccess, stdout, stderr } = await startDBMigration(targetScript, DBMigrationType.DOWN);
+
+		res.status(isSuccess ? 200: 500).send(`
```
bgvozdev, line 21: Would JSON be easier?

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+export const getTargetScript = (req: Request) => {
+	let targetScript = (req.body || {}).targetScript;
+	if (!targetScript) {
+		throw {
+			statusCode: 400,
+			message: `"targetScript" is mandatory in the request body, but found none.`
+		};
+	}
+	//just make sure the script name matching the ones in db down the track.
```
bgvozdev, line 24: /nit comment duplicating code

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,90 @@
...
+import { isNodeDev, isNodeTest, isNodeProd, getNodeEnv } from "utils/is-node-env";
+import { getLogger } from "config/logger";
+import { exec as execOrigin } from "child_process";
+import { promisify } from "util";
+
+const exec = promisify(execOrigin);
+const logger = getLogger("DBMigration");
+
+/**
+ * Return the common param in req.body `targetScript` to migrate up or rollback
```
bgvozdev, line 13: /nit `@return` - I think the convention is to use java-like doc

bgvozdev: Why "migrate up" only? What about "down"?

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,90 @@
...
+ * Note: Intentially to make this param case sensitive to be more safe
```
bgvozdev, line 14: /nit I'd've rephrased that to something like "Param is case-sensitive for better safety"

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,90 @@
...
+	if (!targetScript.endsWith(".js")) targetScript = targetScript + ".js";
+	return targetScript;
+};
+
+/**
+ * Make sure the script to migrate up or rollback is the lastest scripts in this build.
+ * So following scenarios will failed.
+ * Person A merge script 1, person B merge scripts 2. Now we CANNOT migrate anymore. Need to revert one PR first inorder to migrate.
+ *
+ * In short, encourage ppl to over-communicate when it comes to db migrations.
```
bgvozdev, line 34: /nit I'd've double-checked English in these comments 😛

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,90 @@
...
+ */
+export const validateScriptLocally = async (targetScript: string) => {
+
+	const scripts = await fs.promises.readdir(path.resolve(process.cwd(), "db/migrations"));
+	//Sort by name, asc, so filename order has to be in order now.
```
bgvozdev, line 39: /nit comment duplicating code

src/routes/api/db-migrations/db-migration-utils.ts
bgvozdev: Can it be expressed as a unit-test instead of a comment?

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,90 @@
...
+
+};
+
+export enum DBMigrationType {
+	UP = "UP",
+	DOWN = "DOWN"
+}
+
+export const startDBMigration = async (targetScript: string, ops: DBMigrationType) => {
+	logger.info(`All validation pass, now executing db migration script ${targetScript} for ${ops}`);
```
bgvozdev, line 61: /nit implies someone did the validation... this log line shouldn't be here but should be in the code that is calling this method

src/routes/api/db-migrations/db-migration-down.ts
gxueatlassian: Tried that locally just now, realise since the major consume is human/developer who run this script, the following screenshot (with json output) is not very user friendly, compare to the one in this PR's description.

thoughts?

image

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,96 @@
...
+		isSuccess,
+		stdout,
+		stderr
+	};
+};
+
+const getDBMigrateEnv = () => {
+	if (isNodeDev()) return "development";
+	if (isNodeTest()) return "test";
+	if (isNodeProd()) return "production-migrate";
```
bgvozdev, line 90: Is staging same as prod here?

src/routes/api/db-migrations/db-migration-utils.ts
gxueatlassian: Let me think about the test. But on the other hand, this is more like a "got-cha" that developer needs to remember the scripts need to be named in alphabetical order. 

I can, however, remove the restriction on this one, but that means you may "accidentally" run other's db migration scripts as well. A bit hard to explain in pure text, though.

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,96 @@
...
+	const env = getDBMigrateEnv();
+	if (ops !== DBMigrationType.UP && ops !== DBMigrationType.DOWN) {
+		throw {
+			statusCode: 500,
+			message: `Fail to execute db migration type ${ops}`
+		};
+	}
+	const cmd = ops === DBMigrationType.UP ?
+		`./node_modules/.bin/sequelize db:migrate --env ${env}`
+		: `./node_modules/.bin/sequelize db:migrate:undo:all --to ${targetScript} --env ${env}`;
```
bgvozdev, line 71: So, targetScript is always the top from the dir, right? How would rollback work in this case? Rollback to the lastest means do nothing, right?

src/routes/api/db-migrations/db-migration-up.ts
```diff
@@ -0,0 +1,54 @@
...
+
+	} catch (e){
+		logger.error("Error doing db migration up", e);
+		res.status(e.statusCode || 500);
+		res.send(safeJsonStringify(e));
+	}
+
+};
+
+const validateScriptAgainstDB = async (targetScript: string) => {
```
bgvozdev, line 37: tests?

src/routes/api/db-migrations/db-migration-utils.ts
gxueatlassian: 👍 thanks mate, just double checked and fixed things.

src/routes/api/db-migrations/db-migration-down.ts
```diff
@@ -0,0 +1,59 @@
...
+
+	} catch (e){
+		logger.error("Error doing db migration down", e);
+		res.status(e.statusCode || 500);
+		res.send(safeJsonStringify(e));
+	}
+
+};
+
+const validateScriptAgainstDB = async (targetScript: string) => {
```
bgvozdev, line 37: tests?

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,96 @@
...
+export const getTargetScript = (req: Request) => {
+	let targetScript = (req.body || {}).targetScript;
+	if (!targetScript) {
+		throw {
+			statusCode: 400,
+			message: `"targetScript" is mandatory in the request body, but found none.`
+		};
+	}
+	//just make sure the script name matching the ones in db down the track.
+	if (!targetScript.endsWith(".js")) targetScript = targetScript + ".js";
```
bgvozdev, line 25: /nit I'd've renamed the method to "extractTargetScript" because this is not a plain getter anymore but getter + some logic...

src/routes/api/db-migrations/db-migration-utils.ts
atrigueiro: As long as we have a convention where the filename starts with the current date, for instance `2022-09-20`, we should be ok.

src/routes/api/db-migrations/db-migration-utils.ts
gxueatlassian: Yup, sounds good. Fixed.

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,96 @@
+import { Request } from "express";
```
bgvozdev, line 1: Tests?

src/routes/api/db-migrations/db-migration-utils.ts
gxueatlassian: Got this line in `github-for-jira.sd.yml`, so I think `ddev` and `stg-west` and `prod-west` all use `production` as `NODE_ENV`
```yaml
config:
  environmentVariables:
    NODE_ENV: production
```

src/routes/api/db-migrations/db-migration-utils.ts
gxueatlassian: `targetScript` is always the last one (alphabetically). And the `sequelize db:migrate:undo:all --to <some_scripts.js>` is quite misleading. Turns out it is "inclusive". So this line will rollback the `some_script.js`, make the db back to the state before that script.

src/routes/api/db-migrations/db-migration-utils.ts
gxueatlassian: Sorry what do you mean?

src/routes/api/db-migrations/db-migration-utils.ts
gxueatlassian: Fixed, but I ended up removing the logic here, ( and also `targetScript.toLowercase() ===` in another place as well)

So the idea is keep it safe and straightforward, no logic, just use what user input.

src/routes/api/db-migrations/db-migration-utils.ts
gxueatlassian: fixed. `rollback` means `down`. But anyway I converted all `rollback` to `down` to be consistent.

src/routes/api/db-migrations/db-migration-utils.ts
bgvozdev: @gxueatlassian I just don't like comments in code ;) If this is something important, I think this should be covered by a test in case someone modifies it the test will break.

Order in this case means "latest". If you think it is not clear from the top comment to the function itself, I'd've added more explanations there; at least the documentation would be in one place, more cohesive IMHO

src/routes/api/db-migrations/db-migration-utils.ts
bgvozdev: This either requires a comment or a test, or (better) both.

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,91 @@
...
+};
+
+export enum DBMigrationType {
+	UP = "UP",
+	DOWN = "DOWN"
+}
+
+export const startDBMigration = async (targetScript: string, ops: DBMigrationType) => {
+	const env = getDBMigrateEnv();
+	if (ops !== DBMigrationType.UP && ops !== DBMigrationType.DOWN) {
```
bgvozdev, line 63: This validation needs to be done somewhere else, where the string is converted to enum. Once it is converted, type-safety should guarantee that this never happens.

gxueatlassian: `string is converted to enum` <-- there's no string converted to enum in the code. The code is using it as enum directly.

But but, I just make sure it is safe, as at the end of day, js doesn't have enum, just string.

src/routes/api/db-migrations/db-migration-utils.ts
bgvozdev: Maybe better to rename it to something like `getDbConfigEnvForMigration()`? Would it make it a little bit more clear?

gxueatlassian: done.

src/routes/api/db-migrations/db-migration-utils.ts
bgvozdev: Ah right, I didn't paid attention to it. Anyway, I find it awkward that we are doing validation of something that is always true. 

If we worry that someone might add a new DBMigrationType that won't be supported by this function (which means the ternary operator will fallback to "down"), I would've changed that to 
```
switch (ops) {
   case UP: blah
   case DOWN: yada
   default: throw "Should never happen"
} 
```

src/routes/api/db-migrations/db-migration.test.ts
```diff
@@ -0,0 +1,101 @@
+import fs from "fs";
```
gxueatlassian, line 1: @bgvozdev tests !!

src/routes/api/db-migrations/db-migration-down.ts
gxueatlassian: added.

src/routes/api/db-migrations/db-migration-up.ts
gxueatlassian: added

src/routes/api/db-migrations/db-migration-utils.ts
gxueatlassian: added

src/routes/api/db-migrations/db-migration-utils.ts
gxueatlassian: @bgvozdev sounds good, let's chat tomorrow to pair on this comment.

.sequelizerc
```diff
@@ -1,9 +1,13 @@
...
+const migrationsDir =
+  process.env.IS_TESTING_DB_MIGRATION === "true" ?
+  "migrations-test" : "migrations";
+
 module.exports = {
   "config": path.resolve('./db/config.json'),
   "models-path": path.resolve('./src/models'),
   "seeders-path": path.resolve('./db/seeders'),
-  "migrations-path": path.resolve('./db/migrations')
+  "migrations-path": path.resolve(`./db/${migrationsDir}`),
```
bgvozdev, line 12: Would it be better to have it  "test/db"?

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,101 @@
...
+import { getLogger } from "config/logger";
+import { exec as execOrigin } from "child_process";
+import { promisify } from "util";
+
+const exec = promisify(execOrigin);
+const logger = getLogger("DBMigration");
+
+/**
+ *
+ * Functiont to get the parem for db migration.
```
bgvozdev, line 14: /nit param

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,101 @@
...
+ *
+ * Intentially to NOT make this param case insensitive to be more safe
```
bgvozdev, line 16: I'd've removed the double-negation. Maybe something like "Intentially making this param case sensitive for better safety"?

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,101 @@
...
+
+/* Make sure the script to migrate up or down is the latest (alphabetically) script in this build.
+ * So the following scenarios will fail.
+ * Person A merge script 1, person B merge scripts 2. Now we CAN NOT migrate anymore. Need to revert one PR first to migrate.
+ *
+ * In short, encourage ppl to over-communicate when it comes to DB migrations.
+ */
+export const validateScriptLocally = async (targetScript: string) => {
+
+	const checkDirSurFix = isNodeTest() ? "-test" : "";
```
bgvozdev, line 39: SurFix?...

I mentioned that in another place: should we rather move it to tests dir?

.sequelizerc
```diff
@@ -1,9 +1,13 @@
 #!/usr/bin/env node
 const path = require('path')
 
+const migrationsDir =
+  process.env.IS_TESTING_DB_MIGRATION === "true" ?
```
bgvozdev, line 5: Should we rather have the whole path in the env variable and fall back to the default value (which is the production one)?

.sequelizerc
gxueatlassian: I think I like `db/migrations` and `db/migrations-test`, feels like it is trivial, so ignoring it now.
Please let me know if you think it is quite important.

.sequelizerc
gxueatlassian: Okay, I've refactor the test a bit, so this file still the same as here, but the other one, db-migration-utils.ts now does NOT have any test purpose code.

The reason to NOT mock the exec is, it is quite hard as the code also check against db, so need to mock many of those db query as well, make the test quite hard to write (and read).

db/migrations/20220919123300-test-db-migration.js
bgvozdev, line 1: Do we still need this file?

src/routes/api/db-migrations/db-migration-up.ts
```diff
@@ -0,0 +1,61 @@
...
```
bgvozdev, line 44: I thought you were going to move it to utils and cover with tests?...

src/routes/api/db-migrations/db-migration-utils.test.ts
```diff
@@ -0,0 +1,75 @@
...
+			let err = null;
+			try {
+				await validateScriptLocally(MIGRATION_SCRIPT_FIRST);
+			} catch (e) {err = e;}
+			expect(err).toEqual(expect.objectContaining({
+				statusCode: 400
+			}));
+		});
+	});
+	describe("DB migration util", () => {
```
bgvozdev, line 59: shouldn't it be `startDbMigration` ?

src/routes/api/db-migrations/db-migration-utils.ts
```diff
@@ -0,0 +1,99 @@
...
+	logger.info(`Target script match latest scripts in repo ${targetScript}, validation passed`);
+
+};
+
+export enum DBMigrationType {
+	UP = "UP",
+	DOWN = "DOWN",
+}
+
+export const startDBMigration = async (targetScript: string, ops: DBMigrationType) => {
```
bgvozdev, line 58: Btw, shouldn't it be something like `runDbMigration` instead? I feel like "start" means that there should be "finish", but there's none...

src/routes/api/db-migrations/db-migration.test.ts
```diff
@@ -0,0 +1,116 @@
+import express from "express";
```
bgvozdev, line 1: Shouldn't it be called `db-migration-router.test.ts` ?

db/migrations/20220919123300-test-db-migration.js
gxueatlassian: Not sure, it is for demo/once off verification purpose.

src/routes/api/db-migrations/db-migration-up.ts
bgvozdev: Ah, never mind, see you are testing it below

src/routes/api/db-migrations/db-migration.test.ts
bgvozdev: Ah, I see it is for both up and down... Would it make sense to merge them together into one file then to keep the convention? Currently it was tricky for me to relate the tests...

bgvozdev: Or maybe it would make sense to split this 2 file onto 2 (for up and down)?

.sequelizerc
gxueatlassian: Refactor the test again to have unit test and some test on the handler itself to with mock on the exec and some db query.

src/routes/api/db-migrations/db-migration-utils.test.ts
gxueatlassian: Updated.

src/routes/api/db-migrations/db-migration-utils.ts
gxueatlassian: Updated per your proposal. :)

src/routes/api/db-migrations/db-migration.test.ts
gxueatlassian: Yup make sense, split into two files, one for up and one for down.
I'll pass the router test for now, as it should be covered already in each of the up/down handler test.

db/migrations/20220919123300-test-db-migration.js
bgvozdev: ah right

src/routes/api/db-migrations/db-migration-down.test.ts
```diff
@@ -0,0 +1,83 @@
...
+}));
+jest.mock("models/sequelize", ()=>({
+	...jest.requireActual("models/sequelize")
+}));
+
+const DB_MIGRATE_DOWN_URL = "/api/db-migration/down";
+const MIGRATION_SCRIPT_FIRST = "20220101000000-first-script.js";
+const MIGRATION_SCRIPT_LAST = "20220101000001-second-script.js";
+
+describe("DB migration", ()=>{
```
bgvozdev, line 21: DB migration down?

src/routes/api/db-migrations/db-migration-up.test.ts
```diff
@@ -0,0 +1,80 @@
...
+}));
+jest.mock("models/sequelize", ()=>({
+	...jest.requireActual("models/sequelize")
+}));
+
+const DB_MIGRATE_UP_URL = "/api/db-migration/up";
+const MIGRATION_SCRIPT_FIRST = "20220101000000-first-script.js";
+const MIGRATION_SCRIPT_LAST = "20220101000001-second-script.js";
+
+describe("DB migration", ()=>{
```
bgvozdev, line 21: DB migration up?

bgvozdev (approved):

src/routes/api/db-migrations/db-migration-up.test.ts
gxueatlassian: fixed.

src/routes/api/db-migrations/db-migration-down.test.ts
gxueatlassian: fixed.

bgvozdev (approved):

PR #1615, opened 2022-09-20
ARC-1351: Remove OAuth scopes for the user's access_token

atrigueiro:
**What's in this PR?**
We're removing OAuth scopes for the user's access_token

**Why**
Security. There's no need to have that scope - we don't need the repo names to render the UI (we only need the number of repos) and we don’t need to write perms for the user profile either.

**Added feature flags**
n/a

**Affected issues**  
ARC-1351

**How has this been tested?**  
Locally and staging. It has been soaking in staging for over an hour and both pollinators (basic check and backfill) look ok.

**Whats Next?**
Update documentation - https://github.com/atlassian/github-for-jira/blob/main/docs/security.md


.env.development
```diff
@@ -24,5 +24,5 @@ MAINTENANCE_MODE=false
 LOG_LEVEL=debug
 
 # Cryptor
-CRYPTOR_URL=http://cryptor-sidecar:26272
+CRYPTOR_URL=http://cryptor:26272
```
atrigueiro, line 27: This is the same as `main`, not too sure why's here (I also have latest main locally). Can we .gitignore this file?

gxueatlassian: good pickup.

src/routes/github/github-oauth-router.ts
```diff
@@ -11,14 +11,13 @@ import { createHashWithSharedSecret } from "utils/encryption";
 
 const logger = getLogger("github-oauth");
 const appUrl = envVars.APP_URL;
-const scopes = ["user", "repo"];
```
gxueatlassian, line 14: Hmm, actually sorry just realise that, @joshkay10 do you mind have a look at this as well? I am not sure will the `create-branch` require any scope from this user token as well?

joshkay10: We will need repo still, but not user.

atrigueiro: Thanks, @joshkay10. Just adding here for our future reference:

> Access tokens require the repo scope for private repositories.
https://docs.github.com/en/enterprise-cloud@latest/graphql/guides/using-the-graphql-api-for-discussions

gxueatlassian (approved):

joshkay10 (approved):

PR #1616, opened 2022-09-20
CES-6339 sync fix

mboudreau:
**What's in this PR?**
Fixing things preventing sync from happening within a client.

**Why**
Cryptor failures would stop sync in it's track since the introduction of it to be used with DB secrets without reliable checks or retries

**Added feature flags**
None

**Affected issues**  
CES-6339

**How has this been tested?**  
Included unit tests are still passing

**Whats Next?**
Retry sync for customer and see if it passes.


src/routes/healthcheck/healthcheck-get.ts
```diff
@@ -1,19 +1,16 @@
 import { Request, Response } from "express";
 import { getLogger } from "config/logger";
-import { envVars } from "config/env";
 import { deepCheckCryptor } from "./deep-check-cryptor";
 
 const logger = getLogger("healthcheck");
 export const HealthcheckGet = async (_: Request, res: Response): Promise<void> => {
 
-	if (envVars.MICROS_GROUP === "Worker") {
```
gxueatlassian, line 9: Webserver healtcheck will be hit very often, like every minute. Have you considered this?

src/transforms/transform-branch.ts
```diff
@@ -36,22 +36,26 @@ export const transformBranch = async (github: GitHubInstallationClient, webhookP
...
-				id: getJiraId(ref),
-				issueKeys,
-				name: ref,
-				url: `${repository.html_url}/tree/${ref}`,
-				updateSequenceId: Date.now()
-			}
-		],
-		updateSequenceId: Date.now()
-	};
+	try {
```
gxueatlassian, line 39: Can you please explain how is try/catch impacting the result?

src/routes/healthcheck/healthcheck-get.ts
mboudreau: Are you suggesting that the cryptor sidecar can't handle 2 requests per minute?

gxueatlassian: It should be able to handle that, all good.

src/transforms/transform-branch.ts
mboudreau: I've added comments, logs and a better flow

gxueatlassian (approved):

src/worker/startup.ts
```diff
@@ -12,6 +12,8 @@ export async function start() {
 		return;
 	}
 
+	await
```
krazziekay, line 15: Whats this `await` for? Is this something that got missed out?

krazziekay: I think this is causing the 500 Internal server error in the test case. Can you please revert this part?

mboudreau: woops, good catch, it was something I was trying out.

gxueatlassian (approved):

krazziekay (approved):

PR #1617, opened 2022-09-20
ARC-1606 Display user details

Harminder84:
**What's in this PR?**
Display logged-in GitHub user details with an option to change GitHub login. 
When `Change GitHub login` link is clicked, it will clear the `GitHub` token from session and opens up GitHub cloud or enterprise site in a new window to logout from the portal itself.

**Affected issues**  
[ARC-1606]

**How has this been tested?**  
Local

[ARC-1606]: https://softwareteams.atlassian.net/browse/ARC-1606

### DIFF
--- a/src/routes/github/create-branch/github-remove-session.ts
+++ b/src/routes/github/create-branch/github-remove-session.ts
@@ -1,0 +1,15 @@
+import { Request, Response } from "express";
+
+
+export const GithubRemoveSession = (req: Request, res: Response) => {
+	const { gitHubAppConfig } = res.locals;
+
+	if (!gitHubAppConfig) {
+		res.sendStatus(401);
+		return;
+	}
+
+	req.session.githubToken = undefined;
+	res.send({ baseUrl: gitHubAppConfig.hostname });
+
+};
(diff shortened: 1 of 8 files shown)


src/routes/github/create-branch/github-remove-session.ts
```diff
@@ -0,0 +1,15 @@
...
+
+export const GithubRemoveSession = (req: Request, res: Response) => {
+	const { gitHubAppConfig } = res.locals;
+
+	if (!gitHubAppConfig) {
+		res.sendStatus(401);
+		return;
+	}
+
+	req.session.githubToken = undefined;
```
joshkay10, line 12: If i refresh the page after this it recreates my session, you might have to use something like `req.session.destroy`

joshkay10, line 12: Refreshing the page after this recreates my session, you might need to destroy the whole value?
		`req.session = null`

Harminder84: This will not prevent to re-create the session on refresh.  You have explicitly sign out from the GitHub portal.

joshkay10 (approved):

PR #1621, opened 2022-09-21
[ :facepalm: ] [ Free karma, instant release ]  NONE: fix metrics

bgvozdev:
**What's in this PR?**
Removing prefixed " " from tags

**Why**
Because it is not trimmed :facepalm:


bgvozdev:
Ooops

PR #1630, opened 2022-09-27
ARC 1689 - Update the loading states for the Inputs for Repo and Branch

krazziekay:
**What's in this PR?**
- Updated the spinner for repo and branch inputs.
- Added 2 new methods for displaying spinner 
  - Inside the input dropdown list
  - On the input 

**Why**
- For better User Experience

**How has this been tested?**  
- Local

**Video**

https://user-images.githubusercontent.com/13076255/192453049-63468198-c3c8-492e-bde9-ce71b09010d2.mov


static/css/github-create-branch.css
```diff
@@ -51,8 +51,14 @@
   right: 4px;
 }
 .select2-loader {
+  display: flex;
```
joshkay10, line 54: The spinner looks a bit cramped, can we add some padding for a bit of vertical space?
You might need to add extra room for the results too so it doesn't bounce around in size..... maybe throw some screenshot options at jajajade?

joshkay10 (commented):
This feels way nicer to me!!!

static/css/github-create-branch.css
krazziekay: Padding has been added.
And as for bouncing effect, that will happen in most cases because we don't know the number of repos that will be returned. So, lets say if we make the spinner about the same height as the original dropdown, and now if the search result is just 1 repo, it will still do the jumpy effect cause the dropdown's height will decrease after finishing loading.

joshkay10 (approved):

PR #1649, opened 2022-10-07
Backfill targetted

joshkay10:
**What's in this PR?**
During a targetted backfill reset the relevant cursors and status's in reposync table

**Why**
Stops use syncing extra stuff we don't want to or setting states out of sync !


src/models/reposyncstate.ts
```diff
@@ -37,6 +38,7 @@ export class RepoSyncState extends Model {
 
 	get status(): TaskStatus {
 		const statuses = [this.pullStatus, this.branchStatus, this.commitStatus];
+		console.log(statuses);
```
mboudreau, line 41: oh the horror!

src/sync/sync-utils.ts
```diff
@@ -18,22 +18,27 @@ export async function findOrStartSync(
...
 		syncWarning: null
 	});
 
 	logger.info({ subscription, syncType }, "Starting sync");
 
 	if (syncType === "full") {
-		// Remove all state as we're starting anew
-		await RepoSyncState.deleteFromSubscription(subscription);
 		fullSyncStartTime = new Date().toISOString();
+		if (targetTasks && targetTasks.length > 0) {
```
mboudreau, line 31: `if (targetTasks?.length)`

src/sync/sync-utils.ts
```diff
@@ -18,22 +18,27 @@ export async function findOrStartSync(
...
+			await RepoSyncState.resetTasksSyncFromSubscription(targetTasks, subscription);
+		} else {
+			await subscription.update({
+				totalNumberOfRepos: null,
+				repositoryCursor: null,
+				repositoryStatus: null
+			});
+			// Remove all state as we're starting anew
+			await RepoSyncState.deleteFromSubscription(subscription);
+		}
```
mboudreau, line 41: This logic isn't correct.  It's not just for a full backfill, but we want the ability to reset specific things for partial as well.

Hm... maybe we want to stop using the concept of 'full' or 'partial' and just make it "what do you want to reset before I send this backfill message to the queue"..

src/models/reposyncstate.ts
joshkay10: What have i done!

src/routes/api/api-resync-post.ts
```diff
@@ -0,0 +1,54 @@
+import { Request, Response } from "express";
```
joshkay10, line 1: No change in here, just lifted from sync-utils and dumped

src/sync/sync-utils.ts
```diff
@@ -49,6 +54,30 @@ export async function findOrStartSync(
...
+	});
+
+	return RepoSyncState.update({
+		repoUpdatedAt: null,
+		...updateTasks
+	}, {
+		where: {
+			subscriptionId: subscription.id
+		}
+	});
```
mboudreau, line 78: This completely forgets the `repository` task that's on subscription model (not reposyncstate), but also, it will also try to update `repositoryStatus` which doesn't exist on RepoSyncState and I think will error?

joshkay10: ohhh ok, i thought it was on the reposyncstate, but that make sense since its a higher level. copy that.

mboudreau: I don't think we need to split this up into extra functions, especially since they have duplicate code in it.  just a simple if statement for `repository` task should do

src/sync/sync-utils.ts
```diff
@@ -49,6 +54,57 @@ export async function findOrStartSync(
...
+	repositoryCursor?: string | null;
+	repositoryStatus?: string | null;
+}
+
+const resetTargetedTasks = async (subscription: Subscription, syncType?: SyncType, targetTasks?: TaskType[]): Promise<void> => {
+	if (!syncType || !targetTasks?.length) {
+		return;
+	}
+
+	const subscriptionTasks = targetTasks.find(task => task === "repository");
```
mboudreau, line 68: don't you mean `filter`?  or I guess you could change it to `hasRepositoryTask = targetTasks.includes("repository");`

src/sync/sync-utils.ts
```diff
@@ -49,6 +54,57 @@ export async function findOrStartSync(
 	}, 0, logger);
 }
 
+type subscriptionUpdateTasks = {
```
mboudreau, line 57: Types should always start capitalized

src/sync/sync-utils.ts
```diff
@@ -49,6 +54,57 @@ export async function findOrStartSync(
...
+	// Partial sync only resets status (continues from existing cursor)
+	if (subscriptionTasks) {
+		if (syncType === "full") {
+			updateSubscriptionTasks.totalNumberOfRepos = null;
+			updateSubscriptionTasks.repositoryCursor = null;
+		}
+		updateSubscriptionTasks.repositoryStatus = null;
+	}
+
+	await subscription.update({
```
mboudreau, line 103: This should only be called in the if statement above.  You don't need to call this when there's nothing to update.
You also don't need to use the spread operator when it's the only key in the object, just use `updateSubscriptionTasks` as the update object.

src/sync/sync-utils.ts
```diff
@@ -49,6 +54,57 @@ export async function findOrStartSync(
...
+
+const resetTargetedTasks = async (subscription: Subscription, syncType?: SyncType, targetTasks?: TaskType[]): Promise<void> => {
+	if (!syncType || !targetTasks?.length) {
+		return;
+	}
+
+	const subscriptionTasks = targetTasks.find(task => task === "repository");
+	const repoSyncTasks = targetTasks.filter(task => task !== "repository");
+	const updateRepoSyncTasks = {};
+	const updateSubscriptionTasks: subscriptionUpdateTasks = {};
```
mboudreau, line 71: I'd probbly move this in the if statement below and start it with the default key `repositoryStatus: null`

mboudreau (approved):

PR #1652, opened 2022-10-09
ARC-1700 Skeleton loading for create branch

rachellerathbone:
**What's in this PR?**
Skeleton loading for CB form page.

https://user-images.githubusercontent.com/37155488/194750847-fc4fd117-16db-4fd7-bdca-788e7fad8634.mov

**Why**
Currently, when the form for this page loads it looks like this:

Screen Shot 2022-09-28 at 4 32 32 PM

We debated using the loading spinner which is used elsewhere in the app vs skeleton loading which is now used across Jira. We decided it would be better to align with the latter as customers will be navigating to this screen from Jira and it will provide a more seamless experience. Unfortunately, AUI doesn't provide a skeleton component (found out they were only built for Atlaskit) so we've had to handle this ourselves.

**Added feature flags**
CB

**Affected issues**  
ARC-1700

**Whats Next?**
Update other parts of the app to use skeleton loading so that our entire app aligns with what's now used across Atlassian products.


static/css/global.css
```diff
@@ -319,3 +319,45 @@ a:focus {
...
+  ) #ebecf0;
+  background-size: 200% 100%;
+  background-position-x: 180%;
+  border-radius: 4px;
+}
+
+.skeleton,
+.skeleton > label {
+/* Override AUI styling */
+  color: transparent !important;
```
rachellerathbone, line 343: I definitely don't like using this but unfortunately the styling of AUI components really get in the way here and override this style without it.

views/github-create-branch.hbs
```diff
@@ -30,26 +30,28 @@
   <div class="headerImage">
     <img src="/public/assets/jira-and-github.png" alt="Jira and GitHub logos" />
   </div>
+
   <div class="gitHubCreateBranch__title">
     <div class="gitHubCreateBranch__header">Create GitHub Branch</div>
-    <div class="gitHubCreateBranch__subHeader">
-      Creating a branch for <b>{{issue.key}}</b>
-    </div>
+    <div class="gitHubCreateBranch__subHeader skeleton skeleton-text">Creating a branch for <b>{{issue.key}}</b></div>
```
rachellerathbone, line 36: For most elements, simply adding skeleton and skeleton-element will make it possible implement this loading state.

views/github-create-branch.hbs
```diff
@@ -30,26 +30,28 @@
...
     <!--TODO: Add condition to display this only for GHES -->
     <div class="field-group">
-      <aui-label for="server">GitHub Enterprise Server instance</aui-label>
-      <select id="ghServers" name="server" class="full-width-field" placeholder="Select a repository">
-        {{#each servers}}
-          <option value="{{id}}">{{server}} / {{appName}}</option>
-        {{/each}}
-      </select>
+      <aui-label for="server" class="skeleton skeleton-text">GitHub Enterprise Server instance</aui-label>
+       <div class="skeleton skeleton-input"></div>
```
rachellerathbone, line 43: For input elements, we need to handle it a little differently because the AUI styling gets in the way and I didn't want to end up with !important all over the place. Easier solution here is to simply render a separate skeleton element temporarily in place of the real element.

static/js/skeleton.js
```diff
@@ -0,0 +1,9 @@
+$(document).ready(function() {
+
+	setTimeout(function() {
```
rachellerathbone, line 3: I had originally planned to use a handlebars conditional to handle rendering the skeleton element vs the real one but turns out that isn't possible. When the page loads, we already have the data e.g. repos. The issue with the delayed rendering appears to be due to how we're looping over that data within the input fields. So, because the data is already there when the page loads, we're simply displaying the skellie stuffs for half a second before moving it and displaying the real element.

Curious on thoughts are the half second wait time. Seems to work for this page but may differ with more data?

joshkay10: ahhhh thats inconvenient. So is it the select2 elements that are slow?
They might have an event when they are ready?

Since the repo data gets passed with the template(think theres about a 20 repos max) we should see a pretty consistent load..... we could try force 100+ or so and see how it reacts.

rachellerathbone: Nah it's not just the select element. Repository is janky too and that's just a div but I think it's the looping that's slowing the rendering down.

```
{{#each repos}}
        <div class="hidden default-repos">{{node.full_name}}</div>
      {{/each}}
```

rachellerathbone: Looks like the max we can pass is 100. Just tested it out and still looks good.

static/js/skeleton.js
```diff
@@ -0,0 +1,9 @@
...
+		$('.skeleton-container').remove();
+		$('.skeleton-input').remove();
+		$('.skeleton').removeClass('skeleton');
+		$(".loaded").attr('style', 'display: block !important');
+	}, 500);
```
krazziekay, line 8: Are we sure about the 500 ms As you mentioned, it might differ for other components in different pages.
As per this [answer](https://stackoverflow.com/questions/25831679/select2-initialization-callback#answer-25833086), we can get it working by simply removing the skeleton classes after the `.auiSelect2()`.

views/github-create-branch.hbs
```diff
@@ -31,17 +31,17 @@
     <img src="/public/assets/jira-github-connection.svg" class="headerImageLogo" alt="Jira and GitHub logos" />
     <aui-spinner class="gitHubCreateBranch__spinner" size="small"></aui-spinner>
   </div>
+
   <div class="gitHubCreateBranch__title">
     <div class="gitHubCreateBranch__header">Create GitHub Branch</div>
-    <div class="gitHubCreateBranch__subHeader">
-      Creating a branch for <b>{{issue.key}}</b>
-    </div>
+    <div class="gitHubCreateBranch__subHeader skeleton skeleton-element">Creating a branch for <b>{{issue.key}}</b></div>
```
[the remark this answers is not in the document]
joshkay10: bitbucket/Jira were super inconsistent in skellied / not skellied content. I guess showing text we have straight away makes sense. Need a mini rule book :(

static/js/skeleton.js
joshkay10: Guess it depends if we want a blanket skellie or individual. This will work on the indiviual element but could leave us with half a skellie page?

views/github-create-branch.hbs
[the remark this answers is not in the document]
rachellerathbone: Jade sent me this screenshot and asked to hide all of the following:

image

static/js/skeleton.js
rachellerathbone: We can play around with that if needed when updating other pages. I don't follow what you mean: 'we can get it working by simply removing the skeleton classes after the .auiSelect2().' ??

krazziekay: In that case, it's fine. 
It would be better to change this into a function and make `500` as default for a parameter, so that we can set different values if used in other components(optional).

Harminder84: Can we remove skeleton classes on select2 load event?

krazziekay: No, `select2-load` event is triggered only when the dropdown is opened.

static/js/skeleton.js
joshkay10: 👍

static/js/skeleton.js
rachellerathbone: > It would be better to change this into a function and make 500 as default for a parameter, so that we can set different values if used in other components(optional).

We won't know that though until we implement it in other parts of the app. My inclination is to leave as is for and then, when someone picks up the work to switch from the spinner to skeleton, if we determine the 0.5 sec wait time doesn't work, refactor it. I don't want to refactor it now only to find out every location that uses it winds up passing the same value.

krazziekay (approved):

Harminder84 (approved):

PR #1653, opened 2022-10-09
ARC-1731 - limit initial repos list to connected repos

joshkay10:
**What's in this PR?**
Restricting the list of repos to ones that are accessible by the GitHub app

**Why**
Because users don't want to see what they cant use, plus maybe security?

**Added feature flags**
nope

**How has this been tested?**  
Local and staging

**Whats Next?**
Unit tests 

Apply the same rules to the search

limit the search by org when there is just too damn many!


src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -26,11 +30,18 @@ export const GithubCreateBranchGet = async (req: Request, res: Response, next: N
 		return next(new Error(Errors.MISSING_ISSUE_KEY));
 	}
 
-	const branchSuffix = summary ? replaceSpaceWithHyphenHelper(summary as string) : "";
+	const subscriptions = await Subscription.getAllForHost(jiraHost);
+
+	// TODO - this should redirect to a you are not configured page instead.
```
joshkay10, line 35: Will create a new story for this, bascially if they haven't configured we should kick them off to somewhere betters

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -52,3 +64,23 @@ export const GithubCreateBranchGet = async (req: Request, res: Response, next: N
 		jiraHost
 	});
 };
+
+const getEpochFromDateString = (d: string): number => {
```
joshkay10, line 68: don't like this refactoring, 1 moment

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -52,3 +62,27 @@ export const GithubCreateBranchGet = async (req: Request, res: Response, next: N
 		jiraHost
 	});
 };
+
+const sortByDateString = (a, b) => {
+	return new Date(b).valueOf() - new Date(a).valueOf();
+};
+
+const getReposBySubscriptions = async (subscriptions: Subscription[], jiraHost: string, gitHubAppId: number | undefined, logger: Logger): Promise<RepositoryNode[]> => {
```
joshkay10, line 70: This basically does the get repos call for each subscription... this is mostly fine..
of 100000 subscriptions only ~300 have four or more subs. There is a handful of 20's and 30's this should still work.... might be a bit poor performance though, so as a seperate story we can investigate introducing orgs option so that if we are overwhelmed with options we can refine to what they chose. Its an extra step which is lame but we are talking 0.1% of customers

krazziekay (approved):
Looks good...

src/models/subscription.ts
```diff
@@ -190,6 +194,19 @@ export class Subscription extends Model {
 		return results[0] as SyncStatusCount[];
 	}
 
+	/*
+	 * Returns array or orgs that have connected repos
+	 */
+	static async getConnectedOrgs(jiraHost: string): Promise<RepoOwner[]> {
```
Harminder84, line 200: Where is this being used?

joshkay10: well...... its funny you should mention that.
I was originally leaving it there for the next step, for the search but its not going to work as I had hoped because we don't maintain repos :'(
So I shall remove it!

Harminder84 (approved):

PR #1658, opened 2022-10-11
Fix repo listing

Harminder84:
**What's in this PR?**
Filter out subscriptions based on githubAppId. Earlier it was showing repos from all subscriptions which means repo listing for cloud also includes enterprise repos.

**How has this been tested?**  
Local

### DIFF
--- a/src/models/subscription.ts
+++ b/src/models/subscription.ts
@@ -44,16 +44,7 @@ export class Subscription extends Model {
 	static async getAllForHost(jiraHost: string, gitHubAppId?: number): Promise<Subscription[]> {
 		return this.findAll({
 			where: {
-				...(gitHubAppId && { gitHubAppId }), // Add gitHubAppId only if passed
-				jiraHost
-			}
-		});
-	}
-
-	static async getAllForHostAndCloud(jiraHost: string): Promise<Subscription[]> {
-		return this.findAll({
-			where: {
-				gitHubAppId: null,
+				...(gitHubAppId !== undefined && { gitHubAppId }), // Add gitHubAppId only if passed
 				jiraHost
 			}
 		});
--- a/src/models/subscription.ts
+++ b/src/models/subscription.ts
@@ -50,6 +50,15 @@ export class Subscription extends Model {
 		});
 	}
 
+	static async getAllForHostAndCloud(jiraHost: string): Promise<Subscription[]> {
+		return this.findAll({
+			where: {
+				gitHubAppId: null,
+				jiraHost
+			}
+		});
+	}
+
 	static getAllForInstallation(
 		gitHubInstallationId: number,
 		gitHubAppId: number | undefined
--- a/src/routes/github/create-branch/github-create-branch-get.ts
+++ b/src/routes/github/create-branch/github-create-branch-get.ts
@@ -26,13 +26,7 @@ export const GithubCreateBranchGet = async (req: Request, res: Response, next: N
 	if (!key) {
 		return next(new Error(Errors.MISSING_ISSUE_KEY));
 	}
-
-	let subscriptions;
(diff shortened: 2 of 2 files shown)


src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -27,7 +27,12 @@ export const GithubCreateBranchGet = async (req: Request, res: Response, next: N
 		return next(new Error(Errors.MISSING_ISSUE_KEY));
 	}
 
-	const subscriptions = await Subscription.getAllForHost(jiraHost);
+	let subscriptions;
+	if (gitHubAppConfig.gitHubAppId) {
+		subscriptions = await Subscription.getAllForHost(jiraHost, gitHubAppConfig.gitHubAppId);
```
joshkay10, line 32: Will this return cloud values if we pass undefined for gutHubAppId?

joshkay10 (commented):
GOOD FIND!!

src/routes/github/create-branch/github-create-branch-get.ts
krazziekay: I agree, `getAllForHost` already handles the case for both GH cloud and GHE server. So, don't see much point for this `if-else`.

src/models/subscription.ts
```diff
@@ -50,6 +50,15 @@ export class Subscription extends Model {
 		});
 	}
 
+	static async getAllForHostAndCloud(jiraHost: string): Promise<Subscription[]> {
```
krazziekay, line 53: Isn't it better to just modify the existing `getAllForHost`? 
If we change the condition in `where` to do add the `gitHubAppId` if its not undefined, then it should work. 
For cloud, we pass null values: `.getAllForHost(jiraHost, null)`, which makes sense cause cloud apps have `null` `gitHubAppId`
And for enterprise, we pass number: `getAllForHost(jiraHost, numberVal)`.

This way we won't need the if-else check in the `GithubCreateBranchGet`. We can just do `await Subscription.getAllForHost(jiraHost, gitHubAppConfig.gitHubAppId || null)`.

```suggestion
					...(gitHubAppId !== undefined && { gitHubAppId }), // Add gitHubAppId only if passed

```

src/routes/github/create-branch/github-create-branch-get.ts
Harminder84: This will return both cloud and server if we pass undefined for gitHubAppId

src/models/subscription.ts
Harminder84: `getAllForHost` is being used at multiple places. If we change modify the existing implementation, it may impact other flows.

krazziekay: This change won't affect any other places. 
Originally, the second parameter is just another filter to fetch the list of subscriptions for server apps. So, the only use cases for this method so far is one of these two:
- `getAllForHost(jiraHost)` - without the second param, which means its `undefined`.
- `getAllForHost(jiraHost, id)` - with second param, where its a number.

What I'm suggesting is to avoid adding the `gitHubAppId` parameter only if it is `undefined`.  Basically separating `undefined` and `null` cases.

So, for the current case, we need to fetch the subscriptions for cloud or ghe server.

And if we get the subscriptions using: 
`await Subscription.getAllForHost(jiraHost, gitHubAppConfig.gitHubAppId || null)`

Then this will work for both cases:
- For cloud it will pass in null value and the filter becomes  `{ gitHubAppId: null, jiraHost }` 
- And for enterprise, it will pass in a number value and the filter becomes `{ gitHubAppId, jiraHost }`.

And for the already used ones, nothing's really changed.

krazziekay (approved):

PR #1659, opened 2022-10-12
[QUICK-FIX] - Added an ID for console errors

krazziekay:
**What's in this PR?**
- Added a `UUID` for console errors in `logger.ts`
- Added a custom error handler for the errors caught by **Express JS**.
  - Added a `UUID` as `traceId`. The user only receives this `traceId` and from our end we can easily search the Splunk logs using this `traceId`

**Why**
- For better readability in Splunk(and other logging tools maybe)

**Whats Next?**
- Not much, just merge and go...

### DIFF
--- a/src/config/logger.ts
+++ b/src/config/logger.ts
@@ -1,6 +1,7 @@
 import Logger, { createLogger, LogLevel, Serializers, Stream } from "bunyan";
 import { isArray, isString, merge, omit } from "lodash";
 import { SafeRawLogStream, UnsafeRawLogStream } from "utils/logger-utils";
+import { v4 as newUUID } from "uuid";
 
 function censorUrl(url) {
 	if (!url) {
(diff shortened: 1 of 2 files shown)


src/config/logger.ts
```diff
@@ -110,6 +111,7 @@ interface LoggerOptions {
 export const getLogger = (name: string, options: LoggerOptions = {}): Logger => {
 	return createLogger(merge<Logger.LoggerOptions, LoggerOptions>({
 		name,
+		id: newUUID(),
```
bgvozdev, line 114: I think this is not needed because we add one in frontend-middleware-logger, don't we?

krazziekay:
Closing because I could not override the Express JS error handling. The app is probably using `Probot`'s error handling, so leaving it as it is

PR #1660, opened 2022-10-12
ARC-1728 - Success Page after creating a branch

krazziekay:
**What's in this PR?**
- Added the final success state in the same page using jQuery. This is not done in a separate route as intended previously, because:
  - Creating a new route means redirect, which means we lose the animating flow.
  - There’s not much value in creating a separate route for this.
- After the branch has been created successfully, added a 1.5 second timeout to show the action links.

**Why**
- Following the UI with the smooth animation 😎 

**How has this been tested?**  
- Local
- Staging

**Video**

https://user-images.githubusercontent.com/13076255/195258732-f6528e8d-aab3-48da-80e2-94c754bd53cf.mov

### DIFF
--- a/src/routes/github/create-branch/github-create-branch-get.ts
+++ b/src/routes/github/create-branch/github-create-branch-get.ts
@@ -47,6 +47,7 @@ export const GithubCreateBranchGet = async (req: Request, res: Response, next: N
 			key
 		},
 		repos,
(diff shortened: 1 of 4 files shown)


joshkay10 (approved):
I prefer it staying on the page so that a win to me!!

PR #1662, opened 2022-10-12
ARC-1736 create branch ff view

rachellerathbone:
**What's in this PR?**
A temporary view that we'll render if the CB ff is off for customers.

**Why**
Coz it's what the CB team decided to do. Figured it won't be up for long and this option has a much faster turn-around time then trying to battle with the Jira codebase 😱 Plus, we now have this coming soon component if we ever need it again.

**Added feature flags**
Pretty self-explanatory.

**Affected issues**  
ARC-1736

**How has this been tested?**  
Locally: ff off, ff on :)

https://user-images.githubusercontent.com/37155488/195266930-e07dab2b-a725-45df-b986-8345eb4673f3.mov

**Whats Next?**
Wrapping up 1740 - the initial view you see when the ff is on. Mostly done. Just need to do a couple of small things and some clean up.


views/github-create-branch-options.hbs
```diff
@@ -6,64 +6,73 @@
   <meta name="ap-local-base-url" content="{{localBaseUrl}}" />
   <link rel="stylesheet" href="/public/aui/aui-prototyping.css"
     integrity="DTM1Q+8lU7SzJT+FWr0JFisCSZlwfM0GiAKYy7h1s9vIKa/CIh37s9NuOCqIOgK4tmqrjLK4NuWuIPUQNsikHA=="
-    crossorigin="anonymous" referrerpolicy="no-referrer" />
+    crossorigin="anonymous" referrerpolicy="no-     referrer" />
```
joshkay10, line 9: looks like a tab snuck in

krazziekay (approved):

krazziekay (commented):
All good, just need to fix the broken test case.

src/routes/github/create-branch/github-create-branch-options-get.test.ts
```diff
@@ -3,6 +3,16 @@ import supertest from "supertest";
 import { getLogger } from "config/logger";
 import { getFrontendApp } from "~/src/app";
 import { getSignedCookieHeader } from "test/utils/cookies";
+import { booleanFlag, BooleanFlags } from "config/feature-flags";
+import { when } from "jest-when";
+
+jest.mock("config/feature-flags");
+
+const turnFF_OnOff = (newStatus: boolean) => {
```
krazziekay, line 11: /nit
```suggestion
const turnFF_On = (newStatus: boolean) => {
```

krazziekay (dismissed):

src/routes/github/create-branch/github-create-branch-options-get.test.ts
rachellerathbone: The function's job isn't to only turn the ff on otherwise we could just hardcode the value to `true`. The idea is that if you have multiple tests, you call that function and can turn it on/off based on what you pass in.

krazziekay: IMO, if we name the variable to `turnFF_On` and if we pass `true`, that means its **ON** and if we pass `false`, it means its **OFF**.
But when its `turnFF_OnOff` and we pass `true`, its not clear as if the **ON** part is true or the **OFF** part is true. I had to check the code and deduce when to pass true/false values. 
But yes, its def not a blocker, so feel free to leave it as it.

joshkay10 (changes requested):
Blocked while we attempt Monolith solution

rachellerathbone:
We ended up going with a diff option.

PR #1667, opened 2022-10-14
ARC-1755 validate source branch

joshkay10:
**What's in this PR?**
Parent branch gets checked against GitHub that it does indeed exist!

**Why**
If the user manually types in the branch we need to validate it as soon as possible and capture a nice error message. Leaving it to the submit will get an error but is often unhelpful!!!

**How has this been tested?**  
Stage unit tests

**Whats Next?**
Validate the branch name with same endpoint but expect as an inverse test (expect a 404)


static/js/github-create-branch.js
```diff
@@ -118,6 +122,20 @@ $(document).ready(() => {
...
 });
 
+
+const validateSourceBranch = (branchName) => {
+	hideValidationErrorMessage("ghParentBranch");
+	const repo = getRepoDetails();
+	const url = `/github/branch/owner/${repo.owner}/repo/${repo.name}/${branchName}`;
+
+	$.get(url)
+		.fail((error) => {
```
joshkay10, line 132: Only interested in 404
- Success ignore (potentially expand to give a tick or visual validation in future story)
 - all other 4**/5** are ignored for this check, we only care if we try and cant find it, other issue handled later

static/js/github-create-branch.js
```diff
@@ -81,6 +81,10 @@ $(document).ready(() => {
     data: []
   });
 
+	$("#ghParentBranch").on("change", function (e) {
+		validateSourceBranch(e.val);
```
joshkay10, line 85: Originally only intended to test "manual" branch case
Now we check every time. 1 - its simpler. 2 - (pretty edgy) It catches the case if people take a long time between selecting branhes. 3 - It uses the User token, so if there is any mismatch it will be picked up here early and clearer error message

krazziekay (approved):

krazziekay (dismissed):

krazziekay (approved):

PR #1672, opened 2022-10-17
Arc 1662 increase payload limit for enterprise

gxueatlassian:
**What's in this PR?**
Set the express json parse size to 50mb for ddev/stg/prod

**Why**
When we cut over to use new webhook route, the default size limit 1mb is not enough according to [github cloud](https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads) and [enterpriser server doc](https://docs.github.com/en/enterprise-server@3.3/developers/webhooks-and-events/webhooks/webhook-events-and-payloads).

image

**Added feature flags**
N/A

**Affected issues**  
ARC-1662

**How has this been tested?**  
Unit test.

**Whats Next?**
Test on stage and cut over prod traffic for webhook

### DIFF
--- a/src/routes/router.ts
+++ b/src/routes/router.ts
@@ -25,7 +25,9 @@ RootRouter.use(Sentry.Handlers.requestHandler());
 
 // Parse URL-encoded bodies for Jira configuration requests
 RootRouter.use(urlencoded({ extended: false }));
-RootRouter.use(json());
+RootRouter.use(json({
+	limit: "30mb" //set limit according to github doc https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-common-properties
+}));
 RootRouter.use(cookieParser());
 
 // Add pertinent information to logger for all subsequent routes


src/routes/router.ts
```diff
@@ -18,14 +20,18 @@ import { createAppClient } from "~/src/util/get-github-client-config";
...
 
 // The request handler must be the first middleware on the app
 RootRouter.use(Sentry.Handlers.requestHandler());
 
 // Parse URL-encoded bodies for Jira configuration requests
 RootRouter.use(urlencoded({ extended: false }));
-RootRouter.use(json());
+RootRouter.use(json({
+	limit: envVars.EXPRESS_PAYLOAD_LIMIT
+}));
```
mboudreau, line 33: Why do we need this as an environment variable?  It's probably not ever going to change.  Plus, you're not using a default.

Probably don't need to make this bigger than 5mb

mboudreau (changes requested):
No need to overcomplicate.  just set it directly.  5mb should be more than enough.

src/routes/router.ts
gxueatlassian: The env is handy for unit testing of that settings is working. But If that's "over complicated", I can remove the unit test if you like.

For "50mb", I think it is safe to do so, as from github doc (in description of this pr), `25mb` is the max. So I just double it.

mboudreau: ...but why double it if it's the max?

mboudreau: And if you want to test it, there are others ways of doing so without using an environment variable that'll never change.

gxueatlassian: It can be 30mb, 35mb, or whatever number. I just don't want to sit on the edge of 25mb.

gxueatlassian: Okay, let's recap on this:
1. I can remove it from env, I can agree the hardcoded value is simplier.
2. Do you think I should remove that test? Or just test against the hardcoded value (which can be done as well).
3. What number do you suggest? How about 30mb?

mboudreau: I don't think you need to test it.  It's a library that we trust and I'd imagine they already have a plethora of tests for it already.  Plus we know it works as per the issue this morning :P

I really don't think we need to go anything above 25mb, even if it's on the edge, that's just the maximum github will ever send, but it doesn't come close to that 99.999999% of the time since the limit was just 100kb before.

bgvozdev: I agree with Michel; I don't think we should test the underlying libraries. Also don't think it should be an env variable, cause we have plenty of them TBH and each has its own cognitive toll.

Instead, I'd suggest to add a link to GitHub doc explaining why we picked 25mb, wdyt?

bgvozdev:
Awesome finding @gxueatlassian ! 🥇

src/routes/router.ts
bgvozdev: Also yeah, sitting at 25mb might be dodgy (depending on how github calculates it). Should we have 20-25% of a gap just in case? Which is 31.25Mb.

gxueatlassian: Okay thanks both @mboudreau and @bgvozdev. 
I've updated to just 30mb. I see no points arguing about 25mb vs 30mb since only 5mb different.
But if you guys are still thinking 1mb vs 10mb vs 25mb, then we can continuing talking about it.

bgvozdev (approved):

src/routes/router.ts
```diff
@@ -25,7 +25,9 @@ RootRouter.use(Sentry.Handlers.requestHandler());
 
 // Parse URL-encoded bodies for Jira configuration requests
 RootRouter.use(urlencoded({ extended: false }));
-RootRouter.use(json());
+RootRouter.use(json({
+	limit: "30mb" //set limit according to github doc https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-common-properties
```
bgvozdev, line 29: /nit: your comment doesn't explain why it is 30mb, not 25 😛

mboudreau (approved):

PR #1675, opened 2022-10-17
ARC-1757 - Cancel button does not work when on Iframe version

krazziekay:
**What's in this PR?**
- Removed the cancel button

**Why**
- Doesn't work in iframes

**How has this been tested?**  
- Local


joshkay10 (approved):

joshkay10:
😭

PR #1677, opened 2022-10-17
ARC 1752 - Shows search text instead of no results found on repo search

krazziekay:
**What's in this PR?**
- Selection of unavailable repositories has been removed.
- Condition for displaying **Configure GitHub Integration** has been updated.
  - True only when the input has no value selected and the user types in a repo that is not available.
- **No results text** updated to **No search results**.
- Link added for **Configure GitHub Integration**.

**Why**
- Discussed during the catchup.

**How has this been tested?**  
- Local

**Video**

https://user-images.githubusercontent.com/13076255/196088442-d4bd0078-b1df-4e77-8899-2854640033a5.mov

### DIFF
--- a/views/github-create-branch.hbs
+++ b/views/github-create-branch.hbs
@@ -61,8 +61,7 @@
         </a>
         <aui-inline-dialog id="show-no-repo-popup" alignment="bottom left" responds-to="hover">
           <div>You need to configure your GitHub repositories to your Jira site to see them here.</div>
-          <!--TODO: Add URL after confirmation-->
-          <a href="#" class="configure-link">Configure GitHub Integration</a>
+          <a href="{{jiraHost}}/plugins/servlet/ac/com.github.integration.production/github-post-install-page" target="_blank" class="configure-link">Configure GitHub Integration</a>
         </aui-inline-dialog>
       </div>
     </div>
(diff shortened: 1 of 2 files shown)


views/github-create-branch.hbs
```diff
@@ -61,8 +61,7 @@
         </a>
         <aui-inline-dialog id="show-no-repo-popup" alignment="bottom left" responds-to="hover">
           <div>You need to configure your GitHub repositories to your Jira site to see them here.</div>
-          <!--TODO: Add URL after confirmation-->
-          <a href="#" class="configure-link">Configure GitHub Integration</a>
+          <a href="{{jiraHost}}/plugins/servlet/ac/com.github.integration.production/github-post-install-page" target="_blank" class="configure-link">Configure GitHub Integration</a>
```
krazziekay, line 64: This page is either shown inside the iframe or in a separate new tab. If it was inside iframe, we needed to use `AP.navigator` and if its new tab, will have to provide URL. 
So, just hard-coded the prod URL instead.

rachellerathbone: Nit: Integration should be lowercased.

Screen Shot 2022-10-17 at 3 48 17 pm

krazziekay: Casing corrected

rachellerathbone (approved):

PR #1678, opened 2022-10-17
ARC-1734 - Not configured page

krazziekay:
**What's in this PR?**
- Added a new view for non-configured state.
- Redirected to this new non-configured view when there are no apps connected.

**Why**
- To guide users who have no connections

**How has this been tested?**  
- Local
- Staging

**Screenshot**
Screen Shot 2022-10-17 at 5 09 13 pm


src/routes/github/create-branch/github-create-branch-options-get.ts
```diff
@@ -26,6 +26,13 @@ export const GithubCreateBranchOptionsGet = async (req: Request, res: Response,
 
 	try {
 		const url = new URL(`${req.protocol}://${req.get("host")}${req.originalUrl}`);
+		if (githubToken && !servers.hasCloudServer && !servers.gheServerInfos.length) {
```
rachellerathbone, line 29: Looks like this conditional and the one below are checking if there's a GH token and if there is/isn't servers connected. What if there is no GH token? What if there's a token and the customer is only connected to cloud? This means they would see this screen.

I thought we were rendering this page only if there was no GH token otherwise:
- render the form for customers that have cloud/server connected
- render the select a product screen for customers that have cloud + server connected.

src/routes/github/create-branch/github-create-branch-options-get.ts
```diff
@@ -26,6 +26,13 @@ export const GithubCreateBranchOptionsGet = async (req: Request, res: Response,
...
+			res.render("no-configuration.hbs", {
```
rachellerathbone, line 30: You'll need to update the test for this to check we're seeing the correct components based on the conditions.

krazziekay: Test case updated

src/routes/github/create-branch/github-create-branch-options-get.ts
krazziekay: Condition updated to check for the `gitHubToken` OR the number of connections.

joshkay10:
Can we redirect them on the github-create-branch.ts route as well.
Probably shouldn't arrive here unless they try hit it directly 🤷  but probs nicer than what we have.

replace this 
```
	// TODO - this should redirect to a you are not configured page instead.
	if (!subscriptions) {
		return next(new Error(Errors.MISSING_CONFIGURAITON));
	}
```
with a res.redirect maybe
could remove the MISSING_.CONFIGURAITON error as well, was only ever a temp thing

src/routes/github/create-branch/github-create-branch-options-get.ts
```diff
@@ -26,6 +26,13 @@ export const GithubCreateBranchOptionsGet = async (req: Request, res: Response,
 
 	try {
 		const url = new URL(`${req.protocol}://${req.get("host")}${req.originalUrl}`);
+		if (!githubToken || (!servers.hasCloudServer && !servers.gheServerInfos.length)) {
```
joshkay10, line 29: A configured customer without a token is valid.

If they don't have a token, they will get kicked off to authenticate... then they will reload the options page and it should catch the unconfigured state then

```suggestion
		if (!servers.hasCloudServer && !servers.gheServerInfos.length) {
```

joshkay10: actually, hmmmm that might not work.

krazziekay: This route `GitHubCreateBranchOptionsGet` is not inside `GitHubRouter`, so the auth dance doesn't happen at this moment.

joshkay10: ok, in this scenario, we need the githubtoken to exist otherwise we aren't authenticated against server or cloud so we dont know if the apps have been configured or not so its full circle

`githubToken && !servers.hasCloudServer && !servers.gheServerInfos.length`

So the variations as I see them
i've installed cloud but not configured(same for single server)
With token: i come here
without token: i login to github, try hit this page again then come here

With multiple servers or cloud and server - none configure
With token: i come here
without token: i login to github, try hit this page again then come here

krazziekay: Updated and removed the `gitHubToken` check

src/routes/github/create-branch/github-create-branch-options-get.ts
```diff
@@ -26,6 +26,13 @@ export const GithubCreateBranchOptionsGet = async (req: Request, res: Response,
 
 	try {
 		const url = new URL(`${req.protocol}://${req.get("host")}${req.originalUrl}`);
+		if (!servers.hasCloudServer && !servers.gheServerInfos.length) {
```
joshkay10, line 29: This is seperate logic to what the try block is doing, we can move it just below 
`const servers = await getGitHubServers(jiraHost);`

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -32,9 +32,12 @@ export const GithubCreateBranchGet = async (req: Request, res: Response, next: N
 	}
 	const subscriptions = await Subscription.getAllForHost(jiraHost, gitHubAppConfig.gitHubAppId || null);
 
-	// TODO - this should redirect to a you are not configured page instead.
+	// In case if somehow the user lands on this page directly from URL
```
rachellerathbone, line 35: nit: can we reword this? It reads a little funny.

rachellerathbone, line 35: nit: can we reword this? It reads a little funny.

krazziekay: I rushed with this comment. I've the comments

src/routes/github/create-branch/github-create-branch-options-get.ts
krazziekay: Moved...

joshkay10 (approved):

PR #1682, opened 2022-10-18
Skip repo resync if tasks omit it

joshkay10:
**What's in this PR?**
Skip repository backfill if its a targeted backfill that doesn't include repository

**Why**
During the most recent recent backfill for PULL only we are picking up considerable work for repository, which we want to skip at this stage.

Task count for current backfill(from Splunk(for 1000 installations))
pull | 91
repository | 608315

lolololol


src/sync/installation.ts
```diff
@@ -45,7 +45,7 @@ export const getTargetTasks = (targetTasks?: TaskType[]): TaskType[] => {
 const getNextTask = async (subscription: Subscription, targetTasks?: TaskType[]): Promise<Task | undefined> => {
 	const tasks = getTargetTasks(targetTasks);
```
joshkay10, line 46: If no targets tasks are provided then this function(getTargetTasks) returns all tasks including repository.

So the new logic(&& tasks.includes("repository")) should only evaluate true when a subset of tasks is passed that omits repository

src/sync/installation.ts
```diff
@@ -45,7 +45,7 @@ export const getTargetTasks = (targetTasks?: TaskType[]): TaskType[] => {
...
 
-	if (subscription.repositoryStatus !== "complete") {
+	if (subscription.repositoryStatus !== "complete" && tasks.includes("repository")) {
```
gxueatlassian, line 48: /nit
Or alternatively, instead of modifying the allTaskTypes, we could just change this line?
`if (... && targetTasks.includes("repository")) {`

bgvozdev:
Shouldn't we be fixing `/resync` endpoint instead so it won't be changing the `repositoryStatus`?

bgvozdev (changes requested):
Waiting for reply to my comment, I think this is might be a risky/unnecessary change

joshkay10:
> Shouldn't we be fixing `/resync` endpoint instead so it won't be changing the `repositoryStatus`?

Thanks for the comments!!

Last week we made a change to `/resync` to only set repositoryStatus to null if repository is passed in as a target.
When I tried to resync Pulls only again last week it managed to pick up a lot of repository events 1000x more than pulls 😆 

Since the condition is repoStatus !== **complete** I dug into the Subscriptions db and majority are NULL 😢 

This is not unexpected since repo sync is relatively new, so would have existed pre-repo sync functionality.

**pending: 8
complete: 30690
failed: 354
null: 93465**

bgvozdev:
But shouldn't we then create an API endpoint that would set it to "complete" for all such records instead?

bgvozdev (approved):

bgvozdev:
I'm putting it to "approved" only to unblock it, my availability will suffer today. But I find it bizarre that we are changing our production code because our API endpoint doesn't do what we want.

src/sync/installation.ts
bgvozdev: I don't think this change is covered by a test...

joshkay10: @gxueatlassian - allTaskTypes is the default list to backfill, so if we add the condition without adding it to allTaskTypes it will never do repo discovery unless we explicitly state, but we want it as default.

@bgvozdev - Adding test now :)

mboudreau (approved):

bgvozdev:
So, we have chatted and looks like the problem is a bit deeper:
 - http://github.com/atlassian/github-for-jira/blob/fcb0dd55b171b0e3e6f5539f63ba22b531c32017/src/sync/installation.ts#L381-L381 - here we have a generic logic to mark the repo as failed and continue with the next one, for all syncs, including repos
 - while it doesn't make sense at all: repos are not synced per-repo at all (e.g. [here](http://github.com/atlassian/github-for-jira/blob/fcb0dd55b171b0e3e6f5539f63ba22b531c32017/src/sync/installation.ts#L50-L50) we are setting up repositoryId to 0, which means the aforementioned call to `markCurrentRepositoryAsFailedAndContinue` won't do anything
 - the same message will be sent and the same logic will be executed.

Instead of doing what we are doing now, we should throw a error, therefore it will be retried later (SQS). If all SQS attempts will failed, the sync will be stuck, and this will be a legitimate thing. How can we do "sync" when we cannot even fetch repos?...

Ideally, we should also watch for the number of retries and put the sync into failed state if that was the last SQS attempt.

gxueatlassian:
> repositoryId to 0, which means the aforementioned call to `markCurrentRepositoryAsFailedAndContinue` won't do anything

Doesn't it call `updateRepo` and set the `repositoryStatus` to `fail`? And therefore next time it shouldn't pickup the `repository` task again?

joshkay10:
We no longer want to go down this path

PR #1688, opened 2022-10-19
ARC-1758 - Show "connecting to github" on create-branch-options page when authenticating

krazziekay:
**What's in this PR?**
- Added an intermediate(loading) view when there is no `gitHubToken`

**Why**
- When the user is creating a branch for the first time(or when the token is not saved in session yet), the `iframe` window opens another window for the OAuth dance. During this, the `iframe` still remains, so need something to indicate that this window is idle for this case.

**How has this been tested?**  
- Local
- Staging

**Video**

https://user-images.githubusercontent.com/13076255/196568958-c8be696c-c8a5-48ff-8a68-ae28c1b93d22.mov


views/github-create-branch-options.hbs
```diff
@@ -57,6 +58,13 @@
     </form>
   </section>
 
+  <div class="gitHubCreateBranchOptions__loading">
+    {{> loading-screen
+        title="Connecting to GitHub"
+        message="Please close this window when you are done creating a branch!"
```
joshkay10, line 64: Not sure if we need this one?
message="Please close this window when you are done creating a branch!"

krazziekay: An optional text, so that the users would know to close the iframe window if it is still opening somehow.

joshkay10: mmm fair, 
can we make it not break onto a new line with styles or slightly shorter message?

krazziekay: Updated the text to `Please close this window after creating a branch.`

joshkay10 (dismissed):

joshkay10 (approved):

PR #1691, opened 2022-10-19
Arc 1770 fix outbound proxy for oauth router

gxueatlassian:
**What's in this PR?**
Set to use outbound proxy for oauth router.

**Why**
Otherwise enterprise customer who limiting ip will fail to use the app.

**Added feature flags**
use-outbound-proxy-for-oauth-router

**Affected issues**  
ARC-1770

**How has this been tested?**  
NONE [Work in progress]

**Whats Next?**
Deploy and FF on and test.


src/github/client/github-anonymous-client.ts
```diff
@@ -27,4 +27,38 @@ export class GitHubAnonymousClient extends GitHubClient {
...
+					"content-type": "application/json"
+				},
+				responseType: "json"
+			}
+		);
+		return response.data.access_token;
+	}
+
+	public async checkGitHubToken(gitHubToken: string) {
+		await this.axios.get("", {
```
krazziekay, line 57: Out of curiosity, this is an empty string, so will it work? In most places  a `"/"` is used, is the same here?

PR #1693, opened 2022-10-19
Tenanthost

joshkay10:
**What's in this PR?**
WE can now pass the tenanthost(jirahost) from the create branch link, so this PR makes use of that fact and stops nesting everything inside of iframes.

**Why**
iFrames within Jira are very limiting when it comes to doing most things we want to do, we get forced out of them sometime and sometimes not so to keep a consistent experience we can now live on a new tab like all other github pages.

**How has this been tested?**  
Tested heaviliy in stage
Updated unit tests and added additional one to make sure user/installation data coming back is filtered correctly


src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -11,10 +11,12 @@ const MAX_REPOS_RETURNED = 20;
 
 export const GithubCreateBranchGet = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
 	const {
-		jiraHost,
+		jiraHost: jiraHostLocal,
```
joshkay10, line 14: open to opinions on naming...

But the jirahost could come from two places, 1) - the session.. yay 2) query param....

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -11,10 +11,12 @@ const MAX_REPOS_RETURNED = 20;
...
 		githubToken,
 		gitHubAppConfig
 	} = res.locals;
+	const { jiraHost: jiraHostParam } = req.query;
+	const jiraHost = jiraHostLocal || jiraHostParam;
```
joshkay10, line 19: We prioritize the session jiraHost over the param version, simply because we know the session one is always solid params should be but more opportunity to be wrong

src/routes/github/github-router.ts
```diff
@@ -13,11 +13,14 @@ import { UUID_REGEX } from "~/src/util/regex";
 import { GithubCreateBranchRouter } from "routes/github/create-branch/github-create-branch-router";
 import { GithubRepositoryRouter } from "routes/github/repository/github-repository-router";
 import { GithubBranchRouter } from "routes/github/branch/github-branch-router";
+import { GithubInstanceRouter } from "routes/github/instance/github-instance-router";
 
 export const GithubRouter = Router();
 const subRouter = Router({ mergeParams: true });
 GithubRouter.use(`/:uuid(${UUID_REGEX})?`, subRouter);
 
+subRouter.use("/instance", GithubInstanceRouter);
```
joshkay10, line 22: ???? interesting... continue the adventure down to see what the what!

src/routes/github/instance/github-instance-get.ts
```diff
@@ -0,0 +1,90 @@
+import { NextFunction, Request, Response } from "express";
```
joshkay10, line 1: What have we here!!?!?

This is the `/create-branch-options-get` file moved and slighlty modified to route` github/instance-get`

**Opinions encouraged!!!!!**

This moves the route under GitHub 💚 
It is instance agnostic, so it doesnt car if we know the uuid or not its a know nothing at this stage. 💚 
Its re-usable... mostly instead of redirecting to create-branch it takes a redirect query param 💚 

It does feel a little messy in its placement in the router specifically that its part of the subrouter that gets thrown on here `GithubRouter.use(`/:uuid(${UUID_REGEX})?`, subRouter)`  but maybe thats ok???? 🔴

src/routes/github/instance/github-instance-get.ts
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}
+
+	return {
+		hasCloudServer: ghCloudSubscriptions.length,
+		gheServerInfos
+	};
+};
+
+
```
joshkay10, line 90: Tests are coming!

src/routes/github/create-branch/github-create-branch-options-get.test.ts
```diff
@@ -13,12 +13,6 @@ jest.mock("config/feature-flags");
 
 describe("GitHub Create Branch Options Get", () => {
 	let app: Application;
-	const setupGitHubCloudPingNock = () => {
-		githubNock.get("/").reply(200);
-	};
-	const setupGHEPingNock = () => {
-		gheApiNock.get("").reply(200);
-	};
```
joshkay10, line 21: We don't need to validate the github token manually any more yay!!!!

src/routes/github/create-branch/github-create-branch-options-get.ts
```diff
@@ -68,13 +62,9 @@ export const GithubCreateBranchOptionsGet = async (req: Request, res: Response,
...
 
-const validateGitHubToken = async (jiraHost: string, githubToken: string, logger: Logger, gitHubAppId?: number) => {
-	const githubUrl = await getGitHubApiUrl(jiraHost, gitHubAppId, logger);
-	await axios.get(githubUrl, {
-		headers: {
-			Authorization: `Bearer ${githubToken}`
-		}
-	});
+const getJiraHostFromTenantUrl = (jiraHostParam) =>  {
+	const siteName = jiraHostParam?.substring(0, jiraHostParam?.indexOf("."));
```
joshkay10, line 66: The format that tenant url get passed to us from jira is
<JIRAHOSTNAME>.<DOMAIN>

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
```diff
@@ -170,12 +170,12 @@ const modules = {
...
 export const moduleUrls = compact(map([...modules.adminPages, ...modules.generalPages], "url"));
 
-// Remove this function when CREATE_BRANCH flag is complete
+// Remove this function when CREATE_BRANCH flag is complete!!!!
 const addCreateBranchAction = async (modules) => {
 	if (await booleanFlag(BooleanFlags.CREATE_BRANCH, false)) {
 		modules.jiraDevelopmentTool.actions = {
 			createBranch: {
-				templateUrl: `/plugins/servlet/ac/${key}/create-branch-options?ac.issueKey={issue.key}&ac.issueSummary={issue.summary}`
+				templateUrl: `${envVars.APP_URL}/create-branch-options?issueKey={issue.key}&issueSummary={issue.summary}&tenantUrl={tenant.url}`
```
joshkay10, line 178: Excite, this is now throwing us directly to the app instead of iframe hell

static/js/github-create-branch-options.js
```diff
@@ -61,20 +40,3 @@ const ghServerOptionHandler = (event) => {
     $(".gitHubCreateBranchOptions__serversContainer").hide();
   }
 };
-
```
joshkay10, line 64: All rerouting is handled on the express side of the fence!

src/routes/github/create-branch/github-create-branch-options-get.ts
```diff
@@ -24,10 +21,15 @@ export const GithubCreateBranchOptionsGet = async (req: Request, res: Response,
...
+	const jiraHost = jiraHostLocal || getJiraHostFromTenantUrl(tenantUrl);
+
+	// TODO move to middleware or shared for create-branch-get
 	const servers = await getGitHubServers(jiraHost);
 	if (!servers.hasCloudServer && !servers.gheServerInfos.length) {
+		const instance = process.env.INSTANCE_NAME;
 		res.render("no-configuration.hbs", {
-			nonce: res.locals.nonce
+			nonce: res.locals.nonce,
+			configurationUrl: `${jiraHost}/plugins/servlet/ac/com.github.integration.${instance}/github-select-product-page`
```
joshkay10, line 32: since we no longer exist in the iframe we need to be more explicit with where to go, we dont have AP.navigator access on the FE so we pass the path instead

rachellerathbone (approved):

PR #1694, opened 2022-10-19
NONE: add logs about the usage of outbound proxy

bgvozdev:
**What's in this PR?**
More logs

**Why**
Cause it is unclear why the calls might be failing to internal GHE

### DIFF
--- a/src/util/get-github-client-config.ts
+++ b/src/util/get-github-client-config.ts
@@ -57,9 +57,11 @@ async function calculateProxyBaseUrl(jiraHost: string, gitHubBaseUrl: string | u
 			skipOutboundProxy = false;
 		}
 		if (skipOutboundProxy) {
+			logger.warn("Skip outbound proxy");
 			return undefined;
 		}
 	}
+	logger.info("Use outbound proxy");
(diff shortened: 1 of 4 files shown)


src/util/get-github-client-config.ts
```diff
@@ -57,9 +57,11 @@ async function calculateProxyBaseUrl(jiraHost: string, gitHubBaseUrl: string | u
 			skipOutboundProxy = false;
 		}
 		if (skipOutboundProxy) {
+			logger.warn("Skip outbound proxy");
```
gxueatlassian, line 60: Does these two lines include the jira host and ghe server url as well? 
If not, how can these two logging lines help us?

Harminder84 (approved):

src/util/get-github-client-config.ts
bgvozdev: Yes, logger is a child of the request's logger, and it has all the data. Not sure about GHE server URL, but if it is logged on error, this line can be found using request_Id

Harminder84 (approved):

PR #1701, opened 2022-10-21
ARC 1784 - Adding URL for the 403 error message

krazziekay:
**What's in this PR?**
- Added a new method to fetch the subscription for a given `repository`(`name` and `owner`) and `jiraHost`.
  - This new method uses a JOIN query. 
  - Used the new method to get the `gitHubInstallationId`.
- Added a method to generate the **URL** for installations settings in `GitHub` and added this URL for 403 errors. 

**Why**
- To fill out the URL for `Allow access to this repository` in case of 403 errors

**How has this been tested?**  
- Local
- Staging

**Screenshot**
- Cloud
Screen Shot 2022-10-21 at 3 16 33 pm

- Enterprise
Screen Shot 2022-10-21 at 3 16 06 pm

**Video**

https://user-images.githubusercontent.com/13076255/197114463-a0a27eb6-1111-4abd-83c9-1cbe6c861a33.mov


src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -5,20 +5,56 @@ import { AnalyticsEventTypes, AnalyticsTrackEventsEnum } from "interfaces/common
...
-	422: "We couldn’t create this branch, possibly because:<ul>" +
-		"<li>This GitHub branch already exists. Please try a different branch name.</li>" +
-		"<li>The GitHub branch name is not valid.</li>" +
-		"</ul>",
-	404: "We couldn’t create this branch, possibly because the GitHub source branch you entered doesn't exist. Please try a different branch.",
-	500: "We couldn’t create this branch because something went wrong. Please try again."
+const errorMessages = (statusCode: number, url?: string): string => {
+	switch (statusCode) {
+		case 403: {
+			return `We couldn’t create this branch, possibly because this GitHub repository hasn't been configured to your Jira site. <a href="${url}" target="_blank">Allow access to this repository.</a>`;
```
joshkay10, line 42: I'm taking ages to review because I keep coming back to the switch statement....
Id like another opinion I'm on the fence.

I prefer the original format here,  I understand why we changed it but perhaps we leave this as it was and then interpolate the result. Changing the structure of this to solve logic somewhere else too heavy handed.

e.g. if it was all still 403: "blah blah blah maybe blah {url}"...

Then in the if statement below
```
	if (err.status === 403) {
			const gitHubConfigurationLink = await getGitHubConfigurationLink(gitHubAppConfig, gitHubUserClient, jiraHost, owner, repo);
			const forbiddenErrorYo = genericUtilReplaceStringfunctionOfNice(errorMessages(err?.status || 500, gitHubConfigurationLink) // this replaces ? with ...args
						res.status(err.status).json(forbbinErrorYo);
```

src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -52,12 +93,19 @@ export const GithubCreateBranchPost = async (req: Request, res: Response): Promi
 		statsd.increment(metricCreateBranch.created, tags);
 	} catch (err) {
 		req.log.error({ err }, "Error creating branch");
-		res.status(err.status).json(errorMessages[err?.status || 500]);
+
+		if (err.status === 403) {
+			const user = await gitHubUserClient.getUser();
+			const gitHubConfigurationLink = await getGitHubConfigurationLink(user.data.login, jiraHost, gitHubAppConfig.hostname, owner, repo);
+			res.status(err.status).json(errorMessages(err?.status || 500, gitHubConfigurationLink));
```
bgvozdev, line 100: `err?.status` is always defined here, right?

src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -52,12 +93,19 @@ export const GithubCreateBranchPost = async (req: Request, res: Response): Promi
...
+
+		} else {
+			res.status(err.status).json(errorMessages(err?.status || 500));
```
bgvozdev, line 103: `err.status` might be not defined here, right? How would `res.status(undefined)` behave?

src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -5,20 +5,61 @@ import { AnalyticsEventTypes, AnalyticsTrackEventsEnum } from "interfaces/common
...
+	if (userLogin === repoOwner) {
+		url = `${hostname}/settings/installations/`;
+	}
+
+	const subscription = await Subscription.findForRepoNameAndOwner(repoName, repoOwner, jiraHost);
+
+	return url + subscription?.gitHubInstallationId;
+};
+
+const errorMessages = (statusCode: number, url?: string): string => {
```
bgvozdev, line 39: As this is a function, shouldn't the name contains some action? "Get" ? "Map"?

src/models/subscription.ts
```diff
@@ -153,6 +153,22 @@ export class Subscription extends Model {
 		});
 	}
 
+	static async findForRepoNameAndOwner(repoName: string, repoOwner: string, jiraHost: string): Promise<Subscription | null> {
```
bgvozdev, line 156: tests?

src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -5,20 +5,61 @@ import { AnalyticsEventTypes, AnalyticsTrackEventsEnum } from "interfaces/common
...
+/**
+ * Returns the URL for Installation settings page in GitHub
+ *
+ * @param userLogin
+ * @param hostname
+ * @param jiraHost
+ * @param repoOwner
+ * @param repoName
+ */
+const getGitHubConfigurationLink = async (
```
bgvozdev, line 19: I see many changes but 0 tests...

bgvozdev (changes requested):
Please add unit tests 🙏

src/routes/github/create-branch/github-create-branch-post.ts
krazziekay: The original format with the object mapping was easier. But that means we have to modify the message again if its a 403. I prefer getting the modified or unmodified message from the same method, so I think better to use this method.

src/routes/github/create-branch/github-create-branch-post.ts
krazziekay: Updated to `err.status || 500`. If its undefined, it will take the default 500.

src/routes/github/create-branch/github-create-branch-post.ts
krazziekay: True status will always be 403 in this case, so removed 500.

src/routes/github/create-branch/github-create-branch-post.ts
krazziekay: Renamed to `getGitHubConfigurationLink`.

src/routes/github/create-branch/github-create-branch-post.ts
krazziekay: Test case added

src/models/subscription.ts
krazziekay: Test case added

src/models/subscription.test.ts
```diff
@@ -204,6 +237,22 @@ describe("Subscription", () => {
 				})]);
 			});
 		});
+		describe("findForRepoNameAndOwner", () => {
+			it("should fetch the subscription for the given repo name and owner", async () => {
+				const repo0Sub = await Subscription.findForRepoNameAndOwner("repo-0", "atlassian", jiraHost);
+				const repo00Sub = await Subscription.findForRepoNameAndOwner("repo-0-0", "atlassian", jiraHost);
+				const repo1Sub = await Subscription.findForRepoNameAndOwner("repo-1", "atlassian", jiraHost);
+				const repo2Sub = await Subscription.findForRepoNameAndOwner("repo-2", "atlassian", jiraHost);
+				const repoRandomSub = await Subscription.findForRepoNameAndOwner("repo-unknown", "atlassian", jiraHost);
```
bgvozdev, line 246: I think this needs to be a separate test, e.g.

`returns undefined when repo cannot be find` or something like that

src/models/subscription.ts
bgvozdev: Btw why is it `sequelize?` Could it be undefined?

src/models/subscription.ts
```diff
@@ -153,6 +153,22 @@ export class Subscription extends Model {
...
+			"LEFT JOIN \"RepoSyncStates\" rss on s.\"id\" = rss.\"subscriptionId\" " +
+			"WHERE s.\"jiraHost\" = :jiraHost " +
+			"AND rss.\"repoName\" = :repoName " +
+			"AND rss.\"repoOwner\" = :repoOwner",
+			{
+				replacements: { jiraHost, repoName, repoOwner },
+				type: QueryTypes.SELECT
+			}
+		);
+		return results[0] as Subscription;
```
bgvozdev, line 169: if `this.sequelize` is undefined, this will explode in our faces

src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -29,12 +70,12 @@ export const GithubCreateBranchPost = async (req: Request, res: Response): Promi
 	}
 
 	if (!owner || !repo || !sourceBranchName || !newBranchName) {
-		res.status(400).json(errorMessages[400]);
+		res.status(400).json(getErrorMessages(400));
```
bgvozdev, line 73: Correct me if I'm wrong, but `getErrorMessages` returns a string, right? Why do we call `.json()` then?

src/models/subscription.test.ts
krazziekay: Separated in a new test case.

src/routes/github/create-branch/github-create-branch-post.ts
krazziekay: Updated to `send`

bgvozdev: Shouldn't we return json though?...

src/models/subscription.ts
krazziekay: I was using just `sequelize` initially, but then saw `this.sequelize` being used in `syncStatusCounts` method, so followed the same convention.
Will it be undefined though? Not sure of the proper way here.

src/routes/github/create-branch/github-create-branch-post.ts
krazziekay: I don't think that makes much of a difference though, its only returning the messages which are rendered directly in the frontend.

src/models/subscription.ts
bgvozdev: I think you could just use `this.sequelize!` as we pass it into ctor

src/routes/github/create-branch/github-create-branch-post.ts
bgvozdev: Well, in other places we do return JSONs, e.g. http://github.com/atlassian/github-for-jira/blob/fcad689efce720aa95309dba6e6b0a6b63057a8e/src/routes/api/jira/api-jira-verify-post.ts#L25-L25 . I think this is just the way we prefer to do error handling? I'd prefer to keep this uniformity.

bgvozdev: Or here http://github.com/atlassian/github-for-jira/blob/fcad689efce720aa95309dba6e6b0a6b63057a8e/src/routes/github/setup/github-setup-post.ts#L15-L15

src/models/subscription.ts
krazziekay: Updated to `this.sequelize!`

src/routes/github/create-branch/github-create-branch-post.ts
krazziekay: Updated to `json` and returned as `{ error: "" }`

src/models/subscription.ts
```diff
@@ -182,7 +198,7 @@ export class Subscription extends Model {
 	 * Returns array with sync status counts. [ { syncStatus: 'COMPLETED', count: 123 }, ...]
 	 */
 	static async syncStatusCounts(): Promise<SyncStatusCount[]> {
-		const results = await this.sequelize?.query(
+		const results = await this.sequelize!.query(
```
bgvozdev, line 201: 🏅

src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -52,12 +93,18 @@ export const GithubCreateBranchPost = async (req: Request, res: Response): Promi
 		statsd.increment(metricCreateBranch.created, tags);
 	} catch (err) {
 		req.log.error({ err }, "Error creating branch");
-		res.status(err.status).json(errorMessages[err?.status || 500]);
+
+		if (err.status === 403) {
+			const user = await gitHubUserClient.getUser();
+			const gitHubConfigurationLink = await getGitHubConfigurationLink(user.data.login, jiraHost, gitHubAppConfig.hostname, owner, repo);
+			res.status(err.status).json({ error: getErrorMessages(err?.status, gitHubConfigurationLink) });
```
bgvozdev, line 100: err.status is always defined, no need in `?`

src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -52,12 +93,18 @@ export const GithubCreateBranchPost = async (req: Request, res: Response): Promi
...
+		} else {
+			res.status(err.status || 500).json({ error: getErrorMessages(err?.status || 500) });
```
bgvozdev, line 102: no need in `?` here too, because it would be undefined and that's OK
Also, since we have code duplication, I'd suggest to have a temp variable
`const status = err.status || 500`

static/js/github-create-branch.js
```diff
@@ -217,7 +217,7 @@ const createBranchPost = () => {
     })
     .fail((error) => {
       toggleSubmitDisabled(false);
-      showErrorMessage(error.responseJSON);
```
bgvozdev, line 220: So, we did expected a JSON?...

src/routes/github/create-branch/github-create-branch-post.ts
krazziekay: Removed

src/routes/github/create-branch/github-create-branch-post.ts
krazziekay: Removed the `OR` part, cause status is always present.

static/js/github-create-branch.js
krazziekay: It works either way in the frontend, if its just a string we can pull it as `error.responseText`, but if its a JSON, then we pull it as `error.responseJSON` then get the desired property of that object.

bgvozdev (approved):

PR #1702, opened 2022-10-24
Arc 1787 clean up tslint warnings

gxueatlassian:
**What's in this PR?**
1. Turn on errors for some of the rules that's previously only warnings
2. Fix some low hangings fruits for warnings.

**Why**
Remove the noises as much as possible.

**Added feature flags**
N/A

**Affected issues**  
ARC-1787

**How has this been tested?**  
Unit test

**Whats Next?**
Further remove typings warnings.

prev

new

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -40,9 +40,9 @@
 		"@typescript-eslint/no-var-requires": "error",
 		"@typescript-eslint/no-unused-vars": "error",
 		"no-unused-vars": "off",
-		"no-console": "warn",
+		"no-console": "error",
 		"quotes": [
-			"off",
+			"error",
 			"double",
 			{
 				"avoidEscape": true,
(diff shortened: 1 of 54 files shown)


.eslintrc.json
```diff
@@ -59,8 +59,25 @@
...
 		"require-await": "off",
-		"@typescript-eslint/no-floating-promises": ["error"]
-	}
+		"@typescript-eslint/no-floating-promises": ["error"],
+		"jest/expect-expect": [
+			"error",
+			{
+				"assertFunctionNames": [
+					"expect",
+					"supertest.**.expect",
```
gxueatlassian, line 70: this is for making `supertest(app).get('/xxx').status(200)` passing linting.

.eslintrc.json
```diff
@@ -59,8 +59,25 @@
...
+					"*.expect"
```
gxueatlassian, line 71: This is for making `someInternaMethod(xxx).expect(200)` to pass. We'll trust the author as long as they wrote `expect` somewhere.

.eslintrc.json
```diff
@@ -59,8 +59,25 @@
...
+				]
+			}
+		]
+	},
+	"overrides": [{
+		"files": [ "src/**/*.test.ts" ],
+		"rules": {
+			"@typescript-eslint/no-explicit-any": "off",
+			"@typescript-eslint/no-non-null-assertion": "off"
```
gxueatlassian, line 80: This is for making `DatabaseStateCreator.gitHubServerApp!.gitHubAppId` pass eslint.
We do want the `!` after the `gitHubServerApp` for fail-fast. So I think it is better to allow it in tests files.

In runtime, allowing `!` may cause a bug, while in test time, allow `!` should fail-fast (which is better).

src/sync/branch.test.ts
```diff
@@ -119,7 +119,7 @@ describe("sync/branches", () => {
 				});
 		};
 
-		const nockBranchRequest = (response, variables?: Record<string, any>) =>
+		const nockBranchRequest = (response: object, variables?: Record<string, unknown>) =>
```
gxueatlassian, line 122: `unknown` is better than `any` as the ts prevents code from using any fields from that object.

mboudreau (approved):
Looks good :+1:

PR #1705, opened 2022-10-24
[NONE] [EASY KARMA]: extract raw body and add debug logs

bgvozdev:
**What's in this PR?**
We use json.stringify() before calculating the signature, and looks like we are rejecting at least some of the webhooks. I want to check if that's true and should we compare rawBody instead.

### DIFF
--- a/src/routes/github/webhook/webhook-receiver-post.ts
+++ b/src/routes/github/webhook/webhook-receiver-post.ts
@@ -22,7 +22,7 @@ const writeDebugLogsForRawBody = (verification: string, signatureSHA256: string,
 	logger.info(`bgvozdev debugging, signatureSHA256=${signatureSHA256}`);
 
 	try {
-		const verificationRawBody = createHash(JSON.stringify(response.locals.rawBody), webhookSecret);
+		const verificationRawBody = createHash(response.locals.rawBodyd, webhookSecret);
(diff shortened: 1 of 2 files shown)


src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -17,6 +17,23 @@ import { codeScanningAlertWebhookHandler } from "~/src/github/code-scanning-aler
 import { GITHUB_CLOUD_API_BASEURL, GITHUB_CLOUD_BASEURL } from "utils/get-github-client-config";
 import { getLogger } from "config/logger";
 
+const writeDebugLogsForRawBody = (verification: string, signatureSHA256: string, webhookSecret: string, logger, response: Response) => {
+	logger.info(`bgvozdev debugging, verification=${verification}`);
+	logger.info(`bgvozdev debugging, signatureSHA256=${signatureSHA256}`);
+
+	try {
+		const verificationRawBody = createHash(response.locals.rawBodyd, webhookSecret);
```
joshkay10, line 25: rawBodyd extra d?

bgvozdev: 🏅 what a catch!

Harminder84 (approved):

PR #1707, opened 2022-10-25
NONE: calculate signature using rawBody instead of parsed payload

bgvozdev:
**What's in this PR?**
Calculate signature on rawBody, not the parsed payload

**Why**
Because parsing affects the content (e.g. removes spaces) and that affects the signature

**Added feature flags**
Not needed: we had tested it before + we have a pollinator run-once check in staging for webhooks

**How has this been tested?**  
splunk

**Whats Next?**
splunk

### DIFF
--- a/src/routes/github/webhook/webhook-receiver-post.test.ts
+++ b/src/routes/github/webhook/webhook-receiver-post.test.ts
@@ -19,9 +19,9 @@ const NON_EXIST_GHES_UUID = "97da6b0e-ec61-11ec-8ea0-0242ac120003";
 const GHES_WEBHOOK_SECRET = "webhookSecret";
 const CLOUD_WEBHOOK_SECRET = envVars.WEBHOOK_SECRET;
 
-const injectRawBodyToReq = (res: any, req: any) => {
+const injectRawBodyToReq = (req: any) => {
 	req.rawBody = JSON.stringify(req.body);
-	return res;
+	return req;
 };
 
 describe("webhook-receiver-post", () => {
@@ -77,7 +77,7 @@ describe("webhook-receiver-post", () => {
 	it("should throw an error if github app not found", async () => {
 		req = createGHESReqForEvent("push", "", NON_EXIST_GHES_UUID);
 
-		await WebhookReceiverPost(req, injectRawBodyToReq(res, req));
+		await WebhookReceiverPost(injectRawBodyToReq(req), res);
 		expect(res.sendStatus).toBeCalledWith(400);
 
 	});
@@ -85,7 +85,7 @@ describe("webhook-receiver-post", () => {
 	it("should throw an error if signature doesn't match for GHE app", async () => {
 		req = createReqWithInvalidSignature("push", EXIST_GHES_UUID);
 
-		await WebhookReceiverPost(req, injectRawBodyToReq(res, req));
+		await WebhookReceiverPost(injectRawBodyToReq(req), res);
 		expect(res.status).toBeCalledWith(400);
 		expect(res.status().send).toBeCalledWith("signature does not match event payload and secret");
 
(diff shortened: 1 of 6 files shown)


gxueatlassian (approved):

src/routes/github/webhook/webhook-receiver-post.test.ts
```diff
@@ -19,6 +19,14 @@ const NON_EXIST_GHES_UUID = "97da6b0e-ec61-11ec-8ea0-0242ac120003";
 const GHES_WEBHOOK_SECRET = "webhookSecret";
 const CLOUD_WEBHOOK_SECRET = envVars.WEBHOOK_SECRET;
 
+const injectRawBodyToResLocal = (res: any, req: any) => {
```
mboudreau, line 22: Use types

src/routes/router.ts
```diff
@@ -28,14 +28,10 @@ RootRouter.use(urlencoded({ extended: false }));
 RootRouter.use(json({
 	limit: "30mb", //set limit according to github doc https://docs.github.com/en/developers/webhooks-and-events/webhooks/webhook-events-and-payloads#webhook-payload-object-common-properties
 	verify: (_: Request, res: Response, buf) => {
-		try {
-			res.locals.rawBody = buf.toString();
-		} catch (e) {
-			// eslint-disable-next-line no-console
-			console.error("bgvozdev debugging just to make sure this never fails. \"try-catch\" will be removed after the experiment");
-		}
+		res.locals.rawBody = buf.toString();
```
mboudreau, line 31: Wouldn't you want to use this on bodyparser, not json?

Also, it would make more sense to add it to the request and not res.locals.  We can update the Request type to have a `rawBody`

mboudreau: You can update the request type in common.ts

mboudreau (changes requested):
use `req.rawBody` instead of `res.locals`

src/routes/github/webhook/webhook-receiver-post.test.ts
```diff
@@ -70,39 +75,26 @@ describe("webhook-receiver-post", () => {
 	});
 
 	it("should throw an error if github app not found", async () => {
-		res = {
-			sendStatus: jest.fn()
-		};
 		req = createGHESReqForEvent("push", "", NON_EXIST_GHES_UUID);
 
-		await WebhookReceiverPost(req, res);
+		await WebhookReceiverPost(req, injectRawBodyToReq(res, req));
```
mboudreau, line 80: Oh, missed this.  You need to move the `injectRawBodyToReq` to wrap the req only, and leave res alone :)

src/routes/github/webhook/webhook-receiver-post.test.ts
```diff
@@ -19,6 +19,11 @@ const NON_EXIST_GHES_UUID = "97da6b0e-ec61-11ec-8ea0-0242ac120003";
 const GHES_WEBHOOK_SECRET = "webhookSecret";
 const CLOUD_WEBHOOK_SECRET = envVars.WEBHOOK_SECRET;
 
+const injectRawBodyToReq = (res: any, req: any) => {
+	req.rawBody = JSON.stringify(req.body);
+	return res;
```
mboudreau, line 24: need to return req as per below

src/routes/github/webhook/webhook-receiver-post.test.ts
```diff
@@ -19,6 +19,11 @@ const NON_EXIST_GHES_UUID = "97da6b0e-ec61-11ec-8ea0-0242ac120003";
 const GHES_WEBHOOK_SECRET = "webhookSecret";
 const CLOUD_WEBHOOK_SECRET = envVars.WEBHOOK_SECRET;
 
+const injectRawBodyToReq = (res: any, req: any) => {
```
mboudreau, line 22: adding types will remove any chance of fuckups

src/routes/github/webhook/webhook-receiver-post.test.ts
gxueatlassian: Or maybe just remove that inject function. Modify the function the create the request object. Why bother a new function while the req was created for this testing purpose?

mboudreau (dismissed):

mboudreau (approved):

PR #1710, opened 2022-10-26
Log backfill errors

joshkay10:
**What's in this PR?**
Extra logs to see why GitHub is failing the requests and check whether it keeps trying to push repository as the next task after failure


src/github/client/github-client-interceptors.ts
```diff
@@ -138,6 +138,7 @@ export const handleFailedRequest = (logger: Logger) =>
 				logger.error(errorMessage);
 			}
 
+			logger.info({ response, request, error, isWarning }, "joshkay temp logging - githubclient interceptor handleFailedRequest");
```
joshkay10, line 141: Want the actual error on Axios fail

src/sync/installation.ts
```diff
@@ -331,6 +331,7 @@ export const handleBackfillError = async (err,
 	logger: Logger,
 	scheduleNextTask: (delayMs: number) => void): Promise<void> => {
 
+	logger.info({ err, data, nextTask }, "joshkay temp logging - handleBackfillError");
```
joshkay10, line 334: My current guess is that after it tries X times and finally calls the task a fail, its picking up "repository" as the next task again every time and can't get past it. This is here to try prove or disprove that nextTask is stuck on "repository"

src/github/client/github-client-interceptors.ts
gxueatlassian: Is there a reason we use `info` instead of `error` for log level?

joshkay10: purely just was re-using the same method for all of my logs. This one is dumping out everthing it gets its hands on so not just specifically errors. Plus the hidden lines above are throwing .errrors

gxueatlassian (approved):

PR #1711, opened 2022-10-26
ARC-1797 Replace Localstack with ElasticMQ

Harminder84:
**What's in this PR?**
As some of our GHE customers don't want to create a hole in the firewall, but as a workaround, they're happy to run the GitHub app in their own infrastructure.  Those customers should be able to run the app using docker-compose command with minimum number of prerequisite steps. One of the bottlenecks is that we're dependent upon the localstack when we're running the app using docker-compose. The localstack is purely used for development purposes and we're only using localstack's SQS service. Our aim in this PR is to use ElasticMQ instead of relying upon localstack's SQS. 

**Why**
Enterprise customers should able run the app in their own infrastructure

**Affected issues**  
[ARC-1797]

**How has this been tested?**  
Local
Staging

[ARC-1797]: https://softwareteams.atlassian.net/browse/ARC-1797


sqs-purge.ts
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable no-console */
+import axios from "axios";
+import fs from "fs";
+import { parse } from "yaml";
+
+
+const purgeQueues = async () => {
```
gxueatlassian, line 7: Is `purgeQueues` only for local dev and testing? If so, shall we add the `isNodeEnvProd()` here for safety checking?

bgvozdev: It is reading data from docker-compose and explicitly reading URLs from it (which target to some magical hostname); I don't think it would do any harm in prod, even if someone manages to invoke it.

But shall we rename it maybe to be explicit a little bit? `sqs-purge-elasicmq-queues` or something like that?

sqs-purge.ts
```diff
@@ -0,0 +1,39 @@
...
+	const file = fs.readFileSync("./docker-compose.yml", "utf8");
+
+	const envVars = parse(file).services?.app?.environment;
```
bgvozdev, line 10: @gxueatlassian 
image

mboudreau:
@Harminder84 I think we can close this one?

gxueatlassian:
Close it for now as inactive for long time. Can re-open if need to.

PR #1713, opened 2022-10-27
Bugfix stop infinite resync

joshkay10:
**What's in this PR?**
If we fail a repository resync completely (after 4 getrepopage failures) we must stop retrying the same repo.
This stops us trying to sync invalid repos forever.


mboudreau (dismissed):
I still don't think this is the issue.  The task repository shouldn't fail unless the app was uninstalled or something since it's just looping through the repository list given to it.  This logic hasn't changed since implemented 6 months ago.  I think it's a side effect of something else.

Nonetheless, if it means not blocking, I'm happy to approve, but I don't think it's the root cause.

bgvozdev (approved):

src/sync/installation.ts
```diff
@@ -376,16 +377,19 @@ export const handleBackfillError = async (err,
...
 };
 
-export const markCurrentRepositoryAsFailedAndContinue = async (subscription: Subscription, nextTask: Task, scheduleNextTask: (delayMs: number) => void): Promise<void> => {
+export const markCurrentRepositoryAsFailedAndContinue = async (subscription: Subscription, nextTask: Task, task: string,  scheduleNextTask: (delayMs: number) => void): Promise<void> => {
 	// marking the current task as failed
 	await updateRepo(subscription, nextTask.repositoryId, { [getStatusKey(nextTask.task)]: "failed" });
 	const gitHubProduct = getCloudOrServerFromGitHubAppId(subscription.gitHubAppId);
 	statsd.increment(metricTaskStatus.failed, [`type:${nextTask.task}`, `gitHubProduct:${gitHubProduct}`]);
 
+	if (task === "repository" && nextTask.task === "repository") {
```
gxueatlassian, line 389: On around line 212 it has `const { task, cursor, repository } = nextTask;`, so isn't the `task` always === `nextTask.task`?

gxueatlassian (changes requested):
Just one quick comment to address.

gxueatlassian:
In side `getNextTask` it has this line `if (subscription.repositoryStatus !== "complete") {`, so the change in this PR ( of setting `syncStatus` to `FAILED` ) won't prevent the app from re-pickup the repository task again?

src/sync/installation.ts
```diff
@@ -387,6 +386,10 @@ export const markCurrentRepositoryAsFailedAndContinue = async (subscription: Sub
 	const gitHubProduct = getCloudOrServerFromGitHubAppId(subscription.gitHubAppId);
 	statsd.increment(metricTaskStatus.failed, [`type:${nextTask.task}`, `gitHubProduct:${gitHubProduct}`]);
 
+	if (nextTask.task === "repository") {
+		await subscription.update({ syncStatus: SyncStatus.FAILED });
+		return;
```
joshkay10, line 391: @gxueatlassian - correct, the update to failed alone wont do anything, mostly tidy up.
The return on the next line is what skips the task being fired again.

joshkay10:
> I still don't think this is the issue. The task repository shouldn't fail unless the app was uninstalled or something since it's just looping through the repository list given to it. This logic hasn't changed since implemented 6 months ago. I think it's a side effect of something else.
> 
> Nonetheless, if it means not blocking, I'm happy to approve, but I don't think it's the root cause.

Agree, its not root cause, but it clean out the backfill. I do wonder if the installations i kicked off are uninstalled? Do we remove the entry on uninstall?

gxueatlassian (approved):

PR #1714, opened 2022-10-27
fixing problem with issue key parser

mboudreau:
**What's in this PR?**
Fix for  jira issue parser expecting a minimum of 2 numbers in the id and accompanied tests

**Why**
it borked the pollinator tests when on

**Added feature flags**
already had one

**Affected issues**  
None

**How has this been tested?**  
Added tests

**Whats Next?**
Deploy and turn on FF

### DIFF
--- a/src/util/jira-utils.ts
+++ b/src/util/jira-utils.ts
@@ -105,7 +105,7 @@ export const jiraIssueKeyParser = (str: string): string[] => {
 	let regex = /(^|[^\p{L}\p{Nd}])([\p{L}][\p{L}\p{Nd}_]{1,255}-\p{Nd}{1,255})/giu;
 
 	if (issueKeyRegexCharLimitFeature) {
-		regex = /(^|[^A-Z\d])([A-Z][A-Z\d]{1,255}-[1-9]\d{1,255})/giu;
+		regex = /(^|[^A-Z\d])([A-Z][A-Z\d]{1,255}-[1-9]\d{0,255})/giu;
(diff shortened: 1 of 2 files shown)


gxueatlassian (approved):

src/util/jira-utils.ts
```diff
@@ -105,7 +105,7 @@ export const jiraIssueKeyParser = (str: string): string[] => {
 	let regex = /(^|[^\p{L}\p{Nd}])([\p{L}][\p{L}\p{Nd}_]{1,255}-\p{Nd}{1,255})/giu;
 
 	if (issueKeyRegexCharLimitFeature) {
-		regex = /(^|[^A-Z\d])([A-Z][A-Z\d]{1,255}-[1-9]\d{1,255})/giu;
+		regex = /(^|[^A-Z\d])([A-Z][A-Z\d]{1,255}-[1-9]\d{0,255})/giu;
```
gxueatlassian, line 108: some tests would be good, as I don't understand why it was broken and how that `0` fix it.

mboudreau: yeah, I had added tests, but because of the FF and since this is a non-async function, there's no way of consistently added the FF to the test, so I went without it.

The 0 here depics how many characters are needed for the id.  The `[1-9]` is any number except 0, then `\d` is any number as long as it has the length of 0 to 255, where before it was excpecting at least another character so the id has to be 2 numbers for it to be caught by parser.

PR #1715, opened 2022-10-28
[QUICKFIX] - Error link fixed for non-connected repos

krazziekay:
**What's in this PR?**
- Condition added for `findForRepoNameAndOwner` for case where the repo names doesn't exist in our DB.
  - This is for the case when the repo is not synced, meaning does not exist in RepoSyncState, but the repo owner would probably exist. So simply returning the subscription for the repo owner.

**Why**
- To avoid the undefined error when repo names doesn't exist

**How has this been tested?**  
- Test cases

**Whats Next?**
- Merging

### DIFF
--- a/src/models/subscription.ts
+++ b/src/models/subscription.ts
@@ -167,15 +167,18 @@ export class Subscription extends Model {
 				repoName
 			}
 		});
+		let replacements = { jiraHost, repoOwner } as any;
+
 		// Check if the repo exists in our DB
 		if (repo.length) {
 			query += "AND rss.\"repoName\" = :repoName";
+			replacements = { ...replacements, repoName };
 		}
 
 		const results = await this.sequelize!.query(
 			query,
 			{
-				replacements: { jiraHost, repoOwner, repoName },
+				replacements,
 				type: QueryTypes.SELECT
 			}
 		);
--- a/src/models/subscription.ts
+++ b/src/models/subscription.ts
@@ -1,6 +1,7 @@
 import { DataTypes, DATE, Model, Op, QueryTypes, WhereOptions } from "sequelize";
 import { uniq } from "lodash";
 import { sequelize } from "models/sequelize";
+import { RepoSyncState } from "models/reposyncstate";
 
 export enum SyncStatus {
 	PENDING = "PENDING",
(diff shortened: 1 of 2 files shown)


gxueatlassian (commented):
It seems getting strange from code flowing point of view. 
1. We have all the details (`sub`, `gitHubApp`), pull down the `servers` in `create-branch-options`, render the hbs, and allow user to pick.
2. And when user submit the form, we go all the way back from the string base `owner/repoName` to find that `gitHubInstallationId` from `Subscription` again? 

Don't we already know which `Subscription` and `GitHubServerApp` when user pick a repo and submit a new branch?

krazziekay:
> It seems getting strange from code flowing point of view.
> 
> 1. We have all the details (`sub`, `gitHubApp`), pull down the `servers` in `create-branch-options`, render the hbs, and allow user to pick.
> 2. And when user submit the form, we go all the way back from the string base `owner/repoName` to find that `gitHubInstallationId` from `Subscription` again?
> 
> Don't we already know which `Subscription` and `GitHubServerApp` when user pick a repo and submit a new branch?


We have `gitHubAppConfig`, but don't have **subscription**. We only know the **subscription** once the user has selected a **repository** in the create branch form. And even in that case, we only know the **subscription** if the **repository** is connected/synced to our app. 
So, if the user types in a non-synced **repository**, then we don't have any data for that in `RepoSyncState` nor `Subscription`, so it was returning `undefined.` This change in query is to fix that, so that we're sending the **subscription** based on just repo owner for non-connected/non-synced repos.

src/models/subscription.ts
```diff
@@ -154,15 +155,30 @@ export class Subscription extends Model {
...
-			"SELECT * " +
+		let query = "SELECT * " +
 			"FROM \"Subscriptions\" s " +
 			"LEFT JOIN \"RepoSyncStates\" rss on s.\"id\" = rss.\"subscriptionId\" " +
 			"WHERE s.\"jiraHost\" = :jiraHost " +
-			"AND rss.\"repoName\" = :repoName " +
-			"AND rss.\"repoOwner\" = :repoOwner",
+			"AND rss.\"repoOwner\" = :repoOwner ";
+
+		const repo = await RepoSyncState.findAll({
```
gxueatlassian, line 164: `RepoSyncState.findAll` will load all repos into js memory. Do we need that?

gxueatlassian (changes requested):
Feels like an old PR? Do we still need this? Happy to reopen if needed.

PR #1716, opened 2022-10-28
Removing regex feature flags and just use the one, improve tests

mboudreau:
**What's in this PR?**
Removing all FF for regex as we now have a stable one and we're not going to use the "new" one anytime soon and should be moved to an API anyways.

**Why**
Too many FF around regex

**Added feature flags**
I removed them :smiling_imp: 

**Affected issues**  
None

**How has this been tested?**  
Staging and added tests

**Whats Next?**
Move all of this to it's own API as we shouldn't have to do this in _each integration with jira_!


gxueatlassian (approved):

gxueatlassian:
Close it for it has too many conflicts. Will solve it later.

gxueatlassian:
Back log items: https://softwareteams.atlassian.net/browse/ARC-2217

PR #1725, opened 2022-11-02
Arc 1802 fix cross site ghe deletion

gxueatlassian:
**What's in this PR?**
🔐 🐞 Add a new where clause when deleting GHE apps.

**Why**
Currently it seems when user disconnect a GHE server from a jiraHost, all other jiraHosts that also has connection to the same server will be disconnected as well.

**Added feature flags**
N/A

**Affected issues**  
ARC-1802

**How has this been tested?**  
Unit test.

**Whats Next?**
Use can also delete ANY GHE apps as long as they know the uuid (which is public on the url anyway), need to fix that as well.

### DIFF
--- a/src/routes/jira/connect/enterprise/jira-connect-enterprise-delete.ts
+++ b/src/routes/jira/connect/enterprise/jira-connect-enterprise-delete.ts
@@ -1,6 +1,5 @@
 import { NextFunction, Request, Response } from "express";
 import { GitHubServerApp } from "models/github-server-app";
-import { Installation } from "models/installation";
 
 export const JiraConnectEnterpriseDelete = async (
 	req: Request,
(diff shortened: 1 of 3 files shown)


src/routes/jira/connect/enterprise/jira-connect-enterprise-delete.ts
```diff
@@ -1,15 +1,23 @@
...
 	res: Response,
 	next: NextFunction
 ): Promise<void> => {
 	try {
+
 		req.log.debug("Received Jira Connect Enterprise Server DELETE request");
 
-		await GitHubServerApp.uninstallServer(req.body.serverUrl);
+		const { jiraHost }  = res.locals;
+		const installation = await Installation.getForHost(jiraHost);
```
bgvozdev, line 15: Don't we have it in res.locals.installation already?

mboudreau: I believe we do :)

gxueatlassian: > Don't we have it in res.locals.installation already?

Maybe, it has the following in jwt middleware
```
	// TODO: Probably not the best place to set things globally
```

Should I use it?

bgvozdev: I believe yes :) cause it was driven by JWT, which means we could trust that? At least this is my thinking

gxueatlassian: Done. :)

bgvozdev (approved):

mboudreau (approved):

gxueatlassian:
@mboudreau look at that!! How good. 

image

PR #1727, opened 2022-11-03
Arc 1804 fix ghe manual creation token bug

gxueatlassian:
**What's in this PR?**
Fix prod bug that user can't create ghe app manually.

**Why**
why not?

**Added feature flags**
N/A

**Affected issues**  
ARC-1804

**How has this been tested?**  
Local

**Whats Next?**
N/A

### DIFF
--- a/static/js/jira-manual-app-creation.js
+++ b/static/js/jira-manual-app-creation.js
@@ -56,7 +56,7 @@ const gitHubAppPutRequest = (uuid, data) => {
 	});
 };
 
-const gitHubAppPostRequest = (data) => {
+const gitHubAppPostRequest = (data, token) => {
 	$.ajax({
 		type: "POST",
 		url: `/jira/connect/enterprise/app`,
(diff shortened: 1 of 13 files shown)


static/js/jira-manual-app-creation.js
```diff
@@ -104,7 +104,7 @@ $(document).ready(function() {
 			if (isUpdatePage()) {
 				gitHubAppPutRequest(uuid, data);
 			} else {
-				gitHubAppPostRequest(data);
+				gitHubAppPostRequest(data, token);
```
gxueatlassian, line 107: I understand the `data.jwt` is equal to `token`, but prefer this as more explicit as for the reason bellow:

IIRC, $.ajax({data}) will send the data as body, which means jwt is also included in body. Usually, token like stuff is better to be in the Auth header. 

So we might need another PR to cleanup the jwt from the body.

bgvozdev (approved):

PR #1730, opened 2022-11-07
ARC-1810 remove decrypted ghe app secrets from res.locals

gxueatlassian:
**What's in this PR?**
Remove the decrypted secrets in `res.locals.gitHubAppConfig` to avoid risk that it might be deserialized and sent to somewhere like in the splunk or worse frontend.

Replace those secrets with a closure method.

**Why**
It is generally dangerous have something secrets in the json object hanging around for long. 

**Added feature flags**
N/A

**Affected issues**  
ARC-1810

**How has this been tested?**  
Unit test
Tested on Stage.

**Whats Next?**
1. Need more proper typings on the `res.locals`. Already have some ideas. Will raise a PR soon.
2. Also remove the decrypted secrets in `gitHubClient` as well.

### DIFF
--- a/src/middleware/github-server-app-middleware.ts
+++ b/src/middleware/github-server-app-middleware.ts
@@ -2,7 +2,6 @@ import { NextFunction, Request, Response } from "express";
 import { GitHubServerApp } from "models/github-server-app";
 import { Installation } from "models/installation";
 import { envVars } from "config/env";
-import { keyLocator } from "../github/client/key-locator";
 import { GITHUB_CLOUD_BASEURL } from "utils/get-github-client-config";
 
 export const GithubServerAppMiddleware = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
@@ -43,10 +42,7 @@ export const GithubServerAppMiddleware = async (req: Request, res: Response, nex
 			appId: gitHubServerApp.appId,
 			uuid: gitHubServerApp.uuid,
 			hostname: gitHubServerApp.gitHubBaseUrl,
-			clientId: gitHubServerApp.gitHubClientId,
-			getDecryptedGitHubClientSecret: async () => gitHubServerApp.getDecryptedGitHubClientSecret(),
-			getDecryptedWebhookSecret: async () => gitHubServerApp.getDecryptedWebhookSecret(),
-			getDecryptedPrivateKey: async () =>  gitHubServerApp.getDecryptedPrivateKey()
+			clientId: gitHubServerApp.gitHubClientId
 		};
 	} else {
 		req.log.info("Defining GitHub app config for GitHub Cloud.");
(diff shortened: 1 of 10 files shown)


src/middleware/github-server-app-middleware.ts
```diff
@@ -44,9 +44,9 @@ export const GithubServerAppMiddleware = async (req: Request, res: Response, nex
 			uuid: gitHubServerApp.uuid,
 			hostname: gitHubServerApp.gitHubBaseUrl,
 			clientId: gitHubServerApp.gitHubClientId,
-			gitHubClientSecret: await gitHubServerApp.getDecryptedGitHubClientSecret(),
-			webhookSecret: await gitHubServerApp.getDecryptedWebhookSecret(),
-			privateKey: await gitHubServerApp.getDecryptedPrivateKey()
+			getDecryptedGitHubClientSecret: async () => gitHubServerApp.getDecryptedGitHubClientSecret(),
```
bgvozdev, line 47: Do we need `async` keyword here? `.getDecryptedBlahBlah()` returns a promise anyway, right?

src/middleware/github-server-app-middleware.ts
```diff
@@ -56,9 +56,9 @@ export const GithubServerAppMiddleware = async (req: Request, res: Response, nex
 			uuid: undefined, //undefined for cloud
 			hostname: GITHUB_CLOUD_BASEURL,
 			clientId: envVars.GITHUB_CLIENT_ID,
-			gitHubClientSecret: envVars.GITHUB_CLIENT_SECRET,
-			webhookSecret: envVars.WEBHOOK_SECRET,
-			privateKey: await keyLocator(undefined)
+			getDecryptedGitHubClientSecret: async () => envVars.GITHUB_CLIENT_SECRET,
```
bgvozdev, line 59: Shouldn't it be `Promise.resolve(...)`? Otherwise, in one case we are returning a promise, while in the other we return a plain value... or not?

gxueatlassian: js engine will wrap it in Promise anyway (due to the `async` keyword)

src/middleware/github-server-app-middleware.ts
gxueatlassian: So ignoring the TS of method returning type, from JS engine point of view, I'd prefer explicit marking the function as async over implicit rely on the function impl return a Promise. Also marking it as `async` will allow the consumer to do `await`, instead of forcing them to do `xxx().then()`

----

But I could be wrong, just some theory from my js experience. Didn't verify what I said though lol :)

bgvozdev (approved):

src/middleware/github-server-app-middleware.ts
bgvozdev: Pretty sure they can await a sync function that returns a promise... But yeah, doesn't matter since we are removing this code altogether :nothingtodohere:

PR #1731, opened 2022-11-09
ARC-1817 improve failed SQS logging and add DLQ alarms

joshkay10:
**What's in this PR?**
Updated the SQS message attribute to be messageId so its more obvious when Splunking

Sent DLQ errors to the unsafe logging environment so we can use legit data

create DLQ alarms - low prio and the numbers are best guess at this point.

**Why**
We had trouble last time we had an outage with error handling on SQS and was lacking details of the messages

**How has this been tested?**  
Testing Alarms in stage(forced with a lower threshold)
Tested that unsafe logs containing UGC/PII only show up in [env]-unsafe Splunk partition.

**Whats Next?**
When the world collapses and we need to know what some of the failed messages are we wont have a shake our fists.

### DIFF
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -655,7 +655,7 @@ environmentOverrides:
             MetricName: ApproximateNumberOfMessagesVisible
             Namespace: AWS/SQS
             Description: "Message count for backfill dead letter queue too high!"
-            Threshold: 5000 # number is initially high until we start bringing numbers down
+            Threshold: 50 # number is initially high until we start bringing numbers down
             Priority: Low
             Dimensions:
               - Name: QueueName
@@ -672,7 +672,7 @@ environmentOverrides:
             MetricName: ApproximateNumberOfMessagesVisible
             Namespace: AWS/SQS
             Description: "Message count for push dead letter queue too high!"
-            Threshold: 5000 # number is initially high until we start bringing numbers down
+            Threshold: 500 # number is initially high until we start bringing numbers down
             Priority: Low
             Dimensions:
               - Name: QueueName
@@ -689,7 +689,7 @@ environmentOverrides:
             MetricName: ApproximateNumberOfMessagesVisible
             Namespace: AWS/SQS
             Description: "Message count for deployment dead letter queue too high!"
-            Threshold: 5000 # number is initially high until we start bringing numbers down
+            Threshold: 4000 # number is initially high until we start bringing numbers down
             Priority: Low
             Dimensions:
               - Name: QueueName
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -655,13 +655,13 @@ environmentOverrides:
             MetricName: ApproximateNumberOfMessagesVisible
             Namespace: AWS/SQS
             Description: "Message count for backfill dead letter queue too high!"
-            Threshold: 40 #~ Autoscaling Threshold x 2. So we'll be alerted if autoscaling failed to fix the piling messages
+            Threshold: 5000 # number is initially high until we start bringing numbers down
             Priority: Low
             Dimensions:
               - Name: QueueName
                 Value: { "Ref": "QueueName" }
-            EvaluationPeriods: 10
-            Period: 600
+            EvaluationPeriods: 5
+            Period: 300
             ComparisonOperator: GreaterThanThreshold
             Statistic: Maximum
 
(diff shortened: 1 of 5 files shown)


joshkay10:
not ready

github-for-jira.sd.yml
```diff
@@ -647,6 +647,75 @@ environmentOverrides:
...
             Statistic: Maximum
+
+      - type: sqs
+        name: backfill_dlq
+        alarms:
+          AlarmOnTooManyMessages:
+            MetricName: ApproximateNumberOfMessagesVisible
+            Namespace: AWS/SQS
+            Description: "Message count for backfill dead letter queue too high!"
+            Threshold: 5000 # number is initially high until we start bringing numbers down
```
joshkay10, line 658: This is intentionally outrageously large at 5000.
We have massive amounts of messages in our DLQ currently and it averages 2000-3000 at any given time(in deployment dlq mainly). Something probably worth investigating, but so we don't get an alert every 20 mins 5000 was chosen as this seems to be the point where a spike will break out.

Open to options here tho!!

Also, once this is merged we will be monitoring the messaged that get tossed into the dlw bin and investigate how to bring them down.. long story short, this is not the final number we probably need to fix things before this alert becomes signal!

src/sqs/sqs.ts
```diff
@@ -252,7 +252,7 @@ export class SqsQueue<MessagePayload> {
 			message,
 			payload,
 			log: listenerContext.log.child({
-				id: message.MessageId,
+				messageId: message.MessageId,
```
joshkay10, line 255: renamed to messageID for clarity!

src/sqs/sqs.ts
```diff
@@ -284,7 +284,9 @@ export class SqsQueue<MessagePayload> {
 	}
 
 	private async handleSqsMessageExecutionError(err, context: SQSMessageContext<MessagePayload>) {
+		const unsafeLogger = getLogger("message-error-handler-unsafe", { level: "warn", unsafe: true });
```
joshkay10, line 287: Tagging a logger with `unsafe: true` this will push logs to [env]-unsafe where we are allowed to write out UGC and PII.... within reason

src/util/logger-utils.test.ts
```diff
@@ -130,6 +130,17 @@ describe("Logger Utils", () => {
...
+					msg: "More messaging",
+					unsafe: true,
+					orgName: "ORG",
+					level: INFO
+				};
+				await stream._write(testMessage, encoding, next);
+				expect(process.stdout.write).toHaveBeenCalled();
+			});
+
 			it("should not serialize sensitive data", async () => {
```
joshkay10, line 144: Missing test for unsafe NOT being written to standard logger

github-for-jira.sd.yml
rachellerathbone: I think investigating this more sounds like a solid idea. Can you create a ticket for disturbed/ops? (I hear @mboudreau is sorting the backlog and know he will love extra tickets to sort :badpokerface:)

src/sqs/sqs.ts
rachellerathbone: Nice. Small change but makes those logs a lot easier to read.

src/sqs/sqs.ts
rachellerathbone: Within reason? Do we have any idea what that is? Might be worth throwing a HOWTO runbook together to get the team on the same page and so we have somewhere to point future Arc folks to.

github-for-jira.sd.yml
```diff
@@ -647,6 +647,75 @@ environmentOverrides:
             Period: 300
             ComparisonOperator: GreaterThanThreshold
             Statistic: Maximum
+
+      - type: sqs
+        name: backfill_dlq
```
gxueatlassian, line 652: Hi @joshkay10 sorry I am bit lost. Can you tell me is it going to create new sqs queue on prod (since the whole block a new block) ? And how does it connect to the backfill as dlq?

github-for-jira.sd.yml
rachellerathbone: Think we need it to be 5000 for all queues? Just took a look at our DLQ metrics for the past year and it looks like the highest discovery ever reached was ~35 and backfill ~65. Push did get up above 6k but that was back in Dec last year. Makes sense for branch and deployments though (especially the later coz the metrics for that bad boy look like a rollercoaster). Not sure we need a one-size-fits all approach here.

src/sqs/sqs.ts
joshkay10: Fair poinst, there is a few docs floating around, most of them we collected in this page https://hello.atlassian.net/wiki/spaces/OTFS/pages/1649328545/Logging+UGC+PII

but I might reference it or make a mini runbook, so its a bit more digestable.

Basically there is still restricted data like credit cards and such that are big no noes

github-for-jira.sd.yml
joshkay10: Micros automatically provision the dead letter queues since we set MaxReceiveCount
This is just attaching an alarm to those queues, it is confusing since you can't see the dlq creation in here and the name of them i had to go get from AWS directly :(

src/sqs/sqs.ts
rachellerathbone: Makes sense 😄

rachellerathbone (approved):

Issue #1536, opened 2022-08-22
Add more unique deployment names


fullstackzach:
Hi - I was wondering if there was some way to customize the "Deployment" or "Pipeline" fields to show more unique information? Or, if an improvement could be made?

Deployments for our team can mean deploying multiple components separately. A frontend, backend, poller/worker service, etc.

In this screenshot you can see that they currently just show "deploy". But, this isn't too useful since I don't know which is which. Once you click the underlying link it will actually take you to the GitHub action, where you can actually determine which belongs to which. It would be nice to map one of these fields to the name of the deploy instead.

<img width="1013" alt="image" src="https://user-images.githubusercontent.com/13011363/185840415-b4b8e10a-9e9b-4f0d-9a53-e5ec7667b651.png">

Digging into the code here; I suspect this is where they are mapped. `deployment.task` I suspect resolves to the value `deploy`.   But, I'm not too sure what to change in my github action, since it doesn't seem we specify that anywhere. https://github.com/atlassian/github-for-jira/blob/main/src/transforms/transform-deployment.ts#L258-L267

rachellerathbone:
Hey @fullstackzach. Would this feature help? https://github.com/atlassian/github-for-jira/blob/main/docs/deployments.md#environment-mapping

fullstackzach:
Hi @rachellerathbone - I did see that documentation when I was trying to solve this problem, but I don't think it solves my request.

Our deployments are being properly captured as "production" - as seen above, so environment mapping isn't the issue. 

So for example the first item in the table above the GitHub action might be named "Deploy Frontend", and the second item is "Deploy Backend", but only "deploy" appears in Jira - it's not until I click the link that it takes me to Github and I see the name of the component in the Github UI

thombergs:
Hey @fullstackzach,

can you give us a sample of your GitHub Actions workflow file so we can have a look?

fullstackzach:
Sorry it took me a bit to get back to you here. I am a little apprehensive about posting our build config in a public forum, while giving you enough to work with. I inserted <redacted> where there is some items I do not want to share.

```
jobs:
  bump_version:
    name: Bump Version
    if: ${{ github.event_name == 'push' }}
    runs-on: [self-hosted, linux, aws]
    outputs:
      git_tag: ${{ steps.tag_version.outputs.new_tag }}
    steps:
    - name: Bump version and push tag
      id: tag_version
      uses: mathieudutour/github-tag-action@v6.0
[...]

PR #1732, opened 2022-11-09
SALTBAE-454 Update displayName for the deployments

kaganatlassian2:
**What's in this PR?**
At deployment's displayName was always deploy, with this PR it's going to be the latest commit message. 

**Why**
In order to give Jira users more visibility. 

**Added feature flags**

**Affected issues**  
[_Jira Issues_](https://jdog.jira-dev.com/browse/SALTBAE-454)

**How has this been tested?**  
Tested locally and added unit tests for it. 

**Whats Next?**


src/transforms/transform-deployment.ts
```diff
@@ -290,7 +290,7 @@ export const transformDeployment = async (githubInstallationClient: GitHubInstal
 			schemaVersion: "1.0",
 			deploymentSequenceNumber: deployment.id,
 			updateSequenceNumber: deployment_status.id,
-			displayName: deployment.task,
+			displayName: message,
```
bgvozdev, line 293: Shouldn't we truncate the message to make sure we are not hitting API's limits?

benoittgt: For the `pipeline.url`. What about the `log_url` ? 

It might be a breaking change, but we already use target_url on 'url' field when building this payload.

bgvozdev (dismissed):

bgvozdev (approved):

benoittgt:
It seems weird that this PR was merged like that. The initial assumption

> At deployment's displayName was always deploy

Is not correct. You can see [here](https://github.com/atlassian/github-for-jira/issues/1536#issuecomment-1310434669) I was able to change it without much effort. I don't think we should have base on last commit because we will see for most of the text visible a duplication of the Jira ticket if people follow the rule of: a git commit should start with the ID of the Jira issue. Also link to the commit title will point to the `deployment_url` not to the commit itself.

Screen_Shot_2022-11-17_at_10_04_02

I think we should have had a look on[ the other proposals](https://github.com/atlassian/github-for-jira/issues/1536#issuecomment-1310639857) I made, and let people choose if they want the last commit title used or something else (by submitting the commit in `description` on deploy status Github API).

PR #1734, opened 2022-11-09
ARC-1767 change instance button

joshkay10:
**What's in this PR?**
Added the link to go back to change github instance

**Why**
Incase people change their minds! or have existing auth they need to override!


joshkay10:
This was included in a different PR and is merged on main!

PR #1735, opened 2022-11-10
Arc 1823 better ghe typings

gxueatlassian:
**What's in this PR?**
1. Add Response typings, especially for those previous untyped `res.locals`
2. Merge the structure of `gitHubAppConfig` between sqs and webhook, and promote the types to common typing.
3. Fix 1 or 2 potential bugs.

**Why**
We have too many free flow untyped `res.locals` every where. It can cause to bug introduced unnoticed. 

**Added feature flags**
N/A

**Affected issues**  
ARC-1823

**How has this been tested?**  
Unit test.
Stage: [wip]

**Whats Next?**
N/A

### DIFF
--- a/src/common-types/middleware-locals.d.ts
+++ b/src/common-types/middleware-locals.d.ts
@@ -1,0 +1,20 @@
+type CommonResponseLocals = {
+	nonce: string
+}
+
+type JiraHostVerifiedLocals = CommonResponseLocals  & {
+	jiraHost: string,
+};
+
+type JiraJwtVerifiedLocals = CommonResponseLocals & {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	installation: any //TODO: This is a hack, fix it with full definition.
+};
+
+type GitHubAppVerifiedLocals = CommonResponseLocals & {
+	gitHubAppConfig: GitHubAppConfig,
+};
+
+type GitHubUserTokenVerifiedLocals = CommonResponseLocals & {
+	githubToken: string;
+};
--- a/src/middleware/github-server-app-middleware.test.ts
+++ b/src/middleware/github-server-app-middleware.test.ts
@@ -126,7 +126,7 @@ describe("github-server-app-middleware", () => {
 			uuid: UUID,
 			clientId: "lvl.1234",
 			gitHubBaseUrl: "http://myinternalserver.com",
-			gitHubApiUrl: "http://myinternalserver.com"
+			gitHubApiUrl: "http://myinternalserver.com/api/v3"
 		});
 	});
 });
--- a/src/middleware/github-server-app-middleware.test.ts
+++ b/src/middleware/github-server-app-middleware.test.ts
@@ -120,13 +120,13 @@ describe("github-server-app-middleware", () => {
 
 		expect(next).toBeCalledTimes(1);
 
-		expect(res.locals.gitHubAppId).toBe(GIT_HUB_SERVER_APP_ID);
 		expect(res.locals.gitHubAppConfig).toEqual({
 			gitHubAppId: GIT_HUB_SERVER_APP_ID,
 			appId: GIT_HUB_SERVER_APP_APP_ID,
 			uuid: UUID,
 			clientId: "lvl.1234",
-			hostname: "http://myinternalserver.com"
+			gitHubBaseUrl: "http://myinternalserver.com",
+			gitHubApiUrl: "http://myinternalserver.com"
 		});
 	});
 });
--- a/src/middleware/github-server-app-middleware.ts
+++ b/src/middleware/github-server-app-middleware.ts
@@ -15,7 +15,6 @@ type ResponseType =  Response<
 export const GithubServerAppMiddleware = async (req: Request, res: ResponseType, next: NextFunction): Promise<void> => {
 	const { jiraHost } = res.locals;
 	const { uuid } = req.params;
-
 	req.addLogFields({ uuid, jiraHost });
 
 	if (uuid) {
--- a/src/middleware/github-server-app-middleware.ts
+++ b/src/middleware/github-server-app-middleware.ts
@@ -48,7 +48,7 @@ export const GithubServerAppMiddleware = async (req: Request, res: ResponseType,
 			appId: gitHubServerApp.appId,
 			uuid: gitHubServerApp.uuid,
 			gitHubBaseUrl: gitHubServerApp.gitHubBaseUrl,
-			gitHubApiUrl: gitHubServerApp.gitHubBaseUrl,
+			gitHubApiUrl: `${gitHubServerApp.gitHubBaseUrl}/api/v3`,
 			clientId: gitHubServerApp.gitHubClientId
 		};
 	} else {
--- a/src/middleware/github-server-app-middleware.ts
+++ b/src/middleware/github-server-app-middleware.ts
@@ -2,9 +2,17 @@ import { NextFunction, Request, Response } from "express";
 import { GitHubServerApp } from "models/github-server-app";
 import { Installation } from "models/installation";
 import { envVars } from "config/env";
-import { GITHUB_CLOUD_BASEURL } from "utils/get-github-client-config";
+import { GITHUB_CLOUD_BASEURL, GITHUB_CLOUD_API_BASEURL } from "utils/get-github-client-config";
 
-export const GithubServerAppMiddleware = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+type ResponseType =  Response<
+	{
+    message: string
+	},
+	JiraHostVerifiedLocals
+	& GitHubAppVerifiedLocals
+>;
+
+export const GithubServerAppMiddleware = async (req: Request, res: ResponseType, next: NextFunction): Promise<void> => {
 	const { jiraHost } = res.locals;
 	const { uuid } = req.params;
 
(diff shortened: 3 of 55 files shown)


src/middleware/github-server-app-middleware.test.ts
```diff
@@ -120,13 +120,13 @@ describe("github-server-app-middleware", () => {
...
 
-		expect(res.locals.gitHubAppId).toBe(GIT_HUB_SERVER_APP_ID);
 		expect(res.locals.gitHubAppConfig).toEqual({
 			gitHubAppId: GIT_HUB_SERVER_APP_ID,
 			appId: GIT_HUB_SERVER_APP_APP_ID,
 			uuid: UUID,
 			clientId: "lvl.1234",
-			hostname: "http://myinternalserver.com"
+			gitHubBaseUrl: "http://myinternalserver.com",
+			gitHubApiUrl: "http://myinternalserver.com/api/v3"
```
gxueatlassian, line 129: copy from `src/util/get-github-client-config.ts`

src/middleware/github-server-app-middleware.ts
```diff
@@ -36,21 +43,22 @@ export const GithubServerAppMiddleware = async (req: Request, res: Response, nex
...
 
 		//TODO: ARC-1515 decide how to put `gitHubAppId ` inside `gitHubAppConfig`
-		res.locals.gitHubAppId = gitHubServerApp.id;
 		res.locals.gitHubAppConfig = {
 			gitHubAppId: gitHubServerApp.id,
 			appId: gitHubServerApp.appId,
 			uuid: gitHubServerApp.uuid,
-			hostname: gitHubServerApp.gitHubBaseUrl,
+			gitHubBaseUrl: gitHubServerApp.gitHubBaseUrl,
+			gitHubApiUrl: `${gitHubServerApp.gitHubBaseUrl}/api/v3`,
```
gxueatlassian, line 51: copy from src/util/get-github-client-config.ts

src/middleware/github-server-app-middleware.ts
```diff
@@ -36,21 +43,22 @@ export const GithubServerAppMiddleware = async (req: Request, res: Response, nex
...
 			clientId: gitHubServerApp.gitHubClientId
 		};
 	} else {
 		req.log.info("Defining GitHub app config for GitHub Cloud.");
 		res.locals.gitHubAppConfig = {
 			gitHubAppId: undefined,
-			appId: envVars.APP_ID,
+			appId: Number(envVars.APP_ID),
```
gxueatlassian, line 58: This could be a potential bug.

src/middleware/jirahost-middleware.ts
```diff
@@ -68,7 +74,7 @@ export const jirahostMiddleware = async (req: Request, res: Response, next: Next
 			next();
 		});
 	} else {
-		res.locals.jiraHost = req.session.jiraHost;
+		res.locals.jiraHost = req.session.jiraHost as string;
```
gxueatlassian, line 77: This could be some potential bug.
So at runtime, `jiraHostMiddleware` can actually have an empty `res.locals.jiraHost` if session is empty.

But lots of routes after this middleware just assume the `jiraHost` is always there. 

--

This only fix the ts complaints, doesn't fix anything at runtime. Might need to look at it later.

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -31,7 +37,10 @@ export const GithubCreateBranchGet = async (req: Request, res: Response, next: N
 	if (!issueKey) {
 		return next(new Error(Errors.MISSING_ISSUE_KEY));
 	}
-	const subscriptions = await Subscription.getAllForHost(jiraHost, gitHubAppConfig.gitHubAppId || null);
+
+	//TODO  ARC-1823: fix potential subscript get all for host bug
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	const subscriptions = await Subscription.getAllForHost(jiraHost, gitHubAppConfig.gitHubAppId as any || null);
```
gxueatlassian, line 43: I need to review the intention of this `Subscriptioin.getAllForHost` when the `gitHubAppId` is undefined. Should it include `gitHubAppId is null` in the where clause? Here it does, but many other places it doesn't. Quite inconsistent.

gxueatlassian: And possibly raise another PR.

src/routes/github/create-branch/github-create-branch-post.ts
```diff
@@ -88,15 +96,15 @@ export const GithubCreateBranchPost = async (req: Request, res: Response): Promi
 		sendTrackEventAnalytics(AnalyticsTrackEventsEnum.CreateBranchSuccessTrackEventName, jiraHost);
 		const tags = {
 			name: newBranchName,
-			gitHubProduct: getCloudOrServerFromGitHubAppId(gitHubAppConfig.githubAppId)
```
gxueatlassian, line 91: A bug.

src/routes/github/repository/github-repository-get.ts
```diff
@@ -15,12 +24,13 @@ export const GitHubRepositoryGet = async (req: Request, res: Response): Promise<
 	}
 
 	if (!repoName) {
-		res.send(400);
```
gxueatlassian, line 18: 🐞

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-router.ts
```diff
@@ -16,6 +16,9 @@ JiraConnectEnterpriseAppRouter.use("/:uuid", routerWithUUID);
 
 routerWithUUID.use(GithubServerAppMiddleware);
 routerWithUUID.route("")
-	.get(csrfMiddleware, JiraJwtTokenMiddleware, JiraConnectEnterpriseAppCreateOrEdit)
-	.put(JiraContextJwtTokenMiddleware, JiraConnectEnterpriseAppPut)
-	.delete(JiraContextJwtTokenMiddleware, JiraConnectEnterpriseAppDelete);
+	//eslint-disable-next-line @typescript-eslint/no-explicit-any
+	.get(csrfMiddleware, JiraJwtTokenMiddleware as any, JiraConnectEnterpriseAppCreateOrEdit)
```
gxueatlassian, line 20: This is interesting, express typings enforce the `res.locals` has same type across the handlers within same route. And doesn't not distinguish `input type` and `output type` of `res.locals`.

I don't want to attach the type `gitHubAppConfig` to `JiraJwtTokenMiddleware` because it is use somewhere else, while `JiraConnectEnterpriseAppCreateOrEdit` required that `gitHubAppConfig` in `res.locals` for typings.

So just type it to `any` to convert it.

-----

So the way express typings enforce me, make me wonder is it a code smell the way we use `res.locals`?

src/routes/router.ts
```diff
@@ -68,10 +68,14 @@ RootRouter.get("/create-branch-options", GithubCreateBranchOptionsGet);
...
-// On base path, redirect to Github App Marketplace URL
-RootRouter.get("/", async (req: Request, res: Response) => {
-	const { jiraHost, gitHubAppId } = res.locals;
-	const gitHubAppClient = await createAppClient(req.log, jiraHost, gitHubAppId);
+// On base path, redirect to Github App Marketplace URL (This is only for cloud)
+type RedirectResponseType =  Response<
+	never,
+	JiraHostVerifiedLocals
+>;
+RootRouter.get("/", async (req: Request, res: RedirectResponseType) => {
```
gxueatlassian, line 76: Let me know if I am wrong, this route ONLY works for cloud right? We don't have intention for it to work for GHE as we won't have uuid in the path?

src/common-types/middleware-locals.d.ts
```diff
@@ -0,0 +1,20 @@
...
+	installation: any //TODO: This is a hack, fix it with full definition.
+};
+
+type GitHubAppVerifiedLocals = CommonResponseLocals & {
+	gitHubAppConfig: GitHubAppConfig,
+};
+
+type GitHubUserTokenVerifiedLocals = CommonResponseLocals & {
+	githubToken: string;
+};
```
mboudreau, line 20: Wouldn't it be easier to just have a single object with all locals, all of which are optional?

src/middleware/github-server-app-middleware.ts
```diff
@@ -2,12 +2,19 @@ import { NextFunction, Request, Response } from "express";
...
+import { GITHUB_CLOUD_BASEURL, GITHUB_CLOUD_API_BASEURL } from "utils/get-github-client-config";
 
-export const GithubServerAppMiddleware = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+type ResponseType =  Response<
+	{
+    message: string
+	},
+	JiraHostVerifiedLocals
+	& GitHubAppVerifiedLocals
+>;
```
mboudreau, line 13: Seems like a lot of extra work and gives you a false sense of security since all locals should be optional.  There's no guarantee that anything is set for locals and might cause a runtime error if it's not.

src/middleware/jirahost-middleware.ts
mboudreau: This is on purpose.  It's suppose to be undefined if it's missing.  there's a middleware that catches the issue of missing jirahost so that all others can assume it works.

src/routes/error-router.ts
```diff
@@ -76,7 +76,7 @@ ErrorRouter.use((err: Error, req: Request, res: Response, next: NextFunction) =>
 
 	const errorStatusCode = errorCodes[err.message] || 500;
 	const message = messages[err.message];
-	const gitHubProduct = getCloudOrServerFromGitHubAppId(res.locals.gitHubAppId);
+	const gitHubProduct = getCloudOrServerFromGitHubAppId((res.locals.gitHubAppConfig || {}).gitHubAppId);
```
mboudreau, line 79: Why not use optional chaining?

src/routes/github/create-branch/github-create-branch-get.ts
mboudreau: yes, that's the point.  If no id is specified, I needs to get the ones specifically with no id in the database (ie `null`).  If undefined is used, it gets all subscriptions as the where clause wouldn't be added.

src/middleware/github-server-app-middleware.ts
gxueatlassian: Some benefits (from these extra works) I see are:

1. Better typing when we need to use any field from `res.locals`. (And it does actually catch 1 or 2 bugs during this RP).
2. It declares what it relies on (sort of)
3. It might give you false sense of security, but it probably not worse than what it is right now, which is like `just use res.locals.xxx without types whenever you like`. 

---

I agree this PR doesn't address the security feeling/promise of the codebase, I'd prefer a more proper routing setup somehow to guarantee that.

src/common-types/middleware-locals.d.ts
gxueatlassian: So this is based on the ***`Assumption: routing is setup properly and securely.`*** 
With above assumption, handlers can mix and use different `XXXVerifiedLocals`, and having a `non-optional` field in it (because of the assumption).

We do have same assumption right now in prod just not typed it (see my other comments) bellow.

---

As how to address/impl the assumption itself, it can be another PR.

src/middleware/jirahost-middleware.ts
gxueatlassian: `there's a middleware that catches the issue of missing jirahost` <-- can you tell me which one is it? I must over looked it.

src/routes/router.ts
mboudreau: This is the top level route of the domain (`/`). as such, it has no knowledge of GHE instances is it's not under the `/github` route

src/routes/github/create-branch/github-create-branch-get.ts
gxueatlassian: Okay, got it. Although I still feel using `null` vs `undefined` to distinguish behaviour is unfamiliar. And could be dangerous when it hooked up to some deleting routes.

src/routes/github/create-branch/github-create-branch-post.ts
bgvozdev: omg! How come this wasn't caught in a test?! @joshkay10 fyi

gxueatlassian:
Leave it. Will do an optional typings in the future per comments.

PR #1740, opened 2022-11-14
dlq alarms

joshkay10:
**What's in this PR?**
Adding Alarms for the four SQS dead letter queues (branch, pull, deployment, backfill)

**Why**
We want to be alerted when there is a spike of messages failing

**Whats next**
Monitor DLQ logs and try to reduce the noise!

### DIFF
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -596,6 +596,19 @@ environmentOverrides:
             Period: 300
             ComparisonOperator: GreaterThanThreshold
             Statistic: Maximum
+          DLQAlarmOnTooManyMessages:
+            MetricName: ApproximateNumberOfMessagesVisible
+            Namespace: AWS/SQS
+            Description: "Message count for backfill dead letter queue too high!"
+            Threshold: 50
+            Priority: Low
+            Dimensions:
+              - Name: QueueName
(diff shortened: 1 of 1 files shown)


github-for-jira.sd.yml
```diff
@@ -630,6 +656,19 @@ environmentOverrides:
             Period: 300
             ComparisonOperator: GreaterThanThreshold
             Statistic: Maximum
+          DLQAlarmOnTooManyMessages:
+            MetricName: ApproximateNumberOfMessagesVisible
+            Namespace: AWS/SQS
+            Description: "Message count for deployment dead letter queue too high!"
+            Threshold: 5000 # number is initially high until we start bringing numbers down
```
gxueatlassian, line 663: Then maybe we don't have to add the alarm right now? Or we are supposed to add the alarm and be like super panic when the number did reach 5000?

gxueatlassian (approved):
LGTM. Happy to go ahead. Minor comments regarding the nubmers.

github-for-jira.sd.yml
joshkay10: its always got a lot in that queue, sits between 2000-4900 for the last few months, so over 5000 would be a "spike" - but there is a bigger issue here that the queue size is so massive to begin with

PR #1741, opened 2022-11-14
ARC-1405 Update JWT validation mechanism

Harminder84:
**What's in this PR?**
We relied on `jiraHost` passed as a request parameter and validation of JWT using the `sharedSecret`. But as per documentation, it states :
- Parse JWT
- Extract issuer which equals to `clientKey`
- query the installation table using client key and fetch the `jirahost`

Updated the middleware to implemented the above guidelines and set the jiraHost in `res.locals`. Also, removed the `jiraHost` cookie altogether. 

**Why**
To fix the vulnerability 

**Added feature flags**
NEW_JWT_VALIDATION

**Affected issues**  
[ARC-1405]

**How has this been tested?**  
Unit Testing 
Local

[ARC-1405]: https://softwareteams.atlassian.net/browse/ARC-1405


src/middleware/jira-jwt-middleware.ts
```diff
@@ -1,41 +1,104 @@
...
+		next();
 		return;
 	}
 
-	const installation = await Installation.getForHost(jiraHost);
-
+	const installation = await getInstallationAndTokenVerification(token, req, res);
 	if (!installation) {
-		return next(new Error("Not Found"));
+		return;
```
gxueatlassian, line 22: Hmm, this just return but not calling `next?`

src/middleware/jira-jwt-middleware.ts
```diff
@@ -1,41 +1,104 @@
...
 	}
-	// TODO: Probably not the best place to set things globally
+
 	res.locals.installation = installation;
+	res.locals.jiraHost = installation.jiraHost;
+	res.locals.jwtVerified = true;
+
+	if (req.cookies.jwt) {
+		req.session.jiraHost = installation.jiraHost;
```
gxueatlassian, line 30: why this line `req.session.jiraHost = installation.jiraHost;`?

src/middleware/jira-jwt-middleware.ts
```diff
@@ -1,41 +1,104 @@
...
+	}
+
+	// if token is alreafy verified, skip verification
+	if (res.locals.jwtVerified) {
+		next();
+		return;
+	}
+
+	if (!getInstallationAndTokenVerification(token, req, res)) {
+		return;
```
gxueatlassian, line 53: same `return` without calling `next`?

src/middleware/jira-jwt-middleware.ts
```diff
@@ -1,41 +1,104 @@
...
-			installation.clientKey && `${installation.clientKey.substr(0, 5)}***`
+	const token = extractJwtFromRequest(req);
+
+	if (!token) {
+		sendError(res, 401, "Could not find authentication data on request");
+		return;
+	}
+
+	// if token is alreafy verified, skip verification
+	if (res.locals.jwtVerified) {
```
gxueatlassian, line 47: I am not sure `res.locals.jwtVerified`, as the convention is anything within `res.locals` is "verified". 
let's discuss a bit ?

src/middleware/jira-jwt-middleware.ts
Harminder84: `getInstallationAndTokenVerification` method sets proper msg on response object, so just return in case of any error.

src/middleware/jira-jwt-middleware.ts
Harminder84: This is a case when application accessed directly and jwt token migrated as a cookie. In that case jirahost to be kept in session and remove jwt cookie.

src/middleware/jira-jwt-middleware.ts
gxueatlassian: Since `getInstallationAndTokenVerification` is completely private to this file, can you refactor it a bit so that the return is more explicity (like we don't implicitly depends on it to call `res.status().send()`)

It could be error prone implicitly depends on it to do something. I'd prefer having it return more explicit result either success or fail with error code (or throw with error )

src/middleware/jira-jwt-middleware.ts
Harminder84: This is for performance improvement. If token is already verified in `jiraHostMiddleware`, don't verify it again.

src/middleware/jira-jwt-middleware.ts
```diff
@@ -1,41 +1,94 @@
...
-	const { jiraHost } = res.locals;
+export const jiraHostMiddleware = async (req: Request, res: Response, next: NextFunction) => {
 
-	if (!jiraHost) {
-		sendError(res, 401, "Unauthorised");
-		return;
+	const token = extractJwtFromRequest(req);
+
+	if (!token) {
+		if (req.session.jiraHost) {
```
gxueatlassian, line 11: /nit, this can be 
```
if (!token) {
  res.locals.jiraHost = req.session.jiraHost;
  return next();
}
```

gxueatlassian (changes requested):
Just blocking it for now, let's discuss tmr.

gxueatlassian (commented):
just unblock it.

src/middleware/jiraSymmetricJwtMiddleware.ts
```diff
@@ -0,0 +1,140 @@
...
+
+};
+
+const getIssuer = (token: string): string | undefined => {
+
+	let unverifiedClaims;
+	try {
+		unverifiedClaims = decodeSymmetric(token, "", getAlgorithm(token), true); // decode without verification;
+	} catch (e) {
+		throw new Error(`Invalid JWT: ${e.message}`);
```
bgvozdev, line 73: Should we maybe log the error, too?
```
catch (err) {
  logger.warn({ err },  whatever);
  throw ...
}
```

src/middleware/frontend-log-middleware.ts
```diff
@@ -65,3 +66,20 @@ export const LogMiddleware = async (req: Request, res: Response, next: NextFunct
 	});
 	next();
 };
+
+const getUnvalidatedJiraHost = (req: Request): string | undefined =>
+	req.session?.jiraHost || extractUnsafeJiraHost(req);
+
+const checkPathValidity = (url: string) => moduleUrls.some(moduleUrl => matchRouteWithPattern(moduleUrl, url));
+
+const extractUnsafeJiraHost = (req: Request): string | undefined => {
```
bgvozdev, line 75: That's for logging only, right?

I think we could change it to something like
```
const createJiraHostChildLogger = (logger: Logger) => {
    ....

}
```

and encapsulate all the related to jiraHost logic here, removing all confusing words like "validated/unsafe/validity" etc. Wdyt?

src/middleware/jiraSymmetricJwtMiddleware.test.ts
```diff
@@ -0,0 +1,170 @@
...
+		installation = {
+			id: 19,
+			jiraHost,
+			clientKey: "jira-client-key",
+			enabled: true,
+			decrypt: jest.fn(() => testSharedSecret),
+			subscriptions: jest.fn().mockResolvedValue([])
+		};
+
+		jest.mocked(Installation.getForClientKey).mockImplementation((clientKey) => {
```
bgvozdev, line 39: Instead of mocking everything, wouldn't it be easer to create a record in Db? E.g. like we do here:

http://github.com/atlassian/github-for-jira/blob/1fc1bfb87eb30b4a805aa9990461b203d4825575/src/sqs/push.test.ts#L109-L109

src/middleware/jiraSymmetricJwtMiddleware.ts
```diff
@@ -0,0 +1,140 @@
...
+	if (!await booleanFlag(BooleanFlags.NEW_JWT_VALIDATION, false)) {
+		return next();
+	}
+
+	const token = req.query?.["jwt"] || req.cookies?.["jwt"];
+
+	if (token) {
+		const issuer = getIssuer(token);
+		if (!issuer) {
+			req.log.info("JWT claim did not contain the issuer (iss) claim");
```
bgvozdev, line 20: I feel like this should be a warn

src/middleware/jiraSymmetricJwtMiddleware.ts
```diff
@@ -0,0 +1,140 @@
...
+			return res.status(401).send("Unauthorised");
+		}
+
+		const installation = await Installation.getForClientKey(issuer);
+		if (!installation) {
+			req.log.info("No Installation found");
```
bgvozdev, line 26: Same: I feel like this should be a warn

src/middleware/jiraSymmetricJwtMiddleware.ts
```diff
@@ -0,0 +1,140 @@
...
+		}
+
+		const secret = await installation.decrypt("encryptedSharedSecret");
+
+		const path = req.baseUrl == "" ? req.path : req.baseUrl;
+		const tokenType = checkPathValidity(path) ? TokenType.normal : TokenType.context;
+		try {
+			verifySymmetricJwt(token, secret, req, tokenType);
+		} catch (err) {
+			req.log.info(err.message);
```
bgvozdev, line 37: Same: I feel like this should be a warn. Also, would it make sense to add error as the first arg?
```
catch (err) {
req.log.warn({ err }, "Could not verify symmetric JWT");
}
```

src/middleware/jiraSymmetricJwtMiddleware.ts
```diff
@@ -0,0 +1,140 @@
...
+
+		const installation = await Installation.getForClientKey(issuer);
+		if (!installation) {
+			req.log.info("No Installation found");
+			return res.status(401).send("Unauthorised");
+		}
+
+		const secret = await installation.decrypt("encryptedSharedSecret");
+
+		const path = req.baseUrl == "" ? req.path : req.baseUrl;
```
bgvozdev, line 32: shouldn't it be `req.originalURL` ? https://stackoverflow.com/a/68372568

src/middleware/jiraSymmetricJwtMiddleware.test.ts
```diff
@@ -0,0 +1,170 @@
...
+jest.mock("config/feature-flags");
+const testSharedSecret = "test-secret";
+
+describe("jiraSymmetricJwtMiddleware", () => {
+	let res;
+	let next;
+	let installation;
+
+	beforeEach(async () => {
+		res = {
```
bgvozdev, line 20: also, instead of mocking everything, would it make sense to create the app? This way we could test `req.path` stuff, too, couldn't we? (We might make a mistake during mocking..)  Like here:
http://github.com/atlassian/github-for-jira/blob/1fc1bfb87eb30b4a805aa9990461b203d4825575/src/routes/api/api-router.test.ts#L19-L19

src/middleware/jiraSymmetricJwtMiddleware.ts
```diff
@@ -0,0 +1,140 @@
...
+		if (req.cookies.jwt) {
+			res.clearCookie("jwt");
+		}
+		return next();
+
+	} else if (req.session?.jiraHost) {
+
+		const installation = await Installation.getForHost(req.session.jiraHost);
+		if (!installation) {
+			throw new Error("No Installation found");
```
bgvozdev, line 54: should we log a error?

bgvozdev: also, probably we should also return res.status(401) ?

src/middleware/jiraSymmetricJwtMiddleware.ts
```diff
@@ -0,0 +1,140 @@
...
+
+const verifySymmetricJwt = (token: string, secret: string, req: Request, tokenType: TokenType): boolean => {
+	const algorithm = getAlgorithm(token);
+
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	let unverifiedClaims: any;
+	try {
+		unverifiedClaims = decodeSymmetric(token, "", algorithm, true); // decode without verification;
+	} catch (e) {
+		throw new Error(`Invalid JWT: ${e.message}`);
```
bgvozdev, line 87: should we log the error here, too?

src/middleware/jiraSymmetricJwtMiddleware.ts
```diff
@@ -0,0 +1,140 @@
...
+	if (!issuer) {
+		throw new Error("JWT claim did not contain the issuer (iss) claim");
+	}
+
+	/* eslint-disable @typescript-eslint/no-explicit-any*/
+	let verifiedClaims: any; //due to decodeSymmetric return any
+	try {
+		verifiedClaims = decodeSymmetric(token, secret, algorithm, false);
+	} catch (error) {
+		throw new Error(`Unable to decode JWT token: ${error}`);
```
bgvozdev, line 100: should we log the error here, too?

src/middleware/jiraSymmetricJwtMiddleware.ts
```diff
@@ -0,0 +1,140 @@
+import { decodeSymmetric, getAlgorithm } from "atlassian-jwt";
```
bgvozdev, line 1: I think the name of the file is not following the convention, should be
`jira-symmetric-jwt-middleware.ts`

src/middleware/frontend-log-middleware.ts
Harminder84: This file doesn't require any change. Changes reverted.

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -0,0 +1,130 @@
...
+import { moduleUrls } from "~/src/routes/jira/atlassian-connect/jira-atlassian-connect-get";
+import { matchRouteWithPattern } from "~/src/util/match-route-with-pattern";
+
+export const jiraSymmetricJwtMiddleware = async (req: Request, res: Response, next: NextFunction) => {
+
+	if (!await booleanFlag(BooleanFlags.NEW_JWT_VALIDATION, false)) {
+		return next();
+	}
+
+	const token = req.query?.["jwt"] || req.cookies?.["jwt"];
```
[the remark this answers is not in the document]
Harminder84: Do you know any scenario where we're passing  JWT in header?

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -0,0 +1,130 @@
...
+
+	if (token) {
```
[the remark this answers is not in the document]
Harminder84: I don't prefer shorter route cuz of readability but we can break into sub methods if it improves readability

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -0,0 +1,130 @@
...
+
+		const installation = await Installation.getForClientKey(issuer);
+		if (!installation) {
+			req.log.warn("No Installation found");
+			return res.status(401).send("Unauthorised");
+		}
+
+		const secret = await installation.decrypt("encryptedSharedSecret");
+
+		const tokenType = checkPathValidity(req.originalUrl) ? TokenType.normal : TokenType.context;
```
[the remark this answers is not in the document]
Harminder84: This middleware only deals with Symmertric tokens not Asymmertric, so there is no need to determine token type.
Now Symmetric token could be with fixed or computed qsh.  That is decided using `checkPathValidity` method.

src/middleware/jira-symmetric-jwt-middleware.ts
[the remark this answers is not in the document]
gxueatlassian: Nah, just a thought. But usually token thing is safer in the headers, which we should be practicing more. As all middlewares (switch, gateway) will possibly logs the url while ignore logging of headers especially if they are in `Authorization` header.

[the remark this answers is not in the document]
gxueatlassian: All good. Can close this.

src/middleware/jira-symmetric-jwt-middleware.ts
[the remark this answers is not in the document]
gxueatlassian: Okay, I can leave this to you. All good.

src/middleware/jira-symmetric-jwt-middleware.ts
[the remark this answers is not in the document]
gxueatlassian: `Symmetric token could be with fixed or computed qsh` <-- this is the `TokenType` I meant, `normal` or `context`

So I think the determine it is `fixed` or `computed qsh`, we should just look into the token itself, instead of depends on url.

---

However, to determine whether `computed qsh` or `fixed` is allowed, we should look into url. This is a separate thing compare to above.

----

But all goods, since this is the existing logics, we can leave this as it is.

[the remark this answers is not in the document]
gxueatlassian: /nit
maybe rename `tokenType` to `expectedTokenType` or `allowedTokenType`? So that we know this is NOT the type of the token itself, but the type we are expecting (from url?)

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -0,0 +1,130 @@
...
+			req.log.warn("No Installation found");
+			return res.status(401).send("Unauthorised");
+		}
+
+		res.locals.installation = installation;
+		res.locals.jiraHost = installation.jiraHost;
+		return next();
+	}
+
+	req.log.info("No Token found");
```
gxueatlassian, line 63: 1. Should it be `warn`? 
2. Change text to `No token found and session cookie has not jiraHost`?

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -0,0 +1,130 @@
...
+export const jiraSymmetricJwtMiddleware = async (req: Request, res: Response, next: NextFunction) => {
+
+	if (!await booleanFlag(BooleanFlags.NEW_JWT_VALIDATION, false)) {
+		return next();
+	}
+
+	const token = req.query?.["jwt"] || req.cookies?.["jwt"];
+
+	if (token) {
+		const issuer = getIssuer(token, req.log);
```
gxueatlassian, line 19: Should we put a try/catch? If `getIssuer` throw errors, what's the expected output?

src/routes/github/create-branch/github-create-branch-options-get.ts
```diff
@@ -22,6 +22,9 @@ export const GithubCreateBranchOptionsGet = async (req: Request, res: Response,
 	}
 
 	const jiraHost = getJiraHostFromTenantUrl(tenantUrl) || jiraHostQuery;
+	// This is a temporary workaround as jiraHost is currently fetched from tenantUrl which is not secure.
+	// JiraHost will be fetched from JWT instead of tenatUrl. This will be implemented soon.
+	req.session.jiraHost = jiraHost;
```
gxueatlassian, line 27: **IMPORTANT**

This is concerning. Need more details on this.

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-router.ts
```diff
@@ -1,21 +1,22 @@
...
 
 export const JiraConnectEnterpriseAppRouter = Router();
 
-JiraConnectEnterpriseAppRouter.post("/", JiraContextJwtTokenMiddleware, JiraConnectEnterpriseAppPost);
+JiraConnectEnterpriseAppRouter.post("/", JiraContextJwtTokenMiddleware, jiraSymmetricJwtMiddleware, JiraConnectEnterpriseAppPost);
 
 const routerWithUUID = Router({ mergeParams: true });
 JiraConnectEnterpriseAppRouter.use("/:uuid", routerWithUUID);
 
 routerWithUUID.use(GithubServerAppMiddleware);
```
gxueatlassian, line 18: **IMPORTANT**
This is missing the new middleware (for `res.locals.jiraHost`)

gxueatlassian: Tbh, I think we need another refactoring (not this PR).
I spend quite a bit of time and look carefully to find this bug. Feels like:
1. We either missing tests (which is the case, but hard to fix)
2. And this is example of why nesting router doesn't help code readability from security point of view.

gxueatlassian (changes requested):
Some comments remaining.
And maybe we can have another walkthrough?

src/middleware/jira-symmetric-jwt-middleware.ts
Harminder84: Refactored the code, it will return 401 if getIssuer `throws` error

src/middleware/jira-symmetric-jwt-middleware.ts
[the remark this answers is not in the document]
Harminder84: I think this will create ambigituy, leaving as it is for now.

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-router.ts
Harminder84: Added `jiraSymmetricJwtMiddleware` for `JiraConnectEnterpriseAppRouter` router

src/routes/github/github-oauth-router.ts
```diff
@@ -16,7 +16,12 @@ const scopes = ["user", "repo"];
 const callbackPath = "/callback";
 
 const getRedirectUrl = async (req, res, state) => {
-	const { baseUrl } = req;
+	const basePaths = req.originalUrl.split("/");
+	let baseUrl  = basePaths.length > 0 ? `/${basePaths[1]}` : req.originalUrl;
+	if (res.locals?.gitHubAppConfig?.uuid) {
+		baseUrl += `/${res.locals.gitHubAppConfig.uuid}`;
+	}
```
Harminder84, line 23: Since `GithubAuthMiddleware` is now bind with route, so `baseUrl` contains full path. 
For e.g. for route `/github/configuration`, earlier it `baseUrl` contains `/github`, now it has `/github/configuration`

src/middleware/jira-symmetric-jwt-middleware.test.ts
```diff
@@ -0,0 +1,172 @@
...
+
+	it("should set res.locals and req.session when jiraHost set in session", async () => {
+
+		locals = {};
+		session.jiraHost = jiraHost;
+
+		await supertest(app)
+			.get(`/test`)
+			.then((res) => {
+				expect(res.status).toEqual(200);
```
bgvozdev, line 129: nit: that already was tested above (in `should return valid response when jiraHost set in session`)

src/middleware/jira-symmetric-jwt-middleware.test.ts
```diff
@@ -0,0 +1,172 @@
...
+				expect(locals.installation.jiraHost).toEqual(jiraHost);
+				expect(locals.jiraHost).toEqual(jiraHost);
+				expect(session.jiraHost).toEqual(jiraHost);
+			});
+	});
+
+	it("should throw error when invalid jiraHost set in session", async () => {
```
bgvozdev, line 136: nit: In this case, should we also drop the value from session? E.g. let's say for some reason the value got malformed. In this case, if we are not removing it from the session, the user will be stuck...

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	return verifyJwtClaims(verifiedClaims, tokenType, req);
+};
+
+export const verifyJwtClaims = (verifiedClaims: { exp: number, qsh: string }, tokenType: TokenType, req: Request): boolean => {
+	const expiry = verifiedClaims.exp;
+
+	if (expiry && (Date.now() / 1000 >= expiry)) {
+		throw new Error("JWT Verification Failed, token is expired");
```
bgvozdev, line 107: do we have a test for this?

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	if (verifiedClaims.qsh) {
+		let qshVerified: boolean;
+		if (tokenType === TokenType.context) {
+			//If we use context jsw tokens, their qsh will be constant
+			qshVerified = verifiedClaims.qsh === "context-qsh";
```
bgvozdev, line 114: do we have a test for this?

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -0,0 +1,137 @@
...
+		} else {
+			//validate query string hash
+			qshVerified = verifyQsh(verifiedClaims.qsh, req);
```
bgvozdev, line 117: do we have a test for this, when it is not valid?

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -0,0 +1,137 @@
...
+
+		const installation = await Installation.getForClientKey(issuer);
+		if (!installation) {
+			req.log.warn("No Installation found");
+			return res.status(401).send("Unauthorised");
+		}
+
+		const secret = await installation.decrypt("encryptedSharedSecret");
+
+		const tokenType = checkPathValidity(req.originalUrl) && req.method == "GET" ? TokenType.normal : TokenType.context;
```
bgvozdev, line 36: do we have a test for at least one of the URL from the list, to make sure we pick the token type correctly?

src/routes/github/github-oauth-router.ts
bgvozdev: Not sure I follow.... Has this been tested in a test?

src/routes/github/github-oauth-router.ts
```diff
@@ -13,12 +13,18 @@ import { GitHubServerApp } from "models/github-server-app";
...
-const callbackPath = "/callback";
+const callbackSubPath = "/callback";
+const callbackPathCloud = `/github${callbackSubPath}`;
+const callbackPathServer = `/github/<uuid>${callbackSubPath}`;
+
+const getRedirectUrl = async (res, state) => {
+	let callbackPath = callbackPathCloud;
+	if (res.locals?.gitHubAppConfig?.uuid) {
+		callbackPath = callbackPathServer.replace("<uuid>", res.locals.gitHubAppConfig.uuid);
+	}
```
bgvozdev, line 24: Made this change to be explicit what kind of URLs we support, which is only 2:
image

Given that `GitHubAuthMiddleware` middleware can be used in many places, we cannot rely on `req` anymore...

src/middleware/jira-symmetric-jwt-middleware.ts
Harminder84: test case added

src/middleware/jira-symmetric-jwt-middleware.ts
Harminder84: test case added

src/middleware/jira-symmetric-jwt-middleware.ts
Harminder84: test case added

bgvozdev (approved):

PR #1742, opened 2022-11-16
Arc-1806 add app properties

rachellerathbone:
**What's in this PR?**
Added appProperties to the Jira client so we can store a property to reflect if the app has been configured or not. 

**Why**
This will allow us to surface a message in the Jira dev panel.

**Affected issues**  
Arc-1806

**How has this been tested?**  
Locally and via tests 

**Whats Next?**
Using the isConfigured state prop in JFE 🎉 

Shout out to @joshkay10 for the spike and initial work on this bad boy


src/jira/client/jira-client.ts
```diff
@@ -33,7 +33,7 @@ export interface DeploymentsResult {
 // TODO: need to type jiraClient ASAP
 export const getJiraClient = async (
 	jiraHost: string,
-	gitHubInstallationId: number,
+	gitHubInstallationId: number | undefined,
```
rachellerathbone, line 36: Had to add this here coz we don't pass the GH installation id when we call the appProperties put/delete from gh-config-get and the Jira uninstall event. That said, it means I've also had to update the typing for batchedBulkUpdate and sub.getSingleInstallation so happy to have alternative approaches thrown at me.

src/routes/github/configuration/github-configuration-post.ts
```diff
@@ -70,6 +71,7 @@ export const GithubConfigurationPost = async (req: Request, res: Response): Prom
 			gitHubAppId
 		});
 
+		await saveConfiguredAppProperties(jiraHost, gitHubInstallationId, gitHubAppId, req, "true");
```
joshkay10, line 74: /nit could bundle them in a promise.all since they don't rely on each other

src/util/save-app-properties.ts
```diff
@@ -0,0 +1,18 @@
+import { getJiraClient } from "~/src/jira/client/jira-client";
+
+export const getAppKey = (): string => {
```
joshkay10, line 3: /nit this looks like a handy function to have in a generic utils or its own file 🤷

src/jira/client/jira-client.ts
joshkay10: mmmmmm this is tricky isn't it, I don't mind too much but then it forces you to update subsequent functions with the same` | undefined` which become more of a pain

other options off the top my head.  Not necessarily better mind you
- if we can, just pass the githubinstallationId anyway?
- pass -1 as the installation Id to get around it and add come comments for the magical number

rachellerathbone: Is that -1 a little hacky though? I wasn't passing it coz at the points where it's being called it's not available e.g. in the uninstall event, that prop doesn't exist on anything we have access to. At least I don't think it does 🤔

src/util/save-app-properties.ts
rachellerathbone: I only did it this way coz in the uninstall event it's calling the delete method which is diff to what's being called in the main function here. Added this guy to try and at least reduce some of the repeated code there but happy to kill this and just add those lines over there if you think that's better.

src/jira/client/jira-client.ts
joshkay10: sweet, leave it as it is and can deal with it if the need arises

src/util/save-app-properties.ts
joshkay10: same things, leave it there, if we need it we can refactor then, you have already broken it outside of the core function so its easy enough for next person to move it if they need it.

joshkay10 (dismissed):

src/jira/client/jira-client.ts
```diff
@@ -356,6 +356,14 @@ export const getJiraClient = async (
 				logger.info("Sending remoteLinks payload to jira.");
 				await instance.post("/rest/remotelinks/1.0/bulk", payload);
 			}
+		},
+		appProperties: {
+			create: (appKey: string, isConfiguredState: string) =>
```
gxueatlassian, line 361: /nit
I don't think `appKey` gonna be different upon each time this function is called (within same env). So maybe easier to read the code if we call `getAppKey()` within this function instead of receive it as an arg.

src/jira/client/jira-client.ts
```diff
@@ -356,6 +356,14 @@ export const getJiraClient = async (
...
+				instance.put(`/rest/atlassian-connect/1/addons/${appKey}/properties/isConfigured`, {
```
gxueatlassian, line 362: Just by the PR me and K is doing, the url mostly will be [/atlassian-connect/1/addons-gh4j/properties/is-configured](https://stash.atlassian.com/projects/JIRACLOUD/repos/jira/pull-requests/80413/diff#jira-components/jira-plugins/atlassian-connect/core-components/core/src/main/java/com/atlassian/plugin/connect/plugin/rest/addons/GitHubForJiraAddonPropertiesResource.java)

src/routes/jira/events/jira-events-uninstall-post.test.ts
```diff
@@ -34,12 +42,29 @@ describe("Webhook: /events/uninstalled", () => {
 				.mockResolvedValue(subscriptions)
 		};
 
+		frontendApp = express();
```
gxueatlassian, line 45: Since we are calling the `await JiraEventsUninstallPost(req, res);` directly, don't think we need to setup `frontendApp` anyway. Or if we do, use `supertest(frontendApp)` instead.

src/jira/client/jira-client.ts
gxueatlassian: ignore me.

src/jira/client/jira-client.ts
```diff
@@ -356,6 +357,14 @@ export const getJiraClient = async (
 				logger.info("Sending remoteLinks payload to jira.");
 				await instance.post("/rest/remotelinks/1.0/bulk", payload);
 			}
+		},
+		appProperties: {
+			create: (isConfiguredState: string) =>
+				instance.put(`/rest/atlassian-connect/1/addons/${getAppKey()}/properties/isConfigured`, {
```
krazziekay, line 363: Can you please update the camelcased `isConfigured` to `is-configured`? The property is now hard-coded and I think hard-coded strings are hyphenated. 
`is-configured` is being used in both JBE and JFE. 
And also, can we update the version `1` to `latest`?

krazziekay (approved):

PR #1745, opened 2022-11-17
ARC-1827 add db script to add new column for installation table to recover clientKey

gxueatlassian:
**What's in this PR?**
Add new column to recover clientKey

**Why**
Origin clientKey is hashed. We need it for the encryptionContext of GitHubServerApp though. So need this new col to save recovered origin clientKey

**Added feature flags**
N/A

**Affected issues**  
ARC-1827

**How has this been tested?**  
N/A

**Whats Next?**
1. Admin API to recover it
2. Real PR to do encryption context with it.


bgvozdev (dismissed):

db/migrations/20221117154600-add-new-clientn-key-column-to-installation-table.js
```diff
@@ -0,0 +1,13 @@
+"use strict";
+
+module.exports = {
+	up: async (queryInterface, Sequelize) => {
+		await queryInterface.addColumn("Installations", "plainClientKey", { type: Sequelize.STRING, allowNull: true });
```
gxueatlassian, line 5: name suggestion welcome. 
As `models/Installation` has `clientKey`, and `models/Subscription` has `jiraClientKey`. I am out of options.
I want to make variable clearly distinguishable ( at least before we deprecate the hashed client key)

bgvozdev (approved):

PR #1747, opened 2022-11-18
Arc 1827 recover existing client key

gxueatlassian:
**What's in this PR?**
Add admin api to recover origin client key value for all existing installations.
Also update how models/Installation/Subscriptions handle the client keys (has to do it in the same PR as admin api requires it).

**Why**
We need the pre-hashed client key to encrypt GHE app for the encryption context.
Don't want to rely on hash client key for the encryption context.

**Added feature flags**
N/A

**Affected issues**  
ARC-1827

**How has this been tested?**  
Locally.

**Whats Next?**
1. Add client key to GHES app encryptionContext
2. Clean up code once stable, remove all hashed clientKey access.


db/migrations/20221117154600-add-new-clientn-key-column-to-installation-table.js
```diff
@@ -0,0 +1,16 @@
+"use strict";
```
joshkay10, line 1: Thinking out loud... Do we want to seperate this migration as a seperate step? I think since its a new endpoint thats being called we are probably ok? but just threw it out for consideration

gxueatlassian: Haha, yes, YES!!
Therefore I put this PR back to draft. Wait for another one to go first. 
Good pickup.

src/routes/api/client-key/recover-client-key.ts
```diff
@@ -0,0 +1,94 @@
...
+};
+
+const getAndVerifyOriginClientKey = async ({ id, jiraHost, clientKey: hashedClientKeyInDB }: Installation): Promise<string>  => {
+
+	if (!jiraHost) {
+		throw { msg: `Cannot find jiraHost`, id };
+	}
+	if (!jiraHost.endsWith("/")) jiraHost = jiraHost + "/";
+
+	const result = await axios(`${jiraHost}plugins/servlet/oauth/consumer-info`);
```
joshkay10, line 67: Make sure you replace this with outbound proxy!!!
example https://github.com/atlassian/github-for-jira/pull/1690/files

src/routes/api/client-key/recover-client-key.ts
```diff
@@ -0,0 +1,115 @@
...
+		throw { msg: `Empty consumer-info`, jiraHost, id };
+	}
+
+	const [, plainClientKey] = /<key>([0-9a-z-]+)<\/key>/gmi.exec(text) || [undefined, undefined];
+	if (!plainClientKey) {
+		throw { msg: `Client key regex failed`, jiraHost, id, text };
+	}
+
+	const hashedplainClientKey = getHashedKey(plainClientKey);
+	if (hashedClientKeyInDB !== hashedplainClientKey) {
```
bgvozdev, line 104: nice!

src/routes/github/configuration/github-configuration-post.ts
```diff
@@ -64,7 +63,7 @@ export const GithubConfigurationPost = async (req: Request, res: Response): Prom
 		}
 
 		const subscription = await Subscription.install({
-			clientKey: getHashedKey(req.body.clientKey),
+			clientKey: req.body.clientKey,
```
bgvozdev, line 66: Shouldn't it be our next step? I thought first we will be writing both keys in parallel? Can we escape from downtime without it?

gxueatlassian: Because admin api need to update the plain client key
  ---> we don't want to do migration again and again
     --->So the normal `install` method must save plain key to new col
         --> `Subscription.install` 's arg is hashed, so need to change it
           --> :)

src/routes/api/client-key/recover-client-key.ts
gxueatlassian: Actually I replace it with whitelist proxy (so that we can hit the jira-dev.com). Let me know if that's okay.
--> whitelist proxy should be a "super-set" of outbound proxy I believe.

bgvozdev (approved):

joshkay10 (approved):

PR #1754, opened 2022-11-21
fix test

gxueatlassian:
`jest.mock` is fine, just some minor fix on the `when(xxx).callWith`

### DIFF
--- a/src/routes/jira/events/jira-events-uninstall-post.test.ts
+++ b/src/routes/jira/events/jira-events-uninstall-post.test.ts
@@ -61,10 +61,10 @@ describe("Webhook: /events/uninstalled", () => {
 
 	it("Existing Installation", async () => {
 		when(jest.mocked(getJiraClient))
-			.calledWith(jiraHost, undefined, undefined, expect.anything())
+			.calledWith(installation.jiraHost, undefined, undefined, expect.anything())
 			.mockResolvedValue(mockJiraClient);
 
-		const req = {} as any;
+		const req = { log: getLogger("test") } as any;
 		const res = { locals: { installation }, sendStatus: jest.fn() } as any;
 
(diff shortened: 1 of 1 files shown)


src/routes/jira/events/jira-events-uninstall-post.test.ts
```diff
@@ -61,10 +61,10 @@ describe("Webhook: /events/uninstalled", () => {
 
 	it("Existing Installation", async () => {
 		when(jest.mocked(getJiraClient))
-			.calledWith(jiraHost, undefined, undefined, expect.anything())
```
gxueatlassian, line 64: This is where the test wasn't working. `expect.anything()` of the 4th arg is not matching `req.log`.

In previous code, `req.log` is actually `undefined` inside `JiraEventsUninstallPost`  (although you set it up at `frontendApp`, but that's not being hit  coz `JiraEventsUninstallPost` was hit directly.

And because of `req.log` is undefined, `expect.anything()` will not match it. [expect.anything() matches anything but null or undefined](https://jestjs.io/docs/expect#expectanything)

src/routes/jira/events/jira-events-uninstall-post.test.ts
```diff
@@ -75,10 +75,10 @@ describe("Webhook: /events/uninstalled", () => {
 
 	it("Existing Installation, no Subscriptions", async () => {
 		when(jest.mocked(getJiraClient))
-			.calledWith(jiraHost, undefined, undefined, expect.anything())
+			.calledWith(installation.jiraHost, undefined, undefined, expect.anything())
```
gxueatlassian, line 78: Also, I don't think `jiraHost` === `installation.jiraHost`, hence not matching either.

PR #1760, opened 2022-11-22
Arc 1827 ghes app encryption context

gxueatlassian:
**What's in this PR?**
1. Change GEHS App to encrypt/decrypt with jiraHost in encryptionContext
2. Add admin api to re-encrypt existing with jiraHost

**Why**
1. To add one more defence line to the app. As it hold secrets, data compromise will lead to our customer intellectual properties (their codes) leakage to public.

**Added feature flags**
N/A

**Affected issues**  
ARC-1827

**How has this been tested?**  
✅ Tested on stg.

**Whats Next?**
1. Run the admin api after merge.


src/models/github-server-app.ts
```diff
@@ -50,28 +49,40 @@ export class GitHubServerApp extends EncryptedModel {
 	updatedAt: Date;
 	createdAt: Date;
 
-	getEncryptionSecretKey() {
-		return EncryptionSecretKeyEnum.GITHUB_SERVER_APP;
+	getDecryptedGitHubClientSecret(jiraHost: string): Promise<string> {
+		return this.decrypt(jiraHost, "gitHubClientSecret");
```
bgvozdev, line 53: Just an idea: for the time being, while we haven't run the migration, would it make sense to try-catch and then fallback to the old implementation (without context)?

gxueatlassian: Yup the fallback is in place already. Check inside `this.decrypt()`

bgvozdev (approved):

PR #1762, opened 2022-11-23
ARC-1807 - update is configured status script

joshkay10:
**What's in this PR?**
new admin endpoint to update Jira app properties with isConfigured value.
Ideally we will only pass it configured jiraHosts to begin with but double checks just in case.

**Why**
With new isConfigured status being applied on new installs we need to update existing installations

**How has this been tested?**  
Writing test cases ATM


joshkay10:
its lost its marbles and can't be bothered un f-ing it

PR #1764, opened 2022-11-23
ARC-1448 adding api gateway

mboudreau:
**What's in this PR?**
Changing service descriptor to add API gateway in front of our app so we can shove github and jira webhooks directly into an SQS queue to prevent losing information if the app goes down for any reason.

**Why**
Because we would sometimes lose webhooks because of a bad deployment or code issue.  

**Added feature flags**
None

**Affected issues**  
ARC-1448

**How has this been tested?**  
Manual

**Whats Next?**
Relax


nginx.conf
```diff
@@ -0,0 +1,12 @@
+# this is a nginx based configuration for the dev proxy
+server {
+    listen 3128;
+    location / {
+      access_log off;
+      proxy_pass http://g4j-localstack:4566/restapis/gateway/local/_user_request_/;
+#       proxy_set_header X-Real-IP $remote_addr;
```
rachellerathbone, line 7: Do we need these or can we deleted the stuff that's commented out?

mboudreau: yeah, can probably remove that :)

github-for-jira.sd.yml
```diff
@@ -486,7 +528,37 @@ environmentOverrides:
...
+          stageName: live
+          dataType:
+            - UGC/Label                # name of GitHub org / Jira site
+            - PII/IndirectConfidential # name of GitHub org
+            - UGC/Configuration        # data about the installation of the GitHub app into Jira sites
+            - Security/Secret          # shared Connect secret
+            - UGC/PrimaryIdentifier    # references to GitHub entities (commits, pull requests, etc.) and Jira issues
+            - UGC/Primary
+
+      - name: rds
```
gxueatlassian, line 556: I think last time I check it, it will raise some credit error after we deploy (or during deploy).
But worth solve it anyway.

package.json
```diff
@@ -4,7 +4,7 @@
 	"description": "GitHub integration for Atlassian Jira",
 	"repository": "https://github.com/atlassian/github-for-jira.git",
 	"engines": {
-		"node": ">= 14.21 <15",
+		"node": ">= 14.17 <15",
```
gxueatlassian, line 7: Just curious why?
/nit

src/sqs/webhooks-handler.ts
```diff
@@ -0,0 +1,76 @@
+import { MessageHandler, SQSMessageContext, WebhookMessagePayload } from "./sqs.types";
+import { createGithubSignature, getWebhookSecret, webhookRouter } from "routes/github/webhook/webhook-receiver-post";
+import { WebhookContext } from "routes/github/webhook/webhook-context";
+import { envVars } from "config/env";
+import { GITHUB_CLOUD_API_BASEURL, GITHUB_CLOUD_BASEURL } from "~/src/github/client/github-client-constants";
+
+export const webhooksQueueMessageHandler: MessageHandler<WebhookMessagePayload> = async (context: SQSMessageContext<WebhookMessagePayload>): Promise<void> => {
```
bgvozdev, line 7: tests?

src/sqs/webhooks-handler.ts
```diff
@@ -0,0 +1,76 @@
...
+			context.log.debug(context, "Could not parse body of webhook payload");
+		}
+	}
+	const event = context.payload.event;
+	if (event.startsWith("github")) {
+		return await githubWebhookHandler(event, context, data);
+	}
+};
+
+const githubWebhookHandler = async (event: string, context: SQSMessageContext<WebhookMessagePayload>, data: Record<string, any>): Promise<void> => {
```
bgvozdev, line 23: tests?

api-gateway.yml
```diff
@@ -0,0 +1,79 @@
...
+        schema:
+          type: "string"
+    post:
+      responses:
+        204:
+          description: "Default response for POST webhooks"
+      x-amazon-apigateway-integration:
+        credentials: "{{RoleARN}}"
+        httpMethod: "POST"
+        uri: "arn:aws:apigateway:{{WebhooksQueueRegion}}:sqs:path/{{AccountNumber}}/{{WebhooksQueueName}}"
```
bgvozdev, line 24: SQS has a limit of 256 Kb . Are we sure our requests are less than this?

bgvozdev:
What's our migration strategy of the enterprise customers? When this will happen? I believe our GHE server app needs to have Organisation hooks perms, how can we make sure they have it for sure? https://docs.github.com/en/rest/overview/permissions-required-for-github-apps?apiVersion=2022-11-28#organization-hooks

bgvozdev (changes requested):
Asked some questions, blocking till have answers :sorry:

github-for-jira.sd.yml
mboudreau: credit error?

package.json
mboudreau: because there was an issue last week with the snyk scanner which didn't have 14.21 available to do.  Not sure if it was temporary or not, but it completely blocked deployment, so this was the fix as it did have 14.17 available

src/sqs/webhooks-handler.ts
mboudreau: yep, incoming.

api-gateway.yml
mboudreau: As of yet, they have been. I think the largest I've seen was around 10kb.  256kb in text is pretty massive for a webhook.

bgvozdev: Do we have some metrics around this? We should double-check this imho, otherwise the change looks risky

mboudreau:
@bgvozdev for GHE customers, we can update the app webhook url via an admin API.  I haven't implemented it yet because the old webhook are still going to work, so nothing broken.  Wanted to get this out into prod and start moving the production app webhook first, then add that admin API to change the GHE apps.

bgvozdev:
@mboudreau But does the app has enough permissions to update it in GHE?

gxueatlassian:
Close it for now as it is too long. Will discuss later. Should be tracked in this ticket https://softwareteams.atlassian.net/browse/ARC-1448

PR #1768, opened 2022-11-29
ARC-1807 - script and endpoints to set jira isconfigured state

joshkay10:
**What's in this PR?**
Created a Python script that loops through a csv of installation id(in batches), for each id it will call a new endpoint to save the is configured state.

**Why**
We currently set this value during the install process.
We need to update existing installs to have the correct flag so it can show/hide the new config banner

**Added feature flags**
Not required

**How has this been tested?**  
Ran on local
-- Will run on stage before merge

**Whats Next?**
Run the script on Production for all installations with subscriptions


src/util/app-properties-utils.ts
```diff
@@ -0,0 +1,26 @@
...
+	const jiraClient = await getJiraClient(jiraHost, gitHubInstallationId, gitHubAppId, logger);
+
+	try {
+		await jiraClient.appProperties.create(isConfiguredState);
+	} catch (err) {
+		logger.error({ err }, "Set app properties failed");
+	}
+};
+
+export const getConfiguredAppProperties = async (jiraHost: string, gitHubInstallationId: number | undefined, gitHubAppId: number | undefined, logger: Logger) => {
```
joshkay10, line 19: Only this function is new, the rest is green because i renamed the file

src/routes/api/configured/api-configured-post.ts
```diff
@@ -0,0 +1,45 @@
...
+	if (installationIds.length > MAX_INSTALLATIONS_BATCH) {
+		res.status(400);
+		res.send(`Calm down Cowboy, keep it under ${MAX_INSTALLATIONS_BATCH} at a time!`);
+		return;
+	}
+
+	const logger = getLogger("api-sync-configured");
+	const tasks = installationIds.map(async installationId => {
+		// need to confirm that passed in value is configured. Existing on subscription table satisfies this.
+		const subscription = await Subscription.findOneForGitHubInstallationId(installationId, undefined);
```
krazziekay, line 28: This is only gonna fetch the list of GitHub cloud subscriptions, what about the GHE ones?

src/routes/api/configured/api-configured-post.ts
```diff
@@ -0,0 +1,45 @@
+import { Request, Response } from "express";
+import { Subscription } from "models/subscription";
+import { getLogger } from "config/logger";
+import { saveConfiguredAppProperties } from "utils/app-properties-utils";
+
+const MAX_INSTALLATIONS_BATCH = 50;
+/**
+ * This will update JIRA APP Properties for the GH4J installation to set isConfigured;
+ */
+export const ApiConfiguredPost = async (req: Request, res: Response): Promise<void> => {
```
krazziekay, line 10: /nit
I think `AppConfigurationPost` or `AppConfigPost` sounds more appropriate than `ApiConfiguredPost`.

src/routes/api/configured/api-configured-get.ts
```diff
@@ -0,0 +1,32 @@
+import { Request, Response } from "express";
+import { getLogger } from "config/logger";
+import { getConfiguredAppProperties } from "utils/app-properties-utils";
+import { Subscription } from "models/subscription";
+
+/**
+ * Makes a request to Jira to get jiraHosts is-configured app property
+ */
+export const ApiConfiguredGet = async (req: Request, res: Response): Promise<void> => {
```
krazziekay, line 9: Are we using this endpoint anywhere? Where do we need this?

etc/scripts/sync-configured-state-from-csv.py
```diff
@@ -0,0 +1,188 @@
...
+class Environment(NamedTuple):
+    github_for_jira_url: str
+    github_for_jira_auth: object
+
+
+def create_environment(env: str) -> Environment:
+    print("Running for env", env)
+    if env == 'dev':
+        return Environment(
+            github_for_jira_url='https://jkay-tunnel.public.atlastunnel.com',
```
krazziekay, line 94: This hard-coded value might need to be pulled from the `env.local`.

etc/scripts/sync-configured-state-from-csv.py
```diff
@@ -0,0 +1,188 @@
...
+    LOG.debug('Starting rotation of %s with url: %s', installations, url)
+    response = session.post(url,
+      auth=env.github_for_jira_auth,
+      json={
+        "installationIds": list(installations)
+      },
+      headers = {"Content-Type": "application/json"})
+
+    if response.ok:
+        LOG.info('Sync of %s successfully started: (%s).',
```
krazziekay, line 127: At this point, isn't the sync finished? Shouldn't the message be `Sync of %s installations completed successfully`?

etc/scripts/sync-configured-state-from-csv.py
```diff
@@ -0,0 +1,188 @@
+#!/usr/bin/env python
+
+"""
+Script to set isConfigured value against Jira App properties from CSV of JiraHosts.
+
+Takes an installationIds.csv with installationIds.
```
krazziekay, line 6: Just a question here: Does this mean that we need another extra step of getting the list of existing installation's id in a CSV?

src/routes/api/configured/api-configured-get.ts
joshkay10: Only using it manually to verify. Happy to pull it out but treat it like a feature flag. 
If we have everything deployed(the other storys for displaying things) otherwise its just handy to validate that the script worked.

etc/scripts/sync-configured-state-from-csv.py
joshkay10: ooooo good catch!!! that was me testing.

etc/scripts/sync-configured-state-from-csv.py
joshkay10: yup, I have a list already, so we can do that.

Alternatively, we could say run for N then stop  and maintain either and index of where we ended or maintain the output.csv as current and it will skip syncs installs.

Just needed a easy way to stop and start incase of issues basically.

src/routes/api/configured/api-configured-post.ts
joshkay10: yeah I bounced around a few i agree with `AppConfigurationPost` Ill update the rest to suit too

src/routes/api/configured/api-configured-post.ts
joshkay10: Correct, I'm in two minds about GHE, will move this chat to slack cuz I don't think we need to include them for the test.

To work with GHE, it would be a call per GitHubAppId or change the data struct of the installations = [] to installations = [{installationId: x, githubAppId: y}], wither way was adding more complexity that didn't give enough benefit IMO. But .. to slack!! and you can decide whether you want them or not

src/routes/jira/jira-get.ts
```diff
@@ -187,7 +187,7 @@ const renderJiraCloudAndEnterpriseServer = async (res: Response, req: Request):
 
 	const hasConnections =  !!(installations.total || gheServers?.length);
 
-	await saveConfiguredAppProperties(jiraHost, undefined, undefined, req, hasConnections ? "true" : "false");
+	await saveConfiguredAppProperties(jiraHost, undefined, undefined, req.log, hasConnections ? true : false);
```
krazziekay, line 190: /nit Its already boolean. 
```suggestion
	await saveConfiguredAppProperties(jiraHost, undefined, undefined, req.log, hasConnections);
```

src/routes/api/configuration/api-configuration-post.ts
```diff
@@ -0,0 +1,42 @@
...
+		res.send(`Calm down Cowboy, keep it under ${MAX_INSTALLATIONS_BATCH} at a time!`);
+		return;
+	}
+
+	const logger = getLogger("api-sync-configured");
+	const tasks = jiraHosts.map(async jiraHost => {
+		// need to confirm that passed in value is configured. Existing on subscription table satisfies this.
+		const subscriptions = await Subscription.getAllForHost(jiraHost, gitHubAppId);
+		// We could still save isconfiguredstate as false, but null is equivalent so why not save some trees and leave Jira alone
+		return await saveConfiguredAppProperties(jiraHost, undefined, gitHubAppId, logger, subscriptions?.length > 0 ? true : false);
```
krazziekay, line 31:
```suggestion
		return await saveConfiguredAppProperties(jiraHost, undefined, gitHubAppId, logger, subscriptions?.length > 0);
```

PR #1772, opened 2022-12-05
ARC-936 fix api installation bug

gxueatlassian:
**What's in this PR?**
Fix a bug in api route.
As the current route doesn't take the `gitHubAppId`, so need a new route the append the id. 

**Affected issues**  
ARC-936

### DIFF
--- a/src/routes/api/api-router.ts
+++ b/src/routes/api/api-router.ts
@@ -85,14 +85,14 @@ ApiRouter.post("/ping", ApiPingPost);
 
 // TODO: remove once move to DELETE /:installationId/:jiraHost
 ApiRouter.delete(
-	"/deleteInstallation/:installationId/:jiraHost",
+	"/deleteInstallation/:installationId/:jiraHost/github-app-id/:gitHubAppId",
 	param("installationId").isInt(),
 	param("jiraHost").isString(),
 	returnOnValidationError,
 	ApiInstallationDelete
 );
 ApiRouter.delete(
-	"/deleteInstallation/:installationId/:jiraHost/github-app-id/:gitHubAppId",
+	"/deleteInstallation/:installationId/:jiraHost",
 	param("installationId").isInt(),
 	param("jiraHost").isString(),
 	returnOnValidationError,
--- a/src/routes/api/api-router.ts
+++ b/src/routes/api/api-router.ts
@@ -91,6 +91,13 @@ ApiRouter.delete(
 	returnOnValidationError,
 	ApiInstallationDelete
 );
+ApiRouter.delete(
+	"/deleteInstallation/:installationId/:jiraHost/github-app-id/:gitHubAppId",
+	param("installationId").isInt(),
+	param("jiraHost").isString(),
+	returnOnValidationError,
+	ApiInstallationDelete
+);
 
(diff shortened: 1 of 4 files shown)


src/routes/api/api-router.ts
```diff
@@ -91,6 +91,13 @@ ApiRouter.delete(
 	returnOnValidationError,
 	ApiInstallationDelete
 );
+ApiRouter.delete(
+	"/deleteInstallation/:installationId/:jiraHost/github-app-id/:gitHubAppId",
```
joshkay10, line 95: /nit naming - do we need to call it `deleteInstallation` when the method is delete? so have `/installation`??
🤷

gxueatlassian: If you expand the lines above a bit, you will see actually this is a duplicate route of an exisiting route (so that we can have the gitHubAppId)

Alternatively, I can modify the existing route, so that it reads
`"/deleteInstallation/:installationId/:jiraHost/(:gitHubAppId)?"`

But that routes is like `/:xxx/:xxx/:xxx`, doesn't looks very nice to me.

src/routes/api/api-router.ts
```diff
@@ -91,6 +91,13 @@ ApiRouter.delete(
...
+	param("installationId").isInt(),
+	param("jiraHost").isString(),
```
joshkay10, line 97: comment on pre-existing code not this one lollol

since the jirahost includes the protocol do we assume it's been encoded? the delete method appears to just take it as is? I feel like i missed a piece but putting it here just in case.

src/routes/api/api-router.ts
gxueatlassian: And I think the route the bottom `ApiRouter.use("/:installationId", param("installationId").isInt(), returnOnValidationError, ApiInstallationRouter);` is more "prefered".

Although I seriously don't like it, as the route is like `DELETE /api/<any_int>/<any_string>`, which can easily match to any other route if not careful.

src/routes/api/api-router.ts
gxueatlassian: Yup good point 👍 
I think "it is working", if you check the test added in the PR, we are assuming that the `jiraHost` is encoded. And also assuming express decoded it automatically.

ps: and since it is pre-exsit code, working fine with pollinator, so we have a bit confidence.

src/routes/api/api-router.ts
gxueatlassian: Sorry @joshkay10, actually I am closing this one, as the existing route at the bottom of the file is the one supposed to be called. I'll fix pollinator anyway.

gxueatlassian:
Have to reopen this as otherwise the poco policy will be like 
```
	"paths": [
		"/api/*/*" <-- pretty sucks and dangerous.
	],
```

cc @joshkay10

src/routes/api/api-router.ts
joshkay10: yeah agree

src/routes/api/api-router.ts
joshkay10: yea good call better to leave as is at this point!!

PR #1782, opened 2022-12-07
ARC 1850 - Set the app property as soon as the app is installed

krazziekay:
**What's in this PR?**
- Moved the API call for setting app properties from `jira-get` to `jira-events-enabled-post`.
- Added `enabled` & `disabled` lifecycle events in the connect-app-descriptor. 

**Why**
- So that the user's `isConfigured` app property is set to `false` as soon as they install the app, without having the users to go inside the app page.

**How has this been tested?**  
- Local

**Whats Next?**
- Merge & test in prod

### DIFF
--- a/src/routes/jira/events/jira-events-router.ts
+++ b/src/routes/jira/events/jira-events-router.ts
@@ -3,6 +3,7 @@ import { JiraEventsInstallPost } from "routes/jira/events/jira-events-install-po
 import { extractInstallationFromJiraCallback } from "~/src/jira/extract-installation-from-jira-callback";
 import { JiraEventsUninstallPost } from "routes/jira/events/jira-events-uninstall-post";
 import { verifyAsymmetricJwtTokenMiddleware } from "~/src/jira/util/jwt";
+import { saveConfiguredAppProperties } from "utils/app-properties-utils";
 
 export const JiraEventsRouter = Router();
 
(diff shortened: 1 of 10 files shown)


src/routes/jira/events/jira-events-router.ts
```diff
@@ -3,14 +3,24 @@ import { JiraEventsInstallPost } from "routes/jira/events/jira-events-install-po
...
+import { saveConfiguredAppProperties } from "utils/app-properties-utils";
 
 export const JiraEventsRouter = Router();
 
 // TODO: remove enabled and disabled events once the descriptor is updated in marketplace
 JiraEventsRouter.post("/disabled", (_: Request, res: Response) => {
 	return res.sendStatus(204);
 });
-JiraEventsRouter.post("/enabled", (_: Request, res: Response) => {
+JiraEventsRouter.post("/enabled", async (req: Request, res: Response) => {
```
gxueatlassian, line 14: Seems the enabled is triggered multiple times. https://developer.atlassian.com/cloud/jira/platform/connect-app-descriptor/#lifecycle-attribute-example

image

gxueatlassian: Then it will "reset" the config status even after the config properties is set to true.

krazziekay: We can make another API call to simply check if the app property is already set or not. If its already set, do nothing, and if its not set, then we set it to this default `false`.

gxueatlassian: Sounds good.

gxueatlassian (approved):

PR #1784, opened 2022-12-08
Hot 101459

joshkay10:
**What's in this PR?**
Pulling the JiraHost from the query param when we dont have a toke to use

**Why**
We need a Jirahost to be able to make calls and retrieve data.

We are actively working on converting the jirahost in the query to a signed JWT token so this is a temporary solution while we do that.


src/routes/github/github-router.ts
```diff
@@ -51,10 +68,8 @@ subRouter.use("/configuration", jiraSymmetricJwtMiddleware, GithubServerAppMiddl
 // TODO: remove optional "s" once we change the frontend to use the proper delete method
 subRouter.use("/subscriptions?", jiraSymmetricJwtMiddleware, GithubServerAppMiddleware, GithubAuthMiddleware, GithubSubscriptionRouter);
 
-subRouter.use("/create-branch", GithubServerAppMiddleware, GithubAuthMiddleware, GithubCreateBranchRouter);
+subRouter.use("/create-branch", JiraHostFromQueryParamMiddleware, GithubServerAppMiddleware, GithubAuthMiddleware, GithubCreateBranchRouter);
 
 subRouter.use("/repository", jiraSymmetricJwtMiddleware, GithubServerAppMiddleware, GithubAuthMiddleware, GithubRepositoryRouter);
 
 subRouter.use("/branch", GithubServerAppMiddleware, GithubAuthMiddleware, GithubBranchRouter);
```
joshkay10, line 75: I have tested this route(/branch), it only gets called from inside the create-branch page, but at that point the new changes have the jirahost saved to it still works fine.

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -11,21 +11,15 @@ const MAX_REPOS_RETURNED = 20;
...
+		jiraHost
 	} = res.locals;
-	const jiraHost = req.query?.jiraHost as string;
 
 	if (!githubToken) {
 		req.log.warn(Errors.MISSING_GITHUB_TOKEN);
 		return next(new Error(Errors.MISSING_GITHUB_TOKEN));
 	}
 
-	if (!jiraHost) {
```
joshkay10, line 23: This is being handled in the new temp middleware function, so just removing it since its duplicated

src/routes/github/github-router.ts
```diff
@@ -14,6 +14,23 @@ import { GithubCreateBranchRouter } from "routes/github/create-branch/github-cre
...
+const JiraHostFromQueryParamMiddleware = async (req: Request, res: Response, next: NextFunction) => {
+	const jiraHostQuery = req.query?.jiraHost as string;
+	const jiraHostSession = req.session?.jiraHost as string;
+	const jiraHost = jiraHostQuery || jiraHostSession;
+	if (!jiraHost) {
+		req.log.warn(Errors.MISSING_JIRA_HOST);
+		res.status(400).send(Errors.MISSING_JIRA_HOST);
+		return;
+	}
+	res.locals.jiraHost = jiraHost;
```
gxueatlassian, line 30: Isn't this a security hole that `jiraHost` can comes from query string?

src/routes/github/github-router.ts
```diff
@@ -33,7 +50,7 @@ subRouter.post("/webhooks",
 // subRouter.use(GithubServerAppMiddleware);
 
 // OAuth Routes
-subRouter.use(jiraSymmetricJwtMiddleware, GithubServerAppMiddleware, GithubOAuthRouter);
```
gxueatlassian, line 36: I understand to remove the `jiraSymmetricJwtMidleware`, but why remove the `GithubServerAppMiddleware` here? Even it is added inside the router?

src/routes/github/create-branch/github-create-branch-get.ts
gxueatlassian: /nit
But it doesn't harm to have it here, as it can be #fail-fast ? If somehow jiraHost is null and it slip through to bellow `await Subscription.getAllForHost(jiraHost....`, we want it to fail earlier?

src/routes/github/github-router.ts
joshkay10: yes, that why we are urgently getting the JWT to come from JIRA
for the time being since we use a combination of this and the github token the damage is contained.

We only allow access to things that the github token can get.... but yes needs to be addressed asap and will be the top priority out of PIR

src/routes/github/github-router.ts
joshkay10: Good point, it was originally pulled from the first idea but then didnt put it back afterwards good catch!!

src/routes/github/github-router.ts
gxueatlassian: But because we are setting `jiraHost` into `session`, malicious user can call this endpoint once, then go to any other endpoints they want with unverified jiraHost in the session. Impact will be outside of just create-branch.

Not saying we have an imminent security hole, but if we are not careful and somewhere trust that jiraHost in session cookie too much, it may compromise data.

src/routes/github/github-router.ts
gxueatlassian: LOL, I had a chat with @Harminder84 , we thought it needed to be pulled out because if we leave it here, all routes after this will still run the `GitHubServerAppMiddlware` ? If that's the case, then yes we should pull it out from here and put it into sub router.

ps: also, seems `jiraSymmetricJwtMiddleware` is forgotten to be put inside the `GithubOauthRouter`

src/routes/github/create-branch/github-create-branch-get.ts
joshkay10: mmmm ill bring it back in, it will check jirahost from locals not query but you are right its still worth checking.

src/routes/github/github-router.ts
joshkay10: Definitely, so we will patch this as high priority! and put on the actual solution!!!

gxueatlassian (dismissed):

src/routes/github/github-oauth-router.ts
```diff
@@ -208,5 +209,5 @@ const getCloudOrGHESAppClientSecret = async (gitHubAppConfig, jiraHost: string)
 // IMPORTANT: We need to keep the login/callback/middleware functions
 // in the same file as they reference each other
 export const GithubOAuthRouter = Router();
-GithubOAuthRouter.get("/login", GithubOAuthLoginGet);
-GithubOAuthRouter.get(callbackSubPath, GithubOAuthCallbackGet);
+GithubOAuthRouter.get("/login", GithubServerAppMiddleware, jiraSymmetricJwtMiddleware, GithubOAuthLoginGet);
```
gxueatlassian, line 212: I think we should swap the order by `jiraSymmetricJwtMiddleware` and `GithubServerAppMiddleware`

src/routes/github/github-router.ts
joshkay10: added it to the /login path but cannot on callback path without casing infinite loops

src/routes/github/github-oauth-router.ts
```diff
@@ -208,5 +209,5 @@ const getCloudOrGHESAppClientSecret = async (gitHubAppConfig, jiraHost: string)
 // IMPORTANT: We need to keep the login/callback/middleware functions
 // in the same file as they reference each other
 export const GithubOAuthRouter = Router();
-GithubOAuthRouter.get("/login", GithubOAuthLoginGet);
-GithubOAuthRouter.get(callbackSubPath, GithubOAuthCallbackGet);
+GithubOAuthRouter.get("/login", jiraSymmetricJwtMiddleware, GithubServerAppMiddleware, GithubOAuthLoginGet);
```
gxueatlassian, line 212: Remove `jiraSymmetricJwtMiddleware` and `GithubServerAppMiddleware`?

src/routes/github/github-router.ts
```diff
@@ -25,36 +39,35 @@ subRouter.post("/webhooks",
...
-//Have an cover all middleware to extract the optional gitHubAppId
-//subRouter.use(param("uuid").isUUID('all'), GithubServerAppMiddleware);
+// Create-branch is seperated above since it currently relies on query param to extract the jirahost
+subRouter.use("/create-branch", JiraHostFromQueryParamMiddleware, GithubServerAppMiddleware, GithubAuthMiddleware, csrfMiddleware, GithubCreateBranchRouter);
 
-// TODO: use jiraSymmetricJwtMiddleware and GithubServerAppMiddleware for all paths below once "create branch" supports JWT
-// subRouter.use(jiraSymmetricJwtMiddleware);
-// subRouter.use(GithubServerAppMiddleware);
+// OAuth Routes
+subRouter.use(GithubOAuthRouter);
```
gxueatlassian, line 46: Remove this line? It is duplicated.

src/routes/github/github-router.ts
```diff
@@ -25,36 +39,35 @@ subRouter.post("/webhooks",
...
-subRouter.use(jiraSymmetricJwtMiddleware, GithubServerAppMiddleware, GithubOAuthRouter);
+subRouter.use(GithubOAuthRouter);
 
 // CSRF Protection Middleware for all following routes
 subRouter.use(csrfMiddleware);
 
-subRouter.use("/setup", jiraSymmetricJwtMiddleware, GithubServerAppMiddleware, GithubSetupRouter);
+subRouter.use("/setup", GithubSetupRouter);
 
 // TODO: use GithubAuthMiddleware for all paths below once "create branch" supports JWT
```
gxueatlassian, line 59: Maybe remove this comment? Tbh I don't understand it.

src/routes/github/github-router.ts
```diff
@@ -25,36 +39,35 @@ subRouter.post("/webhooks",
...
 // subRouter.use(GithubAuthMiddleware);
 
 // App Manifest flow routes
-subRouter.use("/manifest", jiraSymmetricJwtMiddleware, GithubServerAppMiddleware, GithubManifestRouter);
+subRouter.use("/manifest", GithubManifestRouter);
 
-subRouter.use("/configuration", jiraSymmetricJwtMiddleware, GithubServerAppMiddleware, GithubAuthMiddleware, GithubConfigurationRouter);
+subRouter.use("/configuration", GithubAuthMiddleware, GithubConfigurationRouter);
```
gxueatlassian, line 65: **/nit**

So previously (before jwt middleware) the code is like 
```javascript
// All following routes need Github Auth
subRouter.use(GithubAuthMiddleware);

subRouter.use("/configuration", GithubConfigurationRouter);

// TODO: remove optional "s" once we change the frontend to use the proper delete method
subRouter.use("/subscriptions?", GithubSubscriptionRouter);

subRouter.use("/create-branch", GithubCreateBranchRouter);

subRouter.use("/repository", GithubRepositoryRouter);

subRouter.use("/branch", GithubBranchRouter);

```

Shall we mimic the same behaviour (which is the the `subRouter.use(GithubAuthMiddleware);` before the last couple routes?) <-- tbh I am not too sure about it. Personally I found it "safer" the extract `subRouter.use(GithubAuthMiddleware);`  into its own line. Burry into each routing is "less safe" to me.

src/routes/github/github-router.ts
joshkay10: ahhhhh this comment, snuck back in!  deleting

src/routes/github/github-oauth-router.ts
joshkay10: the old double dipping middleware, good catch!

src/routes/github/github-router.ts
joshkay10: Removed the other one, and will keep the middleware on the router directly since its a mixed bag of needing some

src/routes/github/github-oauth-router.ts
joshkay10: cancel that, this is what we want... for now.

the duplication of the router make it confusing, but this will be above the .use for the two middlewares in question

src/routes/github/github-router.ts
joshkay10: May as well finish the tidy while we are here lololol

src/routes/github/github-router.ts
```diff
@@ -48,26 +48,22 @@ subRouter.use(GithubOAuthRouter);
...
 subRouter.use("/setup", GithubSetupRouter);
 
-// TODO: use GithubAuthMiddleware for all paths below once "create branch" supports JWT
-// subRouter.use(GithubAuthMiddleware);
-
 // App Manifest flow routes
 subRouter.use("/manifest", GithubManifestRouter);
 
-subRouter.use("/configuration", GithubAuthMiddleware, GithubConfigurationRouter);
+subRouter.use(GithubServerAppMiddleware);
```
gxueatlassian, line 59: LOL, it is `GithubAuthMiddleware` instead of `GithubServerAppMiddleware`

joshkay10: oh dear

src/routes/github/github-oauth-router.ts
```diff
@@ -205,8 +206,45 @@ const getCloudOrGHESAppClientSecret = async (gitHubAppConfig, jiraHost: string)
 	return ghesApp.getDecryptedGitHubClientSecret(jiraHost);
 };
 
+const TempGetJiraHostFromStateMiddleware  = async (req: Request, res: Response, next: NextFunction) => {
```
gxueatlassian, line 209: At the moment, even for cloud, `jiraHost` is missing on oauth callback.
This is okay for cloud as cloud doesn't need it during callback.
But for GHE it needs the jiraHost to decrypt the gitHubApp to get secrets.

I am using req.session[state] as this supposed to be securely set during oauth get.

src/routes/github/github-router.test.ts
```diff
@@ -194,19 +194,6 @@ describe("GitHub router", () => {
 					expect.anything()
 				);
 			});
-			it("should not match route, return empty if uuid present but invalid", async ()=>{
```
gxueatlassian, line 197: Turns out this test doesn't make sense. If there's random-gibberish in the path, `GitHubServerAppMiddleware` will treat it as cloud coz uuid is undefined.

src/routes/github/github-oauth-router.ts
```diff
@@ -205,8 +206,45 @@ const getCloudOrGHESAppClientSecret = async (gitHubAppConfig, jiraHost: string)
...
+};
+
 // IMPORTANT: We need to keep the login/callback/middleware functions
 // in the same file as they reference each other
-export const GithubOAuthRouter = Router();
-GithubOAuthRouter.get("/login", GithubOAuthLoginGet);
-GithubOAuthRouter.get(callbackSubPath, GithubOAuthCallbackGet);
+export const GithubOAuthRouter = Router({ mergeParams: true });
+GithubOAuthRouter.get("/login", GithubServerAppMiddleware, GithubOAuthLoginGet);
+GithubOAuthRouter.get(callbackSubPath, TempGetJiraHostFromStateMiddleware, GithubServerAppMiddleware, GithubOAuthCallbackGet);
```
Harminder84, line 250: Only `create-branch` flow contains `jirahost` as a query parameter. Other flows which don't have `jiraHost` query parameter will fail bcz of `TempGetJiraHostFromStateMiddleware`

gxueatlassian: This callback originated from the login get, which supposed to have a jiraHost in the session[stateKey]. So I think it is always there?

gxueatlassian: You are right, let me pick the jiraHost from session if available

Harminder84: I doubt for GH cloud or enterpise flows. Have you tested it ?

gxueatlassian: Okay I fixed the bug.

src/routes/github/github-oauth-router.ts
```diff
@@ -205,8 +206,52 @@ const getCloudOrGHESAppClientSecret = async (gitHubAppConfig, jiraHost: string)
...
+	if (!req.session[stateKey]) {
+		req.log.warn("State is empty in req.session for callback redirect");
+		res.status(400).send(Errors.MISSING_JIRA_HOST);
+		return;
+	}
+
+	let url;
+	try {
+		let redirectUrl = req.session[stateKey];
+		if (!redirectUrl.startsWith("https")) {
```
Harminder84, line 226: Shouldn't  we include `http` in the check as well?

src/routes/github/github-oauth-router.ts
```diff
@@ -38,8 +39,8 @@ const GithubOAuthLoginGet = async (req: Request, res: Response): Promise<void> =
 
 	// Save the redirect that may have been specified earlier into session to be retrieved later
 	req.session[state] =
-		res.locals.redirect ||
-		`/github/configuration${url.parse(req.originalUrl).search || ""}`;
+		appendJiraHostIfNeeded(res.locals.redirect ||
+		`/github/configuration${url.parse(req.originalUrl).search || ""}`, res.locals.jiraHost);
```
Harminder84, line 43: Instead of appending `jiraHost`, I think it would be much cleaner if we get `jirahost` from query parameter or session in `TempGetJiraHostFromStateMiddleware`

joshkay10: will jirahost exist in locals at this point since its a new request?

@Harminder84 - Session would work but I think we were trying to avoid keeping a unsafe jirahost in the session. 

We may want to add a conditional that we only adding this as a param for the damn create-branch route.


Alternatively we could put it in the session then rip it out again before we redirect to create-branch route, but sounds messy just typing it.

gxueatlassian: So this state is stored in session, and for create-branch, it get the jiraHost from query (unsafe), and for normal flow, it get it from jiraHostMiddleware (safe)

The line here is to save the `state` ( redirect url that contains jiraHost ) in session before OAuth callback. The state is empty here, so we can't get it from state.

gxueatlassian: Also state get removed after callback. So only unsafe part is Jira sending jiraHost to our app in query string during create-branch. That's our agreement for unsafe part right?

src/routes/github/github-oauth-router.ts
```diff
@@ -205,8 +206,52 @@ const getCloudOrGHESAppClientSecret = async (gitHubAppConfig, jiraHost: string)
...
+	res.locals.jiraHost = jiraHost;
+	next();
+};
+
+const appendJiraHostIfNeeded = (url: string, jiraHost: string): string => {
+	if (!jiraHost) return url;
+	if (url.indexOf("?") === -1) url = url + "?";
+	if (url.indexOf("&jiraHost") === -1) url = url + "&jiraHost=" + jiraHost;
+	return url;
+};
```
Harminder84, line 251: URL with no query paramtere result something like

`http://somejiraulr.com?&jiraHost=abc.net`

Harminder84 (approved):

PR #1789, opened 2022-12-12
Add missing track event source and type it properly

gxueatlassian:
**What's in this PR?**
Add missing source for the trackEvent.

**Why**
Otherwise it will error in prod env saying `Cannot sendAnalytics event: Error: Value trackEvent.source cannot be undefined`

**Added feature flags**
N/A

**Affected issues**  
ARC-1857

**How has this been tested?**  
N/A

**Whats Next?**
N/A

### DIFF
--- a/src/interfaces/common.ts
+++ b/src/interfaces/common.ts
@@ -48,7 +48,7 @@ export enum AnalyticsTrackEventsEnum {
 }
 
 export enum AnalyticsTrackSource {
-	Default = "default",
+	Cloud = "cloud",
 	GitHubEnterprise = "gitHubEnterprise",
 	CreateBranch = "createBranch"
 }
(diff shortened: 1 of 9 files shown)


src/interfaces/common.ts
```diff
@@ -48,6 +48,7 @@ export enum AnalyticsTrackEventsEnum {
 }
 
 export enum AnalyticsTrackSource {
+	Default = "default",
```
Harminder84, line 51: why we need `default` tracks source?

gxueatlassian: Well, from the splunk log it seems we always need a source for TrackEvent. 
Some of this event can be for cloud or ghe, not sure what to put. So `Default` comes to my mind.

Harminder84: how about `gitHubCloud` instead of `default`?

gxueatlassian: Good point, I've updated to use cloud now.

Harminder84 (approved):

PR #1801, opened 2022-12-19
skip unretryable error statuss

joshkay10:
We dont want to retry failed requests that we expect to fail again.


src/sqs/sqs.ts
```diff
@@ -299,7 +300,7 @@ export class SqsQueue<MessagePayload> {
 			if (isNotAFailure(errorHandlingResult)) {
 				context.log.info("Deleting the message because the error is not a failure");
 				await this.deleteMessage(context);
-			} else if (isNotRetryable(errorHandlingResult)) {
+			} else if (isNotRetryable(errorHandlingResult) || UNRETRYABLE_STATUS_CODES.includes(err.status)) {
```
gxueatlassian, line 303: I see `src/sqs/error-handlers.ts` func `maybeHandleNonRetryableResponseCode` does a similar job. Can you please explain why isn't that sufficient but we have to do it here?

PR #1809, opened 2022-12-21
ARC-1877 convert issue summary to hyphens when creating branch name

joshkay10:
**What's in this PR?**
Convert special characters to - when generating a branchname for create branch

**Why**
Things like / makes the default branchname contain a hierarchy on GitHub which we want to avoid, this is also the default behaviour for BitBucket.

Plus its just generally a whole bunch tidier and GitHub will appreciate the simplicity


src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -77,6 +79,18 @@ export const GithubCreateBranchGet = async (req: Request, res: Response, next: N
 	});
 };
 
+export const generateBranchName = (issueKey: string, issueSummary: string) => {
+	if (!issueSummary) {
+		return `${issueKey}-`;
+	}
+
+	let branchSuffix = issueSummary;
```
joshkay10, line 87: I know this block is a bit unsophisticated lol, but its intention is clear and I didnt want to refactor things or us ,.apply etc not today

src/util/handlebars/handlebar-helpers.ts
```diff
@@ -4,6 +4,7 @@ import { isPlainObject } from "lodash";
 export const concatStringHelper = (...strings: string[]) => strings.filter((arg: unknown) => typeof arg !== "object").join(" ");
 export const toLowercaseHelper = (str?: string) => !isPlainObject(str) && str?.toString?.().toLowerCase() || "";
 export const replaceSpaceWithHyphenHelper = (str?: string) => !isPlainObject(str) && str?.toString?.().replace(/ /g, "-") || "";
+export const replaceSlashWithHyphenHelper = (str?: string) => !isPlainObject(str) && str?.toString?.().replace(/\//g, "-") || "";
```
mboudreau, line 7: instead of doing a `toString`, shouldn't we just check to see if it's actually a string and not do anything if it's not?  It's not the job of this function to convert things to a string.  I'm also not a fan of just returning an empty string and be like "everything's fine"

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -77,6 +79,18 @@ export const GithubCreateBranchGet = async (req: Request, res: Response, next: N
 	});
 };
 
+export const generateBranchName = (issueKey: string, issueSummary: string) => {
+	if (!issueSummary) {
+		return `${issueKey}-`;
```
krazziekay, line 84: Should we add `-` at the end of the branch name?

src/routes/github/create-branch/github-create-branch-get.test.ts
```diff
@@ -71,4 +72,13 @@ describe("GitHub Create Branch Get", () => {
 				});
 		});
 	});
+
+	describe("GenerateBranchName", () => {
```
atrigueiro, line 76: <3 love your work

src/routes/github/create-branch/github-create-branch-get.ts
joshkay10: poooof going

src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -77,6 +78,21 @@ export const GithubCreateBranchGet = async (req: Request, res: Response, next: N
 	});
 };
 
+export const generateBranchName = (issueKey: string, issueSummary: string) => {
+	if (!issueSummary) {
+		return issueKey;
+	}
+
+	let branchSuffix = issueSummary;
+	const validBranchCharactersRegex = /[^a-z0-9.\-_]/gi;
```
krazziekay, line 87: Every special character except for `.` , `_` and `-` gets replaced by a `-`. So if we have multiple special characters at the end, we're gonna end up with multiple `-` at the end, which may look weird for a branch name.
`generateBranchName("ISSUEKEY", "/first-task-by-me-:)")` will return `ISSUEKEY-first-task-by-me--`

joshkay10: added a function to clean it up so only one - at a time, also added a test.
Did it manually because our node version doesn't support the new replaceAll function and didn't want to pull on that thread and start updating things as part of this PR

krazziekay (approved):

PR #1817, opened 2023-01-03
ARC 1886 - [JST-827001] - Fixing Regex to support alphanumeric issue keys

krazziekay:
**What's in this PR?**
- Updating the Regex for `addJiraIssueLinks` to match with the Regex being used for `jiraIssueKeyParser`

**Why**
- The current Regex is outdated and does not support alphanumeric issue keys like `K1-1`.
- Raised up in a DOS ticket - [JST-827001](https://getsupport.atlassian.com/browse/JST-827001)

**How has this been tested?**  
- Added Test case
- Locally

**Whats Next?**
- Merge and notify the customer


atrigueiro (approved):
Nice find!

PR #1823, opened 2023-01-06
add block to allow microscope to replay the DLQ

rachellerathbone:
**What's in this PR?**
A new block in our POCO config to give Microscope permission to replay queues.

**Why**
Our deployments queue has been over the threshold of 5k since late November. To rectify this, we need to replay the queue but it's currently being blocked by POCO.

**Whats Next?**

- Replay the deployment DLQ
- Write a runbook for DLQ alerts for GH4J as we currently don't have one


atrigueiro:
PS: We should also give Saiyans a head up before replaying those since it's a high number.

etc/poco/bundle/main-test.json
```diff
@@ -73,12 +73,64 @@
 			"allowed": false
 		},
 		{
-			"name": "Allow service team members to access Admin API endpoints",
```
atrigueiro, line 76: hmm why are we removing this though?

etc/poco/bundle/main-test.json
```diff
@@ -79,6 +79,66 @@
 			"mechanism": "group",
 			"principals": ["micros-sv--github-for-jira-dl-admins"],
 			"allowed": true
+		},
+		{
+			"name": "allow Microscope to call admin dlq endpoint",
+			"path": "/api/internal/admin/dlq",
```
atrigueiro, line 85: We don't have an implementation for this endpoint, do we? For the DSS services, the panels are the same but each service has its own implementation for replaying messages (i.e. for [CP](https://bitbucket.org/atlassian/dss-commit-processor/src/master/src/main/java/cloud/atlassian/fusion/dss/commitprocessor/rest/internal/admin/QueueResource.java)). From my understanding, we never really implemented that for GH and the panels are there just by default?

atrigueiro: I could be wrong though but worth double checking

PR #1834, opened 2023-01-12
added extra log for request to GHE base url on connect

joshkay10:
**What's in this PR?**
Added a simple log to check response from GHE touch event, convert a debug to a info because its useful in prod


krazziekay (approved):

joshkay10:
Going to do this differently

PR #1841, opened 2023-01-17
ARC-1916 - listen for created repos webhook

joshkay10:
**What's in this PR?**
Started listening for created event from GitHub webhooks and then updating our database with the new entry

**Why**
Because the reposyncstate table becomes stale as soon as installation is complete, maintaining an up to date collection of repos connected opens up simpler implementations for other features(like create branch) since we dont need to rely on the GitHub API.

**Added feature flags**
repo-created-event

**Affected issues**  
ARC-1916

**How has this been tested?**  
Tested locally and on stage, added a new unit test to make sure it behaves as expected

**Whats Next?**
Creating a UI view of all connected repos available from the app config


joshkay10:
Has been moved to a different PR

PR #1850, opened 2023-01-22
ARC-1944 - Reduce GitHub Commits requests on Push webhook

joshkay10:
**What's in this PR?**
The payload for Push sometime has an excessive amount of shas in it for historical commits, I've seen hundreds often, This is a problem because we iterate through them all and make indvidual requests per commit, this is causing the rate limit to be gobbled up very fast for some customers.

We should hopefully see the maximum message age decrease from 5 hours down to something normal!


rachellerathbone (approved):

atrigueiro (approved):
Nice find, Josh!

PR #1852, opened 2023-01-24
ARC-1945: Get reviewers email before sending PR review to Jira.

atrigueiro:
**What's in this PR?**
Adding a `getGithubUser` call after we map the reviews because the reviewers' data we get on the webhook payload doesn’t contain the email for that user.

**Why**
So PR Reviewers can be matched to their Atlassian account IDs.

**Affected issues**  
[ARC-1945]

**How has this been tested?**  
Locally - I'll test in staging now.

[ARC-1945]: https://softwareteams.atlassian.net/browse/ARC-1945

### DIFF
--- a/src/interfaces/jira.ts
+++ b/src/interfaces/jira.ts
@@ -121,10 +121,6 @@ export interface JiraAuthor {
 	url?: string;
 }
 
-export interface JiraReviewer extends JiraReview {
-	login: string;
-}
-
 export interface JiraReview extends JiraAuthor {
 	approvalStatus: string;
 }
--- a/src/interfaces/jira.ts
+++ b/src/interfaces/jira.ts
@@ -121,8 +121,11 @@ export interface JiraAuthor {
 	url?: string;
 }
 
-export interface JiraReview extends JiraAuthor {
+export interface JiraReviewer extends JiraReview {
 	login: string;
+}
+
+export interface JiraReview extends JiraAuthor {
 	approvalStatus: string;
 }
 
--- a/src/interfaces/jira.ts
+++ b/src/interfaces/jira.ts
@@ -115,18 +115,14 @@ export enum JiraCommitFileChangeTypeEnum {
 }
 
 export interface JiraAuthor {
-	avatar?: string;
 	email: string;
-	name: string;
+	name?: string;
+	avatar?: string;
 	url?: string;
 }
 
-export interface JiraReviewer {
+export interface JiraReview extends JiraAuthor {
 	login: string;
-	email?: string;
-}
-
-export interface JiraReview extends JiraReviewer {
 	approvalStatus: string;
 }
 
--- a/src/interfaces/jira.ts
+++ b/src/interfaces/jira.ts
@@ -120,7 +120,13 @@ export interface JiraAuthor {
 	name: string;
 	url?: string;
 }
-export interface JiraReview extends JiraAuthor {
+
+export interface JiraReviewer {
+	login: string;
+	email?: string;
+}
+
+export interface JiraReview extends JiraReviewer {
 	approvalStatus: string;
 }
 
--- a/src/sync/pull-requests.test.ts
+++ b/src/sync/pull-requests.test.ts
@@ -94,7 +94,6 @@ describe("sync/pull-request", () => {
 										{
 											"avatar": "test-pull-request-reviewer-avatar",
 											"name": "test-pull-request-reviewer-login",
-											"login": "test-pull-request-reviewer-login",
 											"email": "test-pull-request-reviewer-login@email.test",
 											"url": "https://github.com/reviewer",
 											"approvalStatus": "APPROVED"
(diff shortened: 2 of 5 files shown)


src/interfaces/jira.ts
```diff
@@ -120,7 +120,13 @@ export interface JiraAuthor {
 	name: string;
 	url?: string;
 }
-export interface JiraReview extends JiraAuthor {
+
+export interface JiraReviewer {
```
atrigueiro, line 124: Adding a new interface here. We were using the same JiraAuthor before but `name`, `url` and `avatar` fields have been deprecated on the devinfo API
Screenshot 2023-01-24 at 1 27 02 pm

src/transforms/transform-pull-request.ts
```diff
@@ -19,29 +19,47 @@ const mapStatus = (status: string, merged_at?: string) => {
...
 		) {
 			usernames[author?.login].approvalStatus = "APPROVED";
 		}
+
 		// Returns the reviews' array with unique users
 		return acc;
 	}, []);
+
+	for (const reviewer of reviewsReduced) {
+		const gitHubUser = await getGithubUser(gitHubInstallationClient, reviewer.login);
```
atrigueiro, line 54: Doing this call after the `reduce` so we only need to call once per user.

src/interfaces/jira.ts
```diff
@@ -115,12 +115,14 @@ export enum JiraCommitFileChangeTypeEnum {
...
-	avatar?: string;
 	email: string;
-	name: string;
+	name?: string;
+	avatar?: string;
 	url?: string;
 }
+
 export interface JiraReview extends JiraAuthor {
+	login: string;
```
mboudreau, line 125: This shouldn't be in this type. `login` is not required by the API and the only place that uses it is for an implementation detail to reduce the users by login first before getting the extra information.

Instead, I'd recommend you add a custom type (extending the base type) nearing to the implementation that uses it.

src/transforms/transform-pull-request.ts
```diff
@@ -19,29 +19,44 @@ const mapStatus = (status: string, merged_at?: string) => {
...
+
 		// Returns the reviews' array with unique users
 		return acc;
 	}, []);
+
+	return Promise.all(reviewsReduced.map(async reviewer => {
+		const gitHubUser = await getGithubUser(gitHubInstallationClient, reviewer.login);
+		return {
+			...reviewer,
+			email: gitHubUser && gitHubUser.email ?  gitHubUser.email : `${reviewer.login}@noreply.user.github.com`
```
mboudreau, line 57: could be simplified to `gitHubUser?.email || `${reviewer.login}@noreply.user.github.com``

src/transforms/transform-pull-request.ts
mboudreau: much nicer instead of the for loop :)

src/sync/pull-requests.test.ts
```diff
@@ -253,6 +255,13 @@ describe("sync/pull-request", () => {
 					.reply(200, pullRequest)
 					.get("/repos/integrations/test-repo-name/pulls/51/reviews")
 					.reply(200, reviewsPayload)
+					.get("/users/test-pull-request-reviewer-login")
+					.reply(200, {
+						login: "test-pull-request-reviewer-login",
+						avatar_url: "test-pull-request-reviewer-avatar",
+						html_url: "test-pull-request-reviewer-url",
+						email: "test-pull-request-reviewer-login@gmail.com"
```
mboudreau, line 263: Please don't use `gmail.com` in any tests.  I've had issues in the past where test emails were sent to a real gmail address.  

Not that this would happen here since we're not sending email, but more of a piece of mind kind of thing.  please use something like `email.test` as `.test` is a reserved TLD that can't be used :)

src/interfaces/jira.ts
mboudreau: The type definition is still here and public, but isn't needed anywhere else but in the transform-pull-request file.  It also causes confusion between this and `JiraReview`

src/transforms/transform-pull-request.ts
```diff
@@ -18,30 +18,55 @@ const mapStatus = (status: string, merged_at?: string) => {
...
 		// Returns the reviews' array with unique users
 		return acc;
 	}, []);
+
+	// Get GitHub user email, so it can be matched to an AAID
+	return Promise.all(reviewsReduced.map(async reviewer => {
+		let login;
+		if (reviewer.login) {
+			login = reviewer.login;
+			delete reviewer.login;
```
atrigueiro, line 62: Please let me know if there's a better way to do this, I've explored a number of options but would like your input :)

mboudreau: well, you could just do 
```
const gitHubUser = await getGithubUser(gitHubInstallationClient, login);
delete reviewer.login;
return {
			...reviewer,
			email: gitHubUser?.email || `${login}@noreply.user.github.com`
		};
```

atrigueiro: Not really because `reviewer.login` is optional and the signature of `getGithubUser` expects login. I don't think we want to change that to not be required?

mboudreau (approved):

PR #1853, opened 2023-01-24
remove all changedFiles query

mboudreau:
**What's in this PR?**
removing problematic graphql query property

**Why**
Because github can't create a consistent API


joshkay10 (approved):

gxueatlassian:
Close it for now, it is already tracked under https://softwareteams.atlassian.net/browse/ARC-2132
Can re-open it once picked up.

PR #1856, opened 2023-01-25
ARC-1948 - removed debug import

joshkay10:
**What's in this PR?**
Removed the not debug filter on unsafe log stream

**Why**
The logs are awful enough without restricting them further

****NOTES !!!!!!!!!!!!!!!!****
While this will stop the hashing of values on the unsafe index we have still omitted huge amounts of data that used to be logged, e.g. err or payload objects. so it will be more useful but logging in general needs a uplift.

If you want to only go to unsafe logs you can create a unsafe logger and throw much trash in there.

*Next*
I will put together a quick guide on how to do normy logger unsafe logger and some general guidance on what level to use and when as it feels a bit loose at the moment.


src/util/logger-utils.ts
```diff
@@ -72,17 +71,8 @@ export class SafeRawLogStream extends RawLogStream {
...
-		return !record.level || isNaN(record.level) || record.level > DEBUG;
-	};
-
 	public async _write(record: ChunkData, encoding: BufferEncoding, next: Callback): Promise<void> {
-		// Skip any log above DEBUG level that isn't tagged unsafe
-		if (this.isAboveDebug(record) && !record.unsafe) {
-			return next();
-		}
-		// Tag the record do it gets indexed to the _unsafe logging environment
+		// Tag the record do it gets indexed to the [env]_unsafe logging environment
```
rachellerathbone, line 75: nit: typo -> do

rachellerathbone (dismissed):

rachellerathbone (approved):

PR #1857, opened 2023-01-26
ARC-1868

joshkay10:
**What's in this PR?**
Adding {jwt} to the create branch action since Jira will now turn this into a JWT

**Why**
Everything else uses a JWT so why not create branch, we can also validate this value so less likely people tamper with things

**Whats Next?**
Start using the token!

### DIFF
--- a/src/routes/github/create-branch/github-create-branch-options-get.ts
+++ b/src/routes/github/create-branch/github-create-branch-options-get.ts
@@ -71,8 +71,7 @@ const getJiraHostFromTenantUrl = (jiraHostParam): string | undefined =>  {
 	if (!jiraHostParam) {
 		return undefined;
 	}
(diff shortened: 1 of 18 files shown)


src/routes/github/create-branch/github-create-branch-options-get.ts
```diff
@@ -71,8 +71,7 @@ const getJiraHostFromTenantUrl = (jiraHostParam): string | undefined =>  {
 	if (!jiraHostParam) {
 		return undefined;
 	}
-	const siteName = jiraHostParam?.substring(0, jiraHostParam?.indexOf("."));
-	return `https://${siteName}.atlassian.net`;
```
joshkay10, line 75: The correct domain is coming from this variable, we need to keep what it was not convert to .atlassian.net

Harminder84 (approved):

PR #1861, opened 2023-01-30
NONE: improve logging for cryptor; retry when cryptor is not there

bgvozdev:
**What's in this PR?**
1. use existing loggers for logging to stitch logs together
2. retry cryptor errors during backfill

**Why**
Long backfills might face some intermittent cryptor errors and we don't want this to break the backfilling

**Added feature flags**
🤠 

**Affected issues**  
CES-10000

**How has this been tested?**  
🤠 

**Whats Next?**
🤠

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -188,14 +188,6 @@ const doProcessInstallation = async (data: BackfillMessagePayload, sentry: Hub, 
 		return;
 	}
 
-	const jiraClient = await getJiraClient(
-		subscription.jiraHost,
-		gitHubInstallationId,
-		data.gitHubAppConfig?.gitHubAppId,
-		logger
-	);
-
-	const gitHubInstallationClient = await createInstallationClient(gitHubInstallationId, jiraHost, logger, data.gitHubAppConfig?.gitHubAppId);
 	const nextTask = await getNextTask(subscription, data.targetTasks);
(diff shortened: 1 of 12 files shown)


src/sync/installation.ts
```diff
@@ -351,13 +351,17 @@ export const handleBackfillError = async (
...
 
 	if (String(err).includes("connect ETIMEDOUT")) {
-		// There was a network connection issue.
-		// Add the job back to the queue with a 5 second delay
 		logger.warn("ETIMEDOUT error, retrying in 5 seconds");
 		scheduleNextTask(5_000);
 		return;
 	}
 
+	if (String(err).includes("connect ECONNREFUSED")) {
```
gxueatlassian, line 359: This could be too broad, as GHE customer might also have this ECONNREFUSED if their firewall decide the block our connections.

Can we do a try/wrap in the `EncryptionClient` file to throw a more specific error, like `err.type === 'Encyprtion'`?

bgvozdev: TBH I don't want to leak knowledge all around the place.

If you check a few lines above, there's already "timeout" clause – you could argue that could be due to firewall rules, too. If we want to make it better, we'd need to change it as well, but I doubt that should be done on this level. This is a global error handler: could be a Jira error, a github error, a cryptor error, just any network error.

Please ping me in slack if you have a strong opinion or would like to discuss in more details :)

gxueatlassian (approved):

PR #1865, opened 2023-01-31
Arc 772 recover plain client key

gxueatlassian:
**What's in this PR?**
1. Update the plainClientKey if the installation record already exist.
2. Allow domain name like clientKey ( some jira instance has clientKey like that )

**Why**
So previously, I created this PR to recover the plain client key for existing installations. I only successfully recovered around half of them. The rest are mostly 404/suspend.

The purpose of this PR is to:
1. Fix one of the regex bug that extract clientKey for jira site. There're quite a few of jira site has that problem.
2. Allowing setting the plainClientKey when updating the installation table. 
    This could happen when the site get "updated". 
3. Also, I updated a description file, hoping this "update" will trigger a backfill even for those jira sites that is 404/suspend.
    Well, to be honest, I am not sure about it. Happy to update/close this PR if you disagree.

**Added feature flags**
N/A

**Affected issues**  
ARC-772

**How has this been tested?**  
Unit test.

**Whats Next?**

### DIFF
--- a/src/models/installation.ts
+++ b/src/models/installation.ts
@@ -3,6 +3,7 @@ import { Subscription } from "./subscription";
 import { getHashedKey, sequelize } from "models/sequelize";
 import { EncryptedModel } from "models/encrypted-model";
 import { EncryptionSecretKeyEnum } from "utils/encryption-client";
+import { getLogger } from "config/logger";
 
 export class Installation extends EncryptedModel {
 	id: number;
@@ -111,6 +112,9 @@ export class Installation extends EncryptedModel {
 	}
 }
 
+const LOGGER_HOOK_BEFORE_SAVE = getLogger("installation-hook-beforeSave");
+const LOGGER_HOOK_BEFORE_BULK_CREATE = getLogger("installation-hook-beforeBulkCreate");
+
 Installation.init({
 	id: {
 		type: DataTypes.INTEGER,
@@ -138,12 +142,12 @@ Installation.init({
 	hooks: {
 		beforeSave: async (instance: Installation, opts) => {
 			if (!opts.fields) return;
-			await instance.encryptChangedSecretFields(opts.fields);
+			await instance.encryptChangedSecretFields(opts.fields, LOGGER_HOOK_BEFORE_SAVE);
 		},
 		beforeBulkCreate: async (instances: Installation[], opts) => {
 			for (const instance of instances) {
 				if (!opts.fields) return;
-				await instance.encryptChangedSecretFields(opts.fields);
+				await instance.encryptChangedSecretFields(opts.fields, LOGGER_HOOK_BEFORE_BULK_CREATE);
 			}
 		}
 	},
--- a/src/models/installation.ts
+++ b/src/models/installation.ts
@@ -83,7 +83,8 @@ export class Installation extends EncryptedModel {
 			await installation
 				.update({
 					encryptedSharedSecret: payload.sharedSecret,
-					jiraHost: payload.host
+					jiraHost: payload.host,
+					plainClientKey: payload.clientKey
 				})
 				.then(async (record) => {
 					const subscriptions = await Subscription.getAllForClientKey(
--- a/src/models/installation.ts
+++ b/src/models/installation.ts
@@ -84,8 +84,7 @@ export class Installation extends EncryptedModel {
 			await installation
 				.update({
 					encryptedSharedSecret: payload.sharedSecret,
-					jiraHost: payload.host,
-					plainClientKey: payload.clientKey
+					jiraHost: payload.host
 				})
 				.then(async (record) => {
 					const subscriptions = await Subscription.getAllForClientKey(
--- a/src/models/installation.ts
+++ b/src/models/installation.ts
@@ -84,7 +84,8 @@ export class Installation extends EncryptedModel {
 			await installation
 				.update({
 					encryptedSharedSecret: payload.sharedSecret,
-					jiraHost: payload.host
+					jiraHost: payload.host,
+					plainClientKey: payload.clientKey
 				})
 				.then(async (record) => {
 					const subscriptions = await Subscription.getAllForClientKey(
--- a/src/routes/api/client-key/client-key-regex.test.ts
+++ b/src/routes/api/client-key/client-key-regex.test.ts
@@ -2,15 +2,31 @@ import { extractClientKey } from "./client-key-regex";
 
 const sampleUuid = "yyyyyyyy-0569-4f33-8413-cccccccc";
 const sampleJiraSite = "sample-site.atlassian.net";
+const xmlTemplate = `
+<consumer>
+<key>$$KEY$$</key>
+<name>JIRA</name>
+<publicKey>
+...SIb3DQEBAQUAA4GNADCBiQK...
+</publicKey>
+<description>Atlassian JIRA at url</description>
+</consumer>
+`;
+
+const getXml = (key: string) => {
+	return xmlTemplate.replace("$$KEY$$", key);
+};
+
 const possibleJiraXmlInfo = [
-	[`blah\n<key>jira:${sampleUuid}</key>\nblah`, `jira:${sampleUuid}`],
-	[`blah\n<key>${sampleUuid}</key>\nblah`, sampleUuid],
-	[`blah\n<key>${sampleJiraSite}</key>\nblah`, sampleJiraSite]
+	[getXml(`jira:${sampleUuid}`), `jira:${sampleUuid}`],
+	[getXml(`${sampleUuid}`), sampleUuid],
+	[getXml(`${sampleJiraSite}`), sampleJiraSite],
+	["random-string", undefined]
 ];
 
 describe("Extract client key from xml", ()=>{
 	it.each(possibleJiraXmlInfo)("should extract client key correctly", (xml, key)=>{
-		expect(extractClientKey(xml)).toBe(key);
+		expect(extractClientKey(xml as string)).toBe(key);
 	});
 });
 
--- a/src/routes/api/client-key/client-key-regex.test.ts
+++ b/src/routes/api/client-key/client-key-regex.test.ts
@@ -1,9 +1,11 @@
 import { extractClientKey } from "./client-key-regex";
 
 const sampleUuid = "yyyyyyyy-0569-4f33-8413-cccccccc";
+const sampleJiraSite = "sample-site.atlassian.net";
 const possibleJiraXmlInfo = [
 	[`blah\n<key>jira:${sampleUuid}</key>\nblah`, `jira:${sampleUuid}`],
-	[`blah\n<key>${sampleUuid}</key>\nblah`, sampleUuid]
+	[`blah\n<key>${sampleUuid}</key>\nblah`, sampleUuid],
+	[`blah\n<key>${sampleJiraSite}</key>\nblah`, sampleJiraSite]
 ];
 
 describe("Extract client key from xml", ()=>{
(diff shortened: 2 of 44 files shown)


src/models/installation.ts
```diff
@@ -84,7 +84,8 @@ export class Installation extends EncryptedModel {
 			await installation
 				.update({
 					encryptedSharedSecret: payload.sharedSecret,
-					jiraHost: payload.host
+					jiraHost: payload.host,
+					plainClientKey: payload.clientKey
```
gxueatlassian, line 88: Let me know whether setting the `plainClientKey` make sense here.

src/routes/api/client-key/client-key-regex.test.ts
```diff
@@ -1,9 +1,11 @@
 import { extractClientKey } from "./client-key-regex";
 
 const sampleUuid = "yyyyyyyy-0569-4f33-8413-cccccccc";
+const sampleJiraSite = "sample-site.atlassian.net";
```
gxueatlassian, line 4: Some of the jira site does have a client key as domain name.

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
```diff
@@ -176,7 +176,7 @@ export const JiraAtlassianConnectGet = async (_: Request, res: Response): Promis
 			"signed-install": true
 		},
 		name: APP_NAME,
-		description: "Connect your code and your project with ease.",
+		description: "Connect your code and your projects with ease.",
```
gxueatlassian, line 179: This is trying to trigger an update for ALL the jira sites. And according to the doc, it will [rotate a new shared secrets](https://developer.atlassian.com/cloud/jira/platform/connect-app-descriptor/#lifecycle).

I am hoping trigger an update for ALL jira site may help recover existing client key for the sites that's 404/unavailable/etc

But I am not sure. Need some advices from you.

ps: Although I don't think there should be any harm though.

src/models/installation.ts
bgvozdev: Hmm, that's the site rename, right? Feels like we are solving 2 problems here: doing the update and populating the plainClientKey. I'd suggest not to do that:
 - that doesn't solve much of a problem cause site rename is a rare operation
 - we are making the code more complicated and breaking single responsibility principle.

However, not a strong opinion :)

src/routes/api/client-key/client-key-regex.test.ts
bgvozdev: :wat:

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
bgvozdev: Yeah, not sure either, but given we did that multiple times already that should be fine I guess. But would be nice to know for sure, maybe ping ecosystem, or just add some debug logging? To know for sure for the future...

bgvozdev (approved):

src/models/installation.ts
gxueatlassian: Totally agree, and "plainClientKey" is a primary-key-ish stuff, we shouldn't update it at all. Was thinking only temporary doing this for the life-cycle-update events.

But I've updated the code to update null plain client key in the install event itself. Can you please review?

src/routes/api/client-key/client-key-regex.test.ts
gxueatlassian: 🤷

bgvozdev (approved):

src/routes/api/client-key/client-key-regex.test.ts
mboudreau: I mean, you could add anything as a site name right? :P

My guess is some kind of import from server to cloud?

gxueatlassian: `import from server to cloud?` make sense, that could be the reason.
From app's point of view, it shouldn't matter, it is just a String. So I think it is okay.

src/routes/api/client-key/client-key-regex.ts
```diff
@@ -1,4 +1,4 @@
 export const extractClientKey = (text: string): string | undefined  => {
-	const [, plainClientKey] = /<key>([0-9a-z-:]+)<\/key>/gmi.exec(text) || [undefined, undefined];
+	const [, plainClientKey] = /<key>([0-9a-z-:.]+)<\/key>/gmi.exec(text) || [undefined, undefined];
```
mboudreau, line 2: I _really_ don't recommend we regex to parse XML.  It's one of the major pitfalls of regex and isn't recommended.

Instead, we should use the (Javascript built in XML parser)[https://developer.mozilla.org/en-US/docs/Web/API/DOMParser].

Furthermore, this utility isn't being used except in a test?  But also, I thought all our API/webhooks are JSON and not xml?

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
mboudreau: Pretty sure that connect automatically go through customers every few weeks.  you could ask the connect team about this, but remember when we were doing the DB enryption updates with cryptor?  Like half our customers had already moved after a couple weeks in prod.

I'm not sure you need to update the connect json for this update to trigger.

src/routes/api/client-key/client-key-regex.ts
gxueatlassian: It is being used [here](https://github.com/atlassian/github-for-jira/blob/fef8a9ebbbec338aaf537b27b91397808bd1357b/src/routes/api/client-key/recover-client-key.ts#L113) to extract the key in api recovery process.

For the DOMParser, I don't think node has it. For simplicity reason, I was origin thinking use a `String.indexOf("<key>")`, but settle on regex.

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
gxueatlassian: Yup, agree. Already ask the connect team and will update this.

src/routes/jira/events/jira-events-install-post.ts
```diff
@@ -10,12 +10,19 @@ export const JiraEventsInstallPost = async (req: Request, res: Response): Promis
...
 
 	const { baseUrl: host, clientKey, sharedSecret } = req.body;
-	await Installation.install({
+	const newOrExistingInstallation = await Installation.install({
 		host,
 		clientKey,
 		sharedSecret
 	});
 
+	if (!newOrExistingInstallation.plainClientKey) {
```
mboudreau, line 19: I'm not sure I understand, why isn't this part of the model itself?  This is model logic, it doesn't matter where the `install` call comes from, it should always add the `plainClientKey` to the DB...

mboudreau (changes requested):
Need to incude client key changes in model

src/routes/api/client-key/client-key-regex.ts
mboudreau: Ah, I'd still use an XML parser and not a regex for this.  xml2json or something so you can just do `data.consumer.key`

src/routes/jira/events/jira-events-install-post.ts
mboudreau: just add it to the install.  Better to have the logic centralized to the model instead of bleeding out in the webhook logic.  Like for instance, how I'm currently changing the webhooks...

src/routes/api/client-key/client-key-regex.ts
```diff
@@ -1,4 +1,9 @@
+import { toJson } from "xml2json";
 export const extractClientKey = (text: string): string | undefined  => {
-	const [, plainClientKey] = /<key>([0-9a-z-:]+)<\/key>/gmi.exec(text) || [undefined, undefined];
-	return plainClientKey;
+	try {
+		const json = toJson(text, { object: true });
+		return (json.consumer as { [key: string]: string }).key;
+	} catch (e) {
+		return undefined;
```
gxueatlassian, line 7: This is the usage.
```
	const plainClientKey = extractClientKey(text);
	if (!plainClientKey) {
		throw { msg: `Client key regex failed`, jiraHost, id, text };
	}
```

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
gxueatlassian: Confirm better make a change to trigger it.

src/models/installation.ts
gxueatlassian: Okay I've put it back @mboudreau :)

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
gxueatlassian: Get feedback from connect team. They can manually re-trigger the events for some/all sites for our app. So reverted the changes here.

src/routes/api/client-key/client-key-regex.ts
gxueatlassian: done.

src/routes/jira/events/jira-events-install-post.ts
```diff
@@ -7,16 +7,17 @@ import { metricHttpRequest } from "config/metric-names";
  * Handle the install webhook from Jira
  */
 export const JiraEventsInstallPost = async (req: Request, res: Response): Promise<void> => {
-	req.log.info("Received installation payload");
 
 	const { baseUrl: host, clientKey, sharedSecret } = req.body;
+	req.log.info({ jiraHost: host, clientKey },  "Received installation payload");
```
gxueatlassian, line 12: `clientKey` is public information, so didn't bother hash it. `jiraHost` is already hashed.

src/routes/api/client-key/client-key-regex.ts
```diff
@@ -1,4 +1,9 @@
+import { toJson } from "xml2json";
 export const extractClientKey = (text: string): string | undefined  => {
-	const [, plainClientKey] = /<key>([0-9a-z-:]+)<\/key>/gmi.exec(text) || [undefined, undefined];
-	return plainClientKey;
+	try {
+		const json = toJson(text, { object: true });
+		return (json.consumer as { [key: string]: string }).key;
```
mboudreau, line 5: Not sure why you need to set the type?  Doesn't the `toJson` just return `any`?

src/routes/api/client-key/client-key-regex.test.ts
```diff
@@ -1,14 +1,32 @@
...
+<publicKey>
+...SIb3DQEBAQUAA4GNADCBiQK...
+</publicKey>
+<description>Atlassian JIRA at url</description>
+</consumer>
+`;
+
+const getXml = (key: string) => {
+	return xmlTemplate.replace("$$KEY$$", key);
+};
```
mboudreau, line 18: why are you adding a string replace when typescript already has templating built-in?

gxueatlassian: Fixed.

src/routes/api/client-key/client-key-regex.ts
mboudreau: Oh, you're using `@types/xml2json`.  That definition is horrible and not genericized.  I'd just remove the dependency and just do `return json?.consumer?.key;`

gxueatlassian: Make sense. Fixed.

mboudreau (approved):

PR #1867, opened 2023-01-31
ARC-1950 - create branch search logs

joshkay10:
We have a few instances of people getting 401s on the search branch for create branch, this will help determine if its one more than the other, not going to solve the problem but will help understand it better hopefully!!


gxueatlassian (approved):
Should it be warn instead of error?

PR #1868, opened 2023-01-31
Updated tests in install in the setup

joshkay10:
Moved install into the global setup event so that the create-branch doesn't have to install before test functionality.

This was because it would fail all the time if it had to do that first, this way the test focuses only on what it needs to.... mostly


test/e2e/constants.ts
```diff
@@ -4,7 +4,7 @@ export const STATE_PATH = "./test/e2e/test-results/states";
 export const SCREENSHOT_PATH = "./test/e2e/test-results/screenshots";
 
 export const TEST_PROJECT_NAME = "E2E - Test - " + Date.now();
-export const TEST_PROJECT_KEY = "E" + Math.floor(Math.random()*9000) + 1000;
```
joshkay10, line 7: Javascript being javascript here and converting all that to string so it was always EXXXX1000

test/e2e/create-branch.e2e.ts
```diff
@@ -9,25 +9,31 @@ test.describe("Create branch", () => {
...
-			await jiraCreateIssue(page);
 			await page.goto(data.urls.testProjectIssue);
 			await (page.locator("a[data-testid='development-summary-common.ui.summary-item.link-formatted-button']")).click();
 			const poppedUpPage = await page.waitForEvent("popup");
 			await poppedUpPage.waitForLoadState();
 			expect(poppedUpPage.getByText("Almost there!")).toBeTruthy();
 		});
 	});
+	// Clean up projects to avoid conflicts in future
+	// test.afterEach(async() => {
```
joshkay10, line 34: couldn't quite get the project delete to work but left it there as Ill have another go at it!

test/e2e/create-branch.e2e.ts
```diff
@@ -9,25 +9,31 @@ test.describe("Create branch", () => {
...
 
 	test.use({
 		storageState: testData.jira.roles.admin.state
 	});
 
 	test.describe("cloud", () => {
+		// Create a fresh issue per test
 		test.beforeEach(async() => {
-			await jiraAppInstall(page);
+			await jiraCreateIssue(page);
```
krazziekay, line 22: What about create/add project? A new jira instance has no projects.

krazziekay (approved):

PR #1871, opened 2023-02-01
Jwt validation abstraction

mboudreau:
**What's in this PR?**
Abstracting out JWT logic so that it's not dependent on Express requests so we can use it with SQS webhook queues instead.

**Why**
JWT validation needs the Express `Request` object, which we obviously do not have if we're trying to validate a JWT coming from an SQS message.

**Added feature flags**
None

**Affected issues**  
_Jira Issues_

**How has this been tested?**  
Unit test, made sure it returns the same data as the `atlassian-jwt` we used before

**Whats Next?**
Merge into API gateway webhook branch so I can finish off my testing.

### DIFF
--- a/src/jira/util/jwt.ts
+++ b/src/jira/util/jwt.ts
@@ -1,11 +1,12 @@
 // Original source code:
 // https://bitbucket.org/atlassian/atlassian-connect-express/src/f434e5a9379a41213acf53b9c2689ce5eec55e21/lib/middleware/authentication.js?at=master&fileviewer=file-view-default#authentication.js-227
 // TODO: need some typing for jwt
-import { createQueryStringHash, decodeAsymmetric, decodeSymmetric, getAlgorithm, getKeyId } from "atlassian-jwt";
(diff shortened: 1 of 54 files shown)


src/jira/util/jwt.ts
```diff
@@ -211,41 +195,136 @@ const isStagingTenant = (req: Request): boolean => {
...
+const CANONICAL_QUERY_SEPARATOR = "&";
+
+enum HASH_ALGORITHM {
+	HS256 = "sha256",
+	HS384 = "sha384",
+	HS512 = "sha512",
+	RS256 = "RSA-SHA256"
+}
+
+export const createQueryStringHash = (req: JWTRequest, checkBodyForParams?: boolean): string => {
```
gxueatlassian, line 257: /nit
I'd suggest have this in a separate file and mention the origin file it comes from. So that next dev will easier to pick it up for figure out which team to looks for (if say they change the way `createCanonicalRequest` works in jira)

gxueatlassian (approved):
Overall,  GOOD STUFF!!! much cleaner on the error/throw part.

PR #1877, opened 2023-02-02
Arc 1422 add api to cleanup reposyncstate

gxueatlassian:
**What's in this PR?**
1. Add an admin api to remove ~270k stale repo sync state records.

**More on it**
It seems dangerous, but I think it should be okay as:
1. The code can take an id to limit how many do we delete.
3. The repo sync state is only being used 
      a. In backfill. If we "lose" them accidentally, user can alway restart the backfill.
      b. Tell how many repos are sync. So if we lost the data that actually still been used, **user won't see the correct sync number in the admin ui**. But webhook will still works as it should be. User will need to resync to get the sync number in the ui again. The impact is purely UI, so while not zero, but low.
4. Database can be restored up to 30 days with milliseconds granularity. 

**Bugs Fixed**
1. As for now, the code didn't delete the RepoSyncState when Subscriptions are removed. Hence the invalid rows in the table.
2. Add we will add fk constraint with CASCADE on delete, so above "bug" in the code won't be a "bug" anymore.

**Planning**
1. Also works on another PR (that add the foreign key constraint
3. Copy existing RepoSyncStates data (~600k rows)
4. Call cleanup api to remove ~270k rows that without valid subscription ids.
5. Call the DB migration to add foreign key constraint.

**Why**
1. We need to create foreign key constraint on that table. 

**Added feature flags**
N/A

**Affected issues**  
ARC-1422

**How has this been tested?**  
Unit test

**Whats Next?**


mboudreau:
Code looks fine, but I'm not sure what you're trying to do with the table copy?  Seems a bit odd to copy a table for 'backup' on a production system within the same DB...

gxueatlassian:
> Code looks fine, but I'm not sure what you're trying to do with the table copy? Seems a bit odd to copy a table for 'backup' on a production system within the same DB...

I don't know tbh. We suppose to have DB backup from micros, just not sure how often they are and the retention period and process to recover. Suspect too much hassle.

Copying it into a table (and we could remove it later on once we are happy ) seems least amount of effort to me.

Or we could just go ahead and clean up the data~ 

Thoughts/suggestions welcome :)

mboudreau:
@gxueatlassian maybe ask the data team to see what's the recommended practice?

gxueatlassian:
> @gxueatlassian maybe ask the data team to see what's the recommended practice?

Okay, got confirmation that:

```
...can restore your RDS to a given time and give you access, then you can then replay each row that needs to be restored via qes.  (we can restore down the millisecond for anytime in the last 30 days...
Yes, we'll restore it to a temporary RDS for you, open it up for VPN access and share creds with you. You'll then need to run queries to figure out what you need to run to restore your data on your real datastore
```

So I don't think we need to worry about backup. Can delete data without worrying.

bgvozdev (approved):

PR #1882, opened 2023-02-03
Use jwt for create branch

joshkay10:
Converted to draft while I update the Signature on the Jira side to be the same as the connect one

**Whats Next?**
Roll out the flag to 100% to avoid using the tenantUrl query param complety then remove all references to it and clean up logic a bit.

Then need to verify the token with a public key!


krazziekay (dismissed):

src/middleware/jirahost-middleware.ts
```diff
@@ -19,10 +19,21 @@ const extractUnsafeJiraHost = (req: Request): string | undefined => {
...
 	}
 	return undefined;
 };
 
+const getJiraHostFromJwtToken = (encodedToken: string) => {
+	try {
+		const decodedToken = JSON.parse(Buffer.from(encodedToken.split(".")[1], "base64").toString());
+		return decodedToken.tenant_url;
+	} catch (error) {
+		return null;
```
Harminder84, line 33: 1. logging missing
2. Shouldn't this throw an error?

Harminder84: I think, this will not set `jirahost` in cookies so subsequent calls such as repository search  or branch validation will fail

src/routes/github/create-branch/github-create-branch-options-get.ts
```diff
@@ -21,8 +21,7 @@ export const GithubCreateBranchOptionsGet = async (req: Request, res: Response,
 		return next(new Error(Errors.MISSING_ISSUE_KEY));
 	}
 
-	const jiraHost = getJiraHostFromTenantUrl(tenantUrl) || jiraHostQuery;
-
+	const jiraHost = prependProtocolToSiteName(res.locals.jiraHost) || prependProtocolToSiteName(tenantUrl) || jiraHostQuery;
 	// TODO move to middleware or shared for create-branch-get
```
Harminder84, line 25: I think , prepend protocol should not be done here. If required, it should be done when reading `jirahost` from token.

test/e2e/create-branch.e2e.ts
```diff
@@ -30,6 +30,7 @@ test.describe("Create branch", () => {
 			expect(poppedUpPage.getByText("Almost there!")).toBeTruthy();
 		});
 	});
+
```
Harminder84, line 33: have you added this file intentionally?

src/middleware/jirahost-middleware.ts
```diff
@@ -19,10 +19,21 @@ const extractUnsafeJiraHost = (req: Request): string | undefined => {
 		return req.body?.jiraHost;
 	} else if (req.cookies.jiraHost) {
 		return req.cookies.jiraHost;
+	} else if (req.query?.jwt) {
+		return getJiraHostFromJwtToken(req.query.jwt as string);
```
gxueatlassian, line 23: I assume this jiraHost comes from the asymmetric? I am afraid it may fail, as later on the same middleware verify it as symmetric?

Maybe we should fix this middleware and the new jwt middleware with asymmetric as well (and base on @mboudreau 's [PR](https://github.com/atlassian/github-for-jira/pull/1871)?

test/e2e/create-branch.e2e.ts
joshkay10: rando space, removing

PR #1888, opened 2023-02-06
NONE: remove more data from logs

bgvozdev:
**What's in this PR?**
Removing Auth header from logs

**Why**
Dislike

### DIFF
--- a/src/config/logger.ts
+++ b/src/config/logger.ts
@@ -110,7 +110,7 @@ const headersSerializer = (headers) => {
 		return headers;
 	}
 
-	const ret = mapKeys({ ...headers }, (_, key) => key.toLowerCase());
+	const ret = mapKeys(headers, (_, key) => key.toLowerCase());
 
 	if (ret["authorization"]) {
 		ret["authorization"] = "CENSORED";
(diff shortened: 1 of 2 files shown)


src/config/logger.ts
```diff
@@ -105,6 +105,20 @@ const censorUrl = (url) => {
...
+	}
+
+	const ret = { ...headers };
+	if (ret["Authorization"] || ret["authorization"]) {
+		delete ret["Authorization"];
+		delete ret["authorization"];
+		ret["Authorization"] = "CENSORED";
+	}
+	return ret;
+};
```
mboudreau, line 120: if you're going to do this, wouldn't it make more sense to "normalize" all headers to be lowercase?

src/config/logger.ts
```diff
@@ -105,6 +105,19 @@ const censorUrl = (url) => {
 	return url;
 };
 
+const headersSerializer = (headers) => {
+	if (!headers) {
+		return headers;
+	}
+
+	const ret = mapKeys({ ...headers }, (_, key) => key.toLowerCase());
```
mboudreau, line 113: why are you recreating an object from the headers?  Headers should already be an object :P

mboudreau (approved):

src/config/logger.ts
bgvozdev: Immutability, they say ;)

bgvozdev: Ah wait a sec, lodash is immtable 🤦  I'll fix that, good catch @mboudreau

mboudreau (approved):

PR #1889, opened 2023-02-07
Arc 1422 add repo sync state fk

gxueatlassian:
**What's in this PR?**
Add a FK to RepoSyncStates to Subscriptions. 

**Why**
Keep data integrity. 

**Added feature flags**
N/A

**Affected issues**  
ARC-1422

**How has this been tested?**  
Local
Unit test

**Whats Next?**
Run DB migration on stg/prod.


src/routes/api/data-cleanup/cleanup-reposyncstates.test.ts
```diff
@@ -21,69 +21,16 @@ describe("Cleanup RepoSyncState", () => {
...
 		await supertest(app).delete(`/api/data-cleanup/repo-sync-states?repoSyncStateId=9999`).set("X-Slauth-Mechanism", "test").expect(200);
 		await supertest(app).delete(`/api/data-cleanup/repo-sync-states?repoSyncStateId=9999&commitToDB=`).set("X-Slauth-Mechanism", "test").expect(200);
 		await supertest(app).delete(`/api/data-cleanup/repo-sync-states?repoSyncStateId=9999&commitToDB=false`).set("X-Slauth-Mechanism", "test").expect(200);
 		await supertest(app).delete(`/api/data-cleanup/repo-sync-states?repoSyncStateId=9999&commitToDB=whatever`).set("X-Slauth-Mechanism", "test").expect(200);
 
-		//check result is correct
-		expect([...await RepoSyncState.findAll()].length).toBe(2);
-	});
-
-	it("should delete orphan RepoSyncState, while leave rest untouch", async () => {
```
gxueatlassian, line 37: These tests can't be run due to the FK. But we still need the api to do a final cleanup right before we run the db migration on prod.

db/migrations/20230202110100-add-repo-sync-state-fk.js
```diff
@@ -0,0 +1,20 @@
...
+module.exports = {
+	up: async (queryInterface) => {
+		await queryInterface.addConstraint("RepoSyncStates", ["subscriptionId"], {
+			type: "foreign key",
+			name: "fk_reposyncst_subs_subid",
+			references: {
+				table: "Subscriptions",
+				field: "id"
+			},
+			onUpdate: "cascade",
```
gxueatlassian, line 12: I am not sure `onUpdate` should be "cascade", but I don't see a harm here.

db/migrations/20230202110100-add-repo-sync-state-fk.js
```diff
@@ -0,0 +1,20 @@
...
+			onDelete: "cascade"
```
gxueatlassian, line 13: Instead of fix codes in the apps, adding tests to make sure RepoSyncStates are deleted when we uninstall the Subscriptions everywhere, I just use `cascade`

bgvozdev (approved):

db/migrations/20230202110100-add-repo-sync-state-fk.js
```diff
@@ -0,0 +1,20 @@
+"use strict";
+
+module.exports = {
+	up: async (queryInterface) => {
+		await queryInterface.addConstraint("RepoSyncStates", ["subscriptionId"], {
+			type: "foreign key",
+			name: "fk_reposyncst_subs_subid",
```
mboudreau, line 7: This name isn't very descriptive.  Do you really need to specify the owning table?  Why not just `fk_subscriptionId_subscriptions`?

db/migrations/20230202110100-add-repo-sync-state-fk.js
mboudreau: It shouldn't be, as we shouldn't update the subscription from the RepoSyncState object.  Could be a dangerous...

db/migrations/20230202110100-add-repo-sync-state-fk.js
mboudreau: Can you make sure that cascade doesn't actually delete the subscription?  Finding documentation on this to show exactly what cascade does isn't great

mboudreau (changes requested):
the migration is all fine and good, but I think you might be forgetting to update the model to reflect the database changes

db/migrations/20230202110100-add-repo-sync-state-fk.js
gxueatlassian: It is bit confusing over the web, so I found this [one](https://stackoverflow.com/a/65206388), it seems if the `constraint` is backed by an index, which is the case for unique constraint, its naming should be global across the schema, hence I added the table name.

gxueatlassian:
> the migration is all fine and good, but I think you might be forgetting to update the model to reflect the database changes

Good question, I thought about this and settle down on NOT having any model changes. With `cascade` on delete and update, it doesn't seems like we need to change any model files?

db/migrations/20230202110100-add-repo-sync-state-fk.js
gxueatlassian: I believe the `cascade`, in this case, should be `if subscription is deleted, all related repo sync state will also be deleted`

gxueatlassian: I can add some tests though.

gxueatlassian: Okay, tests added.

db/migrations/20230202110100-add-repo-sync-state-fk.js
gxueatlassian: So again, I believe the cascade is mean for another direction. 
But still I am not sure we should update `RepoSyncState.subscriptionId` when `Subscription.id` is changed, if it changes. But if we do update the `RepoSyncState.subscriptionId` when `Subscription.id` is changed, I don't see a harm though.

bgvozdev (dismissed):

db/migrations/20230202110100-add-repo-sync-state-fk.js
gxueatlassian: > Finding documentation on this to show exactly what cascade does isn't great...

I use the doc here https://sequelize.org/v5/class/lib/query-interface.js~queryinterface#instance-method-addConstraint

And this is the screenshot of what actual constraint is created.

image

db/migrations/20230202110100-add-repo-sync-state-fk.js
mboudreau: ah right.  Is there a character limit?  I'd rather make it longer and precise than abbreviated and can be confused

db/migrations/20230202110100-add-repo-sync-state-fk.js
mboudreau: right

mboudreau:
@gxueatlassian It's optional, but with sequelize you can get the "related" models automatically within another one.  It's a nicety that's kinda cool because you don't have to do extra calls to the DB, if you get the model, you'll get it's children automatically.

mboudreau (dismissed):
I'll approve it, if you wanted to add model relationship for easier development, you can add it in another PR

mboudreau (approved):

PR #1892, opened 2023-02-07
ARC-1231 Preemptive rate limit

joshkay10:
**What's in this PR?**
Preemptively delay backfill messages if the backfill meets a given threshold, currently sources from a number flag on Launch darkly.

**Why**
We need the backfill to settle down sometimes so webhooks don't get skipped while the backfill is chugging along

**Added feature flags**
https://app.launchdarkly.com/github-for-jira/development/features/preemptive-rate-limit-threshold/targeting

**Affected issues**  
ARC-1231

**How has this been tested?**  
Unit tests have been added for exceeding the threshold and within.
Have run it on Stage with a 1% threshold to see the block happen, then increased to 100% to see the backfill continue about 60 minutes later when the message was visible again.

**Whats Next?**
Play with feature flag values to find the sweet spot


src/github/client/github-installation-client.ts
```diff
@@ -179,6 +179,14 @@ export class GitHubInstallationClient extends GitHubClient {
 		});
 	};
 
+
+	/**
+	 * Returns the current status of the rate limit for all resources types
+	 */
+	public getRateLimit = async (): Promise<AxiosResponse<Octokit.RateLimitGetResponse>> => {
+		return await this.get<Octokit.RateLimitGetResponse>(`/rate_limit`, {}, {});
```
joshkay10, line 187:
```suggestion
		return await this.get<Octokit.RateLimitGetResponse>(`/rate_limit`);
```

src/sqs/backfill.test.ts
```diff
@@ -1,18 +1,17 @@
 import { processInstallation } from "../sync/installation";
-import { mocked } from "ts-jest/utils";
+import { mocked } from "jest-mock";
```
joshkay10, line 2: Mocked in ts-jest/utils is being deprecated and we get warnings all the time so just moving it to the new correct way

src/sqs/backfill.test.ts
```diff
@@ -25,6 +24,41 @@ mocked(Sentry.Hub).mockImplementation(() => ({
 	captureException: sentryCaptureExceptionMock
 } as Sentry.Hub));
 
+const mockSentryGetClient = () => {
+	// eslint-disable-next-line @typescript-eslint/ban-ts-comment
+	// @ts-ignore
```
joshkay10, line 29: Because any time a ts-ignore is there its going to draw eyes.... this is the existing mock and ts-ignore not newly added, i just moved it into a function since i need to re-use it. I did have a look into changing the way its mocked but was causing grief so decided to leave it for now

src/sqs/backfill.test.ts
```diff
@@ -37,15 +71,50 @@ describe("backfill", () => {
 		lastAttempt: false
 	};
 
-	afterEach(() => {
-		sentryCaptureExceptionMock.mockReset();
+	describe("Sentry", () => {
```
joshkay10, line 74: This is the existing test, I just wrapped it in a describe for house keeping and added some mocks in a beforeeach

src/sqs/backfill.ts
```diff
@@ -20,6 +25,14 @@ export const backfillQueueMessageHandler: MessageHandler<BackfillMessagePayload>
 	});
 
 	const backfillData = { ...context.payload };
+	const rateLimitResponse = (await getRateRateLimitStatus(backfillData, context.log))?.data;
+
+	// Check if the rate limit is nearing maximum and delay until refresh
```
joshkay10, line 30: Comments are more confusing than helpful

src/sqs/backfill.ts
```diff
@@ -42,3 +55,25 @@ export const backfillQueueMessageHandler: MessageHandler<BackfillMessagePayload>
...
+	return await gitHubInstallationClient.getRateLimit();
+};
+
+const isRateLimitExceedingSoftLimit = async (rateLimitResponse: Octokit.RateLimitGetResponse, jiraHost) : Promise<boolean> => {
+	const threshold = await numberFlag(NumberFlags.PREEMPTIVE_RATE_LIMIT_THRESHOLD, 100, jiraHost);
+	const { core, graphql } = rateLimitResponse.resources;
+	const usedPercentCore = ((core.limit - core.remaining) / core.limit) * 100;
+	const usedPercentGraphql = ((graphql.limit - graphql.remaining) / graphql.limit) * 100;
+
+	return usedPercentCore >= threshold || usedPercentGraphql >= threshold;
```
joshkay10, line 73: There is 5 or so rate limits, we care about core(REST) and graphql and if either of them are running close then we delay, they are both 5000 from testing my cloud instance.

Incase of interest.... the other three rate limits are search, code_scanning_upload and integration_manifest.

There is also a deprecated rate object that we should not use!

src/sqs/sqs.ts
```diff
@@ -327,7 +327,7 @@ export class SqsQueue<MessagePayload> {
 		return context.receiveCount >= this.maxAttempts;
 	}
 
-	private async changeVisibilityTimeout(message: Message, timeoutSec: number, logger: Logger): Promise<void> {
+	public async changeVisibilityTimeout(message: Message, timeoutSec: number, logger: Logger): Promise<void> {
```
joshkay10, line 330: Create a seperate function delayMessage to and add additional protections?

joshkay10: no

src/sqs/backfill.ts
```diff
@@ -20,6 +25,14 @@ export const backfillQueueMessageHandler: MessageHandler<BackfillMessagePayload>
 	});
 
 	const backfillData = { ...context.payload };
+	const rateLimitResponse = (await getRateRateLimitStatus(backfillData, context.log))?.data;
```
gxueatlassian, line 28: Just for code clarity, do you think combining those two lines is easier to read? Like:

```javascript

  // const rateLimitResponse = (await getRateRateLimitStatus(backfillData, context.log))?.data;
	// Check if the rate limit is exceeding self-imposed limit
	if (await isRateLimitExceedingSoftLimit(backfillData, context.log, jiraHost)) {
		context.log.info("Rate limit internal threshold exceeded, delaying backfilling message.");
		return await sqsQueues.backfill.changeVisibilityTimeout(context.message, getRateResetTime(rateLimitResponse), context.log);
	}



const isRateLimitExceedingSoftLimit = async (backfillData: BackfillMessagePayload, logger: Logger) : Promise<boolean> => {
  const rateLimitResponse = (await getRateRateLimitStatus(backfillData, context.log))?.data;
	const threshold = await numberFlag(NumberFlags.PREEMPTIVE_RATE_LIMIT_THRESHOLD, 100, jiraHost);
	const { core, graphql } = rateLimitResponse.resources;
  ...
}
```

gxueatlassian:
Looks good to me :)

src/sqs/backfill.ts
joshkay10: I agree it would be nice, i kept it out because we re-use the response object to get the reset times.

Since it forces an early return it needed something in that scope........... I could make the whole thing part of the if and if it returns true then return... hmmmmmm


....
// change to the existing handler function
if (await checkPreemptiveRateLimitBlahBlahSomethingSomething(backfillData, context.log, jiraHost)) {
		context.log.info("Rate limit internal threshold exceeded, delaying backfilling message.");
		return;
	}
...

Then the checkPreemptiveRateLimitBlahBlahSomethingSomething function will
- test rate limit v threshold
- return false if she's all good
- delay the message and return true of threshold breached

It makes that core function a lot tidier and isolates all the preemptive stuff to a nice little bubble in the corner?

src/util/preemptive-rate-limit.ts
```diff
@@ -0,0 +1,51 @@
+import { createInstallationClient } from "utils/get-github-client-config";
+import { Octokit } from "@octokit/rest";
+import { numberFlag, NumberFlags } from "config/feature-flags";
+import { SQSMessageContext } from "~/src/sqs/sqs.types";
+import { SqsQueue } from "~/src/sqs/sqs";
+import Logger from "bunyan";
+
+const ALLOWED_QUEUES = ["backfill"];
```
joshkay10, line 8:
```suggestion
const TARGETED_QUEUES = ["backfill"];
```

src/util/preemptive-rate-limit.ts
```diff
@@ -0,0 +1,52 @@
+import { createInstallationClient } from "utils/get-github-client-config";
+import { Octokit } from "@octokit/rest";
+import { numberFlag, NumberFlags } from "config/feature-flags";
+import { SQSMessageContext } from "~/src/sqs/sqs.types";
+import { SqsQueue } from "~/src/sqs/sqs";
+import Logger from "bunyan";
+
+// List of queues we want to apply the preemptive rate limiting on
+const TARGETTED_QUEUES = ["backfill"];
```
joshkay10, line 9: Im pretty sure after my deployment errors look around today I'm going to want to add the deployments queue here too!!

src/util/preemptive-rate-limit.test.ts
```diff
@@ -0,0 +1,104 @@
...
+				}
+			}
+		});
+};
+
+describe("Preemptive rate limit check - Cloud", () => {
+
+	it(`Should return true since the remaining percent is greater than the threshold`, async () => {
+
+		when(numberFlag).calledWith(
```
gxueatlassian, line 32: /nit
wrap this as a function?

src/util/preemptive-rate-limit.test.ts
```diff
@@ -0,0 +1,104 @@
...
+			NumberFlags.PREEMPTIVE_RATE_LIMIT_THRESHOLD,
+			expect.anything(),
+			expect.anything()
+		).mockResolvedValue(50);
+
+		mockGitHubRateLimit(100, 30);
+		const message = {
```
gxueatlassian, line 39: /nit
wrap this as a function?

src/util/preemptive-rate-limit.ts
```diff
@@ -0,0 +1,52 @@
...
+
+// Fetch the rate limit from GitHub API and check if the usages has exceeded the preemptive threshold
+export const preemptiveRateLimitCheck = async (context: SQSMessageContext<any>, sqsQueue: SqsQueue<any>) : Promise<boolean> => {
```
gxueatlassian, line 12: It is just personal idea, I find `isRateLimitExceeded()` more readable. Generally for function returning `boolean`, I would let the function name start with `isXXX` or `hasXXX`

gxueatlassian: Hmm, okay, I see this function does more than just checking, it also change the visiblity of the message. 
In that case, function name seems okay. 

/nit
Although I find it a bit hidden, as the `sqsQueue.changeVisibilityTimeout` is both called in this file, and in `src/sqs/sqs.ts` above.

joshkay10: haha yeah i went through the same name changing thoughts. In an ideal world I would have it as seperate call to check then have the visibility timeout as a result of the bool, but its a bit of a hassle to force it in there.

this was the tidiest option I could find.

gxueatlassian:
So from the doc here: https://docs.aws.amazon.com/AWSSimpleQueueService/latest/SQSDeveloperGuide/sqs-visibility-timeout.html
I think we are good right now, just need to keep it in mind though. Should be fine. 
`For most standard queues (depending on queue traffic and message backlog), there can be a maximum of approximately 120,000 in flight messages`

gxueatlassian (approved):

PR #1894, opened 2023-02-08
[EZY KARMA] NONE: removing more UGC

bgvozdev:
**What's in this PR?**
Killing UGC

**Why**
Because I can


mboudreau (approved):
At this point, wouldn't it just be easier to only log the message to the regular stream (since it holds little value anyways) and log everything else to the unsafe logger with all the information we need?

Harminder84 (approved):

PR #1897, opened 2023-02-09
ARC-1939 Render error message when user can't connect to org.

Harminder84:
**What's in this PR?**
Render error message on UI if user doesn't have admin access or GitHub token gets expired. Moreover, blocking the user from clicking connect button more than once unless first request response  received. 

**Why**
At the moment user doesn't get feedback on UI if something bad happens on connecting to org.

Screen shots - 

Connect button disabled cuz user already clicked on connect button and waiting for its response
Screenshot 2023-02-08 at 4 18 27 pm

When user is not an admin
Screenshot 2023-02-09 at 12 06 35 pm

When Github token gets expired
Screenshot 2023-02-08 at 3 50 39 pm

**Affected issues**  
[ARC-1939]

**How has this been tested?**  
Local

[ARC-1939]: https://softwareteams.atlassian.net/browse/ARC-1939


static/js/github-configuration.js
```diff
@@ -1,17 +1,28 @@
...
+		_csrf: document.getElementById('_csrf').value,
+		clientKey: document.getElementById('clientKey').value
+	}, function (data) {
+		event.currentTarget.removeAttribute("disabled");
+		if (data.err) {
+			showErrorMessage(data.err);
+		}
+		window.close();
+	}).fail((err) => {
+		event.currentTarget.removeAttribute("disabled")
```
gxueatlassian, line 18: /nit
Either more defensive programming like `if(err.resposneJSON)` or use a `try/catch` in case we have error showing error?

gxueatlassian (approved):

gxueatlassian (approved):

PR #1902, opened 2023-02-10
NONE: handle deleted reviewers gracefully

bgvozdev:
**What's in this PR?**
Handle gracefully deleted reviewers

**Why**
Because otherwise it will throw an exception and will block the sync

### DIFF
--- a/src/transforms/transform-pull-request.ts
+++ b/src/transforms/transform-pull-request.ts
@@ -9,7 +9,6 @@ import { generateCreatePullRequestUrl } from "./util/pull-request-link-generator
 import { GitHubInstallationClient } from "../github/client/github-installation-client";
 import { JiraReview } from "../interfaces/jira";
 import { transformRepositoryDevInfoBulk } from "~/src/transforms/transform-repository";
-import { getLogger } from "config/logger";
 
 const mapStatus = (status: string, merged_at?: string) => {
 	if (status === "merged") return "MERGED";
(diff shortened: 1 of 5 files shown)


src/transforms/transform-pull-request.ts
```diff
@@ -75,11 +59,15 @@ const mapReviews = async (reviews: Octokit.PullsListReviewsResponse = [], gitHub
...
 	return Promise.all(reviewsReduced.map(async reviewer => {
-		const gitHubUser = await getGithubUser(gitHubInstallationClient, reviewer.login);
-		return {
-			...omit(reviewer, "login"),
-			email: gitHubUser?.email || `${reviewer.login}@noreply.user.github.com`
+		const mappedReviewer = {
+			...omit(reviewer, "login")
 		};
+		const isDeletedUser = !reviewer.login;
+		if (!isDeletedUser) {
```
atrigueiro, line 66: nit: could just do `if (!reviewer.login)` and add a comment explaining that `login` doesn't exist if the user has been deleted

atrigueiro (approved):

atrigueiro (approved):

atrigueiro (approved):

src/transforms/transform-pull-request.ts
bgvozdev: I prefer to make code self-explanatory and avoid comments

PR #1908, opened 2023-02-16
Fix using proper env for platform sidecar statsd

gxueatlassian:
Refer doc: https://hello.atlassian.net/wiki/spaces/MICROS/blog/2022/08/15/1830221772/PSA+We+re+changing+how+platform+sidecar+addresses+are+derived

**What's in this PR?**
Change to use ENV vars

**Why**
Because the fix hostname is gonna be deprecated

**Added feature flags**
N/A

**Affected issues**  
N/A

**How has this been tested?**  
Stage only

Deployed on stage
Screen Shot 2023-02-16 at 12 00 51 pm

Traffic still coming through
Screen Shot 2023-02-16 at 12 01 17 pm

**Whats Next?**


src/config/env.ts
```diff
@@ -61,7 +61,9 @@ envCheck(
 	"MICROS_AWS_REGION",
 	"GLOBAL_HASH_SECRET",
 	"CRYPTOR_URL",
-	"CRYPTOR_SIDECAR_CLIENT_IDENTIFICATION_CHALLENGE"
+	"CRYPTOR_SIDECAR_CLIENT_IDENTIFICATION_CHALLENGE",
+	"MICROS_PLATFORM_STATSD_HOST",
```
gxueatlassian, line 65: Make sure it fails fast

bgvozdev:
Has it been tested in stg?

bgvozdev:
Do we need a similar change for cryptor?

bgvozdev (approved):

gxueatlassian:
> Has it been tested in stg?

Yup that's the screen shot on description

gxueatlassian:
> Do we need a similar change for cryptor?

May or may not. But cryptor is more explicitly declared, so I guess it is okay.

PR #1915, opened 2023-02-20
ARC-1988 inject api key

bgvozdev:
**What's in this PR?**
Adding ability to inject API key for GHE calls via feature flags

**Why**
To unblock some of our customers who are willing to try this feature ASAP

**Added feature flags**
https://app.launchdarkly.com/github-for-jira/staging/features/ghe-encrypted-api-key/targeting

**Affected issues**  
ARC-1988

**How has this been tested?**  
staging
image

**Whats Next?**
🍿


src/github/client/github-client.nock.test.ts
```diff
@@ -0,0 +1,38 @@
...
+			}
+		}).doTestHttpCall();
+		expect(response).toBeDefined();
+	});
+
+	it("does not explode when no API key was provided", async () => {
+		githubNock.get("/")
+			.reply(200);
+		const response = await new TestGitHubClient(gitHubCloudConfig).doTestHttpCall();
+		expect(response).toBeDefined();
```
gxueatlassian, line 35: Not sure whether if 5xx will also return a `toBeDefined()`

src/github/client/github-client.test.ts
```diff
@@ -15,7 +15,7 @@ class TestGitHubClient extends GitHubClient {
 	}
 }
 
-describe("GitHub Client", () => {
+describe("GitHub Client (mocks)", () => {
```
gxueatlassian, line 18: Why?

src/github/client/github-client.nock.test.ts
bgvozdev: It throws an exception if it is no 200

src/github/client/github-client.test.ts
bgvozdev: Cause in this file we mock the whole axios lib, however I've added one more to use it with nock

src/util/get-github-client-config.ts
```diff
@@ -63,13 +64,38 @@ const calculateProxyBaseUrl = async (jiraHost: string, gitHubBaseUrl: string | u
...
+						EncryptionClient.decrypt(headerEncryptedValue, {
+							jiraHost
+						})
+				}
+			});
+		}
+	} catch (err) {
+		logger.error({ err }, "Cannot calculate API key");
+	}
+	return Promise.resolve(undefined);
```
gxueatlassian, line 88: /nit 
`return undefined` will work as well since there's an `async` in the function signature.

src/util/get-github-client-config.ts
```diff
@@ -63,13 +64,38 @@ const calculateProxyBaseUrl = async (jiraHost: string, gitHubBaseUrl: string | u
 	return envVars.PROXY;
 };
 
+const calculateApiKeyConfig = async (jiraHost: string, logger: Logger): Promise<{ apiKeyConfig: GitHubClientApiKeyConfig } | undefined> => {
+	try {
+		const maybeApiKey = await stringFlag(StringFlags.GHE_API_KEY, "", jiraHost);
+		if (maybeApiKey) {
+			logger.info("Encrypted API key found");
+			const maybeApiKeyParts = maybeApiKey.split(":");
```
gxueatlassian, line 72: If we don't use `:` as the separator, then we don't have to `maybeApiKeyParts.join(":");` later on?
How about `maybeApiKey.split("##KEY_VALUE_SPT##");` or something similar?

gxueatlassian: I found the code hard to read, `split(":")` and `join(":")`

src/util/get-github-client-config.test.ts
```diff
@@ -77,4 +77,19 @@ describe("get-github-client-config", () => {
...
+			.calledWith(StringFlags.GHE_API_KEY, expect.anything(), jiraHost)
+			.mockResolvedValue("ApiKeyHeader:encrypted:super-key");
+
+		const config = await getGitHubClientConfigFromAppId(gitHubServerApp.id, getLogger("test"), jiraHost);
+		expect(config.apiKeyConfig!.headerName).toEqual("ApiKeyHeader");
+		expect(await config.apiKeyConfig!.apiKeyGenerator()).toEqual("super-key");
+	});
+
+	it("does not include API key config when not provided", async () => {
+		const config = await getGitHubClientConfigFromAppId(gitHubServerApp.id, getLogger("test"), jiraHost);
```
gxueatlassian, line 92: Can you also test the `createAnonymousClient ` and `createAnonymousClientByGitHubAppId ` as well? 
As those will be used by teradata to auto exchange token when auto create app.

src/util/get-github-client-config.ts
bgvozdev: Done!

src/util/get-github-client-config.test.ts
gxueatlassian: But should be fine as they all use the same method. This src/util/get-github-client-config.test.ts sucks :( Too convoluted.

src/util/get-github-client-config.ts
bgvozdev: Replaced with JSON.

gxueatlassian (dismissed):

src/util/get-github-client-config.test.ts
bgvozdev: added a test per client

gxueatlassian (approved):

PR #1918, opened 2023-02-21
Arc 1975 show backfill since label and update sync date logic

gxueatlassian:
**What's in this PR?**
1. Show the backfill date on the ui label
2. Added the update `backfillSince` logic
3. Update api endpoints to accept `commitsFromDate`
image

**Caveats**
1. The FF of branch is NOT being used eventually. 

**Why**
Customer would like to know when is the earliest date the backfill is set to.

**Added feature flags**
Reuse backfill-algorithm-incremental

**Affected issues**  
ARC-1975

**How has this been tested?**  
Unit test
local

**Whats Next?**
N/A


src/util/handlebars/handlebar-helpers.ts
```diff
@@ -59,4 +60,9 @@ export const registerHandlebarsHelpers = () => {
 			subscriptionHost !== jiraHost
 	);
 
+	hbs.registerHelper("toDay", (date: Date | null | undefined) => {
+		if (!date) return "";
+		return format(date, "YYYY-MM-DD");
+	});
```
Harminder84, line 66: Is this correct date format we're using at other places? I think we're using `mm-dd-yyy`

Harminder84:
we missed to implement sync date logic for commits in branches.

src/sync/sync-utils.ts
```diff
@@ -3,21 +3,20 @@ import { sqsQueues } from "../sqs/queues";
...
 
-type SyncType = "full" | "partial";
-
 export const findOrStartSync = async (
 	subscription: Subscription,
 	logger: Logger,
-	syncType?: SyncType,
-	commitsFromDate?: Date,
+	syncType: SyncType,
+	commitsFromDate: Date | undefined,
```
Harminder84, line 16: I think previous typing were better. Any advantage of using `| undefined`  over optional?

src/routes/api/installation/api-installation-sync-post.ts
```diff
@@ -22,7 +22,7 @@ export const ApiInstallationSyncPost = async (req: Request, res: Response): Prom
 			return;
 		}
 
-		await findOrStartSync(subscription, req.log, resetType);
+		await findOrStartSync(subscription, req.log, resetType, undefined);
```
Harminder84, line 25: if you make `commitsFromDate` optional, passing `undefined` is not required.

src/routes/jira/sync/jira-sync-post.ts
```diff
@@ -1,15 +1,19 @@
...
 export const JiraSyncPost = async (req: Request, res: Response, next: NextFunction): Promise<void> => {
+
+	const { jiraHost } = res.locals;
+
 	const { installationId: gitHubInstallationId, syncType, appId: gitHubAppId } = req.body;
 
 	// A date to start fetching commit history(main and branch) from.
-	const commitsFromDate = req.body.commitsFromDate ? new Date(req.body.commitsFromDate) : undefined;
+	const commitsFromDate = req.body.commitsFromDate ? new Date(req.body.commitsFromDate) : await getDefaultSyncDate(jiraHost);
 	Sentry.setExtra("Body", req.body);
```
Harminder84, line 17: Do we have any use case where we want to run full backfill? Do we still want a capability to trigger via admin api?

Harminder84: Do want to control the backfill data time frame? e.g. Can't run backfill for more than 1 year time frame

src/sync/backfill-date-calc.ts
```diff
@@ -0,0 +1,40 @@
...
+
+export const calcNewBackfillSinceDate = (
+	existingBackfillSince: Date | undefined,
+	backfillSinceInMsgPayload: string | undefined,
+	syncType: SyncType | undefined,
+	logger: Logger
+): Date | undefined  => {
+
+	if (syncType === "partial" || syncType === undefined) {
+		//do not change anything on partial sync or missing sync type on mgs body ( which means old msg before the prod deployment )
```
Harminder84, line 13: why to add check for partial backfill? I believe when we'r changing the date from UI, we're backfilling commits again. Isn't it a partial backfill?

src/sync/installation.ts
```diff
@@ -118,7 +119,10 @@ export const updateJobStatus = async (
 		scheduleNextTask(0);
 		// no more data (last page was processed of this job type)
 	} else if (!(await getNextTask(subscription, targetTasks))) {
-		await subscription.update({ syncStatus: SyncStatus.COMPLETE });
+		const newBackFillSinceDate = await booleanFlag(BooleanFlags.USE_BACKFILL_ALGORITHM_INCREMENTAL, jiraHost)
+			? calcNewBackfillSinceDate(subscription.backfillSince, data.commitsFromDate, data.syncType, logger)
+			: subscription.backfillSince;
+		await subscription.update({ syncStatus: SyncStatus.COMPLETE, backfillSince: newBackFillSinceDate  });
```
Harminder84, line 125: I think we need to update this when syncing of commits has been completed instead of whole sync process done. Can't we fetch `newBackFillSinceDate` from message rather than computing it again?

src/util/handlebars/handlebar-helpers.ts
gxueatlassian: Tests added. Yup I think 'YYYY-MM-DD' is the correct format https://date-fns.org/v1.29.0/docs/format

src/sync/sync-utils.ts
gxueatlassian: For the `syncType`, mandatory is always better than optional.
For `commitsFromDate`, I'd love to make it mandatory as well. But the logic change is a bit more, so leave it til later.

----------
Meanwhile, it is less bug prone to have explicit param (but can be undfined) compared to optional params. Think about `gitHubAppId`, 
calling a function `getServerApp(someJiraHost, gitHubAppId<--which is undefined)` is better to calling a function that can be `getServerApp(someJiraHost)`. Because for the second one, it could be a bug that someone forgot to passin the `gitHubServerAppId` which may or may not be undefined.

gxueatlassian: But it depends. I can still accept that for truly `optional` param, we take it as `optional`. Just mind that when a function has, say, 4 option params, like `doSomething(a, b?, c?, d?)`, it is quite in flexible, as in order to pass in `c`, you need to pass in `b` first. 

So I prefer using `object`, like
```
findOrStartSync({
  subscription,
  syncType: 'partial',
  commitsFromDate: undefined
});
```

However that's not how the rest of our code looks like. So keep it consistent for this one.

src/routes/jira/sync/jira-sync-post.ts
gxueatlassian: > Do we have any use case where we want to run full backfill? Do we still want a capability to trigger via admin api?

The backfill via the UI is full currently. Unless I misunderstand your questions?

gxueatlassian: > Do want to control the backfill data time frame? e.g. Can't run backfill for more than 1 year time frame

Not sure, could be a new ticket as this is UX enhancement. This change just keeping as it is ( and change the implicit to use the FF on 6months)

src/routes/api/installation/api-installation-sync-post.ts
gxueatlassian: Per my comments above, making it explicit `undefined` is "better" in my mind.
As code is written only once, but read/run millions times, I prefer making it more explicit.

src/sync/backfill-date-calc.ts
gxueatlassian: I thought it is full? haha.

src/routes/jira/sync/jira-sync-post.ts
gxueatlassian: > Do we have any use case where we want to run full backfill? Do we still want a capability to trigger via admin api?

https://github.com/atlassian/github-for-jira/blob/main/static/js/jira-configuration.js#L92

src/sync/backfill-date-calc.ts
gxueatlassian: https://github.com/atlassian/github-for-jira/blob/main/static/js/jira-configuration.js#L92

src/sync/installation.ts
gxueatlassian: Hmm, good questions. I thought the `backfillSinceDate` don't just mean `commits`, but also other tasks as well. Hence I update it only on all job success. Let me ask @Ryan-Yuanqing-Jiang .

`Can't we fetch newBackFillSinceDate` <-- the date in the msg body is the `requested` date. But for existing customers, requested date will be more recent, so the final bakfillSinceDate in DB should still be `undefined`.

----------------

But, **WAIT** !! I think you found a bug!!
For the first time new customers, the existing `backfillSinceDate` is blank, and it should be updated to the requested date. Let me think about it.

src/util/handlebars/handlebar-helpers.ts
Harminder84: This date format is different from we're using in restart backfill window.

Screenshot 2023-03-01 at 9 22 37 am

gxueatlassian: Oh, did you mean UX problem instead of bug?
UX: Yup it is confusing. However I think the ISO format is better than some au format mm-dd-yyy, it won't make sense to us folks.
Bug wise: I checked the date sent from js to backend is ISO format YYYY-MM-DD.

Harminder84:
@gxueatlassian, are we going to implement the same behaviour for commits in non-default branch?

src/util/handlebars/handlebar-helpers.ts
Harminder84: This is how we initialise the current date to restart backfill date input:

```
const initializeBackfillDateInput = function () {
	const dateElement = document.getElementById('backfill-date-picker');
	const date = new Date();
	date.setFullYear(date.getFullYear() - 1);
	dateElement.value = date.toISOString().split('T')[0];
	dateElement.max = new Date().toISOString().split('T')[0];
}
```

src/routes/jira/sync/jira-sync-post.ts
Harminder84: Any idea when are we using partial sync?

gxueatlassian: 1. Some api endpoint can allow us to provide full or partial.
2. First initial sync is partial, but I think that should be equivalent to full.

src/util/handlebars/handlebar-helpers.ts
gxueatlassian: Hey @Harminder84 I've created a followup ticket for the localised ui label. https://softwareteams.atlassian.net/browse/ARC-2008

So closing this now.

gxueatlassian:
> we missed to implement sync date logic for commits in branches.

As discussed, the sync date in branches will same same main branch dates.

src/routes/jira/sync/jira-sync-post.ts
gxueatlassian: So "partial" is not used by user, but only from admin api.

src/sync/installation.ts
gxueatlassian: > But, **WAIT** !! I think you found a bug!! For the first time new customers, the existing `backfillSinceDate` is blank, and it should be updated to the requested date. Let me think about it.

Fixed.

src/sync/installation.ts
```diff
@@ -207,7 +213,12 @@ const doProcessInstallation = async (data: BackfillMessagePayload, sentry: Hub,
 	const gitHubProduct = getCloudOrServerFromGitHubAppId(subscription.gitHubAppId);
 
 	if (!nextTask) {
-		await subscription.update({ syncStatus: "COMPLETE" });
+
+		const newBackFillSinceDate: Date | undefined = await booleanFlag(BooleanFlags.USE_BACKFILL_ALGORITHM_INCREMENTAL, jiraHost)
```
joshkay10, line 217: Can we wrap this into a function?

Already so many things going on in this file would be nice to break into pieces, looks like your doind the same update on line 122 there so could re-use too

gxueatlassian:
Don't know what I am doing. Too many changes!!

Created a new one with only 8 changes. 
https://github.com/atlassian/github-for-jira/pull/1936

PR #1919, opened 2023-02-21
Arc 1975 show backfill since label

gxueatlassian:
**What's in this PR?**
Update the backfill since date in the Subscriptions table

**Why**
We want to show users what's the backfill since date.

**Added feature flags**
N/A

**Affected issues**  
ARC-1975

**How has this been tested?**  
[IN PROGRESS]

**Whats Next?**
N/A

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -221,7 +221,12 @@ const doProcessInstallation = async (data: BackfillMessagePayload, sentry: Hub, 
 
 	const { task, cursor, repository } = nextTask;
 
-	const logger = rootLogger.child({ task: nextTask, gitHubProduct });
+	const logger = rootLogger.child({
+		task: nextTask,
+		gitHubProduct,
+		startTime: data.startTime,
+		commitsFromDate: data.commitsFromDate
+	});
 
 	logger.info("Starting task");
 
@@ -324,8 +329,10 @@ export const handleBackfillError = async (
 	data: BackfillMessagePayload,
 	nextTask: Task,
 	subscription: Subscription,
-	logger: Logger,
+	rootLogger: Logger,
 	scheduleNextTask: (delayMs: number) => void): Promise<void> => {
+
+	const logger = rootLogger.child({ err });
 
 	const isRateLimitError = err instanceof RateLimitingError || Number(err?.headers?.["x-ratelimit-remaining"]) == 0;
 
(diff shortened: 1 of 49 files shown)


src/sync/installation.ts
```diff
@@ -207,7 +209,8 @@ const doProcessInstallation = async (data: BackfillMessagePayload, sentry: Hub,
 	const gitHubProduct = getCloudOrServerFromGitHubAppId(subscription.gitHubAppId);
 
 	if (!nextTask) {
-		await subscription.update({ syncStatus: "COMPLETE" });
+		const newBackFillSinceDate = calcNewBackfillSinceDate(subscription.backfillSince, data.commitsFromDate, rootLogger);
+		await subscription.update({ syncStatus: "COMPLETE", backfillSince: newBackFillSinceDate });
```
gxueatlassian, line 213: We have TWO places for setting `COMPLETE` and metrics seems different! Which one is the one being used?

src/sync/installation.ts
```diff
@@ -118,7 +119,8 @@ export const updateJobStatus = async (
 		scheduleNextTask(0);
 		// no more data (last page was processed of this job type)
 	} else if (!(await getNextTask(subscription, targetTasks))) {
-		await subscription.update({ syncStatus: SyncStatus.COMPLETE });
+		const newBackFillSinceDate = calcNewBackfillSinceDate(subscription.backfillSince, data.commitsFromDate, logger);
+		await subscription.update({ syncStatus: SyncStatus.COMPLETE, backfillSince: newBackFillSinceDate  });
```
gxueatlassian, line 123: We have TWO places for setting COMPLETE and metrics seems different! Which one is the one being used?

src/sync/sync-utils.ts
```diff
@@ -14,10 +14,11 @@ type SyncType = "full" | "partial";
 export const findOrStartSync = async (
 	subscription: Subscription,
 	logger: Logger,
-	syncType?: SyncType,
-	commitsFromDate?: Date,
+	syncType: SyncType,
```
gxueatlassian, line 17: I just think more explicit param is better than implicit param that can be `?`

src/sync/sync-utils.ts
```diff
@@ -54,13 +55,22 @@ export const findOrStartSync = async (
 	}, 0, logger);
 };
 
+export const convertCommitsFromDateStringToDate = (commitsFromDate: string | undefined, logger: Logger) => {
```
gxueatlassian, line 58: This function make sense here because originally the string is converted from a date above `commitsFromDate: commitsFromDate?.toISOString(),`

gxueatlassian: They should be co-locate.

PR #1926, opened 2023-02-27
ARC-968 listen for repo created events and add repos to the reposyncstate table

joshkay10:
**What's in this PR?**
Updated the repo webhook handler to listen for created event as well.

**Why**
We want a more up to date list in our database for clients repos

**Added feature flags**
repo-created-event

**How has this been tested?**  
Added unit tests and ran through staging

**Whats Next?**
Once we are aware of the new repo we need to backfill it. Need to have a think about that and was probably going to blow this PR our too much!


src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -104,9 +104,7 @@ const webhookRouter = async (context: WebhookContext) => {
 			await GithubWebhookMiddleware(deleteBranchWebhookHandler)(context);
 			break;
 		case "repository":
-			if (context.action === "deleted") {
```
Harminder84, line 107: I think we should have both `created` and `deleted` repository action handlers here to have a common pattern.

src/github/repository.ts
```diff
@@ -1,8 +1,46 @@
...
+			repoId: repository.id,
+			repoName: repository.name,
+			repoFullName: repository.full_name,
+			repoOwner: repository.owner.login,
+			repoUrl: repository.html_url,
+			repoUpdatedAt: new Date(repository.updated_at)
+		});
+
+		await subscription.update({ totalNumberOfRepos: (subscription.totalNumberOfRepos || 0) + 1 });
+		await findOrStartSync(subscription, context.log, "partial");
```
Harminder84, line 35: Just for my context, what will happen? will it backfill everything for all repose?

joshkay10: ahhh whoops I was meant to pull this out until I have given partial a good run down. will remove for now
await findOrStartSync(subscription, context.log, "partial");

src/routes/github/webhook/webhook-receiver-post.ts
joshkay10: We still have `created` and `deleted` conditional checks, I have just moved them into `repositoryWebhookHandler`. I reckon its more relevent there and it tidies this function up a bit 🤷

src/github/repository.ts
```diff
@@ -1,8 +1,46 @@
...
+		webhookProcessComplete(context, 200);
+	} catch (err) {
+		context.log.error({ err }, "Error processing create repository webhook");
+		webhookProcessComplete(context, 500);
+	}
+};
+
+export const deleteRepositoryWebhookHandler = async (context: WebhookContext, jiraClient, gitHubInstallationId: number): Promise<void> => {
```
gxueatlassian, line 43: Just realise, should we also DELETE the RepoSyncStates as well? And subtract the Subscriptions.totalRepoNumbers?

gxueatlassian: Coz, otherwise, a full restart backfill will trigger the 404 on the deleted Repo.

src/routes/github/webhook/webhook-receiver-post.ts
gxueatlassian: Just my two cents, I am happy to leave this to @joshkay10 your decision. Seems /nit to me. What @Harminder84 said has his point, however we already have bellow patterns:
```
case "issue_comment":
			if (context.action === "created" || context.action === "edited") {
				await webhookTimeout(GithubWebhookMiddleware(issueCommentWebhookHandler))(context);
			}
			break;
```

gxueatlassian (approved):

src/github/repository.ts
joshkay10: oooooooo Good catch, thats a good point, I just assumed it was whoops.

Sweet I will add that to the story now and do it as a seperate task!

gxueatlassian (dismissed):

src/routes/github/webhook/webhook-receiver-post.ts
joshkay10: Yeah it's a bit messy, we have a combination of logic at this level and then down a level on the handlers themselves. My opinion/thinking is it was more approriate for the condition to exist within the handler, I reckon this function should only have that top level decision of what handler, then the handler can decide the more finer details

PR #1927, opened 2023-02-27
NONE: retry sequelize errors until we don't have DLQ

bgvozdev:
**What's in this PR?**
Replying sequelize errors because we don't have DLQ just yet

**Why**
There are more customers complaining their syncs were "stuck", and splunk shows there was a sequelize error.

image

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -326,8 +326,10 @@ export const handleBackfillError = async (
 	data: BackfillMessagePayload,
 	nextTask: Task,
 	subscription: Subscription,
-	logger: Logger,
+	rootLogger: Logger,
 	scheduleNextTask: (delayMs: number) => void): Promise<void> => {
+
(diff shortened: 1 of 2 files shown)


src/sync/installation.ts
```diff
@@ -365,6 +367,15 @@ export const handleBackfillError = async (
 		return;
 	}
 
+	// TODO: replace with "instanceof" when sequelize version is upgraded to some modern one
+	// Capturing errors like SequelizeConnectionError, SequelizeConnectionAcquireTimeoutError etc
+	// that are not exported from sequelize
+	if (String(err.name).toLowerCase().includes("sequelize")) {
```
gxueatlassian, line 373: `NOT` your fault, `NOT` this PR, but the way we determine the error type sucks! Really missed the java checked exception.

gxueatlassian (approved):

PR #1931, opened 2023-02-28
NONE: improve logging of headers

bgvozdev:
**What's in this PR?**
Adding whitelist for headers. Now for logging, but then we will need it for BTF UI (to check if a particular customer's header is allowed)


src/util/http-headers.ts
```diff
@@ -0,0 +1,282 @@
...
+// 	"proxy-authorization",
+// 	"www-authenticate",
+//
+// 	"cookie",
+// 	"cookie2",
+// 	"set-cookie",
+// 	"set-cookie2",
+// ]);
+
+const knownGitHubSafe = new Set([
```
gxueatlassian, line 266: `x-ratelimit-used` as well.

gxueatlassian (approved):

PR #1935, opened 2023-03-01
ARC-1730 Fix the reposyncstate bulk update duplicate issue

gxueatlassian:
**What's in this PR?**
Convert the buggy bulkUpdateOrCreate to a more simple loop and discovering repos in backfill

**Why**
Partial backfill has some bug that duplicate the RepoSyncStates in some cases.

**Added feature flags**
N/A

**Affected issues**  
ARC-1730

**How has this been tested?**  
Unit test

**Whats Next?**
N/A

### DIFF
--- a/src/models/reposyncstate.ts
+++ b/src/models/reposyncstate.ts
@@ -156,6 +156,30 @@ export class RepoSyncState extends Model {
 			}
 		});
 	}
+
+	static async bulkCreateOrUpdateByRepos(
+		subscriptionId: number,
+		repositories: {
+			repoId: number,
+			repoName: string,
+			repoFullName: string,
+			repoOwner: string,
+			repoUrl: string,
+			repoUpdatedAt: Date
+		}[]
+	): Promise<RepoSyncState[]> {
+		return Promise.all(repositories.map(repo => {
+			return RepoSyncState.findOrCreate({
+				where: {
+					subscriptionId,
+					repoId: repo.repoId
+				},
+				defaults: {
+					...repo
+				}
+			}).then((r: [RepoSyncState, boolean]) => r[0]);
+		}));
+	}
 }
 
 RepoSyncState.init({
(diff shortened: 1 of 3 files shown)


bgvozdev (approved):

src/models/reposyncstate.ts
```diff
@@ -156,6 +156,30 @@ export class RepoSyncState extends Model {
...
+			repoId: number,
+			repoName: string,
+			repoFullName: string,
+			repoOwner: string,
+			repoUrl: string,
+			repoUpdatedAt: Date
+		}[]
+	): Promise<RepoSyncState[]> {
+		return Promise.all(repositories.map(repo => {
+			return RepoSyncState.findOrCreate({
```
mboudreau, line 172: ...why `findOrCreate`?  This is suppose to be an `upsert`...

src/models/reposyncstate.ts
```diff
@@ -156,6 +156,30 @@ export class RepoSyncState extends Model {
 			}
 		});
 	}
+
+	static async bulkCreateOrUpdateByRepos(
+		subscriptionId: number,
```
mboudreau, line 161: This is not the signature convention that we're using in the models

src/models/reposyncstate.ts
```diff
@@ -156,6 +156,30 @@ export class RepoSyncState extends Model {
...
+		repositories: {
```
mboudreau, line 162: We already have a type for this

src/models/reposyncstate.ts
```diff
@@ -156,6 +156,30 @@ export class RepoSyncState extends Model {
 			}
 		});
 	}
+
+	static async bulkCreateOrUpdateByRepos(
```
mboudreau, line 160: Maybe renaming it as `bulkUpsert` would make more sense?

mboudreau (changes requested):
This is not the same functionality that it was doing before.

gxueatlassian:
Will have a look later.

PR #1936, opened 2023-03-01
Arc 1957 udpate backfill since date

gxueatlassian:
**What's in this PR?**
Update the backfillSince data column.

**Why**
So that user can have visibility of when the backfill is up to. 

**Added feature flags**
N/A <-- column is not being used. 

**Affected issues**  
ARC-1957

**How has this been tested?**  
Unit Test
Manual Test: [WIP]

**Whats Next?**
Display the backfill date on UI.

### DIFF
--- a/src/routes/github/configuration/github-configuration-post.ts
+++ b/src/routes/github/configuration/github-configuration-post.ts
@@ -84,7 +84,7 @@ export const GithubConfigurationPost = async (req: Request, res: Response): Prom
 		await Promise.all(
 			[
 				saveConfiguredAppProperties(jiraHost, gitHubInstallationId, gitHubAppId, req.log, true),
-				findOrStartSync(subscription, req.log, "full")
+				findOrStartSync(subscription, req.log, true, "full")
 			]
 		);
 
--- a/src/routes/github/configuration/github-configuration-post.ts
+++ b/src/routes/github/configuration/github-configuration-post.ts
@@ -70,17 +70,21 @@ export const GithubConfigurationPost = async (req: Request, res: Response): Prom
 			return;
 		}
 
-		const subscription = await Subscription.install({
+		const subscription: Subscription = await Subscription.install({
 			hashedClientKey: req.body.clientKey,
 			installationId: gitHubInstallationId,
 			host: jiraHost,
 			gitHubAppId
 		});
 
+		//mark is as current date so that subsequent backfill will override it
+		subscription.backfillSince = new Date();
+		await subscription.save();
+
 		await Promise.all(
 			[
 				saveConfiguredAppProperties(jiraHost, gitHubInstallationId, gitHubAppId, req.log, true),
-				findOrStartSync(subscription, req.log)
+				findOrStartSync(subscription, req.log, "full")
 			]
 		);
 
--- a/src/routes/github/configuration/github-configuration-post.ts
+++ b/src/routes/github/configuration/github-configuration-post.ts
@@ -77,10 +77,6 @@ export const GithubConfigurationPost = async (req: Request, res: Response): Prom
 			gitHubAppId
 		});
 
-		//mark is as current date so that subsequent backfill will override it
-		subscription.backfillSince = new Date();
-		await subscription.save();
-
 		await Promise.all(
 			[
 				saveConfiguredAppProperties(jiraHost, gitHubInstallationId, gitHubAppId, req.log, true),
--- a/src/sync/backfill-since-date-calc.ts
+++ b/src/sync/backfill-since-date-calc.ts
@@ -4,7 +4,7 @@ export const calcNewBackfillSinceDate = (
 	existingBackfillSince: Date | undefined,
 	commitsFromDate: Date | undefined,
 	syncType: SyncType | undefined,
-	isInitialNewSync: boolean | undefined
+	isInitialSync: boolean | undefined
 ): Date | undefined  => {
 
 	//------------ partial sync ---------------
@@ -16,7 +16,7 @@ export const calcNewBackfillSinceDate = (
 	}
 
 	//------------ initial new sync ---------------
-	if (isInitialNewSync === true) {
+	if (isInitialSync) {
 		//for initial new sync, take whatever provided as the commitsFromDate
 		//even it is undefined (which means everything will be synced)
 		return commitsFromDate;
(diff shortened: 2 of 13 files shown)


src/routes/github/configuration/github-configuration-post.ts
```diff
@@ -70,17 +70,21 @@ export const GithubConfigurationPost = async (req: Request, res: Response): Prom
...
-		const subscription = await Subscription.install({
+		const subscription: Subscription = await Subscription.install({
 			hashedClientKey: req.body.clientKey,
 			installationId: gitHubInstallationId,
 			host: jiraHost,
 			gitHubAppId
 		});
 
+		//mark is as current date so that subsequent backfill will override it
+		subscription.backfillSince = new Date();
```
gxueatlassian, line 81: So for all new github org connection ( subscription ), the backfillSinceDate is mark as current date():

1. Before turn on `sync-main-commit-time-limit`, the `commitSince` will be `undefined`, upon successful backfill, it will override this.
2. After turn on `sync-main-commit-time-limit`, the `commitSince` will be `6 months (or so)`, upon successful backfill, it will override this.
3. On the UI, user will NOT see anything until the backfill is complete and success.

So I think this `subscription.backfillSince = new Date();` is safe. 

The purpose is, it help the code to understand, whether the subscription is a new backfill or existing with undefined backfillSinceDate.

src/sync/installation.ts
```diff
@@ -506,3 +514,10 @@ const updateRepo = async (subscription: Subscription, repoId: number, values: Re
 		Object.keys(rest).length && RepoSyncState.updateRepoFromSubscription(subscription, repoId, rest)
 	]);
 };
+
+const getBackfillSince = async (subscription: Subscription, data: BackfillMessagePayload): Promise<Date | null> => {
+	const commitSince = await getCommitSinceDate(data.jiraHost, NumberFlags.SYNC_MAIN_COMMIT_TIME_LIMIT, data.commitsFromDate);
```
Harminder84, line 519: `CommitSince` should be fetched from the message payload. If backfilling takes more than a day, which will happen, then there is  a mismatch with the initial `CommitSince` which is used for backfilling

src/sync/sync-utils.ts
```diff
@@ -3,17 +3,16 @@ import { sqsQueues } from "../sqs/queues";
...
 import { envVars } from "config/env";
 import { GitHubServerApp } from "models/github-server-app";
 import { GITHUB_CLOUD_API_BASEURL, GITHUB_CLOUD_BASEURL } from "~/src/github/client/github-client-constants";
 
-type SyncType = "full" | "partial";
-
 export const findOrStartSync = async (
 	subscription: Subscription,
 	logger: Logger,
+	isInitialNewSync: boolean,
```
joshkay10, line 15: /nit isInitialSync for the name?

src/sync/backfill-since-date-calc.ts
```diff
@@ -0,0 +1,48 @@
+import { SyncType } from "~/src/sync/sync.types";
+
+export const calcNewBackfillSinceDate = (
+	existingBackfillSince: Date | undefined,
+	commitsFromDate: Date | undefined,
+	syncType: SyncType | undefined,
+	isInitialNewSync: boolean | undefined
+): Date | undefined  => {
+
+	//------------ partial sync ---------------
```
joshkay10, line 10: THE COMMENTS ARE VERY MUCH APPRECiATED HERE!! llololol

src/sync/backfill-since-date-calc.ts
```diff
@@ -0,0 +1,48 @@
...
+	if (syncType === "partial" || syncType === undefined) {
+		//do not change anything on partial sync
+		//or missing sync type on mgs body
+		//( which means old msg before the prod deployment )
+		return existingBackfillSince;
+	}
+
+	//------------ initial new sync ---------------
+	if (isInitialNewSync === true) {
```
joshkay10, line 19: /nit `if (isInitialNewSync)`

src/sync/backfill-since-date-calc.ts
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	if (!commitsFromDate) {
+		//backfill is instructed to backfill everything, so use it
+		return commitsFromDate;
+	}
+
+	if (existingBackfillSince.getTime() <= commitsFromDate.getTime()) {
+		//Origin backfill date is either empty or earlier,
+		//So use the origin one.
```
joshkay10, line 40: So use the original one*?

joshkay10 (approved):

src/sync/backfill-since-date-calc.ts
```diff
@@ -0,0 +1,48 @@
+import { SyncType } from "~/src/sync/sync.types";
+
+export const calcNewBackfillSinceDate = (
+	existingBackfillSince: Date | undefined,
+	commitsFromDate: Date | undefined,
+	syncType: SyncType | undefined,
+	isInitialNewSync: boolean | undefined
+): Date | undefined  => {
```
Harminder84, line 8: Do we really need `syncType` here? Can't we decide on the basis of `isInitialNewSync`?

src/sync/installation.ts
gxueatlassian: --------
For restart backfill, there's no problem, the `data.commitsFromDate` is the date user picked from UI

----
For initial sync. The commitSince could be missing in the body
And I don't want to mess up with the msg body again, too much code changes. Maybe in another PR.

-------

So yes, this will be slightly different compare to the actual commit sync date, but it will be earlier, so it is "safe" from data missing point of view.

And current prod behaviour is already like that, the `commitsSince` day will changes everytime the task is reschdule again for initial sync.

src/sync/backfill-since-date-calc.ts
gxueatlassian: If partial, we should ignore it for safe side. So yes, need synctype.

joshkay10: keep synctype!! i want to start using Partial a lot so will be more relevant in time

gxueatlassian: LOL, 👍 
Close this convo

src/sync/sync-utils.ts
gxueatlassian: done.

src/sync/backfill-since-date-calc.ts
gxueatlassian: fixed as `existing backfillSince date`

src/sync/backfill-since-date-calc.ts
gxueatlassian: fixed.

src/sync/installation.ts
gxueatlassian: Let's create another PR for put dates inside the msg body to be more explicit.

Harminder84 (approved):

PR #1937, opened 2023-03-02
ARC-968 view repos

joshkay10:
**What's in this PR?**
Took @Harminder84 PR and just added in some pagination for the repo list

Existing things from Harminders...
We can select the repos from the config page and it will show us a list with their sync status

**Why**
Gives us/customers more visibility on backfill states

**Screenshots yay**
Screenshot 1 - they All repos text is now a link that takes us to a magical land of repos list
Screen Shot 2023-03-08 at 1 19 43 PM

The list itself....we tried to maintain the same appearance as the config page. Only showing the syncstatus for now, more can be added laters!

**Whats Next?**
- Would be useful to have a filter to show only failing repos
- Probably give them the option to sync individual repos too
- decide whether we want to show more info around the failed states
- Way down the line linking a repo to a project we could do that here?

TESTSSSSSSSS!!!!!!


src/routes/jira/jira-get-connected-repos.ts
```diff
@@ -0,0 +1,139 @@
...
+
+		if (!subscriptionId) {
+			req.log.error("Missing Subscription ID");
+			res.status(401).send("Missing Subscription ID");
+			return;
+		}
+
+		const subscription = await Subscription.findByPk(subscriptionId);
+
+		const repoSyncStates = await RepoSyncState.findAllFromSubscription(subscription);
```
Harminder84, line 35: We need to put some limit in the query to restrict the result set. Some orgs have more than 12k repos, I believe this view is not capable of handling that much result set. can you please test the behaviour of UI with large mocked data set?

joshkay10: and get totalCount

rachellerathbone:
We've added new files and made lots of file changes - would be great to add some tests.

rachellerathbone:
Should we feature this? I can see we limit it to 10 per page but it doesn't look like we're limiting the number of repos we're requesting at a time. Worried things could blow up/really slow down. 

Also, have we thought about what the UI should look like if we have more than 4 pages? What if I have 100 pages? Are we planning to render all numbers from 1 to 100? And I can see we hide prev/next if there isn't a previous next - does this mean the numbers move around? i.e. I'm on page 1 and 'Prev' is hidden but then I click page 2 and suddenly 'Prev' appears and all numbers shift to the right? A recording of this would be great.

views/jira-connected-repos.hbs
```diff
@@ -0,0 +1,108 @@
...
+                {{/if}}
+            {{/each}}
+
+            {{#if hasNextPage}}
+            <a href="#" class="next-page page-selector" data-page-num="{{nextPageNum}}">Next</a>
+            {{/if}}
+        </div>
+    </section>
+
+<!--    <span class="pageone" data-subscription-id="{{subscriptionId}}">PAGE 1</span>-->
```
rachellerathbone, line 90: Can we delete this?

src/routes/jira/jira-get-connected-repos.ts
```diff
@@ -0,0 +1,139 @@
...
+	res: Response,
+	next: NextFunction
+): Promise<void> => {
+
+	try {
+		const { jiraHost, nonce } = res.locals;
+		const subscriptionId = Number(req.params.subscriptionId) || Number(req.body.subscriptionId);
+		const page = Number(req.query.page || 1);
+
+		if (!jiraHost) {
```
gxueatlassian, line 21: I reckon we could get away without it? As should rely on the middleware?

src/routes/jira/jira-get-connected-repos.ts
```diff
@@ -0,0 +1,139 @@
...
+
+export const JiraGetConnectedRepos = async (
+	req: Request,
+	res: Response,
+	next: NextFunction
+): Promise<void> => {
+
+	try {
+		const { jiraHost, nonce } = res.locals;
+		const subscriptionId = Number(req.params.subscriptionId) || Number(req.body.subscriptionId);
```
gxueatlassian, line 18: Why do we need both params?

src/routes/jira/jira-get-connected-repos.ts
gxueatlassian, line 35: May something like `await RepoSyncState.findAllFromSubscription(subscription, { order: [["id", "DESC"]] });`?
As otherwise the pagination may not be stable?

joshkay10: I've added ordering to the findAllFromSubscription function itself

views/jira-connected-repos.hbs
joshkay10: deleted

joshkay10: > Should we feature this? I can see we limit it to 10 per page but it doesn't look like we're limiting the number of repos we're requesting at a time. Worried things could blow up/really slow down.
> 
> Also, have we thought about what the UI should look like if we have more than 4 pages? What if I have 100 pages? Are we planning to render all numbers from 1 to 100? And I can see we hide prev/next if there isn't a previous next - does this mean the numbers move around? i.e. I'm on page 1 and 'Prev' is hidden but then I click page 2 and suddenly 'Prev' appears and all numbers shift to the right? A recording of this would be great but good UX for implementing pagination is to disable prev/next as opposed to hiding/displaying them.

The UI can handle as many pages as you can throw at it and it will show a nuch before and after e.g. 2000 pages on page 600 will show prev 598 599 600 601 602 next 

We only send one page of data to the template at a time so the performance should be ok since we retrieve from our db, however, we are changing it to a seperate API fetch so it can be fine tuned a bit better.

prev and next sometime showing, numbers bouncing... yup, agree its pretty lame, ill style it so its a fixed position for the numbers and Sarah has shown us a prettier way to do it.

If you have 329083479 Repos its simply unusable, so will be adding a filter to show/hide based of status and also a search.

src/routes/jira/jira-get-connected-repos.ts
joshkay10: I genuinely cant remember why we need both, i guess one is a safety net potentially coming from different routes.

src/routes/jira/jira-get-connected-repos.ts
joshkay10: goneskies!

src/routes/jira/jira-get.ts
```diff
@@ -175,7 +175,8 @@ const renderJiraCloudAndEnterpriseServer = async (res: Response, req: Request):
 			hasConnections,
 			useNewSPAExperience,
 			APP_URL: process.env.APP_URL,
-			enableRepoConnectedPage: await booleanFlag(BooleanFlags.ENABLE_CONNECTED_REPOS_VIEW, jiraHost),
+			// TODO: uncomment me
+			enableRepoConnectedPage: true, //await booleanFlag(BooleanFlags.ENABLE_CONNECTED_REPOS_VIEW, jiraHost),
```
gxueatlassian, line 179: ???

gxueatlassian (changes requested):
ff

PR #1938, opened 2023-03-02
ARC968 delete repos and sync created

joshkay10:
**What's in this PR?**
On repo delete event we remove it from our reposync state table to maintain a correct list!

When a new repo is added ot reposyncstate via webhook we sync it straight away.

Tested on stage and syncs only the new repo and deletes repos correctly yay


src/github/repository.ts
```diff
@@ -31,14 +32,15 @@ export const createRepositoryWebhookHandler = async (context: WebhookContext, gi
 		});
 
 		await subscription.update({ totalNumberOfRepos: (subscription.totalNumberOfRepos || 0) + 1 });
+		await findOrStartSync(subscription, context.log, "partial");
```
gxueatlassian, line 35: I think you might want to:
1. Passin `false` as the `isInitialSync` flag, coz that flag means is the WHOLE subscription is a new sync, and will take the date as the final backfillSinceDate to DB and show on UI label. But this is just one repo. So its date shouldn't take over.
2. Maybe also see what date shall we pass in? Take the existing backfillSinceDate in the subscription? I haven't think about it yet.

src/github/repository.ts
```diff
@@ -49,6 +51,10 @@ export const deleteRepositoryWebhookHandler = async (context: WebhookContext, ji
 		transformRepositoryId(context.payload.repository?.id, context.gitHubAppConfig?.gitHubBaseUrl)
 	);
 
+	// Update the total number of repos, we can assume atleast 1 repo(just deleted repo) exists for the default value
+	const totalNumberOfRepos = ((subscription.totalNumberOfRepos || 1) - 1);
```
gxueatlassian, line 55: There's a `numberOfSyncedRepos` as well. Maybe should do a count like `select from RepoSyncStates`, as the repo we delete we can't be sure it is success or not?

src/github/repository.ts
```diff
@@ -49,6 +51,10 @@ export const deleteRepositoryWebhookHandler = async (context: WebhookContext, ji
...
+	await subscription.update({ totalNumberOfRepos });
+	await RepoSyncState.deleteRepoForSubscription(subscription, context.payload.repository?.id);
```
gxueatlassian, line 57: I find `context.payload.repository?.id` could be dangerous. If it is `undefined` we are at the mercy of sequlize to do proper job to avoid delete all repos.

Maybe do a defensive check, and only proceed when it is not empty.

src/github/repository.ts
```diff
@@ -49,6 +51,10 @@ export const deleteRepositoryWebhookHandler = async (context: WebhookContext, ji
...
 	webhookProcessComplete(context, jiraResponse?.status);
```
Harminder84, line 58: Can we first delete from `RepoSyncState` before updating the `totalNumberOfRepos` in `subscription` table? I think, we also need to check whether any recored is deleted from `RepoSyncState` before updating count in `Subscription` table.

src/github/repository.ts
joshkay10: ahh your right, i did this before that change lol - I will update. It will take backfillSinceDate so its consistent

src/github/repository.ts
```diff
@@ -31,14 +32,15 @@ export const createRepositoryWebhookHandler = async (context: WebhookContext, gi
 		});
 
 		await subscription.update({ totalNumberOfRepos: (subscription.totalNumberOfRepos || 0) + 1 });
+		await findOrStartSync(subscription, context.log, false, "partial", subscription.backfillSince);
```
gxueatlassian, line 35: Can we have a chat about it? I worry it target too much repos.

gxueatlassian (changes requested):
Can we have a chat about it? I worry it target too much repos.

src/github/repository.test.ts
```diff
@@ -101,6 +146,6 @@ describe("createdRepositoryWebhookHandler", () => {
 			}
 		}), DatabaseStateCreator.GITHUB_INSTALLATION_ID, subscription);
 
-		expect(subscription.update).toBeCalledWith({ totalNumberOfRepos: 100 });
+		expect(subscription.update).toBeCalledWith({ totalNumberOfRepos: 2 });
```
joshkay10, line 149: we check the actual repo count in reposync state so we needed to change this to reflect the mocked reposyncstate count instead of a arbitrary number

src/github/repository.ts
```diff
@@ -3,18 +3,25 @@ import { WebhookContext } from "routes/github/webhook/webhook-context";
...
 	if (context.action === "deleted") {
-		return deleteRepositoryWebhookHandler(context, jiraClient, gitHubInstallationId);
+		return deleteRepositoryWebhookHandler(context, jiraClient, gitHubInstallationId, subscription);
 	}
 	if (context.action === "created" && await booleanFlag(BooleanFlags.REPO_CREATED_EVENT, jiraClient.baseURL)) {
 		return createRepositoryWebhookHandler(context, gitHubInstallationId, subscription);
 	}
 };
 
+const updateRepoCount = async (subscription: Subscription) => {
```
joshkay10, line 19: split update count to its own function so can do the same in delete and create

src/github/repository.ts
```diff
@@ -30,15 +37,16 @@ export const createRepositoryWebhookHandler = async (context: WebhookContext, gi
 			repoUpdatedAt: new Date(repository.updated_at)
 		});
 
-		await subscription.update({ totalNumberOfRepos: (subscription.totalNumberOfRepos || 0) + 1 });
+		await updateRepoCount(subscription);
+		await findOrStartSync(subscription, context.log, false, "partial", subscription.backfillSince,undefined, repository.id);
```
joshkay10, line 41: We pass repository.id here now, this will force only the events repo to update

src/github/repository.ts
```diff
@@ -3,18 +3,25 @@ import { WebhookContext } from "routes/github/webhook/webhook-context";
...
+		return deleteRepositoryWebhookHandler(context, jiraClient, gitHubInstallationId, subscription);
 	}
 	if (context.action === "created" && await booleanFlag(BooleanFlags.REPO_CREATED_EVENT, jiraClient.baseURL)) {
 		return createRepositoryWebhookHandler(context, gitHubInstallationId, subscription);
 	}
 };
 
+const updateRepoCount = async (subscription: Subscription) => {
+	// Update the repos by taking the new post-delete count of repos for the given subscription
+	const totalNumberOfRepos = (await RepoSyncState.findAllFromSubscription(subscription)).length;
```
gxueatlassian, line 21: Actually seems there's a `RepoSyncState.countFromSubscription` ? Maybe we could use that?

src/github/repository.ts
```diff
@@ -30,26 +37,37 @@ export const createRepositoryWebhookHandler = async (context: WebhookContext, gi
 			repoUpdatedAt: new Date(repository.updated_at)
 		});
 
-		await subscription.update({ totalNumberOfRepos: (subscription.totalNumberOfRepos || 0) + 1 });
+		await updateRepoCount(subscription);
+		await findOrStartSync(subscription, context.log, false, "partial", subscription.backfillSince,undefined, repository.id);
```
gxueatlassian, line 41: `subscription.backfillSince,undefined` <-- there should be a space after the comma?

gxueatlassian (approved):

PR #1939, opened 2023-03-03
ARC-1975 promote commitsFromDate into msg body

gxueatlassian:
**What's in this PR?**
Promote the commitsSince date into msg body, so that they are calculated in WebServer instead of Worker.

**Why**
1. Explicit is always better than implicit.
2. We can apply more field, like commitsToDate in to msg body.

**Added feature flags**
N/A

**Affected issues**  
ARC-1975

**How has this been tested?**  
Unit Test

**Whats Next?**
Might want to remove the branch commits date ff.

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -12,7 +12,7 @@ import { getCommitTask } from "./commits";
 import { getBuildTask } from "./build";
 import { getDeploymentTask } from "./deployment";
 import { metricSyncStatus, metricTaskStatus } from "config/metric-names";
-import { isBlocked, NumberFlags } from "config/feature-flags";
+import { isBlocked } from "config/feature-flags";
 import { Deduplicator, DeduplicatorResult, RedisInProgressStorageWithTimeout } from "./deduplicator";
 import { getRedisInfo } from "config/redis-info";
 import { BackfillMessagePayload } from "../sqs/sqs.types";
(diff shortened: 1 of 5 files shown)


src/sync/installation.ts
```diff
@@ -516,7 +515,7 @@ const updateRepo = async (subscription: Subscription, repoId: number, values: Re
 };
 
 const getBackfillSince = async (subscription: Subscription, data: BackfillMessagePayload): Promise<Date | null> => {
-	const commitSince = await getCommitSinceDate(data.jiraHost, NumberFlags.SYNC_MAIN_COMMIT_TIME_LIMIT, data.commitsFromDate);
+	const commitSince = data.commitsFromDate ? new Date(data.commitsFromDate) : undefined;
```
joshkay10, line 518: /nit ternary vs &&... fight!!! I've always leaned towards the `&&` approach, but don't need to change this, just an observation for discussion at some point

`const commitSince = data.commitsFromDate ? new Date(data.commitsFromDate) : undefined;`
`const commitSince = data.commitsFromDate && new Date(data.commitsFromDate);`

joshkay10 (dismissed):
Strong approve, I think this change makes a lot of sense!!

src/sync/installation.ts
gxueatlassian: sorry I want to stick to my origin way of the ternary vs && fight. If `data.commitsFromDate is "" or 0 or NaN`, then `data.commitsFromDate && new Date(data.commitsFromDate)` will evaluated to `"" or 0 or NaN` . However, I want to make sure they evaluated to undefined, hence stick to my origin way. Let me know if you really want me to change it to the &&.

Harminder84 (approved):

PR #1941, opened 2023-03-06
NONE: remove some code duplication

bgvozdev:
**What's in this PR?**
Just removing some code duplication that I introduced on the weekend. Also adding some comments about pagination (tribal knowledge before)

**Why**
Cleaning up my own mess

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -239,6 +239,13 @@ const doProcessInstallation = async (data: BackfillMessagePayload, sentry: Hub, 
 
 	const execute = async (): Promise<TaskPayload> => {
 		const gitHubInstallationClient = await createInstallationClient(gitHubInstallationId, jiraHost, logger, data.gitHubAppConfig?.gitHubAppId);
+		//
+		// We are not using "big" numbers here (e.g. 100) to speed up the process; API returns smaller pages faster and
+		// they will more likely not to break.
+		//
+		// Also this logic is broken for any rest queries (where cursor is the page number), because when the page size
+		// changes the cursor is broken (as it is just an interger specifically for the original page size).
+		//
 		for (const perPage of [20, 10, 5, 1]) {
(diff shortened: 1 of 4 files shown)


src/sync/installation.ts
```diff
@@ -239,6 +239,13 @@ const doProcessInstallation = async (data: BackfillMessagePayload, sentry: Hub,
 
 	const execute = async (): Promise<TaskPayload> => {
 		const gitHubInstallationClient = await createInstallationClient(gitHubInstallationId, jiraHost, logger, data.gitHubAppConfig?.gitHubAppId);
+		//
+		// We are not using "big" numbers here (e.g. 100) to speed up the process; API returns smaller pages faster and
+		// they will more likely not to break.
+		//
+		// Also this logic is broken for any rest queries (where cursor is the page number), because when the page size
```
rachellerathbone, line 246: If this is broken, do we have a ticket in the backlog to address it?

src/sync/pull-request.ts
```diff
@@ -90,9 +89,11 @@ export const getPullRequestTask = async (
 	const pullRequests = (
 		await Promise.all(
 			edgesWithCursor.map(async (pull) => {
-				const issueKeys = jiraIssueKeyParser(`${pull.title}\n${pull.head.ref}\n${pull.body}`);
 
-				if (isEmpty(issueKeys)) {
+				if (isEmpty(extractIssueKeysFromPr(pull))) {
+					logger.info({
+						prId: pull.id
+					}, "Skip PR cause has no issue keys");
```
rachellerathbone, line 96: Nit: cause it has

src/sync/installation.ts
bgvozdev: No, but I think I'll be touching this code soon as a part of hot-102548

rachellerathbone (approved):

PR #1942, opened 2023-03-06
NONE: add coef for page size for builds

bgvozdev:
**What's in this PR?**
Increaseing page size for builds: a sync is going on forever for one particular customer

**Why**
I'd like to see how the changed page size affects the speed of the sync. 

**Added feature flags**
- increase-builds-page-size-coef

### DIFF
--- a/src/sync/build.ts
+++ b/src/sync/build.ts
@@ -5,6 +5,8 @@ import { GitHubInstallationClient } from "../github/client/github-installation-c
 import { transformWorkflow } from "../transforms/transform-workflow";
 import { GitHubWorkflowPayload } from "~/src/interfaces/github";
 import { transformRepositoryDevInfoBulk } from "~/src/transforms/transform-repository";
+import { numberFlag, NumberFlags } from "config/feature-flags";
+
 type BuildWithCursor = { cursor: number } & Octokit.ActionsListRepoWorkflowRunsResponse;
 
 // TODO: add types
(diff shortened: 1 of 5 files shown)


src/sync/build.ts
```diff
@@ -25,17 +27,27 @@ const getTransformedBuilds = async (workflowRun, gitHubInstallationClient, logge
 export const getBuildTask = async (
 	logger: Logger,
 	gitHubInstallationClient: GitHubInstallationClient,
-	_jiraHost: string,
+	jiraHost: string,
 	repository: Repository,
-	cursor: string | number = 1,
-	perPage?: number
+	cursorOrigStr: string | number = 1,
```
rachellerathbone, line 32: Nit: I can see we're now using cursor and perPage in the function block but is there any other names we could use here? perPageOrig, cursorOrigStr, cursorOrig aren't super clear and it gets a little hard keeping track of what's what. I'm okay with letting this through now because it's an experiment but would probably be good to update these for readability sakes if we decide to keep this in the code.

src/sync/sync.types.ts
```diff
@@ -14,9 +14,9 @@ export interface TaskProcessors {
 		gitHubInstallationClient: GitHubInstallationClient,
 		jiraHost: string,
 		repository: Repository,
-		cursor?: string | number,
-		perPage?: number,
-		messagePayload?: BackfillMessagePayload
+		cursor: string | number | undefined,
```
rachellerathbone, line 17: What's the reason for this change? I don't think cursor is a required arg. There are multiple places in the code where we are setting the values for cursor/perPage if it isn't passed e.g. in `getBuildTask` for cursor -> `cursor: string | number = 1`

src/sync/build.ts
bgvozdev: I'm open for suggestions :)

src/sync/sync.types.ts
bgvozdev: Just typing @rachellerathbone , I'm removing optionality: we always pass over perPage and messagePayload

bgvozdev: also, typescript would've yelled if something was wrong, however tests are green ...

rachellerathbone: lol you put a lot of faith in our tests.

rachellerathbone (approved):

PR #1943, opened 2023-03-06
Arc 1995 keep msg in queue for exceed ratelimit

gxueatlassian:
**What's in this PR?**
Fix the bug of preemtively rate limiting.

**Why**
1. It sometimes reschdule the msg immediately. Fixed via added minimum delay. This seems to be something strange from the github side. It return the reset time that's almost exactly same as current time.
2. It use changeMsgVisiblity, which will incurr exceed retry count and lost the msg, customer sync will get stuck. Fixed via sending a new message. 

**Added feature flags**
N/A

**Affected issues**  
ARC-1995

**How has this been tested?**  
Unit test.

On stage:

When setting threshold to 1, rate limit kicked in, new msg was scheduled
image

Reset the threshold back to normal, and after 15min (the maximum new msg delay), backfill success again.
image

image

**Whats Next?**
N/A

### DIFF
--- a/src/sqs/sqs.ts
+++ b/src/sqs/sqs.ts
@@ -82,7 +82,7 @@ export class SqsQueue<MessagePayload extends BaseMessagePayload> {
 
 		const sendMessageResult = await this.sqs.sendMessage(params)
 			.promise();
-		logger.info({ delaySeconds: delaySec, messageId: sendMessageResult.MessageId }, `Successfully added message to sqs queue messageId: ${sendMessageResult.MessageId}`);
+		logger.info({ delaySeconds: delaySec, newMessageId: sendMessageResult.MessageId }, `Successfully added message to sqs queue messageId: ${sendMessageResult.MessageId}`);
 		statsd.increment(sqsQueueMetrics.sent, this.metricsTags);
 		return sendMessageResult;
 	}
--- a/src/sqs/sqs.ts
+++ b/src/sqs/sqs.ts
@@ -6,7 +6,7 @@ import { v4 as uuidv4 } from "uuid";
 import { statsd } from "config/statsd";
 import { Tags } from "hot-shots";
 import { sqsQueueMetrics } from "config/metric-names";
-import { ErrorHandler, ErrorHandlingResult, MessageHandler, QueueSettings, SQSContext, SQSMessageContext, SqsTimeoutError } from "~/src/sqs/sqs.types";
+import { ErrorHandler, ErrorHandlingResult, MessageHandler, QueueSettings, SQSContext, SQSMessageContext, BaseMessagePayload, SqsTimeoutError } from "~/src/sqs/sqs.types";
 import { stringFlag, StringFlags } from "config/feature-flags";
 import { preemptiveRateLimitCheck } from "utils/preemptive-rate-limit";
 
@@ -32,7 +32,7 @@ const isNotRetryable = (errorHandlingResult: ErrorHandlingResult) => {
  *
  * Allows sending SQS messages, as well as listening to the queue messages.
  */
-export class SqsQueue<MessagePayload> {
+export class SqsQueue<MessagePayload extends BaseMessagePayload> {
 	readonly queueUrl: string;
 	readonly queueName: string;
 	readonly queueRegion: string;
@@ -84,6 +84,7 @@ export class SqsQueue<MessagePayload> {
 			.promise();
 		logger.info(`Successfully added message to sqs queue messageId: ${sendMessageResult.MessageId}`);
 		statsd.increment(sqsQueueMetrics.sent, this.metricsTags);
+		return sendMessageResult;
 	}
 
 	/**
@@ -267,8 +268,11 @@ export class SqsQueue<MessagePayload> {
 		try {
 			const messageProcessingStartTime = Date.now();
 
-			if (await preemptiveRateLimitCheck(context, this)) {
-				context.log.info("Preemptive rate limit threshold exceeded");
+			const rateLimitCheckResult = await preemptiveRateLimitCheck(context, this);
+			if (rateLimitCheckResult.isExceedThreshold) {
+				const { MessageId } = await this.sendMessage(payload, rateLimitCheckResult.resetTimeInSeconds, context.log);
+				await this.deleteMessage(context);
+				context.log.info({ NewMessageId: MessageId }, "Preemptive rate limit threshold exceeded, rescheduled new one and deleted the origin msg");
 				return;
 			}
 
--- a/src/sqs/sqs.ts
+++ b/src/sqs/sqs.ts
@@ -82,7 +82,7 @@ export class SqsQueue<MessagePayload extends BaseMessagePayload> {
 
 		const sendMessageResult = await this.sqs.sendMessage(params)
 			.promise();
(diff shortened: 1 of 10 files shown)


src/sync/branch.test.ts
```diff
@@ -135,7 +135,6 @@ describe("sync/branches", () => {
 				.repoSyncStatePendingForBranches()
 				.create();
 
-			mocked(sqsQueues.backfill.sendMessage).mockResolvedValue(Promise.resolve());
```
gxueatlassian, line 138: Test doesn't need anything from the mock result. So remove the mock impl instead of feeding real object, which need to comply to a certain ts structure.

bgvozdev:
Should we test this in stg? Maybe we could force it to occasionally (random) to false-trigger the check?

gxueatlassian:
> Should we test this in stg? Maybe we could force it to occasionally (random) to false-trigger the check?

Good questions. Let's think about it.

src/sqs/sqs.ts
```diff
@@ -82,8 +82,9 @@ export class SqsQueue<MessagePayload> {
 
 		const sendMessageResult = await this.sqs.sendMessage(params)
 			.promise();
-		logger.info(`Successfully added message to sqs queue messageId: ${sendMessageResult.MessageId}`);
+		logger.info({ DelaySeconds: delaySec, MessageId: sendMessageResult.MessageId }, `Successfully added message to sqs queue messageId: ${sendMessageResult.MessageId}`);
```
bgvozdev, line 85: nit: I think the convention is to use camelCase (starting with lower letter) for the properties
Also, probably I'd opted for something more verbose for the message ID (newMessageId)

src/sqs/sqs.ts
```diff
@@ -256,7 +257,10 @@ export class SqsQueue<MessagePayload> {
 				messageId: message.MessageId,
 				executionId: uuidv4(),
 				queue: this.queueName,
-				...(payload?.jiraHost ? { jiraHost: payload.jiraHost } : { })
+				jiraHost: payload?.jiraHost,
+				installationId: payload?.installationId,
+				gitHubAppId: payload?.gitHubAppConfig?.gitHubAppId,
+				webhookId: payload?.webhookId
```
bgvozdev, line 263: nit would these be visible in splunk as "jiraHost: undefined", are we fine with that?

src/sqs/sqs.ts
```diff
@@ -267,8 +271,12 @@ export class SqsQueue<MessagePayload> {
...
 			const messageProcessingStartTime = Date.now();
 
-			if (await preemptiveRateLimitCheck(context, this)) {
-				context.log.info("Preemptive rate limit threshold exceeded");
+			const rateLimitCheckResult = await preemptiveRateLimitCheck(context, this);
+			if (rateLimitCheckResult.isExceedThreshold) {
+				//threshold exceeded, retry via sending a new message.
+				const { MessageId } = await this.sendMessage(payload, rateLimitCheckResult.resetTimeInSeconds, context.log);
+				await this.deleteMessage(context);
+				context.log.info({ NewMessageId: MessageId, DeletedMessageId: message.MessageId }, "Preemptive rate limit threshold exceeded, rescheduled new one and deleted the origin msg");
```
bgvozdev, line 279: nit: same, I think we have a different naming convention for the object properties

src/sqs/sqs.ts
```diff
@@ -267,8 +271,12 @@ export class SqsQueue<MessagePayload> {
 		try {
 			const messageProcessingStartTime = Date.now();
 
-			if (await preemptiveRateLimitCheck(context, this)) {
-				context.log.info("Preemptive rate limit threshold exceeded");
+			const rateLimitCheckResult = await preemptiveRateLimitCheck(context, this);
+			if (rateLimitCheckResult.isExceedThreshold) {
+				//threshold exceeded, retry via sending a new message.
```
bgvozdev, line 276: nit: comment duplicating code

bgvozdev: I'd rather explain why we are doing this, not what we are doing

src/sqs/sqs.types.ts
```diff
@@ -120,45 +120,38 @@ export interface SQSContext {
 	log: Logger;
 }
 
-export type BranchMessagePayload = {
+export type BaseMessagePayload = {
```
bgvozdev, line 123: Very like!

src/sqs/sqs.ts
gxueatlassian: Good question, I think all msg in our queue should have jiraHost. 
And even if it is undefined, is there any concern on it?

src/util/preemptive-rate-limit.ts
```diff
@@ -3,17 +3,23 @@ import { Octokit } from "@octokit/rest";
...
 import { SqsQueue } from "~/src/sqs/sqs";
-import Logger from "bunyan";
+import type { BaseMessagePayload } from "~/src/sqs/sqs.types";
 
 // List of queues we want to apply the preemptive rate limiting on
 const TARGETTED_QUEUES = ["backfill"];
-const DEFAULT_PREEMPTY_RATELIMIT_DELAY_IN_SECONDS = 30 * 60; //30 minutes
+export const DEFAULT_PREEMPTY_RATELIMIT_DELAY_IN_SECONDS = 10 * 60; //10 minutes
+
+type PreemptyRateLimitCheckResult = {
```
bgvozdev, line 12: Nice!

src/util/preemptive-rate-limit.ts
```diff
@@ -25,37 +31,33 @@ export const preemptiveRateLimitCheck = async (context: SQSMessageContext<any>,
...
 	const gitHubInstallationClient = await createInstallationClient(installationId, jiraHost, context.log, gitHubAppId);
 	return await gitHubInstallationClient.getRateLimit();
 };
 
-const getRateResetTime = (rateLimitResponse: Octokit.RateLimitGetResponse, log: Logger): number => {
+const getRateResetTimeInSeconds = (rateLimitResponse: Octokit.RateLimitGetResponse): number => {
 	// Get the furthest away rate reset to ensure we don't exhaust the other one too quickly
-	const resetEpochDateTime = Math.max(rateLimitResponse?.resources?.core?.reset, rateLimitResponse?.resources?.graphql?.reset);
+	const resetEpochDateTimeInSeconds = Math.max(rateLimitResponse?.resources?.core?.reset, rateLimitResponse?.resources?.graphql?.reset);
 	// Get the difference in seconds between now and reset time
```
bgvozdev, line 59: A comment duplicating code

bgvozdev (dismissed):

src/sqs/sqs.ts
gxueatlassian: fixed.

src/sqs/sqs.ts
gxueatlassian: Close this, as undefined is "better"? As it is more explicit what's going? Lmk if you want to reopen it.

src/sqs/sqs.ts
gxueatlassian: fixed.

src/sqs/sqs.ts
gxueatlassian: fixed.

src/util/preemptive-rate-limit.ts
gxueatlassian: fixed.

bgvozdev (approved):

bgvozdev (approved):

PR #1952, opened 2023-03-08
ARC-787: Add SQS operations for Microscope DLQ panel

atrigueiro:
**What's in this PR?**
Adding backend endpoints for managing DLQ in Microscope

**Why**
Because the panel doesn't work without the back end and we need to replay the DLQ :)

**Added feature flags**

**Affected issues**  
ARC-787

**How has this been tested?**  
Deployed and tested in staging
Screenshot 2023-03-24 at 2 16 15 pm

I haven't added a UT but the endpoint is protected by POCO and can only be called via Microscope

**Whats Next?**


src/routes/microscope/operations/microscope-dlq-operations.ts
```diff
@@ -0,0 +1,88 @@
...
+
+const { DlqService } = optionalRequire("@atlassian/sqs-queue-dlq-service", true) || {};
+
+const log: Logger = getLogger("microscope-dlq");
+
+type RequeueMessageRequest = {
+	messageId: string;
+	receiptHandle: string;
+	body: string;
+	messageAttributes: object;
```
rachellerathbone, line 14: Do we know what properties we're expecting for this? Better to pass another type to this if we do.

src/routes/microscope/operations/microscope-dlq-operations.ts
```diff
@@ -0,0 +1,88 @@
...
+	log.info("No dlqService package - skipping dlq operations");
+} else {
+	dlqServiceClient = new DlqService(log, {
+		sqs: {
+			region: process.env.SQS_PUSH_QUEUE_REGION
+		}
+	});
+}
+
+export const microscopeDlqHealthcheck = async (_: Request, res: Response): Promise<void> => {
```
rachellerathbone, line 33: Think we should add tests for these middlewares?

atrigueiro: Sooo testing those are not as straight forward because of the optional package (we can't just mock it). As I mentioned on the PR description, I haven't added a UT but the endpoint is protected by POCO and can only be called via Microscope. I do intend to go back and try to find a way to test those but had a chat with Sass and this has been going on for too long so it shouldn't be a blocker.

src/routes/microscope/operations/microscope-dlq-operations.ts
atrigueiro: We don't - it's whatever the AWS API client call returns as per https://bitbucket.org/atlassian/dlq-manager/src/master/

rachellerathbone (approved):

PR #1957, opened 2023-03-09
NONE: add reset failed task endpoint

bgvozdev:
**What's in this PR?**
Adding admin endpoint to reset failed tasks

**Why**
we have no way now

### DIFF
--- a/src/routes/api/api-reset-subscription-failed-tasks.ts
+++ b/src/routes/api/api-reset-subscription-failed-tasks.ts
@@ -1,10 +1,16 @@
 import { Request, Response } from "express";
 import { RepoSyncState } from "models/reposyncstate";
 import { Subscription } from "models/subscription";
+import { TaskType } from "../../sync/sync.types";
+
+interface OutputLogRecord {
+	repoSyncStateId: number,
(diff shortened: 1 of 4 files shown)


src/routes/api/api-reset-subscription-failed-tasks.ts
```diff
@@ -0,0 +1,39 @@
...
+	}
+
+	const subscription = await Subscription.findByPk(subscriptionId);
+
+	if (!subscription) {
+		res.status(400).send("subscription not found");
+		return;
+	}
+
+	const log: any[] = [];
```
rachellerathbone, line 21: Can we tighten up the typing here? Chucking in any isn't much different to not adding a type at all. Something like:
```
interface Log {
	repoSyncStateId: number,
	targetTask: "pull" | "branch" | "commit" | "build" | "deployment" | string
}
...
const log: Log[] = [];
```
should do the trick

bgvozdev: good point :)

rachellerathbone (approved):

PR #1968, opened 2023-03-13
None refactor parallel fetching into common component

bgvozdev:
**What's in this PR?**
Refactoring parallel fetching logic into a common component that could be used across different entities

**Why**
1. Want to do same for builds
2. We might want to extend it to other big customers

**Added feature flags**
Already behind one.

**Whats Next?**
Use it to fetch builds in parallel

### DIFF
--- a/src/sync/parallel-page-fetcher.ts
+++ b/src/sync/parallel-page-fetcher.ts
@@ -9,7 +9,7 @@ const mergeJiraPayload = (jiraPayload1?: any, jiraPayload2?: any) => {
 	const pullRequests = [...(jiraPayload1?.pullRequests || []), ...(jiraPayload2?.pullRequests || [])];
 	const builds = [...(jiraPayload1?.builds || []), ...(jiraPayload2?.builds || [])];
 
-	if (pullRequests.length == 0 && builds.length == 0) {
+	if (pullRequests.length === 0 && builds.length === 0) {
 		return undefined;
 	}
 
--- a/src/sync/parallel-page-fetcher.ts
+++ b/src/sync/parallel-page-fetcher.ts
@@ -15,8 +15,8 @@ const mergeJiraPayload = (jiraPayload1?: any, jiraPayload2?: any) => {
 
 	return {
 		...(jiraPayload1 || jiraPayload2),
-		... (pullRequests.length > 0 ? { pullRequests } : {}),
-		... (builds.length > 0 ? { builds } : {})
+		...(pullRequests.length > 0 ? { pullRequests } : {}),
+		...(builds.length > 0 ? { builds } : {})
 	};
 };
 
--- a/src/sync/parallel-page-fetcher.ts
+++ b/src/sync/parallel-page-fetcher.ts
@@ -1,0 +1,45 @@
+import { TaskPayload } from "~/src/sync/sync.types";
+
+const mergeJiraPayload = (jiraPayload1?: any, jiraPayload2?: any) => {
+	if (!jiraPayload1 && !jiraPayload2) {
+		return undefined;
+	}
+
+	// TODO: add more
+	const pullRequests = [...(jiraPayload1?.pullRequests || []), ...(jiraPayload2?.pullRequests || [])];
+	const builds = [...(jiraPayload1?.builds || []), ...(jiraPayload2?.builds || [])];
+
+	if (pullRequests.length == 0 && builds.length == 0) {
+		return undefined;
+	}
+
+	return {
+		...(jiraPayload1 || jiraPayload2),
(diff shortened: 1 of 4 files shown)


src/sync/parallel-page-fetcher.ts
```diff
@@ -0,0 +1,45 @@
...
+const mergeJiraPayload = (jiraPayload1?: any, jiraPayload2?: any) => {
+	if (!jiraPayload1 && !jiraPayload2) {
+		return undefined;
+	}
+
+	// TODO: add more
+	const pullRequests = [...(jiraPayload1?.pullRequests || []), ...(jiraPayload2?.pullRequests || [])];
+	const builds = [...(jiraPayload1?.builds || []), ...(jiraPayload2?.builds || [])];
+
+	if (pullRequests.length == 0 && builds.length == 0) {
```
rachellerathbone, line 12: Think we want to change == to ===. If we only use == we are stating we are okay with any value that is 'loosely' equals to meaning a value of '0'/false would return true in this case. ===, strictly equals, ensures this is only true is we receive the numerical 0. #javascript

src/sync/parallel-page-fetcher.ts
```diff
@@ -0,0 +1,45 @@
...
+	const pullRequests = [...(jiraPayload1?.pullRequests || []), ...(jiraPayload2?.pullRequests || [])];
+	const builds = [...(jiraPayload1?.builds || []), ...(jiraPayload2?.builds || [])];
+
+	if (pullRequests.length === 0 && builds.length === 0) {
+		return undefined;
+	}
+
+	return {
+		...(jiraPayload1 || jiraPayload2),
+		... (pullRequests.length > 0 ? { pullRequests } : {}),
```
rachellerathbone, line 18: Nit: remove the space before ... Doesn't really matter but just figured we don't have a space for the first spread operator but do for the 2nd 🤷

src/sync/pull-request.ts
```diff
@@ -70,42 +60,37 @@ export const getPullRequestTask = async (
...
-				scaledPageSize
-			));
+			(scaledPageNoToFetch) =>
+				doGetPullRequestTask(
+					logger, gitHubInstallationClient, jiraHost, repository,
+					scaledPageNoToFetch,
+					scaledPageSize
+				)
+		);
+		if ((data.edges?.length) || 0 > 0) {
```
rachellerathbone, line 83: 0 > 0?

bgvozdev: hehe Yeah, looks like it is always defined. Well spotted!

src/sync/pull-request.ts
```diff
@@ -70,42 +60,37 @@ export const getPullRequestTask = async (
...
-				scaledPageSize
-			));
+			(scaledPageNoToFetch) =>
+				doGetPullRequestTask(
+					logger, gitHubInstallationClient, jiraHost, repository,
+					scaledPageNoToFetch,
+					scaledPageSize
+				)
+		);
+		if ((data.edges?.length || 0) > 0) {
```
rachellerathbone: Kinda makes more sense but wouldn't just `data.edges?.length` evaluate to the same thing? So if edges has at least one item, `data.edges?.length` would evaluate to true, otherwise it would skip this conditional. And the ternary adds a safety check that edges exists. Think we can ditch the other logic here.

bgvozdev: I removed the IF altogether @rachellerathbone , it should be always defined (added `|| []` just in case)

src/sync/pull-request.ts
```diff
@@ -70,42 +60,35 @@ export const getPullRequestTask = async (
...
+			shouldFetchNextPageInParallel ? 2 : 1,
+			scaledCursor,
+			(scaledPageNoToFetch) =>
+				doGetPullRequestTask(
+					logger, gitHubInstallationClient, jiraHost, repository,
+					scaledPageNoToFetch,
+					scaledPageSize
+				)
+		);
+		(data.edges || []).forEach(edge => {
```
rachellerathbone, line 83: Are we saying we want to loop over an empty array if data.edges doesn't exist is empty? Think we need to add a test to see what happens if edges doesn't exist on data/edges is empty.

bgvozdev: We already have such test: http://github.com/atlassian/github-for-jira/blob/b68a8348d0a73ac2a71c5c99bd59a6437cdcff0d/src/sync/pull-requests.test.ts#L408-L408

rachellerathbone: Good stuff.

rachellerathbone (approved):

PR #1969, opened 2023-03-13
NONE: fetch builds in parallel

bgvozdev:
**What's in this PR?**
Fetching builds in parallel for one customers

https://github.com/atlassian/github-for-jira/pull/1968 needs to be merged first (or along)

**Why**
They have plenty of rate limiting quota and not using it, and huge number of builds

**Added feature flags**
Already under one

### DIFF
--- a/src/sync/build.test.ts
+++ b/src/sync/build.test.ts
@@ -129,7 +129,7 @@ describe("sync/builds", () => {
 		expect((await RepoSyncState.findByPk(repoSyncState!.id)).buildCursor).toEqual(String(Number(ORIGINAL_BUILDS_CURSOR) + 5));
 	});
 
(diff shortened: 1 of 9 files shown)


src/sync/build.test.ts
```diff
@@ -39,6 +39,7 @@ describe("sync/builds", () => {
 	};
 
 	let repoSyncState: RepoSyncState;
+	const ORIGINAL_BUILDS_CURSOR = 21;
```
rachellerathbone, line 42: What's the significance of 21?

rachellerathbone (approved):

src/sync/build.test.ts
bgvozdev: Just a random number @rachellerathbone , the position of the builds cursor for the tests

PR #1972, opened 2023-03-14
Arc 1975 fix commits from date var all time

gxueatlassian:
**What's in this PR?**
1. Fix how to determine the commits from date
2. Add admin api to recover existing commits from date.

**Why**
Current in the ff launch darkly, the "disable" state return a large number `1654506600000` <-- which is roughly like 1970-10-08. The origin logic has flaw, it takes it to be the commitsFromDate.
The fix is to use -1 as indicator for disable state so that it is more stable.

**Added feature flags**
Related to sync-main-commit-time-limit and sync-branch-commit-time-limit

**Affected issues**  
ARC-1975

**How has this been tested?**  
Unit test and local db test.

**Whats Next?**
N/A

### DIFF
--- a/src/routes/api/commits-from-date/recover-commits-from-dates.ts
+++ b/src/routes/api/commits-from-date/recover-commits-from-dates.ts
@@ -1,0 +1,62 @@
+import { Request, Response } from "express";
+import { Subscription } from "models/subscription";
+import { Op } from "sequelize";
+import { getLogger } from "config/logger";
+import safeJsonStringify from "safe-json-stringify";
+
+const DEFAULT_BATCH_SIZE = 5000;
+
+export const RecoverCommitsFromDatePost = async (req: Request, res: Response): Promise<void> => {
+
+	const log = getLogger("RecoverCommitsFromDatePost");
+
+	try {
+
+		const startSubscriptionId = Number(req.query.startSubscriptionId) || 0;
+		const batchSize = Number(req.query.batchSize) || DEFAULT_BATCH_SIZE;
+
+		const info = (msg: string) => {
+			log.info(msg);
+			res.write(msg + "\n");
+		};
+
+		res.status(200);
+
+		info(`Start recovering commits from date for Subscription starting from ${startSubscriptionId} with batch size ${batchSize}`);
+
+		const subscriptionToBeUpdated: Subscription[] = await Subscription.findAll({
+			where: {
+				[Op.and]: {
+					"backfillSince": {
+						[Op.lt]: new Date(1980, 1, 1)
+					},
+					"id": {
+						[Op.gt]: startSubscriptionId
+					}
+				}
+			},
+			order: [ ["id", "ASC"] ]
+		});
+		info(`Found ${subscriptionToBeUpdated.length} subscriptions potentially to look at`);
+
+		let count = 0;
+		let lastId: number | undefined;
+		for (const sub of subscriptionToBeUpdated) {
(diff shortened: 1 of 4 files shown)


src/sync/sync-utils.ts
```diff
@@ -110,8 +110,8 @@ export const getCommitSinceDate = async (jiraHost: string, flagName: NumberFlags
 		return new Date(commitsFromDate);
 	}
 	const timeCutoffMsecs = await numberFlag(flagName, NaN, jiraHost);
-	if (!timeCutoffMsecs) {
-		return;
+	if (!timeCutoffMsecs || timeCutoffMsecs === -1) {
+		return undefined;
```
gxueatlassian, line 114: Just to be more explicit to return `undefined`

joshkay10 (approved):

src/routes/api/commits-from-date/recover-commits-from-dates.ts
```diff
@@ -0,0 +1,62 @@
...
+const DEFAULT_BATCH_SIZE = 5000;
+
+export const RecoverCommitsFromDatePost = async (req: Request, res: Response): Promise<void> => {
+
+	const log = getLogger("RecoverCommitsFromDatePost");
+
+	try {
+
+		const startSubscriptionId = Number(req.query.startSubscriptionId) || 0;
+		const batchSize = Number(req.query.batchSize) || DEFAULT_BATCH_SIZE;
```
Harminder84, line 16: `batchSize` is not being used anywhere.

src/routes/api/commits-from-date/recover-commits-from-dates.ts
```diff
@@ -0,0 +1,62 @@
...
+				[Op.and]: {
+					"backfillSince": {
+						[Op.lt]: new Date(1980, 1, 1)
+					},
+					"id": {
+						[Op.gt]: startSubscriptionId
+					}
+				}
+			},
+			order: [ ["id", "ASC"] ]
```
Harminder84, line 38: I think, `limit` is missing here

src/routes/api/commits-from-date/recover-commits-from-dates.ts
gxueatlassian: fixed.

src/routes/api/commits-from-date/recover-commits-from-dates.ts
gxueatlassian: fixed.

Harminder84 (approved):

src/routes/api/commits-from-date/recover-commits-from-dates.ts
```diff
@@ -0,0 +1,63 @@
...
+const DEFAULT_BATCH_SIZE = 5000;
+
+export const RecoverCommitsFromDatePost = async (req: Request, res: Response): Promise<void> => {
+
+	const log = getLogger("RecoverCommitsFromDatePost");
+
+	try {
+
+		const startSubscriptionId = Number(req.query.startSubscriptionId) || 0;
+		const batchSize = Number(req.query.batchSize) || DEFAULT_BATCH_SIZE;
```
joshkay10, line 16: /nit
Do we want a upper limit to batch size to stop people being monsters, 

~~also do we want any validation that the startSubscriptionId is valid...... i guess we are probably ok since it has to be a number and the where caluse will cause it to return nothing.. so thats fine~~

joshkay10 (approved):

PR #1976, opened 2023-03-15
Fix for the Open Redirect Vulnerability in `github-redirect.js`

krazziekay:
**What's in this PR?**
- A new view was added for showing loading when creating automatic apps in the Enterprise flow.
- Removed the extra Ajax call in `github-redirect.js` for fetching the manifest. Instead the manifest is now created in the frontend.
- The `baseUrl` has now been removed from the query parameters at the time of app creation. `baseUrl` is now passed from the js as a part of the child window.
- Route parameter updated for `GitHubManifestCompleteGet` & test cases updated accordingly.

**Why**
- [VULN-1033674](https://asecurityteam.atlassian.net/browse/VULN-1033674)

**How has this been tested?**  
- Staging
- Local


views/session.hbs
```diff
@@ -39,13 +39,9 @@
...
-        $(function() {
-          window.location = "{{{redirectUrl}}}";
-        });
-      </script>
-{{/if}}
+<script nonce="{{nonce}}">
+  $(function() {
+    window.location = "{{{redirectUrl}}}";
+  });
+</script>
```
mboudreau, line 46: scripts need to be inside `<body>` tag

mboudreau (changes requested):
small change

views/session.hbs
krazziekay: Moved inside body tag.

static/js/jira-select-card-option.js
```diff
@@ -65,9 +65,10 @@ $(document).ready(function() {
 			);
 		} else {
 			AP.context.getToken(function(token) {
-				const child = openChildWindow("/session?ghRedirect=to&autoApp=1&baseUrl=" + gitHubServerBaseUrl);
+				const child = openChildWindow("/github-manifest");
 				child.window.jiraHost = jiraHost;
 				child.window.jwt = token;
+				child.window.gheHost = gitHubServerBaseUrl;
```
Harminder84, line 71: I think, `gheHost` is being fetched from `texfield` and can be pointed to mock GHE server URL. Won't it give same VULN? How will you make sure it is legitimate URL?

krazziekay: Yup, that can be another VULN. One idea is to store this `gheHost` temporarily in redis or someplace else, but that will be done in another [ticket.](https://softwareteams.atlassian.net/browse/ARC-2035)
These changes are to fix the redirecting VULN for the moment.

mboudreau: the original vuln was an open endpoint which could redirect.  This particular endpoint is authenticated behind a JWT token (not open) making it essentially impossible to hijack

mboudreau (approved):

PR #1978, opened 2023-03-15
Arc 2033 apply backfill from date on deployment

gxueatlassian:
**What's in this PR?**
Add incremental backfill to deployment tasks

**Why**
Reduce the load of our servers and make it quicker for customers.

**Added feature flags**
backfill-algorithm-incremental

**Affected issues**  
ARC-2033

**How has this been tested?**  
Unit test
Local test

**Whats Next?**
Branches.


src/github/client/github-queries.ts
```diff
@@ -361,12 +362,50 @@ export type getDeploymentsResponse = {
 	}
 };
 
+export const getDeploymentsQueryByCreatedAtDesc = `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String) {
+  repository(owner: $owner, name: $repo){
+    deployments(first: $per_page, after: $cursor, orderBy: { direction: DESC, field: CREATED_AT }) {
```
gxueatlassian, line 367: **IMPORTANT**
This is interesting. The current order of the graphql seems to be ASC. 
However, in order to make the `fromDate` work, I'll have to change it to DESC. 

The problem might be it will affect existing customers. I think let's have a chat on how to deal with it.

src/github/client/github-queries.ts
```diff
@@ -361,12 +362,50 @@ export type getDeploymentsResponse = {
...
+  }
+}`;
+
 export const getDeploymentsQuery = `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String) {
   repository(owner: $owner, name: $repo){
     deployments(first: $per_page, after: $cursor) {
       edges {
         cursor
         node {
+					createdAt
```
gxueatlassian, line 408: Not sure I need to have this, as it is to match the typing. Otherwise typing can be `?`?

src/github/client/github-installation-client.ts
```diff
@@ -263,6 +264,18 @@ export class GitHubInstallationClient extends GitHubClient {
 		return response?.data?.data;
 	}
 
+	public async getDeploymentsPageByCreatedAtDesc(owner: string, repoName: string, perPage?: number, cursor?: string | number): Promise<getDeploymentsResponse> {
```
joshkay10, line 267: Should the cursor always be a string if its a graphql call?
`cursor?: string | number`

gxueatlassian: Make sense. Although I tried to rectify it, but seems involve quite a lots of files. So will keep it to next PR.

src/github/client/github-queries.ts
gxueatlassian: So I added an admin api to reset the cursor to null when the `deploymentStatus` is null or failed or pending.

joshkay10 (approved):

src/sync/deployment.ts
```diff
@@ -50,7 +57,19 @@ const getTransformedDeployments = async (deployments, gitHubInstallationClient:
...
 export const getDeploymentTask = async (logger: Logger, gitHubInstallationClient: GitHubInstallationClient, jiraHost: string, repository: Repository, cursor?: string | number, perPage?: number, data?: BackfillMessagePayload) => {
 	logger.debug("Syncing Deployments: started");
-	const { edges, deployments } = await fetchDeployments(gitHubInstallationClient, repository, cursor, perPage);
+
+	const shouldUseIncrementalBackfill = await booleanFlag(BooleanFlags.USE_BACKFILL_ALGORITHM_INCREMENTAL, jiraHost);
+	const { edges, deployments } = await fetchDeployments(shouldUseIncrementalBackfill, gitHubInstallationClient, repository, cursor, perPage);
+
+	if (shouldUseIncrementalBackfill) {
+		const fromDate = data?.commitsFromDate ? new Date(data?.commitsFromDate) : undefined;
+		if (isAllEdgesEarlierThanFromDate(edges, fromDate)) {
```
Harminder84, line 66: Shouldn't we use graphql  `since` filter rather than manual filtering?

src/routes/api/api-router.ts
```diff
@@ -142,6 +143,7 @@ ApiRouter.post("/recover-client-key", RecoverClientKeyPost);
 ApiRouter.post("/re-encrypt-ghes-app", ReEncryptGitHubServerAppKeysPost);
 ApiRouter.use("/data-cleanup", DataCleanupRouter);
 ApiRouter.post("/recover-commits-from-date", RecoverCommitsFromDatePost);
+ApiRouter.post("/reset-failed-pending-deployment-cursor", ResetFailedAndPendingDeploymentCursorPost);
```
Harminder84, line 147: What is the use case of this admin API? Shouldn't we create a generic API to reset cursor for build, deployment, commit and other tasks?

src/sync/deployment.ts
gxueatlassian: Unfortunately the `deployment` in graphql doesn't have `since` 😞

src/routes/api/api-router.ts
gxueatlassian: Nah, I'd prefer a specific api for this PR. 
I want to remove the api once done. 
Do not want more api along with the existing TWO restart-sync api admin endpoints.

PR #1981, opened 2023-03-16
Arc 2033 apply backfill from date on deployment  turn on desc

gxueatlassian:
**What's in this PR?**
Use the desc in sort order when getting deployments data.

**Why**
I was planning to use FF `USE_BACKFILL_ALGORITHM_INCREMENTAL` to controller the rollout of this `desc ordering` for deployment. But I realised we are going to do percentage rollout of that ff, and I can't really tell which is which when calling the admin api to reset cursor for failed/pending deployment. 
So I'd like to have this turn on and ready now instead of later.

**Added feature flags**
N/A

**Affected issues**  
ARC-2033

**How has this been tested?**  
Unit test and local

**Whats Next?**
N/A

### DIFF
--- a/src/github/client/github-queries.ts
+++ b/src/github/client/github-queries.ts
@@ -362,46 +362,9 @@ export type getDeploymentsResponse = {
 	}
 };
 
-export const getDeploymentsQueryByCreatedAtDesc = `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String) {
-  repository(owner: $owner, name: $repo){
-    deployments(first: $per_page, after: $cursor, orderBy: { direction: DESC, field: CREATED_AT }) {
-      edges {
-        cursor
-        node {
-					createdAt
-          repository {
-            id: databaseId
-            node_id: id
-            name
-            owner {
-              login
-            }
-          }
-          databaseId
-          commitOid
-          task
-          ref {
-            name
-            id
-          }
-          environment
-          description
-          latestStatus {
-            environmentUrl
-            logUrl
-            state
-            id
-            updatedAt
-          }
-        }
-      }
-    }
-  }
-}`;
-
 export const getDeploymentsQuery = `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String) {
   repository(owner: $owner, name: $repo){
-    deployments(first: $per_page, after: $cursor) {
+    deployments(first: $per_page, after: $cursor, orderBy: { direction: DESC, field: CREATED_AT }) {
       edges {
         cursor
         node {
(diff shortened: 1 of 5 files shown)


src/routes/api/commits-from-date/reset-failed-and-pending-deployment-cursors.ts
```diff
@@ -29,10 +29,7 @@ export const ResetFailedAndPendingDeploymentCursorPost = async (req: Request, re
 			where: {
 				[Op.and]: {
 					"deploymentStatus": {
-						[Op.or]: {
-							[Op.is]: undefined,
-							[Op.in]: ["pending", "failed"]
-						}
+						[Op.in]: ["pending", "failed"]
```
bgvozdev, line 32: curious why this was changed?

src/github/client/github-queries.ts
```diff
@@ -362,46 +362,9 @@ export type getDeploymentsResponse = {
...
-        }
-      }
-    }
-  }
-}`;
-
 export const getDeploymentsQuery = `query ($owner: String!, $repo: String!, $per_page: Int!, $cursor: String) {
   repository(owner: $owner, name: $repo){
-    deployments(first: $per_page, after: $cursor) {
+    deployments(first: $per_page, after: $cursor, orderBy: { direction: DESC, field: CREATED_AT }) {
```
bgvozdev, line 367: What's the rollout strategy for this?

bgvozdev (approved):

src/routes/api/commits-from-date/reset-failed-and-pending-deployment-cursors.ts
gxueatlassian: There're 4 millions rows of `null` lol
But when I checked the Db, they are not needed to be reset
```
=> select count(*) from "RepoSyncStates" where ( "deploymentStatus" is null and "deploymentCursor" is not null);
 count
-------
     0
(1 row)
```

I can however to fine tune above sql to include `status` is null and `cursor` is not null, but wonder will that make above code too complicated? Thoughts?

src/github/client/github-queries.ts
gxueatlassian: No strategy, deploy, hit the api.
Or I could add a ff flag. So deploy, turn on ff, hit the api.
Thoughts?

bgvozdev: Right, so you would:
1. get the IDs of the records that are in sync now (by looking at updatedAt?)
2. trigger endpoint with these IDs?

At which point of time would you collect them, because there could be race-conditions:
- you turned on FF
- a sync has stopped // <-- this will be broken
- you triggered endpoint
?

I was thinking of maybe doing something like:
- collect all active IDs before deploy
- collect all active IDs after deploy
- merge them together - these need to be resynced
- drop statuses
- trigger partial resync for the affected installationIds

wdyt?

gxueatlassian: Make sense. 
I've updated the PR to allow RepoSyncStates with specified ids to be reset for above operations.

bgvozdev (approved):

PR #1986, opened 2023-03-17
Arc 2041 metrics for backfill tasks

gxueatlassian:
**What's in this PR?**
Add metrics to record time spent on each tasks

**Why**
We want to track the time spent and justify our backfill improvement.

**Added feature flags**
N/A

**Affected issues**  
ARC-2041

**How has this been tested?**  
Stage
image

**Whats Next?**
More discussion.

### DIFF
--- a/src/sqs/sqs.types.ts
+++ b/src/sqs/sqs.types.ts
@@ -138,6 +138,7 @@ export type BranchMessagePayload = BaseMessagePayload & {
 export type BackfillMessagePayload = BaseMessagePayload & {
 	syncType?: SyncType,
 	startTime?: string,
+	syncRequestedTime?: string, //ISO format
 	commitsFromDate?: string, //main commits from date, ISO string
 	branchCommitsFromDate?: string, //branch commits from date, ISO string
 	targetTasks?: TaskType[]
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -11,7 +11,7 @@ import { getBranchTask } from "./branches";
 import { getCommitTask } from "./commits";
 import { getBuildTask } from "./build";
 import { getDeploymentTask } from "./deployment";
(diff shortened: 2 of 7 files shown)


src/sqs/sqs.types.ts
```diff
@@ -138,6 +138,7 @@ export type BranchMessagePayload = BaseMessagePayload & {
 export type BackfillMessagePayload = BaseMessagePayload & {
 	syncType?: SyncType,
 	startTime?: string,
+	syncRequestedTime?: string, //ISO format
```
gxueatlassian, line 141: Don't want to use `startTime` as it is being used purely for `full` sync only. Afraid it might mess up the existing sfx dashboard.

src/sync/installation.ts
```diff
@@ -324,12 +326,53 @@ const doProcessInstallation = async (data: BackfillMessagePayload, sentry: Hub,
...
 	} catch (err) {
 		await handleBackfillError(err, data, nextTask, subscription, logger, scheduleNextTask);
 	}
 };
 
+const recordTaskProcessDuration = (nextTask: Task, gitHubProduct: string, subscription: Subscription, data: BackfillMessagePayload) => {
+	statsd.increment(metricTaskDuration.duration, Math.floor(new Date().getTime() / 1000) - nextTask.taskStartTimeInSec, 1, {
+		task: nextTask.task,
+		gitHubProduct,
+		repoNums: getRepoNumberBuckets(subscription.totalNumberOfRepos),
```
gxueatlassian, line 340: So for both `repoNums` and `commitsFromDate` I use manual buckets here. The idea is in the dashboard we can easily filter the tag but `repoNum = 0-9` or `repoNum = 1000+`.
I don't think we should put a raw number here, because the field is `tag` and `tag` isn't supposed to be dynamic and grow unlimited. But correct me if I am wrong. 

ps: Happy to discuss the bucket items as well.

PR #1987, opened 2023-03-19
ARC-2033 add col to record success time

gxueatlassian:
**What's in this PR?**
Add db cols for recording success time for each entities. 

**Why**
Even though it is not being used right now, we don't want to lose the information.

**Added feature flags**
N/A

**Affected issues**  
ARC-2033

**How has this been tested?**  
N/A

**Whats Next?**
Impl code to save the time.

### DIFF
--- a/db/migrations/20230320085600-add-commit-from-all-entities.js
+++ b/db/migrations/20230320085600-add-commit-from-all-entities.js
@@ -1,0 +1,17 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn("RepoSyncStates", "branchFrom", { type: Sequelize.DATE, allowNull: true });
+    await queryInterface.addColumn("RepoSyncStates", "pullFrom", { type: Sequelize.DATE, allowNull: true });
+    await queryInterface.addColumn("RepoSyncStates", "buildFrom", { type: Sequelize.DATE, allowNull: true });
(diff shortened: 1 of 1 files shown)


joshkay10 (approved):

db/migrations/20230320085600-add-commit-from-all-entities.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+module.exports = {
+  up: async (queryInterface, Sequelize) => {
+    await queryInterface.addColumn("RepoSyncStates", "branchFrom", { type: Sequelize.DATE, allowNull: true });
+    await queryInterface.addColumn("RepoSyncStates", "pullFrom", { type: Sequelize.DATE, allowNull: true });
+    await queryInterface.addColumn("RepoSyncStates", "buildFrom", { type: Sequelize.DATE, allowNull: true });
+    await queryInterface.addColumn("RepoSyncStates", "deploymentFrom", { type: Sequelize.DATE, allowNull: true });
```
Harminder84, line 8: What about `issueFrom`? is it skipped intentionally?

gxueatlassian: I looked at that, didn't think we have an task for issue? Although we have cols.

Harminder84: We can remove the `issue` column if it is obsolete one; may be in a seperate ticket.

gxueatlassian: Yup. Definitely in another PR.

PR #1988, opened 2023-03-19
ARC-1997: refactor usage of SQS in backfill to remove cyclic dependency

bgvozdev:
**What's in this PR?**
Replacing calling sqs in backfill with dependency injection. It all comes down to the change of signatures of
 - processInstallation
 - doProcessInstallation
 - markCurrentTaskAsFailedAndContinue

The rest is just to make the code work with the new signature...

**Why**
To streamline the dependency path in the code and to avoid cyclic dependencies. To handle backfilling, we need more calls from SQS "package" to backfill, therefore trying to simplify it beforehand.

### DIFF
--- a/src/sqs/queues.ts
+++ b/src/sqs/queues.ts
@@ -13,11 +13,18 @@ const logger = getLogger("sqs-queues");
 
 // TODO: Make this a class
 
-const backfillQueueHolder: { queue: SqsQueue<BackfillMessagePayload> | undefined } = {
-	queue: undefined
+let backfillQueue: SqsQueue<BackfillMessagePayload> | undefined = undefined;
+
+const backfillQueueMessageSender = (message, delaySec, logger) => {
+	if (backfillQueue) {
+		return backfillQueue.sendMessage(message, delaySec, logger);
+	} else {
+		logger.error("Cannot send a message: queue is not initialised yet. Should never happen!");
+		throw new Error("Queue is not ready");
+	}
 };
 
-backfillQueueHolder.queue = new SqsQueue<BackfillMessagePayload>(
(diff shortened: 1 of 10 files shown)


src/sqs/queues.ts
```diff
@@ -12,18 +12,33 @@ const LONG_POLLING_INTERVAL_SEC = 3;
 const logger = getLogger("sqs-queues");
 
 // TODO: Make this a class
-export const sqsQueues = {
-	backfill: new SqsQueue<BackfillMessagePayload>({
+
+let backfillQueue: SqsQueue<BackfillMessagePayload> | undefined = undefined;
+
+const backfillQueueMessageSender = (message, delaySec, logger) => {
```
gxueatlassian, line 18: This block seems "complicated, can we do following instead?

`await Promise.race([this.messageHandler(context, this <-- pass itself in and `this` is guarantee to be initialised!), timeoutPromise]);`

gxueatlassian: So in the `sync/installation.ts` just do `this.sendMessage` instead of `sqs.backfill.sendMessage`

src/sync/installation.ts
```diff
@@ -337,7 +330,6 @@ export const handleBackfillError = async (
 	err,
 	data: BackfillMessagePayload,
 	nextTask: Task,
-	subscription: Subscription,
```
gxueatlassian, line 340: I don't understand the purpose of removing `subscription` and find it again inside the `markCurrentTaskAsFailedAndContinue`. Is it for future to extra the error handler to outside?

gxueatlassian (approved):

PR #1990, opened 2023-03-20
ARC-1995

joshkay10:
**What's in this PR?**
Save failed tasks into predefined groups so we can better understand the issues faced and present retryable issues to customers
 

**Added feature flags**

**Affected issues**  
_Jira Issues_

**How has this been tested?**  

**Whats Next?**

Screen Shot 2023-03-22 at 9 15 38 AM


src/routes/jira/jira-get.ts
```diff
@@ -194,14 +195,26 @@ const renderJiraCloudAndEnterpriseServer = async (res: Response, req: Request):
 	});
 };
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const getFailedSyncErrors = async (subscription: Subscription) => {
+	const RETRYABLE_ERROR_CODES = ["PERMISSIONS_ERROR", "CONNECTION_ERROR"];
+	const failedSyncs = await RepoSyncState.getFailedFromSubscription(subscription);
+	const errorCodes = failedSyncs.map(sync => sync.failedCode);
+	const retryableErrorCodes = errorCodes.filter(errorCode => RETRYABLE_ERROR_CODES.includes(errorCode || ""));
```
gxueatlassian, line 203: For code readability, in `RETRYABLE_ERROR_CODES.includes(errorCode || "")`, what does `|| "")` means as a param in `.includes()`? I find a a bit hard to wrap my head around.

src/routes/jira/jira-get.ts
```diff
@@ -194,14 +195,26 @@ const renderJiraCloudAndEnterpriseServer = async (res: Response, req: Request):
 	});
 };
 
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const getFailedSyncErrors = async (subscription: Subscription) => {
```
gxueatlassian, line 199: /nit
maybe name it `getRetryableFailedSyncErrors`?

src/sync/installation.ts
```diff
@@ -395,12 +397,48 @@ export const handleBackfillError = async (
...
 };
 
-export const markCurrentTaskAsFailedAndContinue = async (subscription: Subscription, nextTask: Task, scheduleNextTask: (delayMs: number) => void): Promise<void> => {
+export const markCurrentRepositoryAsFailedAndContinue = async (subscription: Subscription, nextTask: Task, scheduleNextTask: (delayMs: number) => void, err, logger: Logger): Promise<void> => {
 	// marking the current task as failed
-	await updateRepo(subscription, nextTask.repositoryId, { [getStatusKey(nextTask.task)]: "failed" });
+	const failedCode = getFailedCode(err);
+	logger.warn({ failedCode, err }, "Backfill task failed");
+
+	await updateRepo(subscription, nextTask.repositoryId, { [getStatusKey(nextTask.task)]: "failed", failedCode });
```
gxueatlassian, line 441: Will it break the app if the task is `repository` since the `Subscription` doesn't have the `failedCode` col?

src/sync/installation.ts
```diff
@@ -395,12 +397,48 @@ export const handleBackfillError = async (
...
+	}
+	if (status === 401) {
+		return "AUTHENTICATION_ERROR";
+	}
+	// A generic catch for authorization issues, invalid permissions on the JWT
+	if (status === 403) {
+		return "AUTHORIZATION_ERROR";
+	}
+	// If the user hasn't accepted updated permissions for the app.
+	if (status === 200 && message === "Resource not accessible by integration") {
```
gxueatlassian, line 422: In the same file, in function `execute`, there's a line
```
				if ((err.status === 403 || err.status === undefined) && err.message?.includes("Resource not accessible by integration")) {
					await subscription?.update({ syncWarning: `Invalid permissions for ${task} task` });
					errorLog.error(`Invalid permissions for ${task} task`);
					// Return undefined objects so the sync can complete while skipping this task
					return { edges: undefined, jiraPayload: undefined };
				}
```

Will it capture the `permission` problem, save it to subscription, and stop it from being saved in the RepoSyncStatus failCode col, and therefore `jira-get` won't be displaying it?

views/partials/jira-configuration-error-summary.hbs
```diff
@@ -0,0 +1,43 @@
...
+    <div>
+      <div class="error-summary-errors-header">
+        Description
+      </div>
+      {{#if syncErrors.CONNECTION_ERROR}}
+      <div>
+        <b>{{syncErrors.CONNECTION_ERROR}}</b> tasks have not been backfilled. Please
+        <a href="{{html_url}}"
+           data-installation-link="{{html_url}}"
+           target="_blank">accept permissions</a>
```
gxueatlassian, line 22: For `CONNECTION_ERROR`, why is it `accept permission`?

src/routes/jira/jira-get.ts
```diff
@@ -194,14 +195,26 @@ const renderJiraCloudAndEnterpriseServer = async (res: Response, req: Request):
...
+	const RETRYABLE_ERROR_CODES = ["PERMISSIONS_ERROR", "CONNECTION_ERROR"];
+	const failedSyncs = await RepoSyncState.getFailedFromSubscription(subscription);
+	const errorCodes = failedSyncs.map(sync => sync.failedCode);
+	const retryableErrorCodes = errorCodes.filter(errorCode => RETRYABLE_ERROR_CODES.includes(errorCode || ""));
+
+	if (retryableErrorCodes.length === 0) {
+		return undefined;
+	}
+	return countBy(retryableErrorCodes);
```
Harminder84, line 208: Agree with @gary, method name is not in sync with what it is returning.

src/sync/installation.ts
Harminder84, line 441: Won't it override the `failed` field if multiple tasks get failed for same repo?

src/sync/installation.ts
```diff
@@ -395,12 +397,48 @@ export const handleBackfillError = async (
...
-export const markCurrentTaskAsFailedAndContinue = async (subscription: Subscription, nextTask: Task, scheduleNextTask: (delayMs: number) => void): Promise<void> => {
+export const markCurrentRepositoryAsFailedAndContinue = async (subscription: Subscription, nextTask: Task, scheduleNextTask: (delayMs: number) => void, err, logger: Logger): Promise<void> => {
 	// marking the current task as failed
-	await updateRepo(subscription, nextTask.repositoryId, { [getStatusKey(nextTask.task)]: "failed" });
+	const failedCode = getFailedCode(err);
+	logger.warn({ failedCode, err }, "Backfill task failed");
+
+	await updateRepo(subscription, nextTask.repositoryId, { [getStatusKey(nextTask.task)]: "failed", failedCode });
 	const gitHubProduct = getCloudOrServerFromGitHubAppId(subscription.gitHubAppId);
 	statsd.increment(metricTaskStatus.failed, [`type:${nextTask.task}`, `gitHubProduct:${gitHubProduct}`]);
```
Harminder84, line 444: I didn't see the code when `faileCode` field gets cleared on successful sync after unsuccessful attempt?

src/routes/jira/jira-get.ts
joshkay10: yeah agree, that was not intended originally, it was typescript having a whinge, ill restructure it a bit so that it can drop the || ""

src/sync/installation.ts
joshkay10: Yes this does happen, had a chat with @gary about this and we landed on keeping one field for now, i had better comments about that must have removed them 🤔  will put them back as a known tradeoff

joshkay10: garggg good catch, i had skip repo task but must have cleared it in the conflict resolve :sigh:

src/sync/installation.ts
joshkay10: Have reset on sync

src/sync/installation.ts
joshkay10: Am updating the faildCode here too, this conditional ends the error here

src/sqs/backfill-error-handler.ts
```diff
@@ -15,13 +15,12 @@ const handleTaskError = async (sendSQSBackfillMessage: (message, delaySec, logge
...
 
 	// TODO: add task-related logic: e.g. mark as complete for 404; retry RateLimiting errors;
-
 	if (context.lastAttempt) {
 		// Otherwise the sync will be "stuck", not something we want
 		log.warn("That was the last attempt: marking the task as failed and continue with the next one");
 		await markCurrentTaskAsFailedAndContinue(context.payload, task, async (delayMs) => {
 			return await sendSQSBackfillMessage(context.payload, delayMs / 1000, log);
-		}, log);
+		}, log, cause);
```
joshkay10, line 23: we need to pass the error through so we can try determine the cause during get code

src/sync/sync-utils.ts
```diff
@@ -22,7 +22,8 @@ export const findOrStartSync = async (
 	await subscription.update({
 		syncStatus: SyncStatus.PENDING,
 		numberOfSyncedRepos: 0,
-		syncWarning: null
+		syncWarning: null,
+		failedCode: null
```
gxueatlassian, line 26: Do we have `failedCode` in `Subscription`?

gxueatlassian (commented):
Also, when will be `failedCode` be cleared?

src/sync/sync-utils.ts
joshkay10: destroyed

src/sync/sync-utils.ts
```diff
@@ -39,6 +39,8 @@ export const findOrStartSync = async (
 		await RepoSyncState.deleteFromSubscription(subscription);
 	}
 
+	await resetRepoFailedCodes(subscription, syncType);
```
Harminder84, line 42: I think we need to reset `failedCode` inside `resetTargetedTasks` function. That is place where we reset the status as well as cursor (for full sync only)

views/partials/jira-configuration-error-summary.hbs
```diff
@@ -0,0 +1,43 @@
...
+        You could continue to use the integration and still view the backfilled data in Jira, while the app attempts to reprocess these entities.</p>
+    </div>
+
+    <div class="error-summary-border-bottom">
+      <div class="error-summary-errors-header error-summary-border-bottom">
+        Description
+      </div>
+      {{#if syncErrors.PERMISSIONS_ERROR}}
+      <div>
+        <b>{{syncErrors.PERMISSIONS_ERROR}}</b> tasks have not been backfilled. Please
```
Harminder84, line 19: Should it be repo or task?

Harminder84, line 19: Should we change the text if permission error for 1 repo only? 
`1 task has not been backfilled`

src/sync/installation.ts
```diff
@@ -349,18 +349,22 @@ const findSubscriptionForMessage = (data: BackfillMessagePayload) =>
...
 		log.warn("No subscription found, nothing to do");
 		return;
 	}
 
+	// marking the current task as failed, this value will override any preexisting failedCodes and only keep the last known failed issue.
+	const failedCode = getFailedCode(err);
+
 	// marking the current task as failed
-	await updateRepo(subscription, nextTask.repositoryId, { [getStatusKey(nextTask.task)]: "failed" });
+	await updateRepo(subscription, nextTask.repositoryId, { [getStatusKey(nextTask.task)]: "failed", failedCode });
```
gxueatlassian, line 363: Can you confirm when the nextTask.task === 'repository', this line won't break? 
ps: I check in my local, seems fine though. Just Fyi.

src/sync/sync-utils.ts
```diff
@@ -85,6 +85,7 @@ const resetTargetedTasks = async (subscription: Subscription, syncType?: SyncTyp
 		}
 		updateRepoSyncTasks[`${task}Status`] = null;
 	});
+	updateRepoSyncTasks[`failedCode`] = null;
```
gxueatlassian, line 88: So when the `targetTasks` is empty, this won't be cleared?

gxueatlassian (approved):

src/sync/installation.ts
joshkay10: Yup, this is fine, it was already updating the repotable its just also setting failedCode

views/partials/jira-configuration-error-summary.hbs
joshkay10: Yeah its tricky, probably repo because task is meaningless to the customer right? I took this from the existing warning.... but we will fix copy as ARC-2039 so will keep this flag off until thats sorted

gxueatlassian (approved):

src/sync/sync-utils.ts
Harminder84: I think it will be ok if we can pass the tasks from the front end rather than moving to a seperate function. Now we've two functions `resetTargetedTasks` and `resetFailedCode` which are doing the pretty similar tasks but on different fields which makes the code complex to understand.  It just a thought!

PR #1991, opened 2023-03-20
ARC-1997: move sequelise and unknown error handling to SQS error handler

bgvozdev:
**What's in this PR?**
- Make backfill to throw some error
- Handle them in SQS error handler

**Why**
We are moving away error handling logic to SQS level. This is the first set of errors we want to do.

**Added feature flags**
All errors here either:
 - cause sync to stuck, or
 - extremely rare (sequelize).
Because of that no need in feature flags.

### DIFF
--- a/src/sqs/backfill-error-handler.ts
+++ b/src/sqs/backfill-error-handler.ts
@@ -7,8 +7,12 @@ import Logger from "bunyan";
 
 const handleTaskError = async (backfillQueue: SqsQueue<BackfillMessagePayload>, task: Task, cause: Error, context: SQSMessageContext<BackfillMessagePayload>, rootLogger: Logger
 ) => {
-	const log = rootLogger.child({ task });
-	log.info({ task }, "Handling error task");
+	const log = rootLogger.child({
+		task,
+		receiveCount: context.receiveCount,
+		lastAttempt: context.lastAttempt
+	});
+	log.info("Handling error task");
 
 	// TODO: add task-related logic: e.g. mark as complete for 404; retry RateLimiting errors;
 
--- a/src/sqs/backfill-error-handler.ts
+++ b/src/sqs/backfill-error-handler.ts
@@ -10,7 +10,9 @@ const handleTaskError = async (_: SqsQueue<BackfillMessagePayload>, task: Task, 
 	const log = rootLogger.child({ task });
 	log.info({ task }, "Handling error task");
 
-	// TODO add task-related logic: e.g. mark as complete for 404; retry RateLimiting errors;
+	// TODO: add task-related logic: e.g. mark as complete for 404; retry RateLimiting errors;
+
+	// TODO: if last attempt mark task as failed and continue
 
 	return handleUnknownError(cause, context);
 };
--- a/src/sqs/backfill-error-handler.ts
+++ b/src/sqs/backfill-error-handler.ts
@@ -1,0 +1,41 @@
+import { BackfillMessagePayload, ErrorHandler, ErrorHandlingResult, SQSMessageContext } from "~/src/sqs/sqs.types";
+import { SqsQueue } from "./sqs";
+import { TaskError } from "~/src/sync/installation";
+import { Task } from "~/src/sync/sync.types";
+import { handleUnknownError } from "~/src/sqs/error-handlers";
+import Logger from "bunyan";
+
+const handleTaskError = async (_: SqsQueue<BackfillMessagePayload>, task: Task, cause: Error, context: SQSMessageContext<BackfillMessagePayload>, rootLogger: Logger
+) => {
+	const log = rootLogger.child({ task });
+	log.info({ task }, "Handling error task");
+
+	// TODO add task-related logic: e.g. mark as complete for 404; retry RateLimiting errors;
+
+	return handleUnknownError(cause, context);
+};
+
+export const backfillErrorHandler: (backfillQueueHolder: { queue: SqsQueue<BackfillMessagePayload> | undefined }) => ErrorHandler<BackfillMessagePayload> =
+	(backfillQueueHolder) =>
+		async (err: Error, context: SQSMessageContext<BackfillMessagePayload>): Promise<ErrorHandlingResult> => {
+			const log = context.log.child({ err });
+			log.info("Handling error");
+
+			if (!backfillQueueHolder.queue) {
+				log.warn("Queue was not ready, retry");
+				return {
+					isFailure: true,
+					retryable: true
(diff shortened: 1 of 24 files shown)


src/sqs/backfill-error-handler.ts
```diff
@@ -0,0 +1,44 @@
+import { BackfillMessagePayload, ErrorHandler, ErrorHandlingResult, SQSMessageContext } from "~/src/sqs/sqs.types";
+import { markCurrentTaskAsFailedAndContinue, TaskError } from "~/src/sync/installation";
```
gxueatlassian, line 2: Should we move `markCurrentTaskAsFailedAndContinue` into this file instead? 
I don't think `sync/installation` is using it anymore?

src/sync/installation.ts
```diff
@@ -327,24 +337,22 @@ const doProcessInstallation = async (data: BackfillMessagePayload, sentry: Hub,
...
 
 	const logger = rootLogger.child({ err });
 
-	const isRateLimitError = err instanceof RateLimitingError || Number(err?.headers?.["x-ratelimit-remaining"]) == 0;
-
-	if (isRateLimitError) {
-		const rateLimit = err instanceof RateLimitingError ? err.rateLimitReset : Number(err?.headers?.["x-ratelimit-reset"]);
-		const delay = Math.max(rateLimit * 1000 - Date.now(), 0);
+	// TODO: rethrow as TaskError and handle in SQS error handler
+	if (err instanceof RateLimitingError) {
```
gxueatlassian, line 349: What happen to `|| Number(err?.headers?.["x-ratelimit-remaining"]) == 0;`

src/sync/parallel-page-fetcher.ts
```diff
@@ -31,15 +31,21 @@ export const fetchNextPagesInParallel = async (
...
 	);
-	const fetchedData = await Promise.all(tasks);
-	return fetchedData.reduce((prev, curr) => {
-		return {
-			edges: [...(prev.edges || []), ...(curr.edges || [])],
-			jiraPayload:
-				mergeJiraPayload(prev.jiraPayload, curr.jiraPayload)
-		};
-	}, {
+	const fetchedData = await Promise.allSettled(tasks);
```
gxueatlassian, line 34: Why from `all` to `allSettled`? 
1. That might reduce performance and wastes rate limit quota
2. If only worry about the structure of the error, just a simple try/catch will do. 
3. The `if (curr.status === "fulfilled") {...} else {...throw curr.reason;}` I think reduce the code readability, compared to `try { fetchedData = await Promise.all()... } catch (e) { throw transform(e) }

But okay I don't see a big problem though.

src/sqs/backfill-error-handler.ts
```diff
@@ -0,0 +1,44 @@
...
+import { SQS } from "aws-sdk";
+
+const handleTaskError = async (sendSQSBackfillMessage: (message, delaySec, logger) => Promise<SQS.SendMessageResult>, task: Task, cause: Error, context: SQSMessageContext<BackfillMessagePayload>, rootLogger: Logger
+) => {
+	const log = rootLogger.child({
+		task,
+		receiveCount: context.receiveCount,
+		lastAttempt: context.lastAttempt
+	});
+	log.info("Handling error task");
```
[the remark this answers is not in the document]
bgvozdev: I'd like to have it for not (when the feature is new), we can always remove that

src/sqs/backfill-error-handler.ts
bgvozdev: It uses subscription and reposyncstate, don't want to move these two over here

src/sqs/backfill-error-handler.ts
```diff
@@ -0,0 +1,44 @@
+import { BackfillMessagePayload, ErrorHandler, ErrorHandlingResult, SQSMessageContext } from "~/src/sqs/sqs.types";
+import { markCurrentTaskAsFailedAndContinue, TaskError } from "~/src/sync/installation";
+import { Task } from "~/src/sync/sync.types";
+import { handleUnknownError } from "~/src/sqs/error-handlers";
+import Logger from "bunyan";
+import { SQS } from "aws-sdk";
+
+const handleTaskError = async (sendSQSBackfillMessage: (message, delaySec, logger) => Promise<SQS.SendMessageResult>, task: Task, cause: Error, context: SQSMessageContext<BackfillMessagePayload>, rootLogger: Logger
+) => {
+	const log = rootLogger.child({
```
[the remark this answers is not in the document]
bgvozdev: Suggestions?

src/sync/installation.ts
bgvozdev: This is some old crap when we had octokit. We don't use it for a long time already, in github client all exceptions are wrapped into GitHubClientError and others

src/sync/parallel-page-fetcher.ts
bgvozdev: This makes tests more deteministic, otherwise there could be some calls happening while the test is already over.

src/sqs/backfill-error-handler.ts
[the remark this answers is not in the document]
gxueatlassian: How about in the param, just `logger`? And in this line, `logger = logger.child({}`? 
Any way, /nit, just the `root` thing gets into my mind a bit.

src/sync/parallel-page-fetcher.ts
bgvozdev: Also, mind you, this is only a error case when it might take longer time, which is a corner-case

src/sqs/backfill-error-handler.ts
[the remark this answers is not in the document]
bgvozdev: cannot do logger = logger.child cause params are const :)

src/sqs/backfill-error-handler.ts
gxueatlassian: Okay, I can accept this, but still want to point out that `markCurrentTaskAsFailedAndContinue` is ALL ABOUT handling errors, which is exactly what this file is for. 
But never mind, I am fine with keeping this inside the origin file.

gxueatlassian (approved):

src/sqs/backfill-error-handler.ts
bgvozdev: I think about sync/installation as a class, therefore all manipulations with the state should be encapsulated inside it. What is exported is a public interface. Moving the internals outside means to break encapsulation. IMHO

PR #1994, opened 2023-03-21
ARC-1997: remove small page retries

bgvozdev:
**What's in this PR?**
- Removing retry-with-small-page thing
- as a consequence, Permission check was moved to SQS error handler
-  Some minor renamings/refactorings

**Why**
- Cause it is broken for several task types and there's no evidence it helps. No comments, no explanation, nothing. Starting from zero ground and we will go from there.

### DIFF
--- a/src/models/reposyncstate.ts
+++ b/src/models/reposyncstate.ts
@@ -36,7 +36,7 @@ export class RepoSyncState extends Model {
 	createdAt: Date;
 	commitFrom?: Date;
 
-
+	// TODO: why it is only for pullStatus, branchStatus and commitStatus ?!
 	get status(): TaskStatus {
 		const statuses = [this.pullStatus, this.branchStatus, this.commitStatus];
 		if (statuses.some(s => s === "failed")) {
@@ -50,8 +50,8 @@ export class RepoSyncState extends Model {
 		return "pending";
 	}
 
-	static async countSyncedReposFromSubscription(subscription: Subscription): Promise<number> {
-		return RepoSyncState.countFromSubscription(subscription, {
+	static async countFullySyncedReposForSubscription(subscription: Subscription): Promise<number> {
+		return RepoSyncState.countSubscriptionRepos(subscription, {
 			where: {
 				pullStatus: "complete",
 				branchStatus: "complete",
@@ -62,8 +62,8 @@ export class RepoSyncState extends Model {
 		});
 	}
 
-	static async countFailedReposFromSubscription(subscription: Subscription): Promise<number> {
-		return RepoSyncState.countFromSubscription(subscription, {
+	static async countFailedSyncedReposForSubscription(subscription: Subscription): Promise<number> {
+		return RepoSyncState.countSubscriptionRepos(subscription, {
 			where: {
 				[Op.or]: {
 					pullStatus: "failed",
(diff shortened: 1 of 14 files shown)


src/models/reposyncstate.ts
```diff
@@ -40,7 +40,7 @@ export class RepoSyncState extends Model {
 	updatedAt: Date;
 	createdAt: Date;
 
-
+	// TODO: why it is only for pullStatus, branchStatus and commitStatus ?!
 	get status(): TaskStatus {
```
gxueatlassian, line 44: I tried hard, with my vim, couldn't find the usage. Can you have a look? And :nuke it if not used? 
ps: make sure check handlebar/html as well.

src/models/reposyncstate.ts
```diff
@@ -54,8 +54,8 @@ export class RepoSyncState extends Model {
 		return "pending";
 	}
 
-	static async countSyncedReposFromSubscription(subscription: Subscription): Promise<number> {
-		return RepoSyncState.countFromSubscription(subscription, {
+	static async countFullySyncedReposForSubscription(subscription: Subscription): Promise<number> {
```
gxueatlassian, line 57: I like your English skill. Better than mine. Hah.

src/models/reposyncstate.ts
bgvozdev: Sure, I can do this in my next pr :ifyouknowwhatimean:

src/sqs/backfill-error-handler.test.ts
```diff
@@ -116,4 +118,30 @@ describe("backfillErrorHandler", () => {
...
+		};
+
+		const task = _.cloneDeep(TASK);
+		task.repositoryId = repoSyncState?.repoId || -1;
+
+		const sendMessageMock = jest.fn();
+		const result = await backfillErrorHandler(sendMessageMock)(new TaskError(task, new InvalidPermissionsError({ } as unknown as AxiosError)), context);
+
+		expect(result).toEqual({
+			isFailure: false
```
gxueatlassian, line 138: should it expect `isRetryable: true` as well?

src/sqs/backfill-error-handler.test.ts
```diff
@@ -116,4 +118,30 @@ describe("backfillErrorHandler", () => {
...
+		});
+		expect(sendMessageMock.mock.calls[0][0]).toEqual(
+			{ installationId: DatabaseStateCreator.GITHUB_INSTALLATION_ID, jiraHost }
+		);
+		expect(sendMessageMock.mock.calls[0][1]).toEqual(0);
```
gxueatlassian, line 143: So we don't need exponential for permission retry? Even for jira?

gxueatlassian: Oh wait, we are not retrying are we?

src/sqs/backfill-error-handler.test.ts
bgvozdev: It doesn't matter, because ifFailure means that the error is not considered as a "error" and the message will be deleted.

src/sqs/backfill-error-handler.test.ts
bgvozdev: That's for github only, InvalidPermissionError is a github client error. I'll rename the exception to make it more clear in my next PR

bgvozdev: For the permissions errors from GitHub we are not retrying, correct

src/sync/installation.ts
```diff
@@ -136,38 +143,12 @@ export const updateJobStatus = async (
...
-				gitHubProduct,
-				repos: repoCountToBucket(subscription.totalNumberOfRepos)
-			});
-		}
-
-		logger.info({ startTime, endTime, timeDiff, gitHubProduct }, "Sync status is complete");
-	} else {
-		logger.info("Sync status is pending");
-		scheduleNextTask(0);
+	if ((edges?.length || 0) > 0) {
```
gxueatlassian, line 146: can use variable `isComplete`

gxueatlassian (commented):
lgtm

gxueatlassian (approved):

PR #1996, opened 2023-03-22
ARC-1997 improve error mapping

bgvozdev:
**What's in this PR?**
 - Improve error mapping
 -  Improve name of GitHub error classes

**Why**
Cause currently we are handling different errors differently (eg not found) and some are ignoring whatsoever (e.g. graphql's permissions one)


gxueatlassian (approved):
Good stuff!

PR #1998, opened 2023-03-23
ARC-2053 Update default backfill date to 6 months from the last backfilled date

Harminder84:
**What's in this PR?**
Default to 6 months from the last. backfilled date. If it is not backfilled at all, it would be 6 months from the current date.

**Added feature flags**
backfill-algorithm-incremental

**Affected issues**  
[ARC-2053]

[ARC-2053]: https://softwareteams.atlassian.net/browse/ARC-2053

### DIFF
--- a/static/js/jira-configuration.js
+++ b/static/js/jira-configuration.js
@@ -49,12 +49,20 @@ $(".configure-connection-link").click(function(event) {
 	openChildWindow($(event.target).data("installation-link"));
 });
 
-const initializeBackfillDateInput = function () {
+const initializeBackfillDateInput = function (event) {
 	const dateElement = document.getElementById('backfill-date-picker');
-	const date = new Date();
-	date.setFullYear(date.getFullYear() - 1);
-	dateElement.value = date.toISOString().split('T')[0];
-	dateElement.max = new Date().toISOString().split('T')[0];
+	const isIncrementalBackfillEnabled = $("body").data("is-incremental-backfill-enabled");
+	if(isIncrementalBackfillEnabled) {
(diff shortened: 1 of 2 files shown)


static/js/jira-configuration.js
```diff
@@ -49,12 +49,20 @@ $(".configure-connection-link").click(function(event) {
...
+const initializeBackfillDateInput = function (event) {
 	const dateElement = document.getElementById('backfill-date-picker');
-	const date = new Date();
-	date.setFullYear(date.getFullYear() - 1);
-	dateElement.value = date.toISOString().split('T')[0];
-	dateElement.max = new Date().toISOString().split('T')[0];
+	const isIncrementalBackfillEnabled = $("body").data("is-incremental-backfill-enabled");
+	if(isIncrementalBackfillEnabled) {
+		const newBackfillDate = $(event.target).data('connection-backfill-since') ? new Date($(event.target).data('connection-backfill-since')) : new Date();
+		newBackfillDate.setMonth(newBackfillDate.getMonth() - 6);
```
gxueatlassian, line 57: Are you sure this will work `newBackfillDate.setMonth(newBackfillDate.getMonth() - 6);`? If the result is a negative number?

gxueatlassian: How about move all this logic into the backend `jira-get` and attached it to the `connection` as a field `suggestBackfillDate` ?

Harminder84: I've checked it. Incase of negative number, it will go back one year and set the month correctly.

Harminder84: We can set from the backend if want to drive this value from FF in future or some other computation on suggested backfill date. I think if this is the only case, it is fine to have this logic on the front end.

gxueatlassian (approved):

PR #1999, opened 2023-03-23
ARC-1997: move remnants of error handling from sync/installations to sqs

bgvozdev:
**What's in this PR?**
Removing all error handling from sync/installations to SQS error handler

**Why**
To leverage SQS retries instead of relying on ad-hoc logic and make the code cleaner

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -276,7 +276,7 @@ const doProcessInstallation = async (data: BackfillMessagePayload, sentry: Hub, 
 		if (taskPayload.jiraPayload) {
 			const jiraClient = await getJiraClient(
 				subscription.jiraHost,
-				taskPayload.jiraPayload,
+				gitHubInstallationId,
 				data.gitHubAppConfig?.gitHubAppId,
 				logger
 			);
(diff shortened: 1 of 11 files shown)


src/sync/installation.ts
```diff
@@ -231,115 +256,50 @@ const doProcessInstallation = async (data: BackfillMessagePayload, sentry: Hub,
...
-	if (err instanceof GithubClientRateLimitingError) {
-		const delayMs = Math.max(err.rateLimitReset * 1000 - Date.now(), 0);
-
-		if (delayMs) {
-			// if not NaN or 0
-			logger.info({ delay: delayMs }, `Delaying job for ${delayMs}ms`);
-			await sendBackfillMessage(data, delayMs / 1000, logger);
+		if (nextTask) {
+			logger.info({ err, nextTask }, "rethrowing as a task error");
+			throw new TaskError(nextTask, err);
```
gxueatlassian, line 297: Find it weird to wrap and throw base on the `nextTask`

gxueatlassian (approved):

PR #2000, opened 2023-03-23
Arc-2050 upgrade sequelize to fix vuln

rachellerathbone:
**What's in this PR?**
Bumped sequelize version from 5.21.7 -> 6.19.1. Needed to change:

- when using the addContraint method in a migration you now need to specify that with a [fields](https://github.com/atlassian/github-for-jira/pull/2000/files#diff-fb2623b999d83b75f203da742f7c5dc697a82c81851b4df43c8462eda94db2c5R8) property
- lots of TS changes :dead: Version 6 of Sequelize involved them really buttoning down on their type checking which highlighted a bunch of stuff that we weren't checking adequately. 

**Why**
There was a critical security flaw in the version we had installed.

**Affected issues**  
Arc-2050

**How has this been tested?**  
Tested the following locally and in staging. Covered anything I figured could be impacted by these changes, along with pretty much all other app functionality (as a sanity check):
- installing and uninstalling the app
- full installation flow for both cloud and server (including auto and manual app creation for the latter)
- deleting orgs/servers/etc connections from the config page
- re-running backfill
- webhooks for both server and cloud
- create branch

**Whats Next?**
Onto project work :hooray:


README.md
```diff
@@ -196,10 +196,10 @@ it will automatically trigger the backfilling process for the allowed repositori
 2. All commits from the default branch will be backfilled. The commit message must contain the Jira issue key.
 3. Only the latest 50 commits from non-default branches will be backfilled.
 4. Unreachable commits (e.g. from deleted branches) will NOT be backfilled.
-5. All pull requests, regardless of their statuses, will be backfilled. The Jira issue key should be included either in 
-   the title of the pull request, in the description of the pull request or in the name of the source branch 
+5. All pull requests, regardless of their statuses, will be backfilled. The Jira issue key should be included either in
```
rachellerathbone, line 199: Uh not sure what happened here but okay 🤷‍♀️

src/jira/client/jira-client.test.ts
```diff
@@ -144,12 +144,12 @@ describe("Test getting a jira client", () => {
 
 	describe("Reading encryptedSharedSecret", () => {
 		beforeEach(async ()=>{
-			const inst: Installation = await Installation.findOne({
+			const inst: Installation | null = await Installation.findOne({
```
rachellerathbone, line 147: In our Installation model, when we call the findOne method we define the return type as `Promise<Installation | null>`.

src/models/encrypted-model.test.ts
```diff
@@ -44,12 +44,14 @@ Dummy.init({
 }, {
 	hooks: {
 		beforeSave: async (app, opts) => {
-			await app.encryptChangedSecretFields(opts.fields, getLogger("test"));
+			const optsFields = typeof opts.fields === "string" ? opts.fields : undefined;
```
rachellerathbone, line 47: Turns out the fields property can hold a string, number, or symbol but in `encryptChangedSecretFields` we are setting `fieldsChanged` to `string[] = []`

src/models/encrypted-model.test.ts
```diff
@@ -80,8 +82,8 @@ describe("Encrypted model", () => {
 		const id = newId();
 		await Dummy.create({ id, name: "test", a: "aaa1", b: "bbb1" });
 		const dummy = await Dummy.findOne({ where: { name: "test" } });
-		await dummy.decrypt("a");
-		await dummy.decrypt("b");
+		await dummy?.decrypt("a", getLogger("test"));
```
rachellerathbone, line 85: This second arg is required. We missed this in a few places and it just wasn't being detected before.

db/migrations/20230202110100-add-repo-sync-state-fk.js
```diff
@@ -3,8 +3,9 @@
 module.exports = {
 	up: async (queryInterface) => {
 		//refer to doc https://sequelize.org/v5/class/lib/query-interface.js~queryinterface#instance-method-addConstraint
-		await queryInterface.addConstraint("RepoSyncStates", ["subscriptionId"], {
+		await queryInterface.addConstraint("RepoSyncStates", {
 			type: "foreign key",
+			fields: ["subscriptionId"],
```
rachellerathbone, line 8: Outside of all the update to type checking in Sequelize v6, this was the only other change we needed to make.

src/models/installation.ts
```diff
@@ -142,12 +143,14 @@ Installation.init({
 	hooks: {
 		beforeSave: async (instance: Installation, opts) => {
 			if (!opts.fields) return;
-			await instance.encryptChangedSecretFields(opts.fields, LOGGER_HOOK_BEFORE_SAVE);
+			const optsFields = opts.fields.filter((it): it is string => !!it);
```
bgvozdev, line 146: Curious why do we need this change?

src/routes/api/ghes-app-encryption-ctx/re-encrypt-ghes-app-keys.ts
```diff
@@ -88,11 +88,11 @@ export const ReEncryptGitHubServerAppKeysPost = async (req: Request, res: Respon
 				gitHubClientSecret: originGitHubClientSecret
 			}, jiraHost);
 
-			const updatedApp: GitHubServerApp = await GitHubServerApp.findByPk(app.id);
+			const updatedApp: GitHubServerApp | null = await GitHubServerApp.findByPk(app.id);
```
bgvozdev, line 91: I'd prefer to either use "!" here because we know that the app exists.

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-put.test.ts
```diff
@@ -72,7 +72,7 @@ describe("PUT /jira/connect/enterprise/app/:uuid", () => {
 
 	it("should use existing privateKey if new privateKey is not passed in as body", async () => {
 
-		let existingApp = await GitHubServerApp.install({
+		let existingApp: GitHubServerApp | null = await GitHubServerApp.install({
```
bgvozdev, line 75: Curious why it could be null now? Looking at the implementation of `GitHubServerApp::install`, looks like it always return a result, and I don't see that we've changed it somewhere.

src/models/installation.ts
rachellerathbone: With the beefed up type checking in sequelize v6 we now need to be more explicit here. We're setting `fieldsChanged: string[] = []` in `encryptChangedSecretFields` and without this update, in `beforeSave` and `beforeBulkCreate` we get the following error:

Screenshot 2023-03-31 at 9 44 59 am

This filters out any values that aren't of type string.

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -133,7 +133,7 @@ const getWebhookSecret = async (uuid?: string): Promise<{ webhookSecret: string,
 		if (!gitHubServerApp) {
 			throw new Error(`GitHub app not found for uuid ${uuid}`);
 		}
-		const installation: Installation = await Installation.findByPk(gitHubServerApp.installationId);
+		const installation: Installation | null = await Installation.findByPk(gitHubServerApp.installationId);
```
bgvozdev, line 136: Same here, let's use `!` instead because from the context we know that there must be an installation (cause we have the instance of the server app and they are linked by a foreign key)

src/models/reposyncstate.ts
```diff
@@ -137,15 +137,15 @@ export class RepoSyncState extends Model {
 		} as FindOptions));
 	}
 
-	static async updateFromSubscription(subscription: Subscription, values: Record<string, unknown>, options: Partial<UpdateOptions> = {}): Promise<[number, RepoSyncState[]]> {
+	static async updateFromSubscription(subscription: Subscription, values: Record<string, unknown>, options: Partial<UpdateOptions> = {}): Promise<[affectedCount: number]> {
```
bgvozdev, line 140: Nice! So very likes!

src/routes/api/data-cleanup/cleanup-reposyncstates.ts
```diff
@@ -24,7 +24,7 @@ export const RepoSyncStateCleanUpOrphanDataPost = async (req: Request, res: Resp
 		const cleanUpFromWhereSql = getCleanupSql(repoSyncStateId);
 		const sql = commitToDB === true ? getCommitInDBSql(cleanUpFromWhereSql) : getDryRunSql(cleanUpFromWhereSql);
 		const result = await RepoSyncState.sequelize?.query(sql);
-		const parsedResult = safeParseResult(result, commitToDB);
+		const parsedResult = result && safeParseResult(result, commitToDB);
```
bgvozdev, line 27: should we rather throw an exception explicitly if null/undefined/empty/whatever we are not expecting?

src/routes/api/db-migrations/db-migration-down.ts
```diff
@@ -41,18 +41,24 @@ export const DBMigrationDown = async (req: Request, res: Response): Promise<void
 
 };
 
-const validateScriptAgainstDB = async (targetScript: string) => {
+type LastScript = {
```
bgvozdev, line 44: 🥇

test/utils/database-state-creator.ts
```diff
@@ -6,10 +6,10 @@ import fs from "fs";
 import path from "path";
 
 interface CreatorResult {
-	installation: Installation,
-	subscription: Subscription,
-	gitHubServerApp?: GitHubServerApp,
-	repoSyncState?: RepoSyncState
+	installation: Installation | null;
+	subscription: Subscription | null;
```
bgvozdev, line 10: My understanding is that installation and subscription must always be defined. Should  we maybe use "!" instead of " |null"? Otherwise, all the users of this class will suffer.... 😬

src/routes/api/ghes-app-encryption-ctx/re-encrypt-ghes-app-keys.ts
rachellerathbone: I had tried that initially but unfortunately it doesn't clear the error.

Screenshot 2023-04-03 at 9 58 22 am

The `findByPk` method returns the entry or null:

```
public static findByPk<M extends Model>(
    this: ModelStatic<M>,
    identifier?: Identifier,
    options?: Omit<FindOptions<Attributes<M>>, 'where'>
  ): Promise<M | null>;
```

src/routes/github/webhook/webhook-receiver-post.ts
rachellerathbone: Same deal here, anywhere we call `findByPk` TS now complains about type null not being assignable to X.

src/routes/api/ghes-app-encryption-ctx/re-encrypt-ghes-app-keys.ts
bgvozdev: `await GitHubServerApp.findByPk(app.id)!` ?

test/utils/database-state-creator.ts
rachellerathbone: Same deal here. We're calling `findByPk`:

```
installation: await Installation.findByPk(installation.id),
subscription: await Subscription.findByPk(subscription.id),
gitHubServerApp: gitHubServerApp ? await 
  GitHubServerApp.findByPk(gitHubServerApp.id) : undefined, 
repoSyncState: repoSyncState ? await RepoSyncState.findByPk(repoSyncState.id) : undefined`

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-put.test.ts
rachellerathbone: We call `findByPk` on line 107. Without passing null, we see the same error we do everywhere else we call this method

bgvozdev (approved):

PR #2002, opened 2023-03-24
ARC-1997: minor refactoring and removing UGC from logs

bgvozdev:
**What's in this PR?**
 - simplify try/catch logic in sync/installation a little bit
 - remove more UGC from logs
 - log stacktraces

**Why**
 - cause it simplifies code
 - cause we don't want to log UGC
 - cause stracktraces are helpful

### DIFF
--- a/src/config/logger.ts
+++ b/src/config/logger.ts
@@ -111,6 +111,26 @@ const censorUrl = (url) => {
 	return url;
 };
 
+const MSG_WITH_REGEX = /^(.*) Repository with the name '(.*)\/(.*)'.$/;
+
+const censorMessage = (msg) => {
+	if (!msg) {
+		return msg;
+	}
+	if (typeof msg === "string") {
+		if (msg.match(MSG_WITH_REGEX)) {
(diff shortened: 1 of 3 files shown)


src/config/logger.ts
```diff
@@ -111,6 +111,26 @@ const censorUrl = (url) => {
 	return url;
 };
 
+const MSG_WITH_REPO_NAME_REGEX = /^(.*) Repository with the name '(.*)\/(.*)'.$/;
+
+const censorMessage = (msg) => {
```
rachellerathbone, line 116: nit: would be good to add a return type here if we know what it is

src/config/logger.ts
```diff
@@ -168,6 +188,15 @@ const graphQlErrorsSerializer = (errors: Array<any>) => (
 	}
 );
 
+const sanitiseStackTrace = (stack: string) => {
```
rachellerathbone, line 191: same here

rachellerathbone (approved):

PR #2006, opened 2023-03-27
Arc 1997 increase page size to 100

bgvozdev:
**What's in this PR?**
Changing cursor type to be page-aware

**Why**
So we can seamlessly change page sizes

**Added feature flags**
Not yet, should I? Will rollback if something goes wrong 🤠


src/sync/sync-utils.ts
```diff
@@ -10,6 +10,55 @@ import { SyncType, TaskType } from "~/src/sync/sync.types";
...
+	 * 									the old sync is still in progress with pageNo size equals 20
+	 */
+	constructor(pageNoOrSerialisedCursor: string | number, perPage: number) {
+		if (!pageNoOrSerialisedCursor || Number(pageNoOrSerialisedCursor)) {
+			this.pageNo = Number(pageNoOrSerialisedCursor) || 1;
+			// Historical value we used to have in our code which was used for plain-number cursors.
+			// Ignoring perPage argument for seamless migration
+			this.perPage = 20;
+		} else {
+			const parsed = JSON.parse("" + pageNoOrSerialisedCursor) as PageSizeAwareCounterCursor;
```
gxueatlassian, line 28: How about the graphQL cursor, which is just a string?

bgvozdev: This logic is only for counter-based cursor, which is PRs and Builds. The rest entity types (graphQl) should continue working as they are now

gxueatlassian: How about extract this file? I mistakenly thought this new class is also related to sync-utils.

bgvozdev: Done :)

gxueatlassian (approved):

PR #2008, opened 2023-03-27
ARC-1997: control page size with a FF

bgvozdev:
**What's in this PR?**
 - Control page size with a FF
 - Control parallel fetching of PRs and builds by their own FFs

**Why**
Because it is cool.

### DIFF
--- a/src/sync/commits.ts
+++ b/src/sync/commits.ts
@@ -6,7 +6,6 @@ import { CommitQueryNode } from "../github/client/github-queries";
 import { JiraCommitBulkSubmitData } from "src/interfaces/jira";
 import { BackfillMessagePayload } from "~/src/sqs/sqs.types";
 import { TaskResultPayload } from "~/src/sync/sync.types";
-import { numberFlag, NumberFlags } from "config/feature-flags";
 
 const fetchCommits = async (gitHubClient: GitHubInstallationClient, repository: Repository, commitSince?: Date, cursor?: string | number, perPage?: number) => {
 	const commitsData = await gitHubClient.getCommitsPage(repository.owner.login, repository.name, perPage, commitSince, cursor);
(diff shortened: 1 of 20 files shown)


src/sync/discovery.ts
```diff
@@ -38,18 +39,18 @@ export const getRepositoryTask = async (
 	let edges: RepositoryNode[];
 	let repositories: Repository[];
 	if (await booleanFlag(BooleanFlags.USE_REST_API_FOR_DISCOVERY, jiraHost)) {
```
bgvozdev, line 41: Looks like the FF is off, which means the modified code will never fire in prod. However, tested it locally just in case

image

gxueatlassian (approved):

src/sync/commits.ts
```diff
@@ -22,26 +21,22 @@ const fetchCommits = async (gitHubClient: GitHubInstallationClient, repository:
 export const getCommitTask = async (
 	logger: Logger,
 	gitHubClient: GitHubInstallationClient,
-	jiraHost: string,
+	_jiraHost: string,
```
rachellerathbone, line 24: Looks like `getCommitTask` is only being used in sync/installation. If we don't need `jiraHost` anymore can we just remove this param?

src/sync/discovery.ts
```diff
@@ -7,20 +7,21 @@ import { RepoSyncState } from "models/reposyncstate";
 import { TaskResultPayload } from "~/src/sync/sync.types";
 import { BackfillMessagePayload } from "~/src/sqs/sqs.types";
 import { updateRepoConfigsFromGitHub } from "services/user-config-service";
+import { PageSizeAwareCounterCursor } from "~/src/sync/page-counter-cursor";
 
 export const getRepositoryTask = async (
 	logger: Logger,
-	newGithub: GitHubInstallationClient,
+	githubInstallationClient: GitHubInstallationClient,
```
rachellerathbone, line 14: nit: githubInstallationClient -> gitHubInstallationClient

rachellerathbone:
Merged already... :sigh:

PR #2012, opened 2023-03-29
Arc 2084 fix missing data on pr

gxueatlassian:
**What's in this PR?**
Return the PR data that could be missing if page returned contains data older than backfill from date.

**Why**
We don't want to miss any data.
I think the problem essentially lies in:
For graphQL it should be fine
For rest api, whenever we try to “filtering”, either on our consumer side (pullrequest) or from the github api side (build), it won’t work. It will miss data as the page size can be large, and bump to next page will miss data. Instead, if we “over fetch” instead of “filtering”, then it should be fine. Only drawback is user might see date earlier than their specified backfill date. But I think we can accept that for now.

**Added feature flags**
use incremental flag

**Affected issues**  
ARC-2084

**How has this been tested?**  
Unit test

**Whats Next?**

### DIFF
--- a/src/github/client/github-installation-client.ts
+++ b/src/github/client/github-installation-client.ts
@@ -209,9 +209,9 @@ export class GitHubInstallationClient extends GitHubClient {
 		);
 	};
 
-	public listWorkflowRuns = async (owner: string, repo: string, per_page, cursor?: number, fromDate?: Date): Promise<AxiosResponse<ActionsListRepoWorkflowRunsResponseEnhanced>> => {
+	public listWorkflowRuns = async (owner: string, repo: string, per_page, cursor?: number): Promise<AxiosResponse<ActionsListRepoWorkflowRunsResponseEnhanced>> => {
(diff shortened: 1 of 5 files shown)


bgvozdev (approved):

gxueatlassian:
@bgvozdev let me know if my updated description doesn't make sense. 
```
I think the problem essentially lies in:
For graphQL it should be fine
For rest api, whenever we try to “filtering”, either on our consumer side (pullrequest) or from the github api side (build), it won’t work. It will miss data as the page size can be large, and bump to next page will miss data. Instead, if we “over fetch” instead of “filtering”, then it should be fine. Only drawback is user might see date earlier than their specified backfill date. But I think we can accept that for now
```

bgvozdev:
Yeah, makes sense @gxueatlassian , however it would be nice to have same explanation in the code, otherwise someone might be tempted to "fix" it 😬  wdyt?

Harminder84 (approved):

PR #2020, opened 2023-03-30
update error modal copy and style

joshkay10:
**What's in this PR?**
Updated copy and styling of retry modal

1 error copy
Screen Shot 2023-04-03 at 3 29 33 PM

Multiple errors copy

Screen Shot 2023-04-03 at 3 29 01 PM


static/css/jira-configuration-new.css
```diff
@@ -88,14 +88,12 @@
 }
 
 .error-summary-errors-header {
-    font-size: 12px;
-    color: #6b778c;
+    padding-top: 1rem;
```
rachellerathbone, line 91: rem is typically used for font-size (relative to the root measurement). Use em for padding/margins so the size adjusts as the screen does.

rachellerathbone:
Doesn't look like the content was part of this PR but there's definitely a few things that read funny:

- 'The backfill finished but following...': Think this should be 'The backfill finished but the following...'
- 'This may be because of permission issues, GitHub API rate-limiting and network outages.': Unless the last 2 items in this list happen together I think 'and' should be replaced with 'or'
- 'Learn more...': this seems to be just randomly dropped in the middle of the paragraph. Is it meant to be there? If so, can we shift it to the end of the paragraph or leave it there but move the next sentence to a new paragraph. Also, guessing there's meant to be a link?
- I'd rewrite that whole last sentence to: 'You can continue using this integration, and view backfilled data in Jira, while the app attempts to reprocess these entities.'

views/partials/jira-configuration-error-summary.hbs
```diff
@@ -10,25 +10,27 @@
...
+         (you must be the admin of your organization), then retry backfill.
+      </li>
       {{/if}}
 
       {{#if syncErrors.CONNECTION_ERROR}}
-      <p>
-        We were unable to complete backfill for <b>{{syncErrors.CONNECTION_ERROR}}</b> repositories, because something unexpected occurred, please retry the backfill.
-      </p>
+      <li>
+        We were unable to complete backfill for <b>{{syncErrors.CONNECTION_ERROR}}</b> repositories, because something unexpected occurred, please retry the backfill or rasie a support ticket.
```
rachellerathbone, line 30: A couple of things here:

1. It doesn't look like there's any number in the screenshot you attached so it just says: 'We were unable to complete backfill for repositories'.
2. Edit: We were unable to complete backfill for <b>{{syncErrors.CONNECTION_ERROR}}</b> repositories because something unexpected occurred. Please retry the backfill or raise a support ticket. (removed comma before because, started a new sentence at 'please' and corrected rasie typo)

views/partials/jira-configuration-error-summary.hbs
```diff
@@ -10,25 +10,27 @@
...
-      </div>
+    <div class="error-summary-border-top">
+      <h4 class="error-summary-errors-header">
+        Please resolve the following errors:
+      </h4>
+      <ul>
       {{#if syncErrors.PERMISSIONS_ERROR}}
-      <div>
+      <li>
         <b>{{syncErrors.PERMISSIONS_ERROR}}</b> tasks have not been backfilled. Please
```
rachellerathbone, line 20: nit: If the count here is 1 can we apply some single logic to change tasks to task? 1 tasks reads funny.

views/partials/jira-configuration-error-summary.hbs
joshkay10: ahh yea, that wasnt a fair screenshot since i just forced it to show, so should be a number, ill re-screenshot
but yup agree will conditional on the 1 vs 1+
I accept the copy change too!

views/partials/jira-configuration-error-summary.hbs
joshkay10: will do!

gxueatlassian (approved):

PR #2027, opened 2023-04-03
Use scheduler to use quota in full

bgvozdev:
**What's in this PR?**
Implement parallel syncing

**Why**
Because large customers complaining

**Added feature flags**
Yes

**How has this been tested?**  
staging, prod (one particular customer), unit tests

**Whats Next?**
finish off tests

### DIFF
--- a/src/sqs/backfill.ts
+++ b/src/sqs/backfill.ts
@@ -1,3 +1,4 @@
+import Logger from "bunyan";
 import { processInstallation } from "../sync/installation";
 import * as Sentry from "@sentry/node";
 import { AxiosErrorEventDecorator } from "models/axios-error-event-decorator";
@@ -6,7 +7,7 @@ import { BackfillMessagePayload, MessageHandler, SQSMessageContext } from "./sqs
 import { SQS } from "aws-sdk";
 
 export const backfillQueueMessageHandler =
-	(sendSQSBackfillMessage: (message, delaySec, logger) => Promise<SQS.SendMessageResult>): MessageHandler<BackfillMessagePayload> =>
+	(sendSQSBackfillMessage: (message: BackfillMessagePayload, delaySec: number, logger: Logger) => Promise<SQS.SendMessageResult>): MessageHandler<BackfillMessagePayload> =>
 		async (context: SQSMessageContext<BackfillMessagePayload>) => {
 			const sentry = new Sentry.Hub(Sentry.getCurrentHub().getClient());
 			sentry.configureScope((scope) =>
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -74,7 +74,7 @@ export const updateTaskStatusAndContinue = async (
 	taskResultPayload: TaskResultPayload,
 	task: Task,
 	logger: Logger,
-	sendBackfillMessage: (message, delay, logger) => Promise<unknown>
+	sendBackfillMessage: (message, delaySecs, logger) => Promise<unknown>
 ): Promise<void> => {
 	// Get a fresh subscription instance
 	const subscription = await findSubscriptionForMessage(data);
@@ -241,7 +241,7 @@ const getTaskMetricsTags = (nextTask: Task) => {
 
 const isMainTask = (taskIndex: number) => taskIndex === 0;
 
-const doProcessInstallation = async (data: BackfillMessagePayload, sentry: Hub, rootLogger: Logger, sendBackfillMessage: (message: BackfillMessagePayload, delay: number, logger: Logger) => Promise<unknown>): Promise<void> => {
+const doProcessInstallation = async (data: BackfillMessagePayload, sentry: Hub, rootLogger: Logger, sendBackfillMessage: (message: BackfillMessagePayload, delaySecs: number, logger: Logger) => Promise<unknown>): Promise<void> => {
 	const { installationId: gitHubInstallationId, jiraHost } = data;
 	const subscription = await findSubscriptionForMessage(data);
 
@@ -334,7 +334,7 @@ const findSubscriptionForMessage = (data: BackfillMessagePayload) =>
 		data.gitHubAppConfig?.gitHubAppId
 	);
 
-export const markCurrentTaskAsFailedAndContinue = async (data: BackfillMessagePayload, nextTask: Task, isPermissionError: boolean, sendBackfillMessage: (message, delay, logger, err: Error) => Promise<unknown>, log: Logger, err: Error): Promise<void> => {
+export const markCurrentTaskAsFailedAndContinue = async (data: BackfillMessagePayload, nextTask: Task, isPermissionError: boolean, sendBackfillMessage: (message, delaySecs, logger, err: Error) => Promise<unknown>, log: Logger, err: Error): Promise<void> => {
 	const subscription = await findSubscriptionForMessage(data);
 	if (!subscription) {
 		log.warn("No subscription found, nothing to do");
@@ -397,7 +397,7 @@ const redis = new IORedis(getRedisInfo("installations-in-progress"));
 
 const RETRY_DELAY_BASE_SEC = 60;
 
-export const processInstallation = (sendBackfillMessage: (message: BackfillMessagePayload, delay: number, logger: Logger) => Promise<unknown>) => {
+export const processInstallation = (sendBackfillMessage: (message: BackfillMessagePayload, delaySecs: number, logger: Logger) => Promise<unknown>) => {
 	const inProgressStorage = new RedisInProgressStorageWithTimeout(redis);
 	const deduplicator = new Deduplicator(
 		inProgressStorage, 1_000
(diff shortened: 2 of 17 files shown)


src/sqs/backfill.ts
```diff
@@ -6,7 +7,7 @@ import { BackfillMessagePayload, MessageHandler, SQSMessageContext } from "./sqs
 import { SQS } from "aws-sdk";
 
 export const backfillQueueMessageHandler =
-	(sendSQSBackfillMessage: (message, delaySec, logger) => Promise<SQS.SendMessageResult>): MessageHandler<BackfillMessagePayload> =>
+	(sendSQSBackfillMessage: (message: BackfillMessagePayload, delaySec: number, logger: Logger) => Promise<SQS.SendMessageResult>): MessageHandler<BackfillMessagePayload> =>
```
rachellerathbone, line 10: Nice 🥳

src/sync/scheduler.ts
```diff
@@ -0,0 +1,146 @@
...
+		const rateLimitData = rateLimitResponse.data;
+
+		const availQuota = Math.min(rateLimitData.resources.core.remaining, rateLimitData.resources.graphql.remaining);
+		// We need to reserve some quota for the main task, too
+		const availQuotaForSubtasks = Math.max(0, availQuota - RATE_LIMIT_QUOTA_PER_TASK_RESERVE);
+		const allowedSubtasks = Math.floor(availQuotaForSubtasks / RATE_LIMIT_QUOTA_PER_TASK_RESERVE);
+
+		const nSubTasks = Math.min(allowedSubtasks, MAX_NUMBER_OF_SUBTASKS);
+
+		logger.info({ nSubTasks, rateLimitData }, "Using subtasks: " + nSubTasks);
```
rachellerathbone, line 39: nit: use string templating

src/sync/scheduler.ts
```diff
@@ -0,0 +1,146 @@
...
+			otherTasks.slice(0, nSubTasks).sort(() => Math.random() - 0.5)
+		)];
+	} catch (err) {
+		logger.warn({ err }, "Cannot determine rate limit, return only main task");
+		return [mainTask];
+	}
+};
+
+const mapSyncStateToTasks = (tasks: TaskType[], syncState: RepoSyncState): Task[] => {
+	const ret: Task[] = [];
```
rachellerathbone, line 55: What is ret? A little hard to determine what this is from the name. Maybe `syncStateTask` or something along those lines?

src/sync/scheduler.ts
```diff
@@ -0,0 +1,146 @@
...
+			}
+		}
+	);
+	return ret;
+};
+
+/**
+ *
+ * @param subscription
+ * @param targetTasks - the first task is always same/deterministic! The tail is random from not complete ones tasks.
```
rachellerathbone, line 81: What's the reason for combining JSdocs with TS types? When we migrated from JS -> TS we made an effort to do away with all JSDocs. The additional info for `targetTasks` and following comments are great, but not sure we need to list everything here when we're already typing these in the function signature. Same goes for `calculateTasksUsingGitHubRateLimitQuota`

bgvozdev: Otherwise IDEA will be complaining....

rachellerathbone: Screenshot 2023-04-04 at 11 12 29 am

Just removed the additional params (with no comment) from `calculateTasksUsingGitHubRateLimitQuota` and committed and didn't run into any issues.

src/sync/scheduler.ts
```diff
@@ -0,0 +1,146 @@
...
+		if (otherTasks.length > SUBTASKS_POLL_MAX_SIZE) {
+			break;
+		}
+	}
+
+	if (!mainTask) {
+		return [];
+	}
+
+	if (otherTasks.length == 0) {
```
rachellerathbone, line 141: Change `==` (loosely equals) to `===` (strictly equals)

src/sync/installation.ts
```diff
@@ -109,7 +75,7 @@ export const updateTaskStatusAndContinue = async (
 	taskResultPayload: TaskResultPayload,
 	task: Task,
 	logger: Logger,
-	sendBackfillMessage: (message, delay, logger) => Promise<unknown>
+	sendBackfillMessage: (message, delaySecs, logger) => Promise<unknown>
```
gxueatlassian, line 78: :nice

src/sync/installation.ts
```diff
@@ -326,7 +327,7 @@ const findSubscriptionForMessage = (data: BackfillMessagePayload) =>
 		data.gitHubAppConfig?.gitHubAppId
 	);
 
-export const markCurrentTaskAsFailedAndContinue = async (data: BackfillMessagePayload, nextTask: Task, isPermissionError: boolean, sendBackfillMessage: (message, delay, logger, err: Error) => Promise<unknown>, log: Logger, err: Error): Promise<void> => {
+export const markCurrentTaskAsFailedAndContinue = async (data: BackfillMessagePayload, mainNextTask: Task, isPermissionError: boolean, sendBackfillMessage: (message, delaySecs, logger, err: Error) => Promise<unknown>, log: Logger, err: Error): Promise<void> => {
```
gxueatlassian, line 330: `mainNextTask` <-- why don't we rename it as `currentTask` or `task`?

src/sync/installation.ts
```diff
@@ -412,14 +413,32 @@ export const processInstallation = (sendBackfillMessage: (message, delay, logger
...
 
+			let hasNextMessage = false;
+			let nextMessage: BackfillMessagePayload | undefined = undefined;
+			let nextMessageDelaySecs: number | undefined = undefined;
+			let nextMessageLogger: Logger | undefined = undefined;
+
 			const result = await deduplicator.executeWithDeduplication(
 				`i-${installationId}-${jiraHost}-ghaid-${gitHubAppId || "cloud"}`,
-				() => doProcessInstallation(data, sentry, logger, sendBackfillMessage)
+				() => doProcessInstallation(data, sentry, logger, (message, delaySecs, logger) => {
```
gxueatlassian, line 423: So this is a current prod bug?

src/sync/installation.ts
bgvozdev: I just wanted to highlight that this function should be called only for the main tasks

src/sync/installation.ts
bgvozdev: But-not-bug @gxueatlassian because currently we are processing only one task and schedule the message at the very end. Till the moment the message is read by a worker, we already release the flag.

However, with this change, this is exacerbated because after the main task is over some subtasks might still be in flight.

src/sync/scheduler.ts
```diff
@@ -0,0 +1,144 @@
...
+import { getTargetTasks } from "~/src/sync/installation";
+import { createInstallationClient } from "utils/get-github-client-config";
+import Logger from "bunyan";
+import { booleanFlag, BooleanFlags } from "config/feature-flags";
+
+const getCursorKey = (type: TaskType) => `${type}Cursor`;
+const getStatusKey = (type: TaskType) => `${type}Status`;
+
+// These numbers were obtained experimentally by syncing a large customer with 60K quota from GitHub
+const RATE_LIMIT_QUOTA_PER_TASK_RESERVE = 500;
```
gxueatlassian, line 13: /nit
Some thinkings, from ops' perspective, we want to adjust our apps somehow during runtime on the fly. So those "knobs", should we always use a FF to control it? Or should there be an admin api to control it? Should we use stash to control it (but re-deploy is required)? Or just hard code it in this case. 

We may want to talk about it for improve ops process. But for now I am okay with hardcoded.

bgvozdev: The idea is to have these small/aggressive enough so we would use all quota (not to slug behind GitHub dropping their rate-limiting counters).  Once we reach the ceiling, preemptive rate-limiting will make it chill.

But yeah, it could be adjust later, just don't think it will make too much difference. This would be very-very fine-grained tuning :)

src/sync/installation.ts
gxueatlassian: `Till the moment the message is read by a worker, we already release the flag.` <-- Technically, although very unlikely, that this still a problem of current prod. Like send a msg at the end of the processing, but other worker node pick that msg up, check dedupe and find out it is not released yet.

But all good, you fix it anyway 🤷 👍

src/sync/scheduler.ts
```diff
@@ -0,0 +1,144 @@
...
+
+/**
+ *
+ * @param subscription
+ * @param targetTasks - the first task is always same/deterministic! The tail is random from not complete ones tasks.
+ * 											The caller can call this function after some delay again (e.g. SQS exponential backoff) and
+ * 											it is guaranteed they will receive same task as the very first one (a.k.a. "main task")
+ * @param logger
+ */
+export const getNextTasks = async (subscription: Subscription, targetTasks: TaskType[], logger: Logger): Promise<Task[]> => {
```
gxueatlassian, line 84: /nit
Instead of `Promise<Task[]>`, maybe following can make code more readable.
```javascript
Promise<{
  mainTask: Task,
  otherTasks: Task[]
}>
```

rachellerathbone (approved):

src/sync/installation.ts
bgvozdev: Yeah, pretty sure this happens. Not a catastrophe because it will just cause 1 minute delay. However that was quite a problem for that particular customer cause everything because sloooow.

src/sync/scheduler.ts
```diff
@@ -0,0 +1,144 @@
...
+		const availQuota = Math.min(rateLimitData.resources.core.remaining, rateLimitData.resources.graphql.remaining);
+		// We need to reserve some quota for the main task, too
+		const availQuotaForSubtasks = Math.max(0, availQuota - RATE_LIMIT_QUOTA_PER_TASK_RESERVE);
+		const allowedSubtasks = Math.floor(availQuotaForSubtasks / RATE_LIMIT_QUOTA_PER_TASK_RESERVE);
+
+		const nSubTasks = Math.min(allowedSubtasks, MAX_NUMBER_OF_SUBTASKS);
+
+		logger.info({ nSubTasks, rateLimitData }, `Using subtasks: ${nSubTasks}`);
+
+		if (nSubTasks === 0) {
```
gxueatlassian, line 39: /nit
Move this up to avoid calling rate limit api?

src/sync/scheduler.test.ts
```diff
@@ -0,0 +1,221 @@
...
+		when(booleanFlag).calledWith(
+			BooleanFlags.USE_SUBTASKS_FOR_BACKFILL,
+			expect.anything()
+		).mockResolvedValue(true);
+
+		configureRateLimit(100000, 2000);
+
+		githubUserTokenNock(DatabaseStateCreator.GITHUB_INSTALLATION_ID);
+		const tasks = await getNextTasks(subscription, [], getLogger("test"));
+		expect(tasks.length).toEqual(4);
```
gxueatlassian, line 99: /nit
I dont' get where does the `4` comes from within this test, unless I dig into the impl code, which I don't want to.

gxueatlassian (approved):
Looks good to me. Happy to approve.

PR #2030, opened 2023-04-05
ARC-1440 - Check user admin permissions from Jira - PART 1

joshkay10:
**What's in this PR?**
During the JWT test, we now call to Jira to check that the user has admin privileges.
We store that response in the session.

If the session aleady contains a response we dont call Jira, calling it everytime seems expensive!

This doesnt do anything with that value.... yet, this is the pre work before we test on endpoint for the user to have correct access and will be in a PR next

**How has this been tested?**  
Whole bunch of new unit tests


src/middleware/jira-symmetric-jwt-middleware.test.ts
```diff
@@ -3,13 +3,31 @@ import express, { Application, NextFunction, Request, Response } from "express";
...
+jest.mock("~/src/jira/client/jira-client", () => {
+	return {
+		getJiraClient: jest.fn()
+	};
+});
 
 jest.mock("config/feature-flags");
 const testSharedSecret = "test-secret";
 
+const getToken = ({
```
joshkay10, line 19: This isnt a new function i just moved it up since i introduced a new describe block, felt better but meh

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -5,6 +5,31 @@ import { getJWTRequest, TokenType, validateQsh } from "~/src/jira/util/jwt";
...
+	// We only need to add this to the session if it doesn't exist
+	if (req.session.isJiraAdmin !== undefined) {
+		return;
+	}
+	try {
+		const userAccountId = claims.sub;
+		const jiraClient = await getJiraClient(
+			jiraHost,
+			installationId,
+			undefined
```
joshkay10, line 20: undefined for githubappid.... I dont think it matters, this is for us to ask Jira about Jira permissions so the GitHub app should be irrelevant, the previous args are required for the function though

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -29,8 +54,10 @@ export const jiraSymmetricJwtMiddleware = async (req: Request, res: Response, ne
 		const secret = await installation.decrypt("encryptedSharedSecret", req.log);
 
 		const tokenType = checkPathValidity(req.originalUrl) && req.method == "GET" ? TokenType.normal : TokenType.context;
+		let verifiedClaims;
 		try {
-			verifySymmetricJwt(token, secret, req, tokenType, req.log);
+			verifiedClaims = verifySymmetricJwt(token, secret, req.log);
+			verifyJwtClaims(verifiedClaims, tokenType, req);
```
joshkay10, line 60: I pulled this out `verifyJwtClaims` of `verifySymmetricJwt` function, just because i needed to decode the JWT again and didnt want to rerun that code, this way its available to me for `setJiraAdminPrivileges`

src/jira/client/jira-client.ts
```diff
@@ -327,6 +327,19 @@ export const getJiraClient = async (
...
+		permissions: {
+			checkAdmin: async (accountId: string) => {
+				const payload = {
+					accountId,
+					globalPermissions: [
+						"ADMINISTER"
+					]
+				};
+				logger.info("Requesting user permissions from jira.");
+				const response: AxiosResponse = await instance.post("/rest/api/latest/permissions/check", payload);
```
bgvozdev, line 340: It does't look like we need installationId for this call, right? If we don't need it, I suggest to put this code into models/jira-client instead... wdyt?

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -40,6 +67,9 @@ export const jiraSymmetricJwtMiddleware = async (req: Request, res: Response, ne
 		res.locals.jiraHost = installation.jiraHost;
 		req.session.jiraHost = installation.jiraHost;
 
+		// Check whether logged in user has Jira Admin permissions and save it to the session
+		await setJiraAdminPrivileges(req, verifiedClaims, installation.jiraHost, installation.id);
```
joshkay10, line 71: Updating this method, forgot some types!

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -5,6 +5,31 @@ import { getJWTRequest, TokenType, validateQsh } from "~/src/jira/util/jwt";
...
 import { matchRouteWithPattern } from "~/src/util/match-route-with-pattern";
+import { getJiraClient } from "~/src/jira/client/jira-client";
+
+export const setJiraAdminPrivileges = async (req: Request, claims, jiraHost, installationId) => {
+	// We only need to add this to the session if it doesn't exist
+	if (req.session.isJiraAdmin !== undefined) {
+		return;
+	}
+	try {
+		const userAccountId = claims.sub;
```
gxueatlassian, line 16: `sub` could be empty according to https://developer.atlassian.com/cloud/jira/platform/understanding-jwt-for-connect-apps/#claims
So maybe we should assume if `accountId` is empty, then assume the action is triggered from somewhere non-user, and therefore as "admin"?
image

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -5,6 +5,31 @@ import { getJWTRequest, TokenType, validateQsh } from "~/src/jira/util/jwt";
...
+		const jiraClient = await getJiraClient(
+			jiraHost,
+			installationId,
+			undefined
+		);
+
+		// Make jira call to permissions with userAccountId.
+		const permissions = await jiraClient.permissions.checkAdmin(userAccountId);
+		const hasAdminPermissions = permissions.data.globalPermissions.includes("ADMINISTER");
```
gxueatlassian, line 25: /nit
put that `ADMINISTER` as const str or enum somewhere?

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -5,6 +5,31 @@ import { getJWTRequest, TokenType, validateQsh } from "~/src/jira/util/jwt";
...
+
+		req.session.isJiraAdmin = hasAdminPermissions;
```
gxueatlassian, line 27: How do we refresh this var? Say 1) user figure our they can't do something in our app, 2) they are added as admin 3) they come back still can't do thing since the session contains `isAdmin: fasle`

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -83,19 +113,15 @@ const getIssuer = (token: string, logger: Logger): string | undefined => {
 	return unverifiedClaims.iss;
 };
 
-const verifySymmetricJwt = (token: string, secret: string, req: Request, tokenType: TokenType, logger: Logger): boolean => {
+const verifySymmetricJwt = (token: string, secret: string, logger: Logger) => {
```
gxueatlassian, line 116: Maybe the method name should be getDecodedClaimes , decodeCliams, etc, instead of verifyXXX as it doesn't verify?

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -29,8 +54,10 @@ export const jiraSymmetricJwtMiddleware = async (req: Request, res: Response, ne
 		const secret = await installation.decrypt("encryptedSharedSecret", req.log);
 
 		const tokenType = checkPathValidity(req.originalUrl) && req.method == "GET" ? TokenType.normal : TokenType.context;
+		let verifiedClaims;
 		try {
-			verifySymmetricJwt(token, secret, req, tokenType, req.log);
+			verifiedClaims = verifySymmetricJwt(token, secret, req.log);
```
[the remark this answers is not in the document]
joshkay10: haha yeah it is a bit cheeky isn't it. ill adjust, i reckon I could adjust it to be proper with some name changes u mentioned as well

src/middleware/jira-symmetric-jwt-middleware.ts
joshkay10: ha, yeah I considered this one, then kinda thought it would be ok, they need to start a new session so a fresh browser would work i think, i dunno about a tab? but it is a pretty esoteric way to get it working.

I might be able to set it against the jira login token so they would still need to log out and log in but I think thats pretty standard for perm changes anyway

src/middleware/jira-symmetric-jwt-middleware.ts
joshkay10: if they arent logged in it will be emtpy so thats fine.

As far as missing due to accessing from another methid or some such, we can handle that when we refuse the endpoint. e.g. if(session.isJiraAdmin === undefined || developers_backdoor_muahahahha)

src/middleware/jira-symmetric-jwt-middleware.ts
joshkay10: So.. its bit of a nuisance for this is refresh, I dont want to check everytime as its a bit expensive time wise.

What we can do though, we are going to present failed/rejected users with some sort of messaging. we can tell them there to start it in a fresh login/browser??

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -26,20 +49,20 @@ export const jiraSymmetricJwtMiddleware = async (req: Request, res: Response, ne
 			return res.status(401).send("Unauthorised");
 		}
 
-		const secret = await installation.decrypt("encryptedSharedSecret", req.log);
-
-		const tokenType = checkPathValidity(req.originalUrl) && req.method == "GET" ? TokenType.normal : TokenType.context;
+		let verifiedClaims;
 		try {
-			verifySymmetricJwt(token, secret, req, tokenType, req.log);
+			verifiedClaims = await verifySymmetricJwt(req, token, installation);
```
joshkay10, line 54: now they are actually verified @gxueatlassian lololol

src/models/jira-client.ts
```diff
@@ -52,4 +52,15 @@ export class JiraClient {
 	async appPropertiesDelete() {
 		return await this.axios.delete(`/rest/atlassian-connect/latest/addons/${getAppKey()}/properties/is-configured`);
 	}
+
+	async checkAdminPermissions(accountId: string) {
```
rachellerathbone, line 56: Going off Gary's comment above https://github.com/atlassian/github-for-jira/pull/2030/files#r1157920279, if it turns out sub isn't present because the user isn't logged in, are we expecting an empty string? Or would it be null/undefined? If it's either of the latter we should probably update the expected type to reflect this. Or better yet, add a check before we even call `checkAdminPermissions`

joshkay10: Good call!! have returned early, no userId, no care, have added test case too 👍 ⚡

src/middleware/jira-symmetric-jwt-middleware.ts
joshkay10: Tidied up a bit and moved secret and tokentype into this function

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -5,6 +5,32 @@ import { getJWTRequest, TokenType, validateQsh } from "~/src/jira/util/jwt";
...
+	// We only need to add this to the session if it doesn't exist
+	if (req.session.isJiraAdmin !== undefined) {
+		return;
+	}
+
+	try {
+		const userAccountId = claims.sub;
+		// Can't check permissions without userAccountId
+		if (!userAccountId) {
+			return;
```
gxueatlassian, line 21: So `req.session.isJiraAdmin` will be `undefined`, and in part 2 of this PR, do we allow this request or block it? 

Can I assume the current prod behaviour is "block" since there's no check anyway?

src/middleware/jira-symmetric-jwt-middleware.ts
gxueatlassian: Hmm, actually, I think we might be able to even avoid putting that `isJiraAdmin` into session.
Can you check in our normal workflow within the app (not the create branch), the jwt token will it contains user id? 
If no, then there's no need to put it in session. Coz in create branch we don't check the admin permission. So not much place we really need to check admin permission because this code return earlier if userAccountId is empty.

joshkay10: Create branch has user id on it, just double checked.

gxueatlassian (approved):

PR #2040, opened 2023-04-11
created new jira admin middleware and tests to return 403 when no auth

joshkay10:
**What's in this PR?**
Created new middleware to return 403 for endpoints that require Jira Admin privileges

**Why**
So we can be sure that endpoints are being called by allowed folks.
Plus the create branch JWT token can be any user so we dont want any funny business

**Added feature flags**
JIRA_CHECK_ADMIN

**Affected issues**  
_Jira Issues_

**How has this been tested?**  
Complete covereage unit tests, testing on STAGE with on/off flag

**Whats Next?**
Remove the user token from create branch and turn on JWT for create branch

### DIFF
--- a/src/middleware/jira-admin-permission-middleware.ts
+++ b/src/middleware/jira-admin-permission-middleware.ts
@@ -17,11 +17,9 @@ export const setJiraAdminPrivileges = async (req: Request, claims: Record<any, a
 			return;
 		}
 		const jiraClient = await JiraClient.getNewClient(installation, req.log);
-		// Make jira call to permissions with userAccountId.
 		const permissions = await jiraClient.checkAdminPermissions(userAccountId);
-		const hasAdminPermissions = permissions.data.globalPermissions.includes(ADMIN_PERMISSION);
 
-		req.session.isJiraAdmin = hasAdminPermissions;
+		req.session.isJiraAdmin = permissions.data.globalPermissions.includes(ADMIN_PERMISSION);
 		req.log.info({ isAdmin :req.session.isJiraAdmin }, "Admin permissions set");
 	} catch (err) {
 		req.log.error({ err }, "Failed to fetch Jira Admin rights");
@@ -29,18 +27,18 @@ export const setJiraAdminPrivileges = async (req: Request, claims: Record<any, a
 };
 
 export const jiraAdminPermissionsMiddleware = async (req: Request, res: Response, next: NextFunction): Promise<void | Response>  => {
-	const hasAdminPermissions = req.session;
+	const { isJiraAdmin } = req.session;
 	if (!(await booleanFlag(BooleanFlags.JIRA_ADMIN_CHECK))) {
 		return next();
 	}
 
-	if (hasAdminPermissions === undefined) {
+	if (isJiraAdmin === undefined) {
 		// User permissions could bot be extracted from the Jira JWT, check that the jwt middleware has run
 		req.log.info("No Jira user permissions found");
 		return res.status(403).send("Forbidden - User Jira permissions have not been found.");
 	}
 
-	if (hasAdminPermissions === "false") {
+	if (!isJiraAdmin) {
 		req.log.info("User does not have Jira admin permissions.");
 		return res.status(403).send("Forbidden - User does not have Jira administer permissions.");
 	}
--- a/src/middleware/jira-admin-permission-middleware.ts
+++ b/src/middleware/jira-admin-permission-middleware.ts
@@ -19,7 +19,7 @@ export const fetchAndSaveUserJiraAdminStatus = async (req: Request, claims: Reco
 		const jiraClient = await JiraClient.getNewClient(installation, req.log);
 		const permissions = await jiraClient.checkAdminPermissions(userAccountId);
 
-		req.session.isJiraAdmin = !permissions.data.globalPermissions.includes(ADMIN_PERMISSION);
+		req.session.isJiraAdmin = permissions.data.globalPermissions.includes(ADMIN_PERMISSION);
 		req.log.info({ isAdmin :req.session.isJiraAdmin }, "Admin permissions set");
 	} catch (err) {
 		req.log.error({ err }, "Failed to fetch Jira Admin rights");
--- a/src/middleware/jira-admin-permission-middleware.ts
+++ b/src/middleware/jira-admin-permission-middleware.ts
@@ -3,7 +3,7 @@ import { Installation } from "models/installation";
 import { JiraClient } from "models/jira-client";
 import { booleanFlag, BooleanFlags } from "config/feature-flags";
 
-export const setJiraAdminPrivileges = async (req: Request, claims: Record<any, any>, installation: Installation) => {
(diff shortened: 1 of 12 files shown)


src/middleware/jira-admin-permission-middleware.ts
```diff
@@ -0,0 +1,48 @@
...
+		const hasAdminPermissions = permissions.data.globalPermissions.includes(ADMIN_PERMISSION);
+
+		req.session.isJiraAdmin = hasAdminPermissions;
+		req.log.info({ isAdmin :req.session.isJiraAdmin }, "Admin permissions set");
+	} catch (err) {
+		req.log.error({ err }, "Failed to fetch Jira Admin rights");
+	}
+};
+
+export const jiraAdminPermissionsMiddleware = async (req: Request, res: Response, next: NextFunction) => {
```
joshkay10, line 31: Why do we rely on the session value hasAdminPermissions instead of reading it directyl from the token??

I agree, mostly, this means we a tightly coupled with the JWT middleware and that we need to put this middleware after every time, kinda annoying, but have put in some comments and messaging to help raise that if it happens.

The problem being, if we decide to decouple and decode the token with cert here, its duplicating the code of the JWT middleware so then we would need to refactor that so its reuseable in both places, while I don't disagree with that approach, the effort was too high and this was imo the best trade off. 
Open to opinions tho!

src/middleware/jira-symmetric-jwt-middleware.test.ts
```diff
@@ -228,72 +227,3 @@ describe("jiraSymmetricJwtMiddleware", () => {
 	};
 
 });
-
```
joshkay10, line 231: Just moved these tests to the jira-admin middleware

src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -5,32 +5,7 @@ import { getJWTRequest, TokenType, validateQsh } from "~/src/jira/util/jwt";
 import { Installation } from "~/src/models/installation";
 import { moduleUrls } from "~/src/routes/jira/atlassian-connect/jira-atlassian-connect-get";
 import { matchRouteWithPattern } from "~/src/util/match-route-with-pattern";
-import { JiraClient } from "models/jira-client";
-
-export const setJiraAdminPrivileges = async (req: Request, claims: Record<any, any>, installation: Installation) => {
```
joshkay10, line 10: Just moved this function to the jira-admin middleware

src/middleware/jira-admin-permission-middleware.ts
```diff
@@ -0,0 +1,48 @@
+import { NextFunction, Request, Response } from "express";
+import { Installation } from "models/installation";
+import { JiraClient } from "models/jira-client";
+import { booleanFlag, BooleanFlags } from "config/feature-flags";
+
+export const setJiraAdminPrivileges = async (req: Request, claims: Record<any, any>, installation: Installation) => {
```
rachellerathbone, line 6: nit: can we add a return type here?

src/middleware/jira-admin-permission-middleware.ts
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	try {
+		const userAccountId = claims.sub;
+		// Can't check permissions without userAccountId
+		if (!userAccountId) {
+			return;
+		}
+		const jiraClient = await JiraClient.getNewClient(installation, req.log);
+		// Make jira call to permissions with userAccountId.
```
rachellerathbone, line 20: Do we need this comment? I feel like the line below speaks for itself.

src/middleware/jira-admin-permission-middleware.ts
```diff
@@ -0,0 +1,48 @@
...
+
+		req.session.isJiraAdmin = hasAdminPermissions;
+		req.log.info({ isAdmin :req.session.isJiraAdmin }, "Admin permissions set");
+	} catch (err) {
+		req.log.error({ err }, "Failed to fetch Jira Admin rights");
+	}
+};
+
+export const jiraAdminPermissionsMiddleware = async (req: Request, res: Response, next: NextFunction) => {
+	const hasAdminPermissions = req.session;
```
rachellerathbone, line 32: Should this be `const hasAdminPermissions = req.session.isJiraAdmin`?

src/middleware/jira-admin-permission-middleware.ts
```diff
@@ -0,0 +1,48 @@
...
+		return next();
+	}
+
+	if (hasAdminPermissions === undefined) {
+		// User permissions could bot be extracted from the Jira JWT, check that the jwt middleware has run
+		req.log.info("No Jira user permissions found");
+		return res.status(403).send("Forbidden - User Jira permissions have not been found.");
+	}
+
+	if (hasAdminPermissions === "false") {
```
rachellerathbone, line 43: Doesn't the includes method return a boolean value? Shouldn't this be `if (!hasAdminPermissions)`

src/middleware/jira-admin-permission-middleware.ts
joshkay10: yes, yes we can. good catch!

joshkay10: might almost be time to enforce this eslint rule with an error
`@typescript-eslint/explicit-function-return-type`

src/middleware/jira-admin-permission-middleware.ts
joshkay10: ahhhhh thank you!! this was {hasAdminPermissions}

src/middleware/jira-admin-permission-middleware.ts
```diff
@@ -0,0 +1,46 @@
...
+
+		req.session.isJiraAdmin = permissions.data.globalPermissions.includes(ADMIN_PERMISSION);
+		req.log.info({ isAdmin :req.session.isJiraAdmin }, "Admin permissions set");
+	} catch (err) {
+		req.log.error({ err }, "Failed to fetch Jira Admin rights");
+	}
+};
+
+export const jiraAdminPermissionsMiddleware = async (req: Request, res: Response, next: NextFunction): Promise<void | Response>  => {
+	const { isJiraAdmin } = req.session;
```
bgvozdev, line 30: nit could be moved after FF check

src/middleware/jira-admin-permission-middleware.ts
```diff
@@ -0,0 +1,46 @@
+import { NextFunction, Request, Response } from "express";
+import { Installation } from "models/installation";
+import { JiraClient } from "models/jira-client";
+import { booleanFlag, BooleanFlags } from "config/feature-flags";
+
+export const setJiraAdminPrivileges = async (req: Request, claims: Record<any, any>, installation: Installation): Promise<void> => {
```
bgvozdev, line 6: nit the name sounds like we are rising up the flag.  Maybe something like "fetchAndSaveUserAdminStatus" ? I know it is not ideal too...

bgvozdev (approved):

bgvozdev (approved):

bgvozdev (approved):

PR #2043, opened 2023-04-12
ARC 1351 - [SECURITY] GitHub for Jira app: double-check the requirements for GitHub user OAuth scope

krazziekay:
**What's in this PR?**
- Updated the scopes variable to include string values from a string FF. 

**Why**
- The current scopes user & repo are very huge scopes. The goal is to narrow it to down to only those scopes that are needed.
- This approach of using a String FF is done so that we can test it easily in staging and production without having to do any code changes.

**Added feature flags**
- New String FF `github-scopes`

**How has this been tested?**  
- Locally
- Staging
- Prod

**Whats Next?**
- Test it in my instance first, then run a test blitz with the team


src/config/feature-flags.ts
```diff
@@ -21,6 +21,7 @@ export enum BooleanFlags {
 }
 
 export enum StringFlags {
+	GITHUB_SCOPES = "github-scopes",
```
krazziekay, line 24: The Feature Flag is not ready yet. Having permission issues while creating in LaunchDarkly..
Working on it!

atrigueiro:
LGTM, however, I'm not too sure if the tests here cover all scenarios where the app uses the token, e.g the [scopes specify what data](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes) we can/can't retrieve so I would like to test scenarios involving retrieval of repositories and user data (such as adding reviewers). Happy to approve that but I think we'll need a Confluence page for the testing blitz with that in mind too.

atrigueiro (approved):

atrigueiro:
Out of curiosity: when/if we change the scopes, will the users get an email from GH informing the change, such as when we change the app's permissions or nah?

krazziekay:
> LGTM, however, I'm not too sure if the tests here cover all scenarios where the app uses the token, e.g the [scopes specify what data](https://docs.github.com/en/apps/oauth-apps/building-oauth-apps/scopes-for-oauth-apps#available-scopes) we can/can't retrieve so I would like to test scenarios involving retrieval of repositories and user data (such as adding reviewers). Happy to approve that but I think we'll need a Confluence page for the testing blitz with that in mind too.

I will run a testing blitz once I test this myself first.


> Out of curiosity: when/if we change the scopes, will the users get an email from GH informing the change, such as when we change the app's permissions or nah?

I don't think there is. I've been testing it on my GitHub account and haven't received any emails yet.

PR #2049, opened 2023-04-14
ARC 2092 - Adding Sonarqube Scan in GitHub

krazziekay:
**What's in this PR?**
- Adding SonarQube scan in GitHub

**Why**
- Earlier this was done in the Bitbucket pipelines, during CD process. Doing it in CD process was incorrect, so doing it here in GitHub when PR is build, so that we can catch any code issues during build and not after pushing it.

### DIFF
--- a/.github/workflows/on-push.yml
+++ b/.github/workflows/on-push.yml
@@ -9,6 +9,20 @@ on:
   pull_request:
 
 jobs:
+  sonarqube:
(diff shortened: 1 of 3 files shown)


krazziekay:
SonarQube can not be implemented from Github because the SonarQube Server is not accessible by any external agents.

PR #2057, opened 2023-04-18
VULN 1054674 - Stored XSS via Branch name at [ GitHub plugin ]

krazziekay:
**What's in this PR?**
- Fixing the XSS via branch name when Creating a new branch.

**Why**
- To avoid XSS attacks via GitHub branch names. [VULN TIcket](https://asecurityteam.atlassian.net/browse/VULN-1054674)

**How has this been tested?**  
- Local
- Test cases added


src/routes/github/create-branch/github-branches-get.ts
```diff
@@ -23,8 +24,11 @@ export const GithubBranchesGet = async (req: Request, res: Response): Promise<vo
 		]);
 
 		res.send({
-			branches: branches.data.filter(branch => branch.name !== repository.data.default_branch),
-			defaultBranch: repository.data.default_branch
+			branches: branches.data.filter(branch => branch.name !== repository.data.default_branch).map(branch => ({
+				...branch,
+				name: sanitizeHtml(branch.name)
```
gxueatlassian, line 29: I wonder how about other fields in the `...branch`? Should we sanitize them as well?

gxueatlassian: Coz, someone later one must forgot to sanitise it and use that field.
And from backend to frontend, it might be better if we explicit pick the fields instead of using `...branches`

gxueatlassian (approved):

src/routes/github/create-branch/github-branches-get.ts
krazziekay: Besides the branch name, there are 2 other properties:
- One is `commit` which is an object containing `sha` and `url`, so no need here.
- And the other one is `protected`, which is a boolean, so not needed here too.

PR #2059, opened 2023-04-19
noissue: allow admins to call /queues endpoint

atrigueiro:


gxueatlassian (approved):
how about stg?

atrigueiro:
> how about stg?

Updating the `main` file updates for both staging and prod

PR #2061, opened 2023-04-19
add metrics for backfill tasks

gxueatlassian:
**What's in this PR?**
Add duration time metrics for backfill tasks

**Why**
Want to see how much time does each task take.

**Added feature flags**
N/A

**Affected issues**  
N/A

**How has this been tested?**  
N/A

**Whats Next?**

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -95,6 +95,8 @@ export const updateTaskStatusAndContinue = async (
 
 	const updateRepoSyncFields: { [x: string]: string | Date} = { [getStatusKey(task.task)]: status };
 
+	const duration = task.startTime ? Date.now() - task.startTime : undefined;
+
 	if (isComplete) {
 		//Skip branches as it sync all history
 		if (allTasksExceptBranch.includes(task.task) && data.commitsFromDate) {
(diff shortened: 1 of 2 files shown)


src/sync/installation.ts
```diff
@@ -108,10 +110,10 @@ export const updateTaskStatusAndContinue = async (
 			}
 		}
 
-		statsd.increment(metricTaskStatus.complete, [`type:${task.task}`, `gitHubProduct:${gitHubProduct}`]);
+		statsd.increment(metricTaskStatus.complete, duration, [`type:${task.task}`, `gitHubProduct:${gitHubProduct}`]);
```
bgvozdev, line 113: Shouldn't it be a separate metric? If I understand the signature of this method correctly, you are going to add this number of milliseconds, which probably is not something we want? (I'd expect to use histogram for duration...)

rachellerathbone: Yeah I agree. We use histogram to show time, in 'buckets'. Example https://github.com/atlassian/github-for-jira/blob/main/src/jira/client/axios.ts#L118

gxueatlassian: Thanks @bgvozdev and @rachellerathbone , good call!
I've updated it to histogram.

gxueatlassian: Oh and thanks for the code sample. I didn't use the bucket/histogram however, as I don't know what the bucket should be. So just call that `histogram`  function is should generate the default (count/median/lower_xx/upper_xx), I will start from there.

mboudreau (approved):

src/sync/installation.ts
```diff
@@ -95,6 +95,8 @@ export const updateTaskStatusAndContinue = async (
 
 	const updateRepoSyncFields: { [x: string]: string | Date} = { [getStatusKey(task.task)]: status };
 
+	const duration = task.startTime ? Date.now() - task.startTime : NaN;
```
rachellerathbone, line 98: Why are we defaulting to `NaN` here? The second param needs to be a number `histogram(stat: string | string[], value: number, tags?: Tags, callback?: StatsCb): void;`

gxueatlassian: haha, too late.

rachellerathbone: Can you update in a new PR? I don't think we should be potentially passing a type that isn't supported.

gxueatlassian: > Why are we defaulting to `NaN` here? The second param needs to be a number `histogram(stat: string | string[], value: number, tags?: Tags, callback?: StatsCb): void;`

Sorry, let me explain, I can push more code if need to.
So that typescript required it to be non undefined, has to be a number. And in js, `NaN` is considered to be a number.

My only concern is whether that `NaN` will impact the metrics. I hope not.
Otherwise, I have to say:
```
if (duration) 
  statsd.histogram(xxx, duration, xxx)
else
  statsd.histogram(xxx, xxx)
```
which make the code a bit messy.

bgvozdev: I'd prefer this tbh rather than having javascript witchery

PR #2062, opened 2023-04-19
None add task duration time metrics

gxueatlassian:
**What's in this PR?**
address previous comments.

### DIFF
--- a/src/sync/installation.ts
+++ b/src/sync/installation.ts
@@ -49,10 +49,12 @@ export const getTargetTasks = (targetTasks?: TaskType[]): TaskType[] => {
 
 export class TaskError extends Error {
 	task: Task;
+	duration: number;
 	cause: Error;
-	constructor(task: Task, cause: Error) {
+	constructor(task: Task, duration: number, cause: Error) {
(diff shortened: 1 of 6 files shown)


src/sync/installation.ts
```diff
@@ -110,16 +110,30 @@ export const updateTaskStatusAndContinue = async (
...
 
+const sendTaskMetrics = (metrics: string, task: Task, gitHubProduct: "server" | "cloud", failedCode?: string) => {
+	const duration = task.startTime ? Date.now() - task.startTime : undefined;
+	const tags = {
+		type: task.task,
+		gitHubProduct,
+		failedCode: failedCode || ""
+	};
+	if (duration === undefined) {
+		statsd.increment(metrics, tags);
```
gxueatlassian, line 131: The `histogram` doesn't support an undefined number in its signature, so for undefined value (which isn't likely to happen according to our app flow, but from typescript point of view I can't exclude it. Also don't want to use `!` to force convert it), we will only have the `count` metrics for it.

rachellerathbone (dismissed):

PR #2064, opened 2023-04-19
Adding automatic app install

mboudreau:
**What's in this PR?**
Adding ability to auto install app on Jira if given the correct properties

**Why**
It'll save us heaps of time in the long run, and UPM is _slooooooow_

**Added feature flags**
None - dev loop improvement

**Affected issues**  
None

**How has this been tested?**  
Locally

**Whats Next?**
Get everyone to setup


etc/app-install/app-install.sh
```diff
@@ -0,0 +1,34 @@
...
+echo "The app has been successfully installed."
+echo "
+*********************************************************************************************************************
+IF YOU ARE USING A FREE NGROK ACCOUNT, PLEASE DO THIS STEP FIRST!!!
+Before going to your app, please open this URL first: ${BASE_URL}.
+This will open up the ngrok page, don't worry just click on the Visit button.
+That's it, you're all ready!
+*********************************************************************************************************************
+*********************************************************************************************************************
+Now open your app in this URL: ${ATLASSIAN_URL}/plugins/servlet/ac/${APP_KEY}/get-started
```
krazziekay, line 32: /nit
Not that important, but will probably need to update this info, cause this URL doesn't exist.

krazziekay (approved):

etc/app-install/app-install.sh
mboudreau: Ah, good catch.

krazziekay (dismissed):

krazziekay (approved):

PR #2065, opened 2023-04-20
NONE: reset cookies when session is initalised

bgvozdev:
**What's in this PR?**
Unblocking SSO customers

**Why**
Cause they are in pain

### DIFF
--- a/src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
+++ b/src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
@@ -8,7 +8,8 @@ const instance = envVars.INSTANCE_NAME;
 const isProd = (instance === EnvironmentEnum.production);
 // TODO: implement named routes (https://www.npmjs.com/package/named-routes) to facilitate rerouting between files
 export const postInstallUrl = "/jira";
-export const APP_NAME = `GitHub for Jira${isProd ? "" : (instance ? (` (${instance})`) : "")}`;
+const devSuffix = `${isProd ? "" : (instance ? (` (${instance})`) : "")}`;
+export const APP_NAME = `GitHub for Jira${devSuffix}`;
 export const APP_KEY = `com.github.integration${instance ? `.${instance}` : ""}`;
 
 const adminCondition = [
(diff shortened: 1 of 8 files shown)


static/js/jira-server-url.js
```diff
@@ -41,7 +41,7 @@ const getGHEServerError = (error, url) => {
 		case "GHE_ERROR_GITHUB_CLOUD_HOST":
 			return {
 				title: "GitHub Cloud site",
-				message: `The entered URL is a GitHub Cloud site. <a href="/session/github/configuration&ghRedirect=to" target="_blank">Connect a GitHub Cloud site<a/>.`,
+				message: `The entered URL is a GitHub Cloud site.`,
```
bgvozdev, line 44: This link was always broken: the session page must have window.jwt property set

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
```diff
@@ -149,15 +150,15 @@ const modules = {
 			url: postInstallUrl,
 			conditions: adminCondition,
 			name: {
-				value: "GitHub for Jira"
+				value: `GitHub for Jira${devSuffix}`
```
bgvozdev, line 153: I'm pretty pissed about not knowing which Jira is prod and which is my own

Harminder84 (approved):

src/routes/session/session-get.ts
```diff
@@ -2,6 +2,14 @@ import { Request, Response } from "express";
 import { createUrlWithQueryString } from "utils/create-url-with-query-string";
 
 export const SessionGet = (req: Request, res: Response) => {
+	if (req.query.resetSession) {
+		const cookies = req.cookies;
+		for (const cookieName in cookies) {
+			res.clearCookie(cookieName);
```
gxueatlassian, line 8: I think you are doing right that also clear the jiraHost as well inside the session, so that multi tab on multi jira host can work perfectly.
However, I can think of one edge case that current support but this pr does not.

At the moment, if user stay in the jira-configuration page for a long time (and give github token already present in cookie), they can setup new org in this new page without problem.

But in the PR, since jiraHost is cleared, when they click the link, there will be no new jwt (from `AP.requestToken()`) passed to the backend, and backend will not found a jiraHost (before doing oauth dance).
Previously, since jiraHost is NOT cleared, backend will proceed to oauth dance.

However, I don't think the fault lies in this PR, it lies in our standalone page weirdness. We should revisit our standalone page.

gxueatlassian (approved):

PR #2067, opened 2023-04-20
Add metrics for the real graphQuery

gxueatlassian:
**What's in this PR?**
Add metrics for graphql

**Why**
Because the current metrics only tell it is a graphql query, but didn't further dive into which exactly the graph query is.

**Added feature flags**
N/A

**Affected issues**  
N/A

**How has this been tested?**  
Local

**Whats Next?**
N/A

### DIFF
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -98,13 +98,16 @@ export class GitHubClient {
 		}
 	}
 
-	protected async graphql<T>(query: string, config: AxiosRequestConfig, variables?: Record<string, string | number | undefined>): Promise<AxiosResponse<GraphQlQueryResponse<T>>> {
+	protected async graphql<T>(query: string, config: AxiosRequestConfig, variables?: Record<string, string | number | undefined>, metrics?: Record<string, string>): Promise<AxiosResponse<GraphQlQueryResponse<T>>> {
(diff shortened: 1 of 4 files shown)


src/github/client/github-client.ts
```diff
@@ -98,13 +98,16 @@ export class GitHubClient {
...
 
-	protected async graphql<T>(query: string, config: AxiosRequestConfig, variables?: Record<string, string | number | undefined>): Promise<AxiosResponse<GraphQlQueryResponse<T>>> {
+	protected async graphql<T>(query: string, config: AxiosRequestConfig, variables?: Record<string, string | number | undefined>, metrics?: Record<string, string>): Promise<AxiosResponse<GraphQlQueryResponse<T>>> {
 		const response = await this.axios.post<GraphQlQueryResponse<T>>(this.graphqlUrl,
 			{
 				query,
 				variables
 			},
-			config);
+			Object.assign({}, {
```
bgvozdev, line 107: Curious what's Object.assign for?

gxueatlassian: It is exactly like bellow. But ts wont' complain.
```
{
 ... config,
metrics:
}
```

gxueatlassian: Eventually both will endup with a new object.

bgvozdev (approved):

PR #2072, opened 2023-04-24
ARC-2112 - Fix some 405 method not allow on /devinfo/0.10/bulk

krazziekay:
**What's in this PR?**
- Added a condition in Axios to pass the 405 failed requests to `/rest/deployments/0.1/bulk`

**Why**
- In case of users who have moved from an old Jira site to a new Jira site, this POST request will be redirected and will return a 405 request. This message will NOT be deleted, and remained invisible and retried couple times. But after couple times, it will be transferred to DLQ.
- Replaying that DLQ won’t help as the redirect will happen again. 

**How has this been tested?**  
- Test case added

**Whats Next?**

### DIFF
--- a/src/jira/client/axios.ts
+++ b/src/jira/client/axios.ts
@@ -62,7 +62,7 @@ const getErrorMiddleware = (logger: Logger) =>
 		 * This GET request is now failing with a 405 response and
 		 */
 		if (error.request.method === "GET" && error.request.path === "/rest/devinfo/0.10/bulk") {
-			logger.warn({ error } , "Redirected to GET /rest/deployments/0.1/bulk");
+			logger.warn({ error } , "Ignoring the error when redirected to GET /rest/deployments/0.1/bulk");
 			return Promise.resolve({ result: "SKIP_REDIRECTED" });
 		}
 
--- a/src/jira/client/axios.ts
+++ b/src/jira/client/axios.ts
@@ -7,6 +7,10 @@ import { metricHttpRequest } from "config/metric-names";
 import { urlParamsMiddleware } from "utils/axios/url-params-middleware";
 import { jiraAuthMiddleware } from "utils/axios/jira-auth-middleware";
 
+type SkipRedirect ={
+	result: string;
+}
+
 /**
  * Wrapper for AxiosError, which includes error status
  */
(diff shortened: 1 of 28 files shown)


src/jira/client/axios.ts
```diff
@@ -154,7 +154,19 @@ const instrumentFailedRequest = (baseURL: string, logger: Logger) => {
...
+
 		if (error.response?.status === 503 || error.response?.status === 405) {
+			/**
+			 * EDGE CASE:
+			 * When sending a POST request to `https://OLD.atlassian.net/rest/deployments/0.1/bulk`, it sends a 302 response,
+			 * and redirects to a GET request `https://NEW.atlassian.net/rest/deployments/0.1/bulk`
+			 * This GET request is now failing with a 405 response and
+			 */
+			if (error.request.method === "GET" && error.request.path === "/rest/devinfo/0.10/bulk") {
+				logger.info({ error } , "Redirected to GET /rest/deployments/0.1/bulk");
```
gxueatlassian, line 166: /nit
Maybe a more descriptive error message, and also change `info` to `warn`?

src/jira/client/axios.ts
```diff
@@ -154,7 +154,19 @@ const instrumentFailedRequest = (baseURL: string, logger: Logger) => {
...
+				return Promise.resolve({ message: "Redirected to GET /rest/deployments/0.1/bulk" });
```
gxueatlassian, line 167: I am not sure what's the use of the field `message`. Either remove it or make it a more programatically readable field, like `result: 'SKIP_REDIRECTED'`.

src/jira/client/axios.ts
```diff
@@ -154,7 +154,19 @@ const instrumentFailedRequest = (baseURL: string, logger: Logger) => {
 			logger.info("Ok, looks like still works.");
 		}
 		instrumentRequest(error?.response);
+
 		if (error.response?.status === 503 || error.response?.status === 405) {
+			/**
```
gxueatlassian, line 159: And, if possible, it should be done in the error handling part, not the `instrucmentXXX` part, coz by name instructment If posshould just send metrics instead of update the error (although existing code is also wrong)

```
	instance.interceptors.response.use(
		instrumentRequest,
		instrumentFailedRequest(baseURL, logger)
	);

	instance.interceptors.response.use(
		getSuccessMiddleware(logger),
		getErrorMiddleware(logger)
	);

```

src/jira/client/axios.ts
krazziekay: Logged info updated to warn

src/jira/client/axios.ts
krazziekay: Updated the response.

src/jira/client/axios.ts
krazziekay: Moved the logic inside `getErrorMiddleware`.

src/jira/client/axios.ts
```diff
@@ -50,7 +54,18 @@ const getErrorMiddleware = (logger: Logger) =>
...
-	(error: AxiosError): Promise<Error> => {
+	(error: AxiosError): Promise<Error|SkipRedirect> => {
+		/**
+		 * EDGE CASE:
+		 * When sending a POST request to `https://OLD.atlassian.net/rest/deployments/0.1/bulk`, it sends a 302 response,
+		 * and redirects to a GET request `https://NEW.atlassian.net/rest/deployments/0.1/bulk`
+		 * This GET request is now failing with a 405 response, which is not an error on the app side
+		 * So treating this as a non-error case
+		 */
+		if (error.request.method === "GET" && error.request.path === "/rest/devinfo/0.10/bulk") {
```
gxueatlassian, line 65: Nice, 👍 , and depends on the logs/result, we might want to expand the request.path to include other jira method paths as well if those have 405 as well. But for now, lets just start with that /devinfo/bulk first.

gxueatlassian (approved):

gxueatlassian (approved):

PR #2082, opened 2023-04-28
ARC-2112 add logs for 405

gxueatlassian:
**What's in this PR?**
Add logs to see
1. Why I couldn't find status calls failured at 302 on 405 request
2. Help decide future should we just wrap 405 as 200 or make it to 404 and delete message?

**Why**

**Added feature flags**

**Affected issues**  
ARC-2112

**How has this been tested?**  
Just logs

**Whats Next?**
Fix for all jira api endpoints on 405 (and potential other status)

### DIFF
--- a/src/jira/client/axios.ts
+++ b/src/jira/client/axios.ts
@@ -162,31 +162,33 @@ const instrumentRequest = (response) => {
  */
 const instrumentFailedRequest = (baseURL: string, logger: Logger) => {
 	return async (error: AxiosError) => {
+		logger = logger.child({ jiraHost: baseURL, errorStatus: error.response?.status });
 		if (error.response) {
 			// TODO: pretty sure this whole function is some dead code that we can delete, because in getErrorMiddleware
 			// TODO: the error is mapped to JiraClientError. If that's the case, the log line below should never be logged
 			// TODO: and we are safe to kill the thing.
-			logger.info({ error }, "Ok, looks like still works.");
+			logger.info("Ok, looks like still works.");
 		}
 		instrumentRequest(error?.response);
 
 		if (error.response?.status === 503 || error.response?.status === 405) {
 			try {
-				logger.info({ jiraHost: baseURL }, "Try fetching status for 503 and 405 failure request");
+				logger.info("Try fetching status for 503 and 405 failure request");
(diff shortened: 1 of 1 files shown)


src/jira/client/axios.ts
```diff
@@ -166,13 +166,15 @@ const instrumentFailedRequest = (baseURL: string, logger: Logger) => {
 			// TODO: pretty sure this whole function is some dead code that we can delete, because in getErrorMiddleware
 			// TODO: the error is mapped to JiraClientError. If that's the case, the log line below should never be logged
 			// TODO: and we are safe to kill the thing.
-			logger.info("Ok, looks like still works.");
+			logger.info({ error }, "Ok, looks like still works.");
```
bgvozdev, line 169: should be `{ err }`

src/jira/client/axios.ts
```diff
@@ -183,6 +185,8 @@ const instrumentFailedRequest = (baseURL: string, logger: Logger) => {
 				}
 			}
 		}
+
+		logger.info({ jiraHost: baseURL, error }, "Rejecting final error");
```
bgvozdev, line 189: `{ , err: error }`

bgvozdev (changes requested):
we shouldn't log UGC/PII

src/jira/client/axios.ts
gxueatlassian: Good catch! 
Fixed now.

src/jira/client/axios.ts
```diff
@@ -172,17 +173,22 @@ const instrumentFailedRequest = (baseURL: string, logger: Logger) => {
...
 					error.response.status = 404;
 				} else if (e.response.status === 302) {
-					logger.info({ jiraHost: baseURL },"405 from Jira: Jira instance has been renamed. Returning 404 to our application.");
+					logger.info("405 from Jira: Jira instance has been renamed. Returning 404 to our application.");
 					error.response.status = 404;
 				}
 			}
 		}
+
+		logger.info({ errorStatus: error.response?.status }, "Rejecting final error");
```
bgvozdev, line 191: logging just `{ err }` would've worked fine btw :) in logger config we are removing all possible UGCs

bgvozdev (approved):

PR #2083, opened 2023-04-28
ARC-1424 and ARC-1425 update api resync post

rachellerathbone:
**What's in this PR?**
Added additional checks to send 400 if no installations are found for passed IDs or wrong data type is passed (the latter now fails on an error but still good to be more explicit).

**Why**
To address 2 tickets that has been open in the disturbed/ops backlog for almost a year (2nd one didn't require much additional effort so it piggy backed off the first).

**Affected issues**  
ARC-1424 and ARC-1425

**How has this been tested?**  
Wrote tests for these additions (plus wrote tests for the existing code because there was none!) and tested in stg

**Whats Next?**
Grab another ticket from the backlog

### DIFF
--- a/src/routes/api/api-resync-post.ts
+++ b/src/routes/api/api-resync-post.ts
@@ -4,21 +4,20 @@ import { GitHubServerApp } from "models/github-server-app";
 import { Subscription } from "models/subscription";
 import { findOrStartSync } from "~/src/sync/sync-utils";
 import { serializeSubscription } from "routes/api/api-utils";
-import { Installation } from "models/installation";
 
-const getIdsForExistingInstallations = async(installationIds, logger): Promise<number[]> => {
-	const installations = installationIds.map(async(id: number): Promise<number | null> => {
-		const installation: Installation | null = await Installation.getForId(id);
+const getIdsForExistingSubscriptions = async(gitHubInstallationIds, logger): Promise<number[]> => {
+	const subscriptions = gitHubInstallationIds.map(async(id: number): Promise<number | null> => {
+		const subscription: Subscription | null = await Subscription.findOneForGitHubInstallationId(id, undefined);
 
-		if (installation?.id) {
-			return installation?.id;
+		if (subscription) {
+			return subscription.gitHubInstallationId;
 		} else {
-			logger.warn({ installationId: id }, "No installation found");
+			logger.warn({ gitHubInstallationIds: id }, "No subscription found");
 			return null;
 		}
 	});
-	const existingInstallations = await Promise.all(installations);
-	return existingInstallations.filter(installation => installation !== null);
+	const existingSubscriptions = await Promise.all(subscriptions);
+	return existingSubscriptions.filter(subscription => subscription !== null);
 };
 
 export const ApiResyncPost = async (req: Request, res: Response): Promise<void> => {
@@ -27,7 +26,7 @@ export const ApiResyncPost = async (req: Request, res: Response): Promise<void> 
 	// Defaults to anything not completed
 	const statusTypes = req.body.statusTypes as string[];
 	// Defaults to any installation
-	const installationIds = req.body.installationIds as number[];
+	const gitHubInstallationIds = req.body.installationIds as number[];
 	// Can be limited to a certain amount if needed to not overload system
 	const limit = Number(req.body.limit) || undefined;
 	// Needed for 'pagination'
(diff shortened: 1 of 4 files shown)


src/routes/api/api-resync-post.ts
```diff
@@ -24,7 +40,14 @@ export const ApiResyncPost = async (req: Request, res: Response): Promise<void>
 	const targetTasks = req.body.targetTasks as TaskType[];
 
 	if (!statusTypes && !installationIds && !limit && !inactiveForSeconds) {
-		res.status(400).send("please provide at least one of the filter parameters!");
+		res.status(400).send("Please provide at least one of the filter parameters!");
```
rachellerathbone, line 43: Opted to return a 400 instead of a 404 as the Jira ticket recommended - figured this falls more into the category of a bad request than the requested web page not being available.

src/routes/api/api-resync-post.ts
```diff
@@ -24,7 +40,14 @@ export const ApiResyncPost = async (req: Request, res: Response): Promise<void>
...
+		return;
+	}
+
+	const existingInstallationIds = await getIdsForExistingInstallations(installationIds, req.log);
+
+	if (!existingInstallationIds.length) {
+		res.status(400).json("No installations exist for provided IDs");
```
rachellerathbone, line 50: Opted to return a 400 instead of a 404 as the Jira ticket recommended - figured this falls more into the category of a bad request than the requested web page not being available.

src/routes/api/api-resync-post.ts
```diff
@@ -4,6 +4,22 @@ import { GitHubServerApp } from "models/github-server-app";
 import { Subscription } from "models/subscription";
 import { findOrStartSync } from "~/src/sync/sync-utils";
 import { serializeSubscription } from "routes/api/api-utils";
+import { Installation } from "models/installation";
+
+const getIdsForExistingInstallations = async(installationIds, logger): Promise<number[]> => {
+	const installations = installationIds.map(async(id: number): Promise<number | null> => {
+		const installation: Installation | null = await Installation.getForId(id);
```
gxueatlassian, line 11: I think the `installationIds` means github installation id, instead of the `id` (pk) in the installation table?

gxueatlassian, line 11: Also, if find by the `id` (pk), I think we can do `Installation.findByPk` from the buildin method.

rachellerathbone: Ohhhh right. Good catch. Bloody confusing naming.

src/routes/api/api-resync-post.ts
```diff
@@ -5,13 +5,28 @@ import { Subscription } from "models/subscription";
...
+};
+
 export const ApiResyncPost = async (req: Request, res: Response): Promise<void> => {
 	// Partial by default, can be made full
 	const syncType = req.body.syncType || "partial";
 	// Defaults to anything not completed
 	const statusTypes = req.body.statusTypes as string[];
 	// Defaults to any installation
-	const installationIds = req.body.installationIds as number[];
+	const gitHubInstallationIds = req.body.installationIds as number[];
```
rachellerathbone, line 29: Renamed for clarity.

src/routes/api/api-resync-post.ts
```diff
@@ -23,8 +38,20 @@ export const ApiResyncPost = async (req: Request, res: Response): Promise<void>
...
 	const targetTasks = req.body.targetTasks as TaskType[];
 
-	if (!statusTypes && !installationIds && !limit && !inactiveForSeconds) {
-		res.status(400).send("please provide at least one of the filter parameters!");
+	if (!statusTypes && !gitHubInstallationIds && !limit && !inactiveForSeconds) {
+		res.status(400).send("Please provide at least one of the filter parameters!");
+		return;
+	}
+
+	if (!gitHubInstallationIds.length) {
```
gxueatlassian, line 46: /nit
for `!gitHubInstallationIds.length`, say [1,2,3] are provided as `gitHubInstallationIds`, but only return [2,3], do we want to make it 400 or still allowing it?

gxueatlassian (approved):

src/routes/api/api-resync-post.ts
rachellerathbone: I'd thought about that - it's primarily used by us to resync folks and I figured it would be a pretty bad experience if we had a CSV with a large number of gh installationIds but accidentally entered a couple incorrectly. It would mean it would fail on one, we'd update it and run again, then it would fail on the second one etc. 

We could potentially try and return all incorrect ids so they could be removed in one go. Having said that though, while we don't need to worry about ppl passing in bogus data and getting a successful response (it would fail when we call Subscription.getAllFiltered), we should probably return a 400 if no subscriptions are found.

PR #2087, opened 2023-05-01
ARC-1090 delete builds and deployments for new connections

rachellerathbone:
**What's in this PR?**
Updated the Jira client's devinfo repository delete.

**Why**
Because we weren't deleting builds and deployments data when a repo was deleted from GH.

**Affected issues**  
ARC-1090

**How has this been tested?**  
Locally and in staging (see below) and updated unit tests.

### Testing scenarios
I tested the following both locally and in staging (screenshots are from staging)

-------
**Scenario:** Repo already connected and the repo linked to this issue was **not** deleted. Confirming that deleting another repo in the same org doesn't impact data from other repos.

**Expected result:** There should be no change

Before:
existing-repo-dont-delete-before

After: ✅ 
existing-repo-dont-delete-after

------
**Scenario:** Repo already connected and the repo linked to this issue was deleted but data was added to the ticket before change went in.

**Expected result:** We will see the current behaviour where commit, branch, and pull request data is deleted but builds and deployments are not. **NOTE:** This is because we will start adding the `repoId` prop as part of this change meaning any existing devinfo data won't have this value. I chatted with Sass about this and we agreed that the best option in regards to this would be to leave as is and manually delete this data for any customers that reach out to us with concerns to deleted repo data still being present in Jira.

Before:
existing-delete-data-added-before-change-before

After: ✅ 
existing-repo-delete-data-added-before-change-after

------
**Scenario:**  Repo already connected and the repo linked to this issue was deleted- data was added to the ticket after change went in.

**Expected result:** All data should be deleted from dev info

Before:
existing-delete-data-added-after-change-before

After: ✅ 
existing-repo-delete-data-added-after-change-after

------
**Scenario:** Repo connected after change went into.

**Expected result:** All data should be deleted from dev info

Before:
new-repo-before

After: ✅ 
new-repo-after

------


test/fixtures/workflow-basic.json
```diff
@@ -73,7 +73,7 @@
 			"html_url": "test-repo-url"
 		},
 		"repository": {
-			"id": "test-repo-id",
+			"id": 123,
```
rachellerathbone, line 76: Updated this as the repository id type is a number.

src/github/deployment.ts
```diff
@@ -67,7 +67,6 @@ export const processDeployment = async (
 		gitHubAppId,
 		logger
 	);
-
```
rachellerathbone, line 70: Soooo not sure what it's saying the diff is here but... ok GitHub.

src/github/repository.ts
```diff
@@ -54,13 +52,12 @@ export const deleteRepositoryWebhookHandler = async (context: WebhookContext, ji
 	context.log.info(`Deleting dev info for repo ${context.payload.repository?.id}`);
 
 	try {
-		const jiraResponse = await jiraClient.devinfo.repository.delete(
-			transformRepositoryId(context.payload.repository?.id, context.gitHubAppConfig?.gitHubBaseUrl)
-		);
+		const jiraResponse =  await jiraClient.devinfo.installation.delete(gitHubInstallationId);
```
rachellerathbone, line 55: Man, implemented this so differently to start with but then simplified... and then simplified again when I discovered we already have installation delete and were just calling the wrong endpoint 🤦

src/sync/build.test.ts
```diff
@@ -26,7 +26,7 @@ describe("sync/builds", () => {
 	const makeExpectedJiraResponse = (builds) => ({
 		builds,
 		properties: {
-			"gitHubInstallationId": DatabaseStateCreator.GITHUB_INSTALLATION_ID
+			gitHubInstallationId: DatabaseStateCreator.GITHUB_INSTALLATION_ID
```
rachellerathbone, line 29: Could probably revert this and the following change as these aren't explicitly tied to the PR. Updated these as I initially made changes to these files in my original implementation. That said, they're small enough, and there aren't many changes in this PR, so I think we can leave them.

src/github/deployment.ts
```diff
@@ -68,7 +68,13 @@ export const processDeployment = async (
 		logger
 	);
 
-	const result: DeploymentsResult = await jiraClient.deployment.submit(jiraPayload);
+	const result: DeploymentsResult = await jiraClient.deployment.submit(jiraPayload, webhookPayload.repository.id);
+
+	if (result.rejectedDeployments?.length) {
```
gxueatlassian, line 73: In previous PR I moved this log into the jira-client. Do you mind remove this one here?

src/jira/client/jira-client.ts
```diff
@@ -262,10 +297,12 @@ export const getJiraClient = async (
 					const subscription = await Subscription.getSingleInstallation(jiraHost, gitHubInstallationId, gitHubAppId);
 					await subscription?.update({ syncWarning: issueKeyLimitWarning });
 				}
+
 				const payload = {
 					builds: data.builds,
 					properties: {
-						gitHubInstallationId
+						gitHubInstallationId,
+						repositoryId
```
gxueatlassian, line 305: Why not use tranformed repository id? I think that `transformed repo id` is necessary for GHES app?

src/github/repository.ts
```diff
@@ -51,18 +49,25 @@ export const deleteRepositoryWebhookHandler = async (context: WebhookContext, ji
...
+	}
+
 	context.log.info(`Deleting dev info for repo ${context.payload.repository?.id}`);
 
+	const { id: repositoryId } = context.payload.repository;
+
 	try {
 		const jiraResponse = await jiraClient.devinfo.repository.delete(
-			transformRepositoryId(context.payload.repository?.id, context.gitHubAppConfig?.gitHubBaseUrl)
+			repositoryId,  context.gitHubAppConfig?.gitHubBaseUrl
```
gxueatlassian, line 65: I LIKE passing in the origin repoId and let the method itself do the transforming.

src/jira/client/jira-client.ts
gxueatlassian: Okay since we use `gitHubInstallationId` as well, so GHES might be okay here, I see what you mean. 
I think there's another place do the deletion, have a look see whether we need fix that as well.
```
			installation: {
				delete: async (gitHubInstallationId: string | number) =>
```

src/github/deployment.ts
rachellerathbone: Looks like it's still in the JC but the merge kept this one around for some reason. Have deleted.

src/jira/client/jira-client.ts
rachellerathbone: But the installation endpoint is to delete the entire org connection (handled by checking the passed gitHubInstallationId), this is just for a single repo that gets deleted in GH.

gxueatlassian: Oh i mean say this is submitting with `gitHubInstallationId` and `repositoryId` (after the PR), while that installation endpoint delete only via `gitHubInstallationId`. I am not sure is an AND or OR logic, just curious whether the `installation delete endpoint` still work.

rachellerathbone: Only the the stored property you want to delete by needs to be supplied when deleting (do not need to provider every single property an entity has when deleting. 

Have tested locally and all looks good. More info here https://developer.atlassian.com/cloud/jira/software/rest/api-group-deployments/#api-rest-deployments-0-1-bulk-post

gxueatlassian: ha, nice 👍

src/jira/client/jira-client.ts
```diff
@@ -210,15 +219,42 @@ export const getJiraClient = async (
...
-				delete: (transformedRepositoryId: TransformedRepositoryId) =>
-					instance.delete("/rest/devinfo/0.10/repository/{transformedRepositoryId}", {
-						params: {
-							_updateSequenceId: Date.now()
-						},
-						urlParams: {
-							transformedRepositoryId
-						}
-					}),
+				delete: async (repositoryId: number, gitHubBaseUrl?: string) => {
```
gxueatlassian, line 222: Sorry I know you've tested, but just curious the typing here for repoId is number, while the submit of workflow/deployment are string. Do you know does it matter? Coz that will be used in the `properties` to identify the correct piece of data.

gxueatlassian (approved):

PR #2088, opened 2023-05-01
ARC-1977 - Handling previous webhook secrets

krazziekay:
**What's in this PR?**
- Condition added to check for old webhook secrets 
- Updated webhook secrets added from Vault

**Why**
- The current webhook secrets will be updated by new ones, so we need our code to work with both the new as well as the old webhook secret.

**How has this been tested?**  
- Added Test case
- Tested in staging


github-for-jira.sd.yml
```diff
@@ -213,6 +213,8 @@ config:
     DEPLOYMENT_DATE: ${DEPLOYMENT_DATE}
     PRIVATE_KEY: vault://secret/data/builds/micros-sv--github-for-jira-dl-admins/github-app-private-key-stg
     GITHUB_CLIENT_SECRET: vault://secret/data/builds/micros-sv--github-for-jira-dl-admins/github-app-client-secret-stg
+    WEBHOOK_SECRET: vault://secret/data/builds/micros-sv--github-for-jira-dl-admins/github-app-webhook-secret-stg
```
krazziekay, line 216: TODO: Need to create a new secret and add it in vault.

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -31,10 +31,21 @@ export const WebhookReceiverPost = async (request: Request, response: Response):
...
+		const { webhookSecrets, gitHubServerApp } = await getWebhookSecrets(uuid);
+		const isVerified = webhookSecrets.some((secret, index) => {
+			const matchesSignature = createHash(request.rawBody, secret) === signatureSHA256;
+			/**
+			 * The latest updated webhook secret will be at index 0,
+			 * Once we stop receiving logs with index other than 0,
+			 * can then completely remove the old webhook secrets.
+			 */
+			if (matchesSignature) {
+				logger.info("Matched with the webhook of index: ", index);
```
Harminder84, line 43: If it matches with secret's index other than 0, add logging in that case only. Otherwise it will clutter the logs.

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -138,10 +149,23 @@ const getWebhookSecret = async (uuid?: string): Promise<{ webhookSecret: string,
 			throw new Error(`Installation not found for gitHubApp with uuid ${uuid}`);
 		}
 		const webhookSecret = await gitHubServerApp.getDecryptedWebhookSecret(installation.jiraHost);
-		return { webhookSecret, gitHubServerApp };
+		/**
+		 * If we ever need to rotate the webhook secrets for Enterprise Customers,
+		 * we can add it in the array: ` [ webhookSecret ]`
+		 */
+		return { webhookSecrets: [ webhookSecret ], gitHubServerApp };
```
Harminder84, line 156: How to add multiple secrets for enterprise customers? Right now `webhookSecret` DB column is of type varchar.

src/routes/github/webhook/webhook-receiver-post.ts
krazziekay: I think its ok to log all indexes for the webook secrets cause it gives us an idea of which webhook secret are in use, otherwise there is no way to figure it out. And once we figure out which ones are ont in use, we can then proceed to remove it forever.

src/routes/github/webhook/webhook-receiver-post.ts
krazziekay: We do not have to worry about the enterprise customers for now cause they have their own webhook secret. Using this as an array here to match the types. 
If we ever need to update webhook secret for any enterprise customer, we can store the webhook secret as an array similar to how we've stored in vault. So have something like "['newSecret','oldSecret']", which can still be saved in the DB. Might need to change this to parse it, but that can be done later.

Harminder84: Make sense. I think this whole thing needs to be put behind FF

krazziekay: When we change the webhook secret in Github's app settings, our previous code won't work in that case. So didn't add any FF.

.github/workflows/on-push.yml
```diff
@@ -41,7 +41,7 @@ jobs:
           echo "NODE_ENV=test" >> .env
           echo "APP_NAME=jira" >> .env
           echo "APP_ID=${{ secrets.E2E_GITHUB_APP_ID }}" >> .env
-          echo "WEBHOOK_SECRET=${{ secrets.E2E_GITHUB_WEBHOOK_SECRET }}" >> .env
+          echo "WEBHOOK_SECRETS=${{ secrets.E2E_GITHUB_WEBHOOK_SECRET }}" >> .env
```
mboudreau, line 44: I would probably just add a new secret here called webhook_secrets.  Did you actually change the value of of the secret?  Because if you did, you just broke all e2e tests for every other runs that isn't on this branch.

krazziekay: Yep, will create a new one and update this. 
However, this one is a bit of oddball, I haven't changed anything here. Yet it passes the test cases.

src/routes/github/webhook/webhook-receiver-post.ts
rachellerathbone: Can we update this to follow the logging pattern we use: object + message e.g. `logger.info({ index }, "Matched webhook index");`

src/config/env.ts
```diff
@@ -83,7 +83,7 @@ export interface EnvVars {
 	APP_ID: string;
 	APP_URL: string;
 	APP_KEY: string;
-	WEBHOOK_SECRET: string;
+	WEBHOOK_SECRETS: string;
```
gxueatlassian, line 86: /nit

I think you can put `JSON.parse(envVars.WEBHOOK_SECRETS);` inside the `const transforms: Transforms<EnvVars> = {` above so that the consumer doesn't need to `JSON.parse` everytime it is called. Especially webhook will be called a lots of time.

src/config/env.ts
```diff
@@ -47,7 +47,7 @@ envCheck(
 	"APP_ID",
 	"APP_URL",
 	"APP_KEY",
-	"WEBHOOK_SECRET",
+	"WEBHOOK_SECRETS",
```
gxueatlassian, line 50: **IMPORTANT**
One thing, since you change the env name, you have to put the current value of the `WEBHOOK_SECRET` into `WEBHOOK_SECRETS` of the vault (or stash).

But the thing is, how do you 100% SURE that the value in lastpass is the value that being used in prod? I can't think of a way to verify it. And if that value in lastpass is wrong, we may lost some webhooks, and need to rush to generate a new secrete and put it in vault, very rushy.

gxueatlassian: Alternative, you can keep the `WEBHOOK_SECRET`, but have a new env like `WEBHOOK_SECRET_VAULT`, and in the code try both.

Another benefits is you don't have to do `JSON.parse`.

krazziekay: Yep, `WEBHOOK_SECRETS` is a string which has both the old and new webhook secrets.
I agree, there is no direct way to check the current webhook secret. However, I checked the history in Lastpass and it has not been changed since `2022-06-16 01:12:35`, so pretty sure its still the same. And I don't think we have any lost webhooks yet, don't have any complaints yet.
And as for the alternative, its more scalable with one variable which holds multiple secrets instead of having 2 separate variables. I have discussed with @mboudreau  and agreed not to use two variables.

.github/workflows/on-push.yml
```diff
@@ -41,7 +41,7 @@ jobs:
           echo "NODE_ENV=test" >> .env
           echo "APP_NAME=jira" >> .env
           echo "APP_ID=${{ secrets.E2E_GITHUB_APP_ID }}" >> .env
-          echo "WEBHOOK_SECRET=${{ secrets.E2E_GITHUB_WEBHOOK_SECRET }}" >> .env
+          echo "WEBHOOK_SECRETS=${{ secrets.WEBHOOK_SECRETS }}" >> .env
```
mboudreau, line 44: With the prefix please.  This webhook is only specific for e2e - please follow the convention.

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -138,10 +153,31 @@ const getWebhookSecret = async (uuid?: string): Promise<{ webhookSecret: string,
...
 	}
-	if (!envVars.WEBHOOK_SECRET) {
-		throw new Error("Environment variable 'WEBHOOK_SECRET' not defined");
+	if (allowGhCloudWebhookSecrets) {
+		if (!envVars.WEBHOOK_SECRETS) {
```
gxueatlassian, line 163: Dont' need this, as we can put that in the env.ts check, fail-fast, make the app crash during startup.

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -138,10 +153,31 @@ const getWebhookSecret = async (uuid?: string): Promise<{ webhookSecret: string,
...
+		}
+		try {
+			/**
+			 * The environment WEBHOOK_SECRETS should be a JSON array string in the format: ["key1", "key1"]
+			 * Basically an array of the new as well as any old webhook secrets
+			 */
+			const parsedWebhookSecrets = JSON.parse(envVars.WEBHOOK_SECRETS);
+			return { webhookSecrets: parsedWebhookSecrets };
+		} catch (e) {
+			throw new Error("Couldn't parse 'WEBHOOK_SECRETS', it is not properly defined!");
```
gxueatlassian, line 174: I think it is better to parse that in the env.ts, fail-fast, make app crash if that JSON is invalid.
Otherwise, if it fail here, we lose webhook msg.

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -138,10 +153,31 @@ const getWebhookSecret = async (uuid?: string): Promise<{ webhookSecret: string,
...
+		}
+	} else {
+		// TODO: Remove after testing
+		if (!envVars.WEBHOOK_SECRET) {
+			throw new Error("Environment variable 'WEBHOOK_SECRET' not defined");
```
gxueatlassian, line 179: Again, leave it inside env.ts, make app crash during startup if not found.

src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -127,7 +139,10 @@ export const createHash = (data: BinaryLike | undefined, secret: string): string
 		.digest("hex")}`;
 };
 
-const getWebhookSecret = async (uuid?: string): Promise<{ webhookSecret: string, gitHubServerApp?: GitHubServerApp }> => {
+const getWebhookSecrets = async (uuid?: string): Promise<{ webhookSecrets: Array<string>, gitHubServerApp?: GitHubServerApp }> => {
+	// TODO: Remove after testing
+	const allowGhCloudWebhookSecrets = await booleanFlag(BooleanFlags.ALLOW_GH_CLOUD_WEBHOOKS_SECRETS);
```
gxueatlassian, line 144: I think you missed the jiraHost here, otherwise you can't enable the flag just for your own instance.

---
Hmm, however, we might not be able to get the jiraHost ( or we shouldn't )? As it is pretty earlier in the stage, and even before signature verified, app doesn't pull DB to grab subscription ( and therefore the jiraHost).

krazziekay: Thats a good point, we don't have jiraHost at this point.

src/routes/github/webhook/webhook-receiver-post.ts
krazziekay: removed

src/routes/github/webhook/webhook-receiver-post.ts
krazziekay: Updated `env.ts` and removed from here

src/routes/github/webhook/webhook-receiver-post.ts
krazziekay: Removed from ehre

.github/workflows/on-push.yml
krazziekay: Updated

src/config/env.ts
krazziekay: updated

src/routes/github/webhook/webhook-receiver-post.ts
krazziekay: Added a FF, using the Github User Login as the key.

gxueatlassian (approved):

PR #2096, opened 2023-05-02
NONE - update missing jiraHost warnings for consistency

rachellerathbone:
**What's in this PR?**
Updated missing jiraHost warnings to use the Errors enum

**Why**
To clean up the errors/warnings that were doing their own thing / for consistency.

**Whats Next?**
Probs do the same with the other Errors messages.


gxueatlassian (approved):
Nice!!

PR #2099, opened 2023-05-03
Arc 2144 check why jira rejected issues

gxueatlassian:
**What's in this PR?**
1. Add debugging info for investigate why so many rejected
2. Fix some minor name must be required.

**Why**

image

Lots of requests rejected. Need more investigation.

**Added feature flags**
N/A

**Affected issues**  
ARC-2144

**How has this been tested?**  
N/A

**Whats Next?**


src/github/deployment.ts
```diff
@@ -70,8 +70,23 @@ export const processDeployment = async (
 
 	const result: DeploymentsResult = await jiraClient.deployment.submit(jiraPayload);
 	if (result.rejectedDeployments?.length) {
+
+		let extraDebugInfo = {};
+		try {
+			const deploymentData = jiraPayload.deployments.length > 0 ? jiraPayload.deployments[0] : undefined;
```
rachellerathbone, line 76: Is this temporary while debugging or permanent? If it's the latter, I say we pull this logic into a function, export it, and add a unit test.

gxueatlassian: Good point, let me pull that into a separate function. Tbh it is for debugging only but I suspect it could be left there for long time.

src/github/deployment.ts
```diff
@@ -69,11 +69,6 @@ export const processDeployment = async (
 	);
 
 	const result: DeploymentsResult = await jiraClient.deployment.submit(jiraPayload);
-	if (result.rejectedDeployments?.length) {
```
gxueatlassian, line 72: Moved into the jira client. So that we can also capture the error logs for deployments in backfill.

rachellerathbone (approved):

PR #2101, opened 2023-05-04
Skip resetting fail code is target task is empty

gxueatlassian:
**What's in this PR?**
Skip resetting the failed code if target task is emtpy.

**Why**
We are about to turn on ff for create repo (and kick off partial sync) on repo_created webhooks.
However existing logic reset the failCode for all existing repos. We should skip that.

**Added feature flags**

**Affected issues**  
ARC-968

**How has this been tested?**  
Unit test.

**Whats Next?**

### DIFF
--- a/src/sync/sync-utils.ts
+++ b/src/sync/sync-utils.ts
@@ -40,7 +40,7 @@ export const findOrStartSync = async (
 	}
 
 	// reset failedCode for Partial or targetted syncs
-	await resetFailedCode(subscription, syncType, targetTasks);
+	await resetFailedCode(subscription, targetTasks);
 
 	const gitHubAppConfig = await getGitHubAppConfig(subscription, logger);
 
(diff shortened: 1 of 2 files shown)


joshkay10 (approved):

src/sync/sync-utils.ts
```diff
@@ -110,9 +110,8 @@ const resetTargetedTasks = async (subscription: Subscription, syncType?: SyncTyp
 
 };
 
-const resetFailedCode = async (subscription: Subscription, syncType?: SyncType, targetTasks?: TaskType[]) => {
-	// a full sync without target tasks has reposyncstates removed so dont update.
-	if (syncType === "full" && !targetTasks?.length) {
+const resetFailedCode = async (subscription: Subscription, targetTasks?: TaskType[]) => {
+	if (!targetTasks?.length) {
```
bgvozdev, line 114: Curious if it affects the behaviour of https://hello.atlassian.net/wiki/spaces/OTFS/pages/1824654901/HOW+TO+Run+a+backfill+resync+endpoint ?

gxueatlassian: This sync-utils is too complicated, too many branches. 
From what I can tell, this change should only affecting `sync=partial && targetTask == empty`. Current behaviour for that branch is do reset the failCode. And I think the current behaviour is wrong. The fix will make it NOT reset the failCode. So regardless of the runbooks, this fix should works correctly.

--
I can double check the runbook, but doubt any runbook relies on the "wrong" behaviour (that reset failCode for partial and empty target task).

bgvozdev: Yes, that's why I"m asking :) IMHO we should update our runbook with the change, because it is super compliated and must be documented... wdyt?

gxueatlassian: Will update the doc.

PR #2103, opened 2023-05-04
Arc 2112 catch 405 on other jira api 2

gxueatlassian:
**What's in this PR?**
1. Try/catch 405 for other similar jira api as well.
2. Merge into the previous logic to convert it into 404 instead of silently mark it as success
3. Modify the backfill to accept 404 as a success

**Why**
That try/catch 405 works well on the devinfo, would like to apply it to other jira api.

**Added feature flags**
N/A

**Affected issues**  
N/A

**How has this been tested?**  
Unit test.

**Whats Next?**

### DIFF
--- a/src/jira/client/axios.ts
+++ b/src/jira/client/axios.ts
@@ -55,17 +55,6 @@ const getErrorMiddleware = (logger: Logger) =>
 	 * @returns {Promise<Error>} The rejected promise
 	 */
 	(error: AxiosError): Promise<Error|SkipRedirect> => {
-		/**
-		 * EDGE CASE:
-		 * When sending a POST request to `https://OLD.atlassian.net/rest/deployments/0.1/bulk`, it sends a 302 response,
-		 * and redirects to a GET request `https://NEW.atlassian.net/rest/deployments/0.1/bulk`
-		 * This GET request is now failing with a 405 response, which is not an error on the app side
-		 * So treating this as a non-error case
-		 */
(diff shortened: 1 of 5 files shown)


src/jira/client/axios.ts
```diff
@@ -173,20 +162,24 @@ const instrumentFailedRequest = (baseURL: string, logger: Logger) => {
 		}
 		instrumentRequest(baseURL)(error?.response);
 
-		if (error.response?.status === 503 || error.response?.status === 405) {
+		if (error?.request?.method === "GET" && error?.response?.status === 405) {
+			/**
+			 * When sending a POST request to `https://OLD.atlassian.net/rest/deployments/0.1/bulk` or similar api on that jirahost,
+			 * it sends a 302 response, and redirects to a GET request `https://NEW.atlassian.net/rest/deployments/0.1/bulk`
+			 * This GET request is now failing with a 405 response, which is not an error on the app side
+			 * So treating this as a non-error case
```
bgvozdev, line 170: I was wondering if it makes sense to reconfigure axios not to follow redirects and to POST using the new hostname?...

gxueatlassian: Yup, thought about that, make sense, but in this case, the jira host is redirected, not much point of still sending data to old jirahost.

gxueatlassian: And if we follow the redirect and post, we will ends up sending twice the copy of data into new jira host, not sure what will that mean.

gxueatlassian:
Will raise another one.

PR #2118, opened 2023-05-12
noissue: add a Feature Flag

atrigueiro:
**What's in this PR?**
Just adding a FF

**Why**
This could have big performance repercussions if DD is not ready for it (also because Terry asked for it) 

**Added feature flags**
`bump-issue-key-limit`

### DIFF
--- a/src/jira/client/jira-client.ts
+++ b/src/jira/client/jira-client.ts
@@ -23,7 +23,7 @@ import { TransformedRepositoryId, transformRepositoryId } from "~/src/transforms
 import { getDeploymentDebugInfo } from "./jira-client-deployment-helper";
 
 // Max number of issue keys we can pass to the Jira API
-export const ISSUE_KEY_API_LIMIT = 100;
+export const ISSUE_KEY_API_LIMIT = 500;
 const issueKeyLimitWarning = "Exceeded issue key reference limit. Some issues may not be linked.";
 
 export interface DeploymentsResult {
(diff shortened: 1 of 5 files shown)


krazziekay (dismissed):

TerryAg (approved):

src/jira/client/jira-client.ts
```diff
@@ -21,9 +21,18 @@ import { uniq } from "lodash";
...
 
 // Max number of issue keys we can pass to the Jira API
-export const ISSUE_KEY_API_LIMIT = 500;
+const getIssueKeyLimit = async () => {
+	let ISSUE_KEY_API_LIMIT = 100;
+	if (await booleanFlag(BooleanFlags.ISSUE_KEY_API_LIMIT_IS_500)) {
+		ISSUE_KEY_API_LIMIT = 500;
+	}
+
+	return Promise.resolve(ISSUE_KEY_API_LIMIT);
```
gxueatlassian, line 33: /nit
Just fyi `Promise.resolve` is not needed as `async` will auto wrap it with a `Promise`.

src/jira/client/jira-client.ts
```diff
@@ -545,6 +556,6 @@ const updateIssueKeyAssociationValuesFor = (resources: JiraRemoteLink[], mutatin
 };
 
 /**
- * Truncates to 100 elements in an array
+ * Truncates to 500 elements in an array
  */
-const truncate = (array) => array.slice(0, ISSUE_KEY_API_LIMIT);
+const truncate = (array) => array.slice(0, getIssueKeyLimit());
```
gxueatlassian, line 561: **IMPORTANT**
This probably don't work, as the `getIssueKeyLimit` return a `Promise`, `array.slice` is expecting a number though.

Please also have a look at the test why this bug is not catched?

atrigueiro: @gxueatlassian I think it's already being caught because the tests are failing, but I'm rejecting this PR now, it looks like DD is handling the load well.

PR #2120, opened 2023-05-15
Arc 2144 deployment  add dynamo

gxueatlassian:
**What's in this PR?**
1. Add dynamo db resource
2. Save to dynamo on success deployment
3. Read from dynamo db first to find last success deployment

**Why**
To speed up deployment processing and potentially enable parallel process

Other than some performance boost, I will expect all api calls in the red block (sum every 6h at peak hours) will be gone. 
ps: the screenshot only capture webhook, not backfill. (Refactoring backfill need more steps, see DACI)
image

**Added feature flags**
use-dynamodb-for-deployment-webhook

**Affected issues**  
ARC-2144

**How has this been tested?**  
Unit test, ddev

**Whats Next?**
Enable parallel processing.


src/config/env.test.ts
```diff
@@ -22,6 +22,7 @@ describe("environment variables", () => {
 	it("Should not be able to change variables in envVars", () => {
 		process.env.NODE_ENV = EnvironmentEnum.production;
 		expect(envVars.NODE_ENV).toBe(EnvironmentEnum.test);
+		process.env.NODE_ENV = EnvironmentEnum.test;
```
gxueatlassian, line 25: A fix the allow one another test to pass (`purgeDBItems for dynamo`)

github-for-jira.sd.yml
```diff
@@ -149,6 +149,16 @@ resources:
         # Name of the RDS resource from above
         resource: rds
 
+  - name: deployment-history
+    type: dynamo-db
+    attributes: &table-attributes
+      HashKeyName: Id
```
bgvozdev, line 155: nit: could it be something more descriptive?

src/github/deployment.test.ts
```diff
@@ -4,8 +4,16 @@ import { getLogger } from "config/logger";
...
 import { sqsQueues } from "../sqs/queues";
 import { GITHUB_CLOUD_API_BASEURL, GITHUB_CLOUD_BASEURL } from "~/src/github/client/github-client-constants";
+import { saveDeploymentInfo } from "models/deployment-service";
+import deployment_status from "fixtures/deployment_status-basic.json";
+import { booleanFlag, BooleanFlags } from "config/feature-flags";
+import { when } from "jest-when";
+import { WebhookPayloadDeploymentStatus } from "@octokit/webhooks";
 
 jest.mock("../sqs/queues");
+jest.mock("models/deployment-service");
```
bgvozdev, line 14: Do we really need to mock it? Could we just check dynamodb?

src/github/deployment.ts
```diff
@@ -4,12 +4,30 @@ import { getJiraClient, DeploymentsResult } from "../jira/client/jira-client";
...
 import { WebhookContext } from "routes/github/webhook/webhook-context";
+import { saveDeploymentInfo } from "models/deployment-service";
+import { statsd } from "config/statsd";
+import { metricDeploymentPersistent } from "config/metric-names";
+import { getCloudOrServerFromGitHubAppId } from "utils/get-cloud-or-server";
 
 export const deploymentWebhookHandler = async (context: WebhookContext, jiraClient, _util, gitHubInstallationId: number): Promise<void> => {
+
+	if (context.payload.deployment_status.state === "success") {
+		if (await booleanFlag(BooleanFlags.USE_DYNAMODB_FOR_DEPLOYMENT_WEBHOOK, jiraHost)) {
```
bgvozdev, line 19: GitHub says jiraHost is undefined, and looks like it is saying that correctly

src/github/deployment.ts
```diff
@@ -4,12 +4,30 @@ import { getJiraClient, DeploymentsResult } from "../jira/client/jira-client";
...
+			await persistentSuccessDeploymentStatusToDynamoDB(
+				jiraHost,
```
bgvozdev, line 21: undefined - some fake value from tests?

src/github/deployment.ts
```diff
@@ -80,3 +98,36 @@ export const processDeployment = async (
 		gitHubAppId
 	);
 };
+
+const persistentSuccessDeploymentStatusToDynamoDB = async (
```
bgvozdev, line 102: curious why is it here and not in `deployment-service` ?

bgvozdev: actually I guess that makes sense to have mapping here...

src/github/deployment.ts
```diff
@@ -80,3 +98,36 @@ export const processDeployment = async (
...
+	const info = { jiraHost };
+
+	try {
+		statsd.increment(metricDeploymentPersistent.toCreate, tags, info);
+		await saveDeploymentInfo({
+			gitHubBaseUrl,
+			gitHubInstallationId,
+			repositoryId: webhookPayload.repository.id,
+			commitSha: webhookPayload.deployment.sha,
+			description: webhookPayload.deployment.description || "",
```
bgvozdev, line 122: do we really need description? It might contain UGC/PII...

src/github/deployment.ts
```diff
@@ -80,3 +98,36 @@ export const processDeployment = async (
...
+			env: webhookPayload.deployment_status.environment,
```
bgvozdev, line 123: should we maybe hash to avoid storing UGC/PII?

src/models/deployment-service.test.ts
```diff
@@ -0,0 +1,118 @@
...
+				gitHubInstallationId, repositoryId,
+				env: "production", currentDate: createdAt2
+			}, logger);
+			expect(result).toEqual({
+				repositoryId,
+				commitSha: "create-1",
+				createdAt: createdAt1
+			});
+		});
+		it("should fetch THE last success deployment when more than one previous success deployments", async () => {
```
bgvozdev, line 97: Could we please also add a test for `findLastSuccessDeployment` when currentDate is not exactly equals the stored value?

src/models/deployment-service.ts
```diff
@@ -0,0 +1,100 @@
...
+		Item: {
+			Id: { "S": getKey(deploymentInfo) },
+			StatusCreatedAt: { "N": String(deploymentInfo.createdAt.getTime()) },
+			GitHubInstallationId: { "N": String(deploymentInfo.gitHubInstallationId) },
+			RepositoryId: { "N": String(deploymentInfo.repositoryId) },
+			CommitSha: { "S": deploymentInfo.commitSha },
+			Description: { "S": deploymentInfo.description },
+			Env: { "S": deploymentInfo.env },
+			Status: { "S": deploymentInfo.status },
+			ExpiredAfter: { "N": String(Math.floor((deploymentInfo.createdAt.getTime() + ONE_YEAR_IN_MILLISECONDS) / 1000)) }
```
bgvozdev, line 32: What about leap years? kidding

src/models/deployment-service.ts
```diff
@@ -0,0 +1,100 @@
...
+ * -- future, I guess we can even share data across installation, coz deployment status data is per repo, regardless of which app installation it comes from.
+ * But this only benefit multi app within GHES, so can relax for now.
+ */
+const getKey = (opts: {
+	gitHubBaseUrl: string;
+	gitHubInstallationId: number;
+	repositoryId: number;
+	env: string;
+}) => {
+	return `ghurl_${opts.gitHubBaseUrl}_ghid_${opts.gitHubInstallationId}_repo_${opts.repositoryId}_env_${opts.env}`;
```
bgvozdev, line 99: nice!

bgvozdev: but again, should we hash env here?

src/models/dynamodb.ts
```diff
@@ -0,0 +1,39 @@
+import AWS from "aws-sdk";
```
bgvozdev, line 1: Curious if it needs to be in models? I think models are only about sequelize. Should we maybe create `dynamodb` dir and pull all dynamodb-related stuff there?

src/routes/github/webhook/webhook-context.ts
```diff
@@ -21,7 +21,7 @@ export class WebhookContext<E = any> {
 	timedout?: number;
 	webhookReceived?: number;
 
-	gitHubAppConfig?: GitHubAppConfig; // TODO: must become required once we remove probot: it is populated for both Cloud and Server in the new webhook handler
```
bgvozdev, line 24: sweet!

src/transforms/transform-deployment.ts
```diff
@@ -48,30 +53,90 @@ const getLastSuccessfulDeployCommitSha = async (
...
+	if (!lastSuccessfullyDeployedCommit) {
+		// Grab the last 10 deployments for this repo
+		const deployments: Octokit.Response<Octokit.ReposListDeploymentsResponse> | AxiosResponse<Octokit.ReposListDeploymentsResponse> =
+			await githubInstallationClient.listDeployments(owner, repoName, currentDeployEnv, 10);
+
+		// Filter per current environment and exclude itself
+		const filteredDeployments = deployments.data
+			.filter(deployment => deployment.id !== currentDeployId);
+
+		// If this is the very first successful deployment ever, return nothing because we won't have any commit sha to compare with the current one.
```
bgvozdev, line 133: amen!

bgvozdev: Just to understand that correctly, if item is not found, we are falling back to the old logic, right?

src/transforms/transform-deployment.ts
```diff
@@ -229,17 +294,28 @@ const mapJiraIssueIdsCommitsAndServicesToAssociationArray = (
 	return associations;
 };
 
-export const transformDeployment = async (githubInstallationClient: GitHubInstallationClient, payload: WebhookPayloadDeploymentStatus, jiraHost: string, metrics: {trigger: string, subTrigger?: string}, logger: Logger, gitHubAppId: number | undefined): Promise<JiraDeploymentBulkSubmitData | undefined> => {
+export const transformDeployment = async (
+	githubInstallationClient: GitHubInstallationClient,
+	payload: WebhookPayloadDeploymentStatus,
+	jiraHost: string,
+	type: "backfill" | "webhook",
+	metrics: {trigger: string, subTrigger?: string},
```
bgvozdev, line 302: lint will be pissed off

src/models/deployment-service.ts
gxueatlassian: We forget things in leap years, so is TTL.

src/transforms/transform-deployment.ts
gxueatlassian: Yup.

gxueatlassian (commented):
blah

src/github/deployment.ts
gxueatlassian: fixed.

src/github/deployment.ts
gxueatlassian: fixed.

src/github/deployment.test.ts
gxueatlassian: tests added.

src/github/deployment.ts
gxueatlassian: 🤷

src/github/deployment.ts
gxueatlassian: removed. good point.

src/models/deployment-service.test.ts
gxueatlassian: added.

src/models/dynamodb.ts
gxueatlassian: Well, I am against put it under a dir name `dynamodb` as I think dir should be divided by 'features' instead of 'technologies`.

But putting under `models` might not make sense either. This is more like a services utils. And from the name of the methods, it doesn't even imply "dynamo" as the technologies.

I'll put it under the `services` dir?

src/transforms/transform-deployment.ts
gxueatlassian: 🤷

github-for-jira.sd.yml
bgvozdev: Yeah, I was thinking of some more explicit naming, e.g. InstallationIdEnvRepoId. I don't think you can easily migrate the records from one logic to another, so I'd not have bothered about flexibility. Also, the name of the range key you are using is quite descriptive 😛

gxueatlassian: `migrate the records from one logic to another` <--- less about this (data migration), more about ask micros/aws to change the partition key from `InstallationIdEnvRepoId` to `AnotherIdEnvRepoId`? Is it possible?

gxueatlassian: > Yeah, I was thinking of some more explicit naming, e.g. InstallationIdEnvRepoId. I don't think you can easily migrate the records from one logic to another, so I'd not have bothered about flexibility. Also, the name of the range key you are using is quite descriptive 😛

I almost change the `StatusCreatedAt` to `CreatedAt`.

bgvozdev: >  more about ask micros/aws to change the partition key from InstallationIdEnvRepoId to AnotherIdEnvRepoId

I reckon for a migration like that you'll need to create a new table anyway, wdyt?

src/models/dynamodb.ts
gxueatlassian: done.

github-for-jira.sd.yml
gxueatlassian: I agree, and that's why I make the ID column generic. But also agree no easy way to migrate data even I can getaway with the generic ID col. 

In short, to change the semantic, nothing easy.

src/models/deployment-service.ts
gxueatlassian: fixed. Key hashed for privacy and bonus data size consistency

src/github/deployment.ts
gxueatlassian: fixed removed unused data.

github-for-jira.sd.yml
gxueatlassian: Okay I've tested, there's no easy way to change the partition key schema on existing table.
After another thought, I still prefer to have a generic Id column, so that when we change our mind in the future, we can:
1. Save the db with new key value
2. Ignore the existing data as a) they will not conflicts with new value (hash + new semantic) and b) they will expired by themselves. 
3. We don't have to deploy a new table with new name + use ff to switch over + cli on prod ( + stg/dev) to remove unused table. 
Too much effort involved to change to use another table should we change our mind to use another value for key.

.env.test
```diff
@@ -21,6 +21,10 @@ SQS_BRANCH_QUEUE_REGION=us-west-1
 GLOBAL_HASH_SECRET=testsecret
 DEBUG=nock.*
 
+#Dyamno for deployment status
+DYNAMO_DEPLOYMENT_HISTORY_TABLE_REGION=us-west-1
+DYNAMO_DEPLOYMENT_HISTORY_TABLE_NAME=deployment-history-cache
```
bgvozdev, line 26: Should we use "cache" in all places? I see now it is just the name of the table, however "cache" keyword is not propagated in the code....

github-for-jira.sd.yml
```diff
@@ -149,6 +149,16 @@ resources:
         # Name of the RDS resource from above
         resource: rds
 
+  - name: deployment-history-cache
+    type: dynamo-db
+    attributes: &table-attributes
+      HashKeyName: Id
+      HashKeyType: "S"
+      RangeKeyName: StatusCreatedAt
```
bgvozdev, line 157: Should we also change that to be generic?

src/models/dynamodb.ts
bgvozdev: Sorry, I might miss something but the file is still in models dir, isn't it? :badpokerface:

src/transforms/transform-deployment.ts
```diff
@@ -48,30 +53,89 @@ const getLastSuccessfulDeployCommitSha = async (
...
-		return undefined;
+	if (type === "webhook" && await booleanFlag(BooleanFlags.USE_DYNAMODB_FOR_DEPLOYMENT_WEBHOOK, jiraHost)) {
+		lastSuccessfullyDeployedCommit = await getLastSuccessDeploymentShaFromDyanmoDB(type, jiraHost, repoId, currentDeployEnv, currentDeployDate, githubInstallationClient, logger);
 	}
 
-	const lastSuccessfullyDeployedCommit = await getLastSuccessfulDeployCommitSha(owner, repoName, githubInstallationClient, filteredDeployments, logger);
+	if (!lastSuccessfullyDeployedCommit) {
+		// Grab the last 10 deployments for this repo
+		const deployments: Octokit.Response<Octokit.ReposListDeploymentsResponse> | AxiosResponse<Octokit.ReposListDeploymentsResponse> =
+			await githubInstallationClient.listDeployments(owner, repoName, currentDeployEnv, 10);
```
bgvozdev, line 126: I was thinking about this... If we treat the Db as a cache, and if the last deployment is not there for some reason (e.g. we dropped the cache) and there was a latency in processing, which makes first 10 items irrelevant... should we fix this logic a little bit?

 - if the last successful deployment is not cached
 - if the CURRENT deployment is not among last 10 deployments OR the last successful deployment is not in the last 10 items
 --- then just use issue key from commit message/branch name (<--- new logic)
 --- otherwise proceed with the current logic
 
wdyt?

src/util/hash-utils.ts
```diff
@@ -0,0 +1,9 @@
+import { createHash } from "crypto";
```
bgvozdev, line 1: I was wondering if we maybe should put it to the existing file with out hashing function? https://github.com/atlassian/github-for-jira/blob/eeeff2f86a9c94d8dc8baf0dcaed42273f1d074f/src/util/encryption.ts#L4-L4 

There we have `createHashWithSharedSecret` , should we have some `createHashWithoutSharedSecret`, to 
1. keep things uniform
2. explicitly separate those

wdyt?

src/models/dynamodb.ts
gxueatlassian: Sorry I missed this one. I moved the deployment-service to under /services. 
For this one, I split it into two now. The `db connection` moved to "config", and the `purgeItem` into the test/setup/setup.ts coz it is only used by test and only in that location.

Now should looks better.

.env.test
gxueatlassian: Good catch!! I should have renamed all those now.

github-for-jira.sd.yml
gxueatlassian: Changed.

src/transforms/transform-deployment.ts
gxueatlassian: I am thinking another PR, another day ;ol.
`--- then just use issue key from commit message/branch name (<--- new logic)` this can be unexpected, we can chat more in detail. Basically calling github api `/compare/deployment_sha...<whatever_sha_from_anywhere>` will yield much more info than to simply `just use issue key from commit message`. 
We already calling that api right now for whatever reason, changing this will need more thinking. I suspect previously I compare the deployment only to futures and skip the sha compare api coz some customer tickets.

src/util/hash-utils.ts
gxueatlassian: Thought of that. But the folder name is `encryption`, but the hash is not about encryption. Still a bit inconsistent.

src/config/metric-names.ts
```diff
@@ -62,3 +62,13 @@ export const metricCreateBranch = {
 	created: `${server}.create-branch.created`,
 	failed: `${server}.create-branch.failed`
 };
+
+export const metricDeploymentPersistent = {
+	toCreate: `${server}.persistent.entry.to-create`,
```
bgvozdev, line 67: btw we don't have anything related to "deployment' in the name of the metric... wondering if that's intentionally?

src/config/metric-names.ts
```diff
@@ -62,3 +62,13 @@ export const metricCreateBranch = {
 	created: `${server}.create-branch.created`,
 	failed: `${server}.create-branch.failed`
 };
+
+export const metricDeploymentPersistent = {
```
bgvozdev, line 66: `metricDeploymentCache`  maybe?

src/github/deployment.ts
```diff
@@ -80,3 +98,32 @@ export const processDeployment = async (
...
+	logger: Logger
+) => {
+
+	const gitHubProduct = getCloudOrServerFromGitHubAppId(gitHubAppId);
+	const tags = { gitHubProduct };
+	const info = { jiraHost };
+
+	try {
+		statsd.increment(metricDeploymentPersistent.toCreate, tags, info);
+		await saveDeploymentInfo({
```
bgvozdev, line 116: how about `cacheSuccessfulDeploymentInfo`?

src/services/deployment-service/deployment-service.ts
```diff
@@ -0,0 +1,87 @@
+import Logger from "bunyan";
```
bgvozdev, line 1: should it be `src/services/deployment-cache-service/deployment-cache-service.ts` maybe?

src/services/deployment-service/deployment-service.ts
```diff
@@ -0,0 +1,87 @@
...
+import { envVars } from "config/env";
+import { getLogger } from "config/logger";
+import { dynamodb as ddb } from "config/dynamodb";
+import { hash } from "utils/hash-utils";
+
+const defaultLogger = getLogger("DeploymentDynamoLogger");
+
+const ONE_YEAR_IN_MILLISECONDS = 365 * 24 * 60 * 60 * 1000;
+
+export const saveDeploymentInfo = async (deploymentInfo : {
```
bgvozdev, line 11: How about `cacheSuccessfulDeploymentInfo`?

src/services/deployment-service/deployment-service.ts
```diff
@@ -0,0 +1,87 @@
...
+		throw result.$response.error;
+	}
+};
+
+export type LastSuccessfulDeployment = {
+	commitSha: string;
+	createdAt: Date;
+}
+
+export const findLastSuccessDeployment = async(
```
bgvozdev, line 38: how aobut `findCachedLastSuccessfulDeploymentInfo`?

src/util/hash-utils.ts
bgvozdev: Well, sha256 is all all about the [encryption/cryptography](https://en.wikipedia.org/wiki/Cryptographic_hash_function) (as opposed to data structures' one [that is used in hash table](https://en.wikipedia.org/wiki/Hash_function))...  IMHO putting it into encryption utils makes more sense tbh...

src/config/metric-names.ts
gxueatlassian: Good catch, I've added some specific name to the metrics itself.

src/services/deployment-service/deployment-service.ts
gxueatlassian: It could be too narrow, what if we want to put more function in this file? I am not too against the idea, but no too fancy about it.

src/services/deployment-service/deployment-service.ts
gxueatlassian: Updated to `findLastSuccessDeploymentFromCache`

src/util/hash-utils.ts
gxueatlassian: Maybe we can rename the folder name in another PR. `Encryption` should means you can `decrypt` it later. But right now all it has are just `hash`, which we can't un-hash.
Anyway I've taken your advice and put it into that folder.

src/services/deployment-service/deployment-service.ts
gxueatlassian: Okay, bring service one level up and renamed to `deployment-cache-service.ts`

src/github/deployment.ts
```diff
@@ -80,3 +98,32 @@ export const processDeployment = async (
 		gitHubAppId
 	);
 };
+
+const persistentSuccessDeploymentStatusToDynamoDB = async (
```
bgvozdev, line 102: nit: I feel like "dynamodb" suffix is implementation detail, should we maybe drop it?

src/transforms/transform-deployment.ts
```diff
@@ -48,30 +53,89 @@ const getLastSuccessfulDeployCommitSha = async (
 	return deployments[deployments.length - 1].sha;
 };
 
-const getCommitsSinceLastSuccessfulDeployment = async (
+const getLastSuccessDeploymentShaFromDyanmoDB = async (
```
bgvozdev, line 56: nit: I feel like "dynamodb" suffix is implementation detail, should we maybe drop it? `getLastPersisted(/Cached)SuccessfulDeployemtnSha` ?

bgvozdev (approved):

bgvozdev (approved):

PR #2124, opened 2023-05-16
None remove parallel sync ff and make scheduler configurable

bgvozdev:
**What's in this PR?**
Control number of backfill subtasks with a FF

**Why**
This would allow to control the load on Data Depot

**Added feature flags**
backfill-max-subtasks

**Affected issues**  
nah

**How has this been tested?**  
unit tests

**Whats Next?**


joshkay10 (approved):
mmmm

src/sync/installation.test.ts
```diff
@@ -193,6 +205,7 @@ describe("sync/installation", () => {
 		it("should update cursor and continue sync", async () => {
 			const sendSqsMessage = jest.fn();
 			githubUserTokenNock(DatabaseStateCreator.GITHUB_INSTALLATION_ID);
+			githubUserTokenNock(DatabaseStateCreator.GITHUB_INSTALLATION_ID);
```
rachellerathbone, line 208: Why do we need to call this twice? 😕

bgvozdev: Because I removed the FF that was controlling rate_limit endpoint in scheduler. But actually changed that back because don't need to call this when number of subtasks is 0 (and it also broke a lot of tests, which was annoying)

joshkay10 (approved):

PR #2126, opened 2023-05-17
None block by jira host

gxueatlassian:
**What's in this PR?**
1. Expand blocking to all webhooks and backfill
2. Add jiraHost to blocking so that we can easily setup the ff (The current variant setup in ff is awkward)

**Why**
We have some offending instance that consume lots of our resources.

**Added feature flags**
Current feature flag state:
image

image

image

**Affected issues**  

**How has this been tested?**  
Unit test

**Whats Next?**


src/config/feature-flags.ts
```diff
@@ -81,9 +81,9 @@ export const onFlagChange = (flag: BooleanFlags | StringFlags | NumberFlags, lis
 	launchdarklyClient.on(`update:${flag}`, listener);
 };
 
-export const isBlocked = async (installationId: number, logger: Logger): Promise<boolean> => {
+export const isBlocked = async (jiraHost: string, installationId: number, logger: Logger): Promise<boolean> => {
 	try {
-		const blockedInstallationsString = await stringFlag(StringFlags.BLOCKED_INSTALLATIONS, "[]");
+		const blockedInstallationsString = await stringFlag(StringFlags.BLOCKED_INSTALLATIONS, "[]", jiraHost);
```
gxueatlassian, line 86: Without the `key` in the ff, it is really hard to setup the ff in launchdarkly for targeting an installation/jira

src/github/deployment.ts
```diff
@@ -39,11 +38,6 @@ export const processDeployment = async (
 		webhookReceived: webhookReceivedDate
 	});
 
-	if (await isBlocked(gitHubInstallationId, logger)) {
```
gxueatlassian, line 42: No need to put it here as the webhook middleware will block at the WebServer side before entering into queue.

src/middleware/github-webhook-middleware.ts
```diff
@@ -203,6 +203,11 @@ export const GithubWebhookMiddleware = (
 				continue;
 			}
 
+			if (await isBlocked(jiraHost, gitHubInstallationId, context.log)) {
+				context.log.info({ jiraHost, gitHubInstallationId }, `Block webhook processing`);
+				continue;
```
rachellerathbone, line 208: Why are we continuing here? If we want to block the webhook from processing shouldn't we return?

rachellerathbone (dismissed):

krazziekay (approved):

PR #2132, opened 2023-05-24
Allow Jira users to navigate directly to a deployed environment from Deployments widget

aldenquimby:
**What's in this PR?**
- Allow Jira users to navigate directly to a deployed environment from Deployments widget
- Uses the existing `environment_url` webhook field to parallel the `View Deployment` button from Github Deployments:
image

**Why**
- Our team uses transient/ephemeral environments per pull request
- When someone is trying to QA a Jira ticket, it's very helpful to be able to quickly navigate to the correct environment
- Here is what we see today, for example:
image

- The `Pipeline` and `Deployment` columns both link to the exact same place - the Github Action job run that created the Github Deployment entry
- We would love for the `Environment name` column to link directly to the environment, but this doesn't appear to be possible
- Alternatively, what this PR proposes, is to have the `Deployment` column link to the actual deployed environment, rather than the Github Action job run (which `Pipeline` still links to)

**Added feature flags**
- none

**Affected issues**  
- None
- I tried to create a Github Issue, but that [redirects here](https://jira.atlassian.com/secure/Dashboard.jspa)
- I tried to create a product suggestion under Jira Software Cloud, but that redirected me to _Give Feedback_ within Jira

**How has this been tested?**  
- considered adding a test for this, but it felt unnecessary based on the level of tests in `transform-deployment.test.ts`. Happy to add one if I should

### DIFF
--- a/src/sync/deployment.ts
+++ b/src/sync/deployment.ts
@@ -38,6 +38,7 @@ const getTransformedDeployments = async (deployments, gitHubInstallationClient: 
 				environment: deployment.environment,
 				id: deployment.databaseId,
 				target_url: deployment.latestStatus?.logUrl,
+				environment_url: deployment.latestStatus?.environmentUrl,
 				updated_at: deployment.latestStatus?.updatedAt,
 				state: deployment.latestStatus?.state
 			}
--- a/src/transforms/transform-deployment.ts
+++ b/src/transforms/transform-deployment.ts
@@ -372,7 +372,7 @@ export const transformDeployment = async (
 			deploymentSequenceNumber: deployment.id,
 			updateSequenceNumber: deployment_status.id,
 			displayName: (message || String(payload.deployment.id) || "").substring(0, 255),
-			url: deployment_status.target_url || deployment.url,
+			url: deployment_status.environment_url || deployment_status.target_url || deployment.url,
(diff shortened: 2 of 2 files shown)


aldenquimby:
Re-opening b/c #2162 was closed (Jira API doesn't support URLs on the environment)

gxueatlassian:
Hi @aldenquimby thank you for the PR. Since this will effect all our other users as well, I'll need to ping our team first. Can give update later.

aldenquimby:
@gxueatlassian ok thank you. I can fix up the merge conflicts and get tests passing if you give the all clear here.

derylseale:
Checking in to see if there's any progress here. Transient testing environments are increasingly common, and making it easier to link to them would bring a lot of joy to our QA process.

aldenquimby:
Seems like this PR is dead, so I am going to close it. Wishing someone else better luck in pushing this though! Happy holidays 🎄

rmott-littler:
And this is how you can tell Atlassian never uses their own product. Yet another simple, needed improvement allowed to wither and die. @gxueatlassian why is this so hard?

derylseale:
> And this is how you can tell Atlassian never uses their own product. Yet another simple, needed improvement allowed to wither and die. 

Agreed, so disappointing. Looks like you can do this in Shortcut with a little elbow grease, time for another look at their product.

PR #2133, opened 2023-05-24
Upgrade @octokit/webhooks-types

aldenquimby:
**What's in this PR?**
- Upgrade `@octokit/webhooks` to the latest `@octokit/webhooks-types`

**Why**
- Existing types are outdated. E.g. see #2132

**Added feature flags**
- No

**Affected issues**  
- None

**How has this been tested?**  
- Since this is a TypeScript-only change, as long as the build passes it should be safe

**Whats Next?**
- nothing?

### DIFF
--- a/src/github/issue.ts
+++ b/src/github/issue.ts
@@ -1,10 +1,10 @@
 import { emitWebhookProcessedMetrics } from "utils/webhook-utils";
-import { WebhookPayloadIssues } from "@octokit/webhooks";
+import type { IssuesOpenedEvent, IssuesEditedEvent } from "@octokit/webhooks-types";
 import { GitHubIssue, GitHubIssueData } from "interfaces/github";
 import { createInstallationClient } from "utils/get-github-client-config";
 import { WebhookContext } from "../routes/github/webhook/webhook-context";
 
-export const issueWebhookHandler = async (context: WebhookContext<WebhookPayloadIssues>, jiraClient, util, gitHubInstallationId: number): Promise<void> => {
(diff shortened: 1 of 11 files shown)


yarn.lock
```diff
@@ -719,6 +719,11 @@
     graphql "^15.0.0"
     graphql-tag "^2.10.3"
```
aldenquimby, line 721: I am unfamiliar with yarn, so if I messed something up here please let me know. This diff came from:
```
yarn remove @octokit/webhooks
yarn add -D @octokit/webhooks-types@latest
```

yarn.lock
```diff
@@ -831,12 +831,10 @@
   dependencies:
     "@octokit/openapi-types" "^7.2.3"
```
aldenquimby, line 833: I am unfamiliar with yarn, so if I messed something up here please let me know. This diff came from:
```
yarn remove @octokit/webhooks
yarn add -D @octokit/webhooks-types@latest
```

src/github/issue.ts
```diff
@@ -28,7 +28,9 @@ export const issueWebhookHandler = async (context: WebhookContext<WebhookPayload
 	// TODO: need to create reusable function for unfurling
 	let linkifiedBody;
 	try {
-		linkifiedBody = await util.unfurl(issue.body);
+		if (issue.body !== null) {
+			linkifiedBody = await util.unfurl(issue.body);
+		}
```
aldenquimby, line 33: This is the only non-typescript change b/c the latest TS types allow `issue.body` to be null

gxueatlassian (approved):
Hi @aldenquimby thank you very much for your effort in this PR. I am happy to merge it in.

PR #2134, opened 2023-05-25
ARC-2192 migrate btf beta customers

bgvozdev:
**What's in this PR?**
- script to migrate Beta BTF customers from FF to Db
- remove unsafe logs

**Why**
- because we are moving away from FF API keys
- because we haven't checked unsafe logs for the past 6 months

### DIFF
--- a/src/routes/api/api-btf-migrate-post.ts
+++ b/src/routes/api/api-btf-migrate-post.ts
@@ -45,9 +45,9 @@ export const ApiBtfMigratePost = async (req: Request, res: Response): Promise<vo
 	});
 
 	const apps = await GitHubServerApp.findAll();
-	await Promise.all(apps.map(async (app) =>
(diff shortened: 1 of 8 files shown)


src/routes/api/api-btf-migrate-post.ts
```diff
@@ -0,0 +1,55 @@
...
+	log.info("Migrated :clapclap: !");
+};
+
+export const ApiBtfMigratePost = async (req: Request, res: Response): Promise<void> => {
+	const log = req.log.child({
+		btfMigration: true
+	});
+
+	const apps = await GitHubServerApp.findAll();
+	for (let appIdx = 0; appIdx < apps.length; appIdx++) {
```
krazziekay, line 48: \nit
Better to use `forEach`

krazziekay (approved):

PR #2136, opened 2023-05-25
ARC 2068 - Handling SSO Login in UI

krazziekay:
**What's in this PR?**
- Identified the SSO login users who can not connect their orgs.
- Displaying information to those users to correctly connect their orgs.

**Why**
- SSO login users can not click on the `CONNECT` button and have no idea why.

**How has this been tested?**  
- Locally

**Screenshot**
image

### DIFF
--- a/src/routes/github/configuration/github-configuration-get.ts
+++ b/src/routes/github/configuration/github-configuration-get.ts
@@ -10,7 +10,7 @@ import { AppInstallation } from "config/interfaces";
 import { envVars } from "config/env";
 import { GitHubUserClient } from "~/src/github/client/github-user-client";
 import { isUserAdminOfOrganization } from "utils/github-utils";
-import { GithubClientBlockedIpError } from "~/src/github/client/github-client-errors";
+import { GithubClientBlockedIpError, GithubClientError } from "~/src/github/client/github-client-errors";
 import {
 	createInstallationClient,
 	createUserClient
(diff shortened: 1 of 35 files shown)


src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -101,6 +101,7 @@ const getInstallationsWithAdmin = async (
 			...installation,
 			numberOfRepos,
 			isAdmin,
+			isSso: !!errors.find(err => err?.cause?.response?.headers["x-github-sso"]),
```
bgvozdev, line 104: It doesn't look too reliable tbh... Could we maybe modify isUserAdminOfOrganization to return some result object/enum that returns more details?

src/routes/github/github-oauth-router.ts
```diff
@@ -141,6 +141,9 @@ export const GithubAuthMiddleware = async (req: Request, res: Response, next: Ne
 	try {
 		const { query, originalUrl } = req;
 		if (query && query["resetGithubToken"]) {
+			req.session.githubToken = undefined;
+			req.session.githubRefreshToken = undefined;
```
bgvozdev, line 145: Curious why the tests were not failing? I think we should them test cause they don't look reliable...

src/routes/github/configuration/github-configuration-get.ts
krazziekay: Created a new Error instance `GithubClientSSOLoginError` and matched the type of error with this type

src/util/github-utils.ts
```diff
@@ -14,6 +14,8 @@ export const isUserAdminOfOrganization = async (githubClient: GitHubUserClient,
 		return role === "admin";
 	} catch (err) {
 		logger.warn({ err, orgName, username }, `could not determine admin status of user in org`);
+		err.requiresSsoLogin = err?.cause?.response?.headers["x-github-sso"];
```
bgvozdev, line 17: Shouldn't error already be an instance of GithubClientSSOLoginError ? Why do we need this flag?

I think in github client interceptors we should check for:
 - status (=401?) 
 - the header

If these match, we rethrow the error as GithubClientSSOLoginError and then just use instanceof everywhere. Wdyt?

krazziekay: Updated as suggested.

bgvozdev (approved):

src/routes/github/github-oauth-router.ts
krazziekay: Test case added

bgvozdev (approved):

PR #2138, opened 2023-05-29
[EZY KARMA]: remove outbound proxy skiplist FF

bgvozdev:
**What's in this PR?**
Removing FF

**Why**
Not used anymore: for the internal testing instance we should use API gateway

### DIFF
--- a/src/util/get-github-client-config.ts
+++ b/src/util/get-github-client-config.ts
@@ -20,39 +20,6 @@ interface GitHubClientConfig extends GitHubConfig {
 	gitHubClientSecret: string;
 }
 
-/**
- * Decides whether to use the proxy URL and which one if so.
- *
- * @param jiraHost - jiraHost from context
- * @param gitHubBaseUrl - for cloud, either undefined or "github.com"'s base url
- * @param logger
- *
- * @return proxy URL or undefined (do not use)
- */
-const calculateProxyBaseUrl = async (jiraHost: string, gitHubBaseUrl: string | undefined, logger: Logger): Promise<string | undefined> => {
-	if (gitHubBaseUrl && gitHubBaseUrl != GITHUB_CLOUD_BASEURL) {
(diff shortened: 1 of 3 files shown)


src/util/get-github-client-config.ts
```diff
@@ -123,14 +90,14 @@ const buildGitHubClientServerConfig = async (gitHubServerApp: GitHubServerApp, j
 	}
 );
 
-const buildGitHubClientCloudConfig = async (jiraHost: string, logger: Logger): Promise<GitHubClientConfig> => {
+const buildGitHubClientCloudConfig = async (jiraHost: string): Promise<GitHubClientConfig> => {
```
krazziekay, line 93: This won't be async too.

src/util/get-github-client-config.ts
```diff
@@ -88,18 +55,18 @@ const buildGitHubServerConfig = async (githubServerBaseUrl: string, jiraHost: st
...
 		apiUrl: `${githubServerBaseUrl}/api/v3`,
 		graphqlUrl: `${githubServerBaseUrl}/api/graphql`,
-		proxyBaseUrl: await calculateProxyBaseUrl(jiraHost, githubServerBaseUrl, logger),
+		proxyBaseUrl: envVars.PROXY,
 		... await calculateApiKeyConfig(jiraHost, logger, apiKeyConfig)
 	};
 };
 
-const buildGitHubCloudConfig = async (jiraHost: string, logger: Logger): Promise<GitHubConfig> => {
+const buildGitHubCloudConfig = async (): Promise<GitHubConfig> => {
```
krazziekay, line 63: This doesn't need to be async anymore.

krazziekay (commented):
Minor changes

src/util/get-github-client-config.ts
bgvozdev: This one has to because of
`const privateKey = await keyLocator(undefined, jiraHost);`

krazziekay (approved):

PR #2142, opened 2023-05-29
Extend `staging` environment mapping with `sta`

krzysztof-pyrzyk-ah:
**What's in this PR?**
This PR extends default environment name mapping of `staging` environment with a commonly used `sta` keyword.

**Why**
`sta` is a less common abbreviation for the word `staging` than `stg`, but it still seems to occur quite frequently.

**Added feature flags**
_no_

**Affected issues**  
_none_

**How has this been tested?**  
_Unit test case added_

**Whats Next?**
_n/a_


gxueatlassian:
Thank you @krzysztof-pyrzyk-ah for your support. Looks good to me, should be auto merged once test pass.

gxueatlassian (approved):

PR #2149, opened 2023-05-31
Arc-2115 search connected orgs

rachellerathbone:
**What's in this PR?**
Middleware for /jira/workspaces/search

**Why**
To support searchConnectedWorkspaces for generic containers https://hello.atlassian.net/wiki/spaces/OTFS/pages/2581193453/Extending+GitHub+for+Jira+for+generic+containers#searchConnectedWorkspaces

**Affected issues**  
ARC-2115

**How has this been tested?**  
Locally and unit tests (as mentioned in ARC-2114, we will test from Jira at the end of June once the work being carried out by the Incredibles has been completed).

**Whats Next?**
Once this has been merged I'll open up ARC-2116


src/models/reposyncstate.ts
```diff
@@ -208,6 +208,20 @@ export class RepoSyncState extends Model {
 		}));
 	}
 
+	static async findByOrgNameAndSubscriptionId(subscriptions: Subscription[], orgName: string): Promise<FlatArray<Awaited<RepoSyncState[]>[], 2>[]> {
+		const result = await Promise.all(subscriptions.map(async sub => {
+			return await RepoSyncState.findAll({
+				where: {
+					subscriptionId: sub.id,
+					repoOwner: {
+						[Op.iLike]: `%${orgName}%`
```
rachellerathbone, line 217: I chatted with Darcy about this. Expected behaviour is that users will type either part of an org name, or the whole thing, so we need to do partial matches. I had some concerns around race conditions so asked him what they were doing to try and avoid that. He told me they use Lodash to debounce requests from the client and ensures that the last request made is the one that will be shown to the user.

src/routes/jira/workspaces/jira-workspaces-get.test.ts
```diff
@@ -0,0 +1,248 @@
...
+			success:true,
+			workspaces: [
+				{
+					id: sub.id.toString(),
+					name: "atlas",
+					canCreateContainer: false
+				},
+				{
+					id: sub2.id.toString(),
+					name: "atlassian",
```
krazziekay, line 227: Is this correct? The search parameter `atlas` should return both `atlassian` and `anotheratlasmatch`, isn't it? Why just `atlassian`?

src/routes/jira/workspaces/jira-workspaces-get.ts
```diff
@@ -1,3 +1,74 @@
...
+		const { subscriptionId, repoOwner } = org;
+
+		return {
+			id: subscriptionId.toString(),
+			name: repoOwner,
+			// default to false until support is added for createContainer
+			canCreateContainer: false
+		};
+	})
+		.filter((value, index, self) =>
```
krazziekay, line 30: So if there are multiple repositories with the same `subscriptionId`, it returns only one match. Is this correct? 
Whats the reason behind this? Shouldn't we be returning all the possible matches?

src/models/reposyncstate.ts
```diff
@@ -208,6 +208,20 @@ export class RepoSyncState extends Model implements RepoSyncStateProperties {
 		}));
 	}
 
+	static async findByOrgNameAndSubscriptionId(subscriptions: Subscription[], orgName: string): Promise<FlatArray<Awaited<RepoSyncState[]>[], 2>[]> {
```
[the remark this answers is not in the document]
gxueatlassian: +1, a bit confused in the return type. Can't it be `Promise<RepoSyncState[]>`?

src/models/reposyncstate.ts
gxueatlassian: /nit
Can you put something in your list that `May need to put index on col repoOwner to speed up performance`? 
But I think we may not need that until really future.

src/routes/jira/workspaces/jira-workspaces-get.test.ts
```diff
@@ -0,0 +1,248 @@
...
+
+	it("Should return multiple matches but only return each org once", async () => {
+		app = express();
+		app.use((req, _, next) => {
+			req.log = getLogger("test");
+			next();
+		});
+		app.use(getFrontendApp());
+
+		const sub2 = await Subscription.install({
```
gxueatlassian, line 158: Can you extract some method to condense this test? It will make test more readble.

src/routes/jira/workspaces/jira-workspaces-get.ts
gxueatlassian: +1 a bit confusing in the logic here
Assume the logic is to return a unique org name, we can do a `distinct` query in postgress on the `repoOwner` column right? Postgres will be easier to process it.

src/routes/jira/workspaces/jira-workspaces-get.ts
```diff
@@ -1,3 +1,74 @@
...
+
+	return matchedOrgs;
+};
+
+export const JiraWorkspacesGet = async (req: Request, res: Response): Promise<void> => {
+	req.log.info({ method: req.method, requestUrl: req.originalUrl }, "Request started for GET workspaces");
+
+	const { jiraHost } = res.locals;
+
+	if (!jiraHost) {
```
gxueatlassian, line 44: I think we can ignore its, since it will/should be verified somewhere up the router?

src/routes/jira/workspaces/jira-workspaces-get.ts
```diff
@@ -1,3 +1,74 @@
...
+	if (!orgName) {
+		req.log.warn(MISSING_ORG_NAME);
+		res.status(400).send(MISSING_ORG_NAME);
+		return;
+	}
+
+	const matchedOrgs = await findMatchingOrgs(subscriptions, orgName);
+
+	if (!matchedOrgs?.length) {
+		res.status(400).send(NO_MATCHING_WORKSPACES);
```
gxueatlassian, line 69: 1. Should it be 404 isntead of 400?
2. Have you checked with saiyans? Should it be 200 with `[]` or `4xx`?

rachellerathbone: You were wrong on both attempts: Incredibles :laughcry:

Good call out though. They didn't specify in the doc they gave me so I just kind made an assumption :disappears: Will confirm with them.

src/routes/jira/workspaces/jira-workspaces-get.test.ts
```diff
@@ -0,0 +1,293 @@
...
+import { Subscription } from "models/subscription";
+import { getFrontendApp } from "~/src/app";
+import supertest from "supertest";
+import { RepoSyncState } from "models/reposyncstate";
+import { Installation } from "models/installation";
+import { encodeSymmetric } from "atlassian-jwt";
+import { Errors } from "config/errors";
+import { DEFAULT_LIMIT, Workspace } from "routes/jira/workspaces/jira-workspaces-get";
+
+const createRepos = async(subscriptions: Subscription[]) => {
```
rachellerathbone, line 12: For the test where we're passing a searchQuery param, I wanted a little more control over the repoOwner naming so created this function just for that use case. createRepositories is a generic function used to create 1 repo for each subscription (basically where we don't care what the repoOwner name is).

src/util/paginate-response.ts
```diff
@@ -0,0 +1,7 @@
+import { Workspace } from "../routes/jira/workspaces/jira-workspaces-get";
+
+export const paginatedRepositories = (page: number, limit: number, payload: Workspace[]) => {
```
rachellerathbone, line 3: I need this in 2116 so figured it should have its own home.

src/routes/jira/workspaces/jira-workspaces-get.test.ts
```diff
@@ -0,0 +1,235 @@
...
+	it("Should return all orgs that match provided searchQuery", async () => {
+		app = express();
+		app.use((req, _, next) => {
+			req.log = getLogger("test");
+			next();
+		});
+		app.use(getFrontendApp());
+
+		const subscriptions = await createSubscriptions(jiraHost, 3);
+		await createRepos(subscriptions);
```
rachellerathbone, line 185: I'm using `createRepositories` for tests where I don't care what the repoOwner name is (in the event no searchQuery is passed and we return all orgs). However, in this case, where I want more control over naming, I created a separate function so I can be more explicit.

src/routes/jira/workspaces/jira-workspaces-get.ts
```diff
@@ -1,3 +1,62 @@
...
+	canCreateContainer: boolean
+}
+
+const DEFAULT_PAGE_NUMBER = 1; // Current page
+export const DEFAULT_LIMIT = 20; // Number of items per page
+
+const findMatchingOrgs = async (subscriptions: Subscription[], orgName?: string): Promise<Workspace[]>  => {
+	let matchingRepos;
+
+	if (orgName) {
```
krazziekay, line 19: The condition can be put inside the loop, to avoid code repetition.
```suggestion
const matchingRepos = await Promise.all(subscriptions.map(async (subscription: Subscription) => {
			return orgName ? await RepoSyncState.findOneFromSubscription(subscription) : await RepoSyncState.findOneFromSubscription(subscription);
		}));
```

src/routes/jira/workspaces/jira-workspaces-get.ts
```diff
@@ -1,3 +1,62 @@
...
+			return await RepoSyncState.findByOrgNameAndSubscriptionId(subscription, orgName);
+		}));
+	} else {
+		matchingRepos = await Promise.all(subscriptions.map(async (subscription: Subscription) => {
+			return await RepoSyncState.findOneFromSubscription(subscription);
+		}));
+	}
+
+	const matchedOrgs = matchingRepos
+		.filter(org => org !== null)
```
krazziekay, line 30: \nit I suggest to use the [Boolean trick](https://michaeluloth.com/filter-boolean) for such null checks, its easier to read and use.
```suggestion
		.filter(Boolean)
```

src/util/paginate-response.test.ts
```diff
@@ -0,0 +1,48 @@
...
+		name: "workspace7",
+		canCreateContainer: false
+	}
+];
+
+describe("Paginate response data", () => {
+	it("Should return the correct data when paginated", () => {
+		expect(paginatedResponse(1, 3, mockWorkspaces)).toEqual(mockWorkspaces.slice(0, 3)); // first 3 items in mockWorkspaces
+		expect(paginatedResponse(2, 3, mockWorkspaces)).toEqual(mockWorkspaces.slice(3, 6)); // second 3 items in mockWorkspaces
+		expect(paginatedResponse(3, 3, mockWorkspaces)).toEqual(mockWorkspaces.slice(6, 7)); // second 3 items in mockWorkspaces
```
krazziekay, line 45:
```suggestion
		expect(paginatedResponse(3, 3, mockWorkspaces)).toEqual(mockWorkspaces.slice(6, 7)); // last 2 items in mockWorkspaces
```

test/utils/create-repositories.ts
```diff
@@ -0,0 +1,30 @@
...
+import { RepoSyncState } from "models/reposyncstate";
+
+const generateUniqueRepoOwner = (): string => {
+	const prefix = "repo-owner";
+	const uniqueId = Math.floor(Math.random() * 1000); // Generate a random number
+
+	return `${prefix}-${uniqueId}`;
+};
+
+export const createRepositories = async (subscriptions: Subscription[]): Promise<RepoSyncState[]> => {
```
gxueatlassian, line 11: /nit
Up for discussion, I personally find sharing data across multiple tests is not very flexible in long term. I'd prefer "creating the prepare data" right within the test, and even better, right within the test function `it`. 
One big problem I have with sharing data is whenever I try to change something in the data, I may broken another test even though I am not touching that bit. So the data become very carefully constructed and a bit fragile across the tests. Are these functions expected to be used by other tests cases?

ps: @bgvozdev previously created database-state-creator.ts to create some subscriptions/repos, not sure those can/should be used in your above test cases.

src/routes/jira/workspaces/jira-workspaces-get.ts
rachellerathbone: If we use that we get TS errors when trying to deconstruct `const { subscriptionId, repoOwner } = org;` So we either go with something like this or use a non-null assertion operator `const { subscriptionId, repoOwner } = org!; // Use non-null assertion operator`. I'd prefer to filter out null values as non-null assertions have the potential to lead to runtime errors.

src/routes/jira/workspaces/jira-workspaces-get.ts
```diff
@@ -1,3 +1,62 @@
...
+			const { subscriptionId, repoOwner } = org;
+
+			return {
+				id: subscriptionId.toString(),
+				name: repoOwner,
+				canCreateContainer: false
+			};
+		});
+
+	return matchedOrgs;
```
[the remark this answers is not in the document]
gxueatlassian: Like if we return 10 records with same repoOwner, will the jira UI show 10 items in the drop down?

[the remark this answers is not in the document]
rachellerathbone: That's why we're doing findOne. So if someone types in an org name and there's, let's say 650 repos, we only find the first matching repo and map the org name from that.

test/utils/create-repositories.ts
rachellerathbone: Good call out. Will move them.

rachellerathbone: Moved them into the test file. Had a look at what we already have in database-state-creator.ts but it doesn't really give me the results I want so sticking with these util functions.

src/routes/jira/workspaces/jira-workspaces-get.ts
```diff
@@ -1,3 +1,62 @@
...
+
+const DEFAULT_PAGE_NUMBER = 1; // Current page
+export const DEFAULT_LIMIT = 20; // Number of items per page
+
+const findMatchingOrgs = async (subscriptions: Subscription[], orgName?: string): Promise<Workspace[]>  => {
+	let matchingRepos;
+
+	if (orgName) {
+		matchingRepos = await Promise.all(subscriptions.map(async (subscription: Subscription) => {
+			return await RepoSyncState.findByOrgNameAndSubscriptionId(subscription, orgName);
```
gxueatlassian: /nit
How about `findOneByOrgNameAndSubscriptionId` instead of `findByOrgNameAndSubscriptionId` to be consistent with bellow?

src/routes/jira/workspaces/jira-workspaces-get.ts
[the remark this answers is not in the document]
gxueatlassian: Yup make sense 👍

src/routes/jira/workspaces/jira-workspaces-get.ts
```diff
@@ -1,3 +1,62 @@
...
+		res.status(400).send(Errors.MISSING_SUBSCRIPTION);
+		return;
+	}
+
+	const orgName = req.query?.searchQuery as string;
+	const page = Number(req.query?.page) || DEFAULT_PAGE_NUMBER;
+	const limit = Number(req.query?.limit) || DEFAULT_LIMIT;
+	const matchedOrgs = await findMatchingOrgs(subscriptions, orgName);
+
+	res.status(200).json({ success: true, workspaces: paginatedResponse(page, limit, matchedOrgs) });
```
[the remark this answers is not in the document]
gxueatlassian: Ignore me, actually we don't have that many `orgName`, just many `repo`. So this should not be an issue.

src/routes/jira/workspaces/jira-workspaces-get.ts
```diff
@@ -1,3 +1,56 @@
...
+
+const findMatchingOrgs = async (subscriptions: Subscription[], orgName?: string): Promise<Workspace[]> => {
+	const matchingRepos = await Promise.all(subscriptions.map(async (subscription: Subscription) => {
+		return orgName ?
+			await RepoSyncState.findByOrgNameAndSubscriptionId(subscription, orgName) :
+			await RepoSyncState.findOneFromSubscription(subscription);
+	}));
+
+	const matchedOrgs = matchingRepos
+		.filter((org): org is RepoSyncState => org !== null)
```
gxueatlassian, line 24: /nit
Nothing to do, this is only a reminder to me how does this `org is RepoSyncState =>` work. Never saw it before. Interesting (and nice 👍 )

gxueatlassian (approved):
👍 to me.

PR #2151, opened 2023-05-31
Arc-2116 search connected repos

rachellerathbone:
What's in this PR?
New endpoint /jira/workspaces/repositories/search

Why
To support searchRepositories for generic containers https://hello.atlassian.net/wiki/spaces/OTFS/pages/2581193453/Extending+GitHub+for+Jira+for+generic+containers#searchRepositories

Affected issues
ARC-2116

How has this been tested?
Locally and unit tests (as mentioned in ARC-2114, we will test from Jira at the end of June once the work being carried out by the Incredibles has been completed).

Whats Next?
Once this has been merged I'll crack open ARC-2117 for review.


src/models/reposyncstate.ts
```diff
@@ -208,6 +208,18 @@ export class RepoSyncState extends Model {
 		}));
 	}
 
+	static async findRepositoriesBySubscriptionIdAndRepoName(subscriptionId: number, repoName: string): Promise<RepoSyncState[] | null> {
+		return RepoSyncState.findAll({
+			where: {
+				subscriptionId,
+				repoName: {
+					[Op.iLike]: `%${repoName}%`
```
rachellerathbone, line 216: See the comment I left in this PR explaining why we need to do partial matching https://github.com/atlassian/github-for-jira/pull/2149/files#r1212443079

src/routes/jira/workspaces/repositories/jira-workspaces-repositories-get.ts
```diff
@@ -1,3 +1,108 @@
...
+	}
+
+	if (!repos?.length) {
+		req.log.warn(NO_MATCHING_REPOSITORIES);
+		res.status(400).send(NO_MATCHING_REPOSITORIES);
+		return;
+	}
+
+	const repositories: WorkspaceRepo[] = repos.map((repo) => {
+		const repoId = transformRepositoryId(repo.repoId);
```
rachellerathbone, line 95: Currently using `transformRepositoryId` but it's effectively just returning the repositoryId when base url is passed as in this case. For most customers this won't be an issue but for any customers who have connected both cloud and server / multiple servers we could potentially run into an issue with non-unique repo ids.

We can't return to id from RepoSyncState, or prefix this onto the repoId, as this value changes anytime a customer runs a backfill. We could prefix with the subscriptionId as this value remains unchanged during backfill and if a user connects and disconnects. It will only change if a user removes permissions for the org in GH or uninstalls & reinstalls the app but think this may be our best option. Thoughts?

src/models/reposyncstate.ts
```diff
@@ -218,6 +218,35 @@ export class RepoSyncState extends Model implements RepoSyncStateProperties {
...
+		repoName?: string
+	): Promise<RepoSyncState[] | null> {
+
+		const whereClause: any = {
+			subscriptionId
+		};
+
+		if (repoName) {
+			whereClause.repoName = {
+				[Op.iLike]: `%${repoName}%`
```
gxueatlassian, line 234: btw, should it be case insensitive?

src/routes/jira/workspaces/repositories/jira-workspaces-repositories-get.ts
```diff
@@ -1,3 +1,85 @@
...
+
+	if (!subscriptions.length) {
+		return null;
+	}
+
+	const reposArray: RepoSyncState[][] = await Promise.all(
+		subscriptions.map(async (subscription) => {
+			const subscriptionRepos = await RepoSyncState.findRepositoriesBySubscriptionIdAndRepoName(
+				subscription.id,
+				page,
```
gxueatlassian, line 42: Will the `page` / `limit` can work nicely with the `subscriptions.map`? I suspect they will conflicts with each other.

gxueatlassian: I suspect you might end up needs to do something like 
```sql
select repoName, repoId from RepoSyncState rss 
where exists ( 
  select sub.id from "Subscirptions" sub 
      where sub.id = rss.subscriptionId 
           && sub.jiraHost = :jiraHost 
           && sub.orgName like '%orgName' 
) 
order by XXX asc limit :limit offset :offset
```

src/routes/jira/workspaces/repositories/jira-workspaces-repositories-get.ts
```diff
@@ -1,3 +1,85 @@
...
+	const connectedOrgId = Number(req.query?.workspaceId);
+	const repoName = req.query?.searchQuery as string;
+	const page = Number(req.query?.page) || DEFAULT_PAGE_NUMBER;
+	const limit = Number(req.query?.limit) || DEFAULT_LIMIT;
+
+	const repos = connectedOrgId ?
+		await getReposForWorkspaceId(jiraHost,connectedOrgId, page, limit, repoName) :
+		await getAllRepos(jiraHost, page, limit, repoName);
+
+	if (repos === null) {
```
gxueatlassian, line 66: /nit
For the code 
```
if (repos === null) {
		req.log.warn(Errors.MISSING_SUBSCRIPTION);
```
It is a bit confusing to read that.

I don't have an suggestion on how actually to change it thought.

src/util/paginate-response.ts
```diff
@@ -1,6 +1,7 @@
 import { Workspace } from "../routes/jira/workspaces/jira-workspaces-get";
+import { WorkspaceRepo } from "routes/jira/workspaces/repositories/jira-workspaces-repositories-get";
 
-export const paginatedResponse = (page: number, limit: number, payload: Workspace[]) => {
+export const paginatedResponse = (page: number, limit: number, payload: Workspace[] | WorkspaceRepo[]) => {
```
gxueatlassian, line 4: /nit
time to do some generic here :) 
export const paginatedResponse =`<T>` (page: number, limit: number, payload: T[]) => {

src/routes/jira/workspaces/repositories/jira-workspaces-repositories-get.ts
gxueatlassian: Actually, how about the uniqueness of the workspace? Is it by id or name(str)? 
As for GHES, the orgName can be duplicate from different GHES server, even the workspaceId is not unique.

We have a similar thing when sending data to jira. We then use `TransformedRepositoryId` inside, `return ${calculatePrefix(gitHubBaseUrl)}-${repositoryId} as TransformedRepositoryId;`

I suspect we can do something like this? 

But might need to ask <insert_that_proper_team_name_here> for suggestions.

src/models/reposyncstate.ts
rachellerathbone: It is. In the test "Should return all matching repos across multiple orgs when repoName is provided but workspaceId is no" `new` is passed as a query param in lower case. The response, defined in `createReposWhenRepoNameIsPassedAsParam`, returns 3 repos across 2 separate orgs: `new-repo`, `another-new-rep` and `imNew`

src/routes/jira/workspaces/repositories/jira-workspaces-repositories-get.ts
rachellerathbone: I'll modify the logic a little so it makes more sense.

src/routes/jira/workspaces/repositories/jira-workspaces-repositories-get.ts
```diff
@@ -1,3 +1,80 @@
...
+	);
+
+	return reposArray.flat().slice(0, limit); // Apply limit after flattening the array
+};
+
+export const JiraWorkspacesRepositoriesGet = async (req: Request, res: Response): Promise<void> => {
+	req.log.info({ method: req.method, requestUrl: req.originalUrl }, "Request started to GET repositories");
+
+	const { jiraHost } = res.locals;
+	const connectedOrgId = Number(req.query?.workspaceId);
```
gxueatlassian, line 50: Is this `workspaceId` passed in by the generic container service? Is it the `repoOwnerName` in our `RepoSyncState`?

views/jira-configuration.hbs
```diff
@@ -31,7 +31,6 @@
     <div class="jiraConfiguration">
       <header class="jiraConfiguration__header">
         <h1 class="jiraConfiguration__header__title">GitHub configuration</h1>
-
```
gxueatlassian, line 34: /nit
If you have chance, just revert this file and make it not in the git diff

src/models/reposyncstate.ts
```diff
@@ -218,6 +218,35 @@ export class RepoSyncState extends Model implements RepoSyncStateProperties {
 		});
 	}
 
+	static async findRepositoriesBySubscriptionIdAndRepoName(
```
gxueatlassian, line 221: How about some unit tests for this method?

src/routes/jira/workspaces/repositories/jira-workspaces-repositories-get.ts
```diff
@@ -1,3 +1,80 @@
...
+	id: string,
+	name: string,
+	workspaceId: string
+}
+
+const DEFAULT_PAGE_NUMBER = 1; // Current page
+export const DEFAULT_LIMIT = 20; // Number of items per page
+
+const getReposForWorkspaceId = async (
+	connectedOrgId: number,
```
gxueatlassian, line 16: I assume this `connectedOrgId` is the `subscriptionId`. I'd say name this param as `subscriptionId` make more sense.

src/routes/jira/workspaces/repositories/jira-workspaces-repositories-get.ts
```diff
@@ -1,3 +1,80 @@
...
+	const repos = connectedOrgId ?
+		await getReposForWorkspaceId(connectedOrgId, page, limit, repoName) :
+		await getAllRepos(subscriptions, page, limit, repoName);
+
+	const repositories: WorkspaceRepo[] = repos ? repos.map((repo) => {
+		const { repoId, repoName, subscriptionId } = repo;
+		return {
+			id: repoId.toString(),
+			name: repoName,
+			workspaceId: subscriptionId.toString()
```
gxueatlassian, line 72: /nit
Just curious what happen if user disconnect and reconnect the subscription and `subscriptionId` changed, what happen to the generic container ui? <-- I don't expect you have an answer though as the UI doesn't work yet.

src/routes/jira/workspaces/repositories/jira-workspaces-repositories-get.ts
gxueatlassian, line 50: So this is actually `subscriptionId` right? Shall we name the variable so?

src/routes/jira/workspaces/repositories/jira-workspaces-repositories-get.ts
```diff
@@ -1,3 +1,80 @@
...
+export const JiraWorkspacesRepositoriesGet = async (req: Request, res: Response): Promise<void> => {
+	req.log.info({ method: req.method, requestUrl: req.originalUrl }, "Request started to GET repositories");
+
+	const { jiraHost } = res.locals;
+	const connectedOrgId = Number(req.query?.workspaceId);
+	const repoName = req.query?.searchQuery as string;
+	const page = Number(req.query?.page) || DEFAULT_PAGE_NUMBER;
+	const limit = Number(req.query?.limit) || DEFAULT_LIMIT;
+
+	const subscriptions = await Subscription.getAllForHost(jiraHost);
```
gxueatlassian, line 55: Is there a reason we don't use the `connectedOrgId` here for finding the subscription?

src/routes/jira/workspaces/repositories/jira-workspaces-repositories-get.ts
rachellerathbone: Updated for this. Will need to chat to the incredibles about how they are going to handle the changing workspaceIds

gxueatlassian (approved):

PR #2157, opened 2023-06-02
Arc 2119 refactor backfill deployment

gxueatlassian:
**What's in this PR?**
This PR is refactor our backfill logic for deployments.

1. Save the deployment infos in dynamo db when fetched from GraphQL
2. Look up last success from dynamo first, and fallback to current behaviour if not found.

**Why**
1. Our backfill logic for deployment is incorrect. Each time processing a deployment, it try to call rest api to fetch last 100 success deployment from today. Which causes it to compare itself to "future" deployment. By saving deployments in dynamo during the graphql fetch, we can look it up, and hopefully find the correct "last" success deployment. `NOTE`: This is PR is **NOT** to make the logic 100% correct. As 1) it is hard to guarantee that. Lots of edge cases. 2) This PR however will "significantly", <-- I hope, improve the accuracy of finding a last success deployments.
2. Our backfill processing for deployments is SLOOOOOWWW (unefficient to be specific). It first fetch 100 from graphql, and each of them, will call a rest api to 100 latest deployments of that env, which in turns each call status api to find a success one. So the time complexity of this logic if O(n^3) 😱 By using dynamo, I will say most of the time we can find "last successful deployment" in the dynamodb. But this need to be verified in the metrics.

**Added feature flags**
use-dynamodb-for-deployment-backfill

**Affected issues**  
ARC-2119

**How has this been tested?**  
Unit tested
stg tested

**Whats Next?**
Depends on the metrics and how far we want to go, we can:
1. Refactor the process furthur more to split backfill deployment into separate two stage. Stage 1 scan all deployment in whole history (and save in dynamo), stage 2 process each of them. <----- however this "a little bit" conflicting with our incremental/stagger backfill. Need more thinking.
2. Right now in this PR app fetch an extra page to try to improve efficiency. But the cursor of this extra page is not saved in RepoSyncStates(deploymentCursor) <--- we can make it cursor 1) json 2) compatible with existing 3) record both the scaning/extra fetch and the current processing cursor. <------- however, this seems much more effort, and I'd like to have this PR merged in first, look at the metrics and make decision later.

TLDR; Thank you for reading to end. Such a loooon desc.

### DIFF
--- a/src/github/client/github-installation-client.ts
+++ b/src/github/client/github-installation-client.ts
@@ -17,6 +17,7 @@ import {
 	ViewerRepositoryCountQuery,
 	getDeploymentsResponse,
 	getDeploymentsQuery,
+	getDeploymentsQueryWithStatuses,
 	SearchedRepositoriesResponse, getBranchesQueryWithoutCommits
 } from "./github-queries";
 import {
@@ -355,8 +356,12 @@ export class GitHubInstallationClient extends GitHubClient {
 		return response?.data?.data;
 	}
 
-	public async getDeploymentsPage(owner: string, repoName: string, perPage?: number, cursor?: string | number): Promise<getDeploymentsResponse> {
-		const response = await this.graphql<getDeploymentsResponse>(getDeploymentsQuery,
+	public async getDeploymentsPage(jiraHost: string, owner: string, repoName: string, perPage?: number, cursor?: string | number): Promise<getDeploymentsResponse> {
+
+		const useDyanmoForBackfill = await booleanFlag(BooleanFlags.USE_DYNAMODB_FOR_DEPLOYMENT_BACKFILL, jiraHost);
+		const graphQuery = useDyanmoForBackfill ? getDeploymentsQueryWithStatuses : getDeploymentsQuery;
+
+		const response = await this.graphql<getDeploymentsResponse>(graphQuery,
 			await this.installationAuthenticationHeaders(),
 			{
 				owner,
--- a/src/github/client/github-installation-client.ts
+++ b/src/github/client/github-installation-client.ts
@@ -33,6 +33,13 @@ import { cloneDeep } from "lodash";
 import { BooleanFlags, booleanFlag } from "config/feature-flags";
 import { runCurl } from "utils/curl/curl-utils";
 
+// Unfortunately, the type is not exposed in Octokit...
+// https://docs.github.com/en/rest/pulls/review-requests?apiVersion=2022-11-28#get-all-requested-reviewers-for-a-pull-request
+export type PullRequestedReviewersResponse = {
+	users: Array<Octokit.PullsUpdateResponseRequestedReviewersItem>,
+	teams: Array<Octokit.PullsUpdateResponseRequestedTeamsItem>,
+}
+
 /**
  * A GitHub client that supports authentication as a GitHub app.
  * API is specific to an organization (e.g. can get all repos for an org)
@@ -88,6 +95,14 @@ export class GitHubInstallationClient extends GitHubClient {
 	 */
 	public async getPullRequestReviews(owner: string, repo: string, pullNumber: string | number): Promise<AxiosResponse<Octokit.PullsListReviewsResponse>> {
 		return await this.get<Octokit.PullsListReviewsResponse>(`/repos/{owner}/{repo}/pulls/{pullNumber}/reviews`, {}, {
+			owner,
+			repo,
+			pullNumber
+		});
+	}
+
+	public async getPullRequestRequestedReviews(owner: string, repo: string, pullNumber: string | number): Promise<AxiosResponse<PullRequestedReviewersResponse>> {
+		return await this.get<PullRequestedReviewersResponse>(`/repos/{owner}/{repo}/pulls/{pullNumber}/requested_reviewers`, {}, {
 			owner,
 			repo,
 			pullNumber
--- a/src/sync/deployment.test.ts
+++ b/src/sync/deployment.test.ts
@@ -63,6 +63,7 @@ describe("sync/deployments", () => {
 		const createDeploymentsEntities = (size: number) => {
 			return "*".repeat(size).split("").map((_, idx)=> {
 				const clone = JSON.parse(JSON.stringify(deploymentNodesFixture.data.repository.deployments.edges[0]));
+				clone._seq = idx + 1;
 				clone.cursor = `cursor:${idx + 1}`;
 				clone.node.createdAt = `2023-01-0${idx + 1}T10:00:00Z`;
 				clone.node.databaseId = `dbid-${idx + 1}`;
@@ -141,6 +142,18 @@ describe("sync/deployments", () => {
 				expect(result.Item).toEqual({ CommitSha: { "S": deployment.node.commitOid } });
 			}
 		};
+		const expectEdgesAndPayloadMatchToDeploymentCommits = (result, deployments) => {
+			expect(result).toEqual({
+				edges: deployments.map(d =>
+					expect.objectContaining({ cursor: d.cursor, node: expect.objectContaining({ commitOid: d.node.commitOid }) })
+				),
+				jiraPayload: {
+					deployments: deployments.map(d => expect.objectContaining({
+						associations: [{ "associationType": "issueIdOrKeys", values: [ `JIRA-${d._seq}`] }] })
+					)
+				}
+			});
+		};
 		/*
 		const nockCommitShaCompare = (sha1, sha2) => {
 			const compareMsgIssueKey = `ISSUEKEY${sha1}${sha2}-100`;
@@ -169,7 +182,8 @@ describe("sync/deployments", () => {
 				when(booleanFlag).calledWith(BooleanFlags.USE_DYNAMODB_FOR_DEPLOYMENT_BACKFILL, jiraHost).mockResolvedValue(true);
 			});
 			describe("for empty demployment cursor", () => {
-				it.only("should  fetch deployments from begining and return correct cursor", async () => {
+				// eslint-disable-next-line jest/expect-expect
+				it("should fetch deployments from begining", async () => {
 					const deployments = createDeploymentsEntities(4);
 					nockFetchingDeploymentgPagesGraphQL(DEPLOYMENT_CURSOR_EMPTY, [deployments[3], deployments[2]]);
 					nockDeploymentListingApi(deployments, REPEAT_ONCE);
@@ -178,28 +192,16 @@ describe("sync/deployments", () => {
 					const result = await getDeploymentTask(logger, gitHubClient, jiraHost, repoFromRepoSyncState(repoSyncState), DEPLOYMENT_CURSOR_EMPTY, PAGE_SIZE__TWO_ITEMS, msgPayload());
 
 					await expectDeploymentEntryInDB([deployments[3], deployments[2]]);
-					expect(result).toEqual({
-						edges: [
-							expect.objectContaining({ cursor: "cursor:4", node: expect.objectContaining({ commitOid: "SHA4" }) }),
-							expect.objectContaining({ cursor: "cursor:3", node: expect.objectContaining({ commitOid: "SHA3" }) })
-						],
-						jiraPayload: {
-							deployments: [
-								expect.objectContaining({ associations: [{ "associationType": "issueIdOrKeys", values: [ "JIRA-4"] }] }),
-								expect.objectContaining({ associations: [{ "associationType": "issueIdOrKeys", values: [ "JIRA-3"] }] })
-							]
-						}
-					});
+					expectEdgesAndPayloadMatchToDeploymentCommits(result, [deployments[3], deployments[2]]);
 				});
 			});
-			describe("for existing legacy (string) deployment cursor", () => {
-			});
-			describe("for new (json-object) deployment cursor", () => {
+			describe("for existing deployment cursor", () => {
 			});
 		});
 		describe("when ff is off", () => {
 			const REPEAT_LOTS_OF_TIME = 20;
 			describe("for empty demployment cursor", () => {
+				// eslint-disable-next-line jest/expect-expect
 				it("should fetch deployments from begining", async () => {
 
 					const deployments = createDeploymentsEntities(4);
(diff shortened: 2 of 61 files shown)


src/sync/deployment.test.ts
```diff
@@ -30,6 +37,187 @@ describe("sync/deployments", () => {
...
+						.reply(200, {
+							commit: {
+								author: { name: "random", email: "random", date: new Date() },
+								message: `commit message [JIRA-${item._seq}]`
+							}, html_url: "random"
+						});
+				});
+			});
+		};
+		const expectDeploymentEntryInDB = async (deployments) => {
```
rachellerathbone, line 130: nit: can we add a new line between each variable declaration and each test? There's a lot of code here and the readability really drops when everything is clumped together without any spaces in between.

gxueatlassian: Have add spaces in between, shuffle them a round, put some functions inline.
Hopefully it can be easier to read.

src/sync/deployment.test.ts
```diff
@@ -30,6 +37,221 @@ describe("sync/deployments", () => {
...
+
+					expectEdgesAndPayloadMatchToDeploymentCommits(result, [deployments[3], deployments[2]]);
+				});
+			});
+
+			describe("for existing deployment cursor", () => {
+				// eslint-disable-next-line jest/expect-expect
+				it("should fetch deployments from existing cursor", async () => {
+
+					const deployments = createDeploymentEntities(4);
```
rachellerathbone, line 128: Can we convert the 'magic number' of 4 to a variable so it's more explicit and gives more context? e.g.

```
 const deploymentCount = 4;
  const deployments = createDeploymentEntities(deploymentCount);
```

src/sync/deployment.test.ts
```diff
@@ -30,6 +37,221 @@ describe("sync/deployments", () => {
...
+								message: `commit message [JIRA-${item._seq}]`
+							}, html_url: "random"
+						});
+				});
+			});
+		};
+
+		const expectDeploymentEntryInDB = async (deployments) => {
+			for (const deployment of deployments) {
+				const key = createHashWithoutSharedSecret(`ghurl_${gitHubCloudConfig.baseUrl}_repo_${deployment.node.repository.id}_env_${deployment.node.environment}`);
```
rachellerathbone, line 226: To tighten this up a little in terms of readability can we throw in some destructuring at the start

```
const {
    repository: { id: repoId },
    environment,
    commitOid
  } = deployment.node;
```

src/sync/deployment.test.ts
```diff
@@ -30,6 +37,221 @@ describe("sync/deployments", () => {
...
+				const successStatusDate = deployment.node.statuses.nodes.find(n=>n.state === "SUCCESS")?.updatedAt;
+				const result = await ddb.getItem({
+					TableName: envVars.DYNAMO_DEPLOYMENT_HISTORY_CACHE_TABLE_NAME,
+					Key: {
+						"Id": { "S": key },
+						"CreatedAt": { "N": String(new Date(successStatusDate).getTime()) }
```
rachellerathbone, line 232: Could probably move this out into a variable as well? `const createdAtTimestamp = successStatusDate && new Date(successStatusDate).getTime();` Then `"CreatedAt": { "N": String(createdAtTimestamp) }`?

src/sync/deployment.ts
```diff
@@ -49,12 +72,39 @@ const getTransformedDeployments = async (deployments, gitHubInstallationClient:
...
-		.filter(deployment => !!deployment)
+	return (transformedDeployments
+		.filter(deployment => !!deployment) as JiraDeploymentBulkSubmitData[])
 		.map(deployment => deployment.deployments)
 		.flat();
 };
 
+const saveDeploymentsForLaterUse = async (deployments: FetchDeploymentResponse["deployments"], gitHubBaseUrl: string, logger: Logger) => {
+	try {
+		const successDeployments: FetchDeploymentResponse["deployments"] = deployments.filter(d => (d.statuses?.nodes.some(n => n.state === "SUCCESS")));
```
rachellerathbone, line 83: nit: same thing here with spacing. There's a lot of lines of code here and it gets really hard to separate the logic when it's all crammed together.

src/sync/deployment.ts
```diff
@@ -49,12 +72,39 @@ const getTransformedDeployments = async (deployments, gitHubInstallationClient:
...
+		logger.info({ deploymentsCount: deployments.length, successDeploymentsCount: successDeployments.length }, "Try to save deployments for later use");
+		const result = await Promise.allSettled(successDeployments.map(dep => {
+			const successStatusDate = dep.statuses?.nodes.find(n=>n.state === "SUCCESS")?.updatedAt;
+			if (!successStatusDate) {
+				logger.warn("Should find a success status date, but found none");
+				throw "Save failure";
```
rachellerathbone, line 89: Can we change this to throw new Error("Save failure")? It's generally better to throw an instance of the Error object rather than throwing a simple string as the Error object provides more info about the error. This will provide more context about what went wrong. Also, the error object will capture the stack trace and we can leverage it for better error handling.

src/sync/deployment.test.ts
```diff
@@ -30,6 +37,221 @@ describe("sync/deployments", () => {
...
+			});
+		};
+
+		const nockFetchingDeploymentgPagesGraphQL = (query, cursor, deployments) => {
+			githubNock.post("/graphql", { query, variables: { owner: repoSyncState.repoOwner, repo: repoSyncState.repoName, per_page: PAGE_SIZE__TWO_ITEMS, cursor } })
+				.query(true).reply(200, { "data": { "repository": { "deployments": { "edges": deployments } } } });
+		};
+
+		const nockDeploymentListingApi = (deployments, repeatTimes) => {
+			"*".repeat(repeatTimes).split("").forEach(() => {
```
rachellerathbone, line 176: Can we change `"*".repeat(repeatTimes).split("")` to `Array(repeatTimes).fill()` as the latter is more performant. The .fill() method will create an array directly with the desired length, whereas `repeat()` followed by `split()` involves additional string operations that can be slower. Might not matter too much coz this is just a test but maybe worth considering ¯\_(ツ)_/¯

src/sync/deployment.test.ts
```diff
@@ -30,6 +37,221 @@ describe("sync/deployments", () => {
...
+				githubNock.get(`/repos/test-repo-owner/test-repo-name/deployments?environment=prod&per_page=10`)
+					.reply(200, deployments.map((item, idx) => ({
+						"id": item.node.databaseId,
```
rachellerathbone, line 179: Can we remove the unnecessary quotations marks around the object property names? These are all valid identifiers and should keep this consistent with the rest of our codebase.

src/sync/deployment.test.ts
rachellerathbone: Same with the quotation marks too.

src/sync/deployment.test.ts
gxueatlassian: fixed.

src/sync/deployment.test.ts
gxueatlassian: fixed (to my best effort 🤷 )

src/sync/deployment.test.ts
gxueatlassian: fixed.

src/sync/deployment.ts
gxueatlassian: fixed (to my best effort 🤷 )

src/sync/deployment.ts
gxueatlassian: Good call!!
Fixed.

src/sync/deployment.test.ts
gxueatlassian: Thanks for picking up this. I tried that no luck though. Seems the `Array(number)` create an something empty (although with `length` equal to that number), so when do `Array(number).forEach()` it won't actually loop as it consider it is still an empty array.

Leave this unchanged.

src/sync/deployment.test.ts
gxueatlassian: Fixed.

src/sync/deployment.ts
rachellerathbone: Love your work sir.

src/sync/deployment.test.ts
```diff
@@ -30,6 +37,225 @@ describe("sync/deployments", () => {
...
+
+		describe("when ff is off", () => {
+
+			beforeEach(() => {
+				when(booleanFlag).calledWith(BooleanFlags.USE_DYNAMODB_FOR_DEPLOYMENT_BACKFILL, jiraHost).mockResolvedValue(false);
+			});
+
+			describe("for empty demployment cursor", () => {
+				// eslint-disable-next-line jest/expect-expect
+				it("should fetch deployments from begining", async () => {
```
rachellerathbone, line 107: nit: typo -> beginning

src/sync/deployment.test.ts
rachellerathbone: actually, `Array.from({ length: repeatTimes }).forEach(() => {` should do the trick

src/sync/deployment.test.ts
```diff
@@ -30,6 +37,221 @@ describe("sync/deployments", () => {
...
+		//--helper funcs--
+		const msgPayload = () => ({
+			jiraHost,
+			installationId: DatabaseStateCreator.GITHUB_INSTALLATION_ID,
+			commitsFromDate: "2023-01-01T00:00:00Z"
+		});
+
+		//size up to 9 entities
+		const createDeploymentEntities = (size: number) => {
+			return "*".repeat(size).split("").map((_, idx)=> {
```
rachellerathbone: We can improve performance here too: `return Array.from({ length: size }, (_, idx) => {`

src/sync/deployment.test.ts
```diff
@@ -30,6 +37,225 @@ describe("sync/deployments", () => {
...
+					.reply(200, deployments.map((item, idx) => ({
+						id: item.node.databaseId,
+						sha: item.node.commitOid,
+						ref: "random",
+						task: `task for deployment ${idx + 1}`,
+						payload: {},
+						original_environment: item.node.environment,
+						environment: item.node.environment,
+						description: `description for deployment ${idx + 1}`,
+						creato: { login: "test-repo-owner", id: 1, type: "User" },
```
rachellerathbone, line 190: a couple of typos after removing the quotes:

- creato -> creator
- production_environmen -> production_environment

src/sync/deployment.test.ts
```diff
@@ -30,6 +37,225 @@ describe("sync/deployments", () => {
...
+					n.createdAt = clone.node.createdAt;
+					n.updatedAt = clone.node.updatedAt;
+				});
+				return clone;
+			});
+		};
+
+		const nockFetchingDeploymentgPagesGraphQL = (query, cursor, deployments) => {
+			githubNock.post("/graphql", { query, variables: { owner: repoSyncState.repoOwner, repo: repoSyncState.repoName, per_page: PAGE_SIZE__TWO_ITEMS, cursor } })
+				.query(true).reply(200, { "data": { "repository": { "deployments": { "edges": deployments } } } });
```
rachellerathbone, line 175: quotations -> :killitwithfire:

src/sync/deployment.test.ts
```diff
@@ -30,6 +37,225 @@ describe("sync/deployments", () => {
...
+						statuses_url: "random",
+						repository_url: "random",
+						transient_environment: false,
+						production_environmen: true
+					})));
+			});
+		};
+
+		const nockDeploymentStatusApi = (deployments, repeatTimes) => {
+			"*".repeat(repeatTimes).split("").forEach(() => {
```
rachellerathbone, line 202: performance improvement: Array.from({ length: repeatTimes }).forEach(() => {

src/sync/deployment.test.ts
```diff
@@ -30,6 +37,225 @@ describe("sync/deployments", () => {
...
+						.reply(200, [
+							{ id: idx * 1000 + 1, state: "inactive", description: "random", environment: item.node.environment },
+							{ id: idx * 1000 + 2, state: "success", description: "random", environment: item.node.environment }
+						]);
+				});
+			});
+		};
+
+		const nockDeploymentCommitGetApi = (deployments, repeatTimes) => {
+			"*".repeat(repeatTimes).split("").forEach(() => {
```
rachellerathbone, line 214: Samesies: Array.from({ length: repeatTimes }).forEach(() => {

src/sync/deployment.test.ts
gxueatlassian: Nice one!! I wasn't aware of this `Array.from`. Was only looking at the `Array()` constructor. TIL. 👍

rachellerathbone (approved):

rachellerathbone:
Looking forward to this going in. Hopefully we see some improvements in deployments :blobpray:

src/sync/deployment.ts
```diff
@@ -49,12 +72,48 @@ const getTransformedDeployments = async (deployments, gitHubInstallationClient:
...
+	return (transformedDeployments
+		.filter(deployment => !!deployment) as JiraDeploymentBulkSubmitData[])
 		.map(deployment => deployment.deployments)
 		.flat();
 };
 
+const saveDeploymentsForLaterUse = async (deployments: FetchDeploymentResponse["deployments"], gitHubBaseUrl: string, logger: Logger) => {
+	try {
+
+		const successDeployments: FetchDeploymentResponse["deployments"] = deployments.filter(d => (d.statuses?.nodes.some(n => n.state === "SUCCESS")));
```
joshkay10, line 84: /nit Your can drop the type for `const successDeployments: FetchDeploymentResponse["deployments"]` since its inferred from the filter response.

src/sync/deployment.ts
```diff
@@ -66,7 +125,12 @@ export const getDeploymentTask = async (
 ) => {
 	logger.debug("Syncing Deployments: started");
 
-	const { edges, deployments } = await fetchDeployments(gitHubInstallationClient, repository, cursor, perPage);
+	const { edges, deployments, extraDeployments } = await fetchDeployments(jiraHost, gitHubInstallationClient, repository, logger, cursor, perPage);
+
+	const useDyanmoForBackfill = await booleanFlag(BooleanFlags.USE_DYNAMODB_FOR_DEPLOYMENT_BACKFILL, jiraHost);
+	if (useDyanmoForBackfill) {
```
joshkay10, line 131: `useDyanmoForBackfill` to `useDynamoForBackfill`

src/github/client/github-installation-client.ts
```diff
@@ -355,8 +356,12 @@ export class GitHubInstallationClient extends GitHubClient {
 		return response?.data?.data;
 	}
 
-	public async getDeploymentsPage(owner: string, repoName: string, perPage?: number, cursor?: string | number): Promise<getDeploymentsResponse> {
-		const response = await this.graphql<getDeploymentsResponse>(getDeploymentsQuery,
+	public async getDeploymentsPage(jiraHost: string, owner: string, repoName: string, perPage?: number, cursor?: string | number): Promise<getDeploymentsResponse> {
```
joshkay10, line 359: Would the cursor always be a string in this case?

src/sync/deployment.ts
gxueatlassian: Good call, it must be my glitch on my vim ts settigs that doesn't show up correctly, so I added the extra types.
Sorry since this is already deployed, I am leaving this as it is right now. I may remember to remove it in next PR.

src/sync/deployment.ts
gxueatlassian: 😞 crap. I'll leave this right now, hopefully the ff will be removed once 100% and this var will be removed in the followup PR.

src/github/client/github-installation-client.ts
gxueatlassian: Yup I think so, 
1. The typing I believe it is because we use rest api and was page no before? So typing to be backward compatibility?
2. And can confirm all deploymentCursor is string (not number as string) in prod db. I'll remember to remove this on ff cleanup.

PR #1736, opened 2022-11-10
Provide an url when possible on deployment environment section

benoittgt:
We want to be able to easily access to an environment when possible from the deployment page.

Capdture_d’écran_2022-11-10_à_15_11_10

It needs to be provided as environment_url when making deployment status call. https://docs.github.com/en/rest/deployments/statuses#create-a-deployment-status

For people who use the classic Github Action to mark deployment as suggested by Jira. It's here:

https://github.com/chrnorm/deployment-action/blob/880668c419e283adc6b1cc092d52eb5e558fec4d/action.yml#L31-L33

Related:
  - https://github.com/atlassian/github-for-jira/issues/1536
  - https://github.com/atlassian/github-for-jira/pull/1732

PR #2162, opened 2023-06-06
Provide environment URL when possible

aldenquimby:
**What's in this PR?**
- revamped version of #2132 following #1736 and fixes it's merge conflicts

**Why**
- Goal is to allow users to quickly navigate to a deployed environment:
image

**Added feature flags**
- no

**Affected issues**  
- none

**How has this been tested?**  
- updated unit tests

**Whats Next?**
- nothing


src/github/client/github-queries.ts
```diff
@@ -469,15 +470,16 @@ export const getDeploymentsQueryWithStatuses = `query ($owner: String!, $repo: S
               createdAt
               updatedAt
               state
-							logUrl
+              logUrl
+              environmentUrl
```
aldenquimby, line 474: This is the only GraphQL changes. Others were my editor auto-fixing spacing of new fields to be consistent with others. I can revert that to clean the diff here if you'd prefer

src/transforms/transform-deployment.ts
```diff
@@ -387,6 +387,7 @@ export const transformDeployment = async (
 			environment: {
 				id: deployment_status.environment,
 				displayName: deployment_status.environment,
+				url: deployment_status.environment_url,
```
gxueatlassian, line 390: Unfortunately, according to this public doc https://developer.atlassian.com/cloud/jira/software/rest/api-group-deployments/#api-rest-deployments-0-1-bulk-post
The `url` is not supported in the payload, and confirmed with internal team, the environment tab is not clickable at the moment.

And this is new to features to the ui team, so they don't have a roadmap on making it clickable so far, no ETA yet.

If you like, can you please submit a feature request to Jira software?

Thanks.

gxueatlassian:
Hi @aldenquimby 
Thank you for your commits, unfortunately this is not possible at the moment. I'll close this one for now. Happy to re-open it once the clickable url in ui is possible in the future.

aldenquimby:
Ok thank you @gxueatlassian . In that case could you take a look at #2132 instead?

MainasuK:
Inspired by #2271. There is a workaround before the Jira update. Due to default link for deployment is the `log_url`. So we just replace the `log_url` to `environment_url` or anything else. 

I test deploy the build with date as tag name. And set the URL for the link to tag as `log_url`. You can open the "Log" in "…" menu but redirect to the link for tag. Also, click the link in the jira deployment not open the log but the tag link I set. 

Hope this helps.

PR #2163, opened 2023-06-07
ARC-2196 fix incorrect reviewer state

rachellerathbone:
**What's in this PR?**
A bugfix 

**Why**
When a user changed their review state e.g. approved -> changes requested we were sending the incorrect state to Jira. See [ARC-2196](https://softwareteams.atlassian.net/jira/software/projects/ARC/boards/385/backlog?selectedIssue=ARC-2196) for more details

**Affected issues**  
ARC-2196

**How has this been tested?**  
Added unit test, locally, staging

In the video below, you'll see I'm reviewing a PR from 2 separate accounts:

- rachelletestsgithub: initially in a state of APPROVED but then requests changes and state changes to UNAPPROVED
- atlassian-test-account: initially in state of UNAPPROVED but then approves and state changes to APPROVED

**Whats Next?**
Continue avoiding GC? Maybe watch some more YouTube videos?

[ARC-2196]: https://softwareteams.atlassian.net/browse/ARC-2196


src/transforms/transform-pull-request.ts
```diff
@@ -52,8 +53,8 @@ const mapReviews = async (
 
 			acc.push(usernames[reviewerUsername]);
 
-		} else if (usernames[reviewerUsername].approvalStatus !== STATE_APPROVED && review.state === STATE_APPROVED) {
-			usernames[reviewerUsername].approvalStatus = STATE_APPROVED;
+		} else {
+			usernames[reviewerUsername].approvalStatus = review.state === STATE_APPROVED ? STATE_APPROVED : STATE_UNAPPROVED;
```
rachellerathbone, line 57: The else if logic we had before was basically saying if we'd previously seen your review, we don't give a damn if the state has changed... PR APPROVED!

src/transforms/transform-pull-request.ts
```diff
@@ -87,44 +88,60 @@ export const transformPullRequest = async (
...
+		comments,
+		base,
+		number,
+		updated_at,
+		head,
+		state,
+		merged_at,
+		title,
+		html_url
+	} = pullRequest;
```
rachellerathbone, line 104: We use destructuring in most of our codebase and was agreed upon in our former coding conventions sessions: "Preference for destructuring as it’s clearer and part of ES6. Should try and use the standards set by ES6, as most of our app is using it." That said we agreed not to use if object was nested so only pulled keys from the first layer. Overall though, I'd say it improves the payload readability.

src/transforms/transform-pull-request.ts
```diff
@@ -30,7 +30,7 @@ const mapReviews = async (
 	gitHubInstallationClient: GitHubInstallationClient
 ): Promise<JiraReview[]> =>
 {
-	const sortedReviews = orderBy(reviews, "submitted_at", "desc");
+	const sortedReviews = orderBy(reviews, "submitted_at", "asc");
```
rachellerathbone, line 33: Basically there were 2 issues. When GitHub sends all the review data we were essentially reversing the order so we were acknowledging the first review state for a reviewer, not the latest.

src/transforms/transform-pull-request.ts
```diff
@@ -30,13 +30,12 @@ const mapReviews = async (
 	gitHubInstallationClient: GitHubInstallationClient
 ): Promise<JiraReview[]> =>
 {
-	const sortedReviews = orderBy(reviews, "submitted_at", "desc");
```
rachellerathbone, line 33: Turns out we don't want to sort the reviews. When GH sends us the reviews they're sending them in the order we want. By reversing them, we wind up sending the original review instead of the most recent.

src/transforms/transform-pull-request.ts
```diff
@@ -52,8 +50,8 @@ const mapReviews = async (
 
 			acc.push(usernames[reviewerUsername]);
 
-		} else if (usernames[reviewerUsername].approvalStatus !== STATE_APPROVED && review.state === STATE_APPROVED) {
-			usernames[reviewerUsername].approvalStatus = STATE_APPROVED;
+		} else {
+			usernames[reviewerUsername].approvalStatus = review.state === STATE_APPROVED ? STATE_APPROVED : STATE_UNAPPROVED;
```
rachellerathbone, line 54: If we've already seen a reviewer, all we want to do is make sure approval status is updated to their most recent review.state

src/transforms/transform-pull-request.ts
```diff
@@ -87,40 +85,57 @@ export const transformPullRequest = async (
 	log?: Logger
 ) =>
 {
-	const { head } = pullRequest;
+	const {
+		id,
+		user,
+		comments,
+		base,
+		number,
```
gxueatlassian, line 93: /nit
`number` looks a bit confusing. Could be `pullRequestNumber` <-- But that draw the question, is this `destructing` really help the readability?
For me when I saw the code bellow, i am confused.
```javascript
pullRequests: [
			{
				...
				id: number,
        ...
      }
]
```
I'd rather see
```javascript
pullRequests: [
			{
				...
				id: pullRequest.number,
        ...
      }
]
```

gxueatlassian: But `/nit`, I understand there's a code convention and consistent sake. So just random thing at 5pm Wed night. Nothing super important here.

src/transforms/transform-pull-request.ts
gxueatlassian: Otherwise reader might wonder why we don't do `asc` sorting instead.

gxueatlassian (approved):

gxueatlassian (approved):

src/transforms/transform-pull-request.ts
rachellerathbone: The definition of chronological order is: "Chronological order refers to the arrangements of events according to the time they occurred (i.e., from first to last)." In the case of reviews, this means GH sends us the reviews starting from the first through to the last. So it's then the last review that we check for each reviewer and ship off to Jira.

Previously, GH was sending these from first to last but then we were flipping them around from last to first. This meant we mapped over all the reviews and sent Jira the review state of the **first** review for each reviewer. 

We don't need to explicitly sort by ascending because that's how the data is already sent to us from GH.

src/transforms/transform-pull-request.ts
rachellerathbone: Before we were basically checking if we had already seen a reviewer before or not. If not, we created a usernames object. Otherwise, if we had seen a reviewer before, we only updated the review state if the approvalStatus was UNAPPROVED and had since changed to APPROVED. If it was the opposite e.g. first reviewer APPROVED (state stored to approvalStatus) but then requested changes (UNAPPROVED) we did... nothing. Which means the approvalStatus remained unchanged.

Essentially, what we want to do is:

- check if we've seen the reviewer before: if not, create the usernames object and store APPROVED/UNAPPROVED in approvalStatus
- otherwise, update approvalStatus to the most recent review.state

src/transforms/transform-pull-request.ts
```diff
@@ -87,40 +85,57 @@ export const transformPullRequest = async (
...
-				lastUpdate: pullRequest.updated_at,
-				reviewers: await mapReviews(reviews, gitHubInstallationClient),
-				sourceBranch: pullRequest.head.ref || "",
-				sourceBranchUrl: `${pullRequest.head.repo.html_url}/tree/${pullRequest.head.ref}`,
-				status: mapStatus(pullRequest.state, pullRequest.merged_at),
-				timestamp: pullRequest.updated_at,
-				title: pullRequest.title,
-				url: pullRequest.html_url,
+				lastUpdate: updated_at,
+				reviewers: reviewers.reverse(),
```
rachellerathbone, line 132: By reversing the data here instead of in the mapping function, our reviewers still render as they do today (most recent first) whilst also sending the correct approval state.

gxueatlassian (approved):

PR #2166, opened 2023-06-08
NONE: allow OAuth tail to finish for non-jira-admins

bgvozdev:
**What's in this PR?**
- Making OAuth flow "tail" to work without jwt & github middlewares

**Why**
- To implement org "approval" flow, we must allow non-Jira admins to get the user token.

**Added feature flags**
Nah

**Affected issues**  

**How has this been tested?**  
- unit tests
- locally for cloud
- locally for server (still in progress)

**Whats Next?**
- Implement the flow itself

### DIFF
--- a/src/models/github-server-app.ts
+++ b/src/models/github-server-app.ts
@@ -43,7 +43,7 @@ type SECRETE_FIELD = "gitHubClientSecret" | "webhookSecret" | "privateKey" | "en
 export class GitHubServerApp extends Model {
 	id: number;
 	uuid: string;
-	appId: number;
+	appId: number; // TODO: rename to gitHubAppId
 	gitHubBaseUrl: string;
 	gitHubClientId: string;
 	gitHubClientSecret: string; // TODO: rename column to prefix with "encrypted"
@@ -107,16 +107,16 @@ export class GitHubServerApp extends Model {
 		return { jiraHost };
 	}
 
-	static async getForGitHubServerAppId(
-		gitHubServerAppId: number
+	static async getByIdPk(
+		idPk: number
 	): Promise<GitHubServerApp | null> {
-		if (!gitHubServerAppId) {
+		if (!idPk) {
 			return null;
 		}
 
 		return this.findOne({
 			where: {
-				id: gitHubServerAppId
+				id: idPk
 			}
 		});
 	}
--- a/src/routes/github/github-oauth.ts
+++ b/src/routes/github/github-oauth.ts
@@ -16,8 +16,7 @@ import { Installation } from "models/installation";
 const logger = getLogger("github-oauth");
 const appUrl = envVars.APP_URL;
 export const OAUTH_CALLBACK_SUBPATH = "/callback";
-const callbackPathCloud = `/github${OAUTH_CALLBACK_SUBPATH}`;
-const callbackPathServer = `/github/<uuid>${OAUTH_CALLBACK_SUBPATH}`;
+const fullCallbackPath = `/github${OAUTH_CALLBACK_SUBPATH}`;
 
 interface OAuthState {
 	postLoginRedirectUrl: string;
@@ -27,17 +26,11 @@ interface OAuthState {
 }
 
 const getRedirectUrl = async (res: Response, state: string) => {
-	// TODO: revert this logic and calculate redirect URL from req once create branch supports JWT and GitHubAuthMiddleware is a router-level middleware again
-	let callbackPath = callbackPathCloud;
-	if (res.locals?.gitHubAppConfig?.uuid) {
-		callbackPath = callbackPathServer.replace("<uuid>", res.locals.gitHubAppConfig.uuid);
-	}
-
 	const definedScopes = await stringFlag(StringFlags.GITHUB_SCOPES, "user,repo", res.locals.jiraHost);
 	const scopes = definedScopes.split(",");
 
 	const { hostname, clientId } = res.locals.gitHubAppConfig;
-	const callbackURI = `${appUrl}${callbackPath}`;
+	const callbackURI = `${appUrl}${fullCallbackPath}`;
 	return `${hostname}/login/oauth/authorize?client_id=${clientId}&scope=${encodeURIComponent(scopes.join(" "))}&redirect_uri=${encodeURIComponent(callbackURI)}&state=${state}`;
 };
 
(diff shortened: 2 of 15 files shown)


src/models/github-server-app.ts
```diff
@@ -107,16 +107,16 @@ export class GitHubServerApp extends Model {
 		return { jiraHost };
 	}
 
-	static async getForGitHubServerAppId(
-		gitHubServerAppId: number
+	static async getByIdPk(
```
gxueatlassian, line 110: /nit
Isn't this provided by the sequelize already?

src/routes/github/github-router.ts
```diff
@@ -27,6 +31,14 @@ subRouter.post("/webhooks",
 	returnOnValidationError,
 	WebhookReceiverPost);
 
+// We want to restrict the tail of OAuth flow to the same scope as the starting point had (where GitHubOAuthMiddleware
+// fired), therefore we are not including neither jira*middlewares nor github*middlewares. GithubOAuthCallbackGet will
+// get all the data needed from session to obtain the token and then redirect to the original URL with all the
+// restrictions that are in place there.
+// We don't want to artificially limit ourselves by including those middlewares, because there are scenarios when
+// OAuth flow was triggered outside of Jira admin scope (e.g. create-branch, or approve-connection).
+subRouter.use(OAUTH_CALLBACK_SUBPATH, GithubOAuthCallbackGet);
```
gxueatlassian, line 40: We can move callback router to after the `jiraSymmetricJwtMiddleware` and `GithubServerAppMiddleware` but before `jiraAdminPermissionsMiddleware`,  so that still allow non jira-admin user for callback.

src/routes/github/github-router.ts
```diff
@@ -36,7 +48,7 @@ subRouter.use("/branch", csrfMiddleware, GithubBranchRouter);
 
 subRouter.use(jiraAdminPermissionsMiddleware); // This must stay after jiraSymmetricJwtMiddleware
 
-subRouter.use(GithubOAuthRouter);
+subRouter.use("/login",  GithubOAuthLoginGet);
```
gxueatlassian, line 51: Are you sure we need this? I couldn't remember where we have a login link in the ui?

src/routes/github/github-oauth.ts
```diff
@@ -33,37 +41,45 @@ const getRedirectUrl = async (res, state) => {
...
 	// Save the redirect that may have been specified earlier into session to be retrieved later
-	req.session[state] =
-		res.locals.redirect ||
-		`/github/configuration${parsedOriginalUrl.search || ""}`;
+
+	const state: OAuthState = {
+		postLoginRedirectUrl: res.locals.redirect ||
+		`/github/configuration${parsedOriginalUrl.search || ""}`,
+		installationIdPk: res.locals.installation.id,
+		gitHubServerUuid: res.locals.gitHubAppConfig.uuid,
```
gxueatlassian, line 58: Base on the gitHubRouter comments, we may not even need this? As the uuid should be in the callback url?

gxueatlassian (commented):
/nit
Also a minor suggestion, this PR contains `naming thing for the xxxIdPk` and the moving of the `oauth-middleware`.
I know both are important, but I find it confusing in the PR. Would love to have separate PR to make it more clean. 

Like if for PR 1 just contain renaming, I can easily / blindly approve it, while spend more time thinking about PR 2 of moving oauth-router

src/routes/github/github-router.ts
bgvozdev: This puts more limitation, to have jiraHost in session (cause otherwise it will be rejected). The callback is just finishing the flow that has been started and the state is passed over, therefore IMHO it is not needed and may prohibit us doing certain things (e.g. when authorization is granted not based on JWT or jiraHost but some random token.

src/models/github-server-app.ts
bgvozdev: pretty sure it is provided, but don't to remove this now

src/routes/github/github-router.ts
bgvozdev: No idea ;)

src/routes/github/github-oauth.ts
bgvozdev: We cannot trust the URL without validation, however that's a good point! I don't think we need UUID for the callback whatsoever. Will try to remove UUID from that particular URL.

gxueatlassian:
Hey @bgvozdev I am a bit concern about this PR at the moment, as security is in the picture. So can you please:
1. Create a separate PR to just rename stuff (if you still want that)? That will reduce "noises" in this "real" PR.
2. Talk about the `"approval" flow` first from end-to-end (before raising this PR) ? As I find it hard to reason about this PR without realise the whole picture, end-to-end. I am not aware what ultimately we want to achieve in the approval stuff. Ideally we should discuss with @Ryan-Yuanqing-Jiang @sasank-xyz as the `Approval` flow sounds like user flow change.
3. Will also need more tests (unit/integrate) as well.

src/routes/github/github-router.ts
gxueatlassian: Okay I know what you are doing. In the code, I find it hard to co-relate.

In the callback, it has following code.
```
const jiraHost = (await Installation.findByPk(state.installationIdPk))?.jiraHost;
```

For me, this can be a **security risk**, as the state contains info for important to security. The code in PR will **work**, but it depends on we have to carefully on constructing the oauth redirect url duirng loginGet, we need to verify user is a jiraAdmin before we put that thing in state. I mean the security of the `Callback` (outside of the `jiraSymmetricJwtMiddleware` ) relies on the `GithubOAuthLoginGet` (being protected by `jiraSymmetricJwtMiddleware`). <-- this cause the confusing in understanding security wise on these middlewares.

Further more, I think we could be using something wrong in the state from this link:
https://auth0.com/docs/secure/attack-protection/state-parameters#:~:text=The%20primary%20reason%20for%20using,matches%20the%20one%20you%20sent.

```
The primary reason for using the state parameter is to mitigate CSRF attacks by using a unique and non-guessable value associated with each authentication request about to be initiated. That value allows you to prevent the attack by confirming that the value coming from the response matches the one you sent.
```

gxueatlassian: But again, let's chat about the big picture of this "Approval flow". It will make more sense once we understand it.

bgvozdev:
Discussed in person:
 - remove renaming
 - add some more tests (done already)
Closing so far

PR #2171, opened 2023-06-11
ARC-2131 - unified logging for backfill tasks

joshkay10:
**What's in this PR?**
Making all the start/complete logging for the 5 backfill tasks the same!!!

**Why**
Makes Splunking a lot easier, and it give us way easier debugging when trying to follow the backfill.

You can now search Splunk with "Backfill task *" and you will see when each task started and completed and whether there was data. JiraHost, messageId and githubInstallationId to filter it further.

**Whats Next?**
Probably a small runbook to help

### DIFF
--- a/src/sync/branches.ts
+++ b/src/sync/branches.ts
@@ -6,7 +6,7 @@ import { BackfillMessagePayload } from "~/src/sqs/sqs.types";
 
 // TODO: better typings
 export const getBranchTask = async (
-	logger: Logger,
+	parentLogger: Logger,
 	gitHubClient: GitHubInstallationClient,
 	_jiraHost: string,
 	repository: Repository,
@@ -14,16 +14,19 @@ export const getBranchTask = async (
 	perPage: number,
 	messagePayload: BackfillMessagePayload) => {
 
-	logger.info("Syncing branches: started");
+	const logger = parentLogger.child({ backfillTask: "Branch" });
+	const startTime = Date.now();
+
+	logger.info({ startTime }, "Backfill task started");
 
 	const commitSince = messagePayload.commitsFromDate ? new Date(messagePayload.commitsFromDate) : undefined;
 	const result = await gitHubClient.getBranchesPage(repository.owner.login, repository.name, perPage, commitSince, cursor as string);
 	const edges = result?.repository?.refs?.edges || [];
 	const branches = edges.map(edge => edge?.node);
 
-	logger.debug("Syncing branches: finished");
-
 	const jiraPayload = transformBranches({ branches, repository }, messagePayload.gitHubAppConfig?.gitHubBaseUrl);
+
+	logger.info({ endTime: Date.now() - startTime, jiraPayloadLength: jiraPayload?.branches?.length }, "Backfill task complete");
 
 	return {
 		edges,
--- a/src/sync/branches.ts
+++ b/src/sync/branches.ts
@@ -26,7 +26,7 @@ export const getBranchTask = async (
 
 	const jiraPayload = transformBranches({ branches, repository }, messagePayload.gitHubAppConfig?.gitHubBaseUrl);
 
-	logger.info({ endTime: Date.now() - startTime, jiraPayloadLength: jiraPayload?.branches?.length }, "Backfill task complete");
+	logger.info({ processingTime: Date.now() - startTime, jiraPayloadLength: jiraPayload?.branches?.length }, "Backfill task complete");
 
 	return {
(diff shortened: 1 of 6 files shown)


sasank-xyz (approved):

src/sync/branches.ts
```diff
@@ -6,25 +6,28 @@ import { BackfillMessagePayload } from "~/src/sqs/sqs.types";
...
 	repository: Repository,
 	cursor: string | undefined,
 	perPage: number,
 	messagePayload: BackfillMessagePayload) => {
 
-	logger.info("Syncing branches: started");
+	const logger = parentLogger.child({ backfillTask: "Branch" });
+	const startTime = Date.now();
+
+	logger.info({ startTime }, "Backfill task started");
```
gxueatlassian, line 20: Put it as `Branch backfill task started`? As if just `Backfill task started` it is a bit confusing.
And also I don't think we need this line, as we should have already something in a upper level? In the coordinator of all the tasks? So instead of doing it in each, leave it to parent?

src/sync/branches.ts
```diff
@@ -6,25 +6,28 @@ import { BackfillMessagePayload } from "~/src/sqs/sqs.types";
...
 	const commitSince = messagePayload.commitsFromDate ? new Date(messagePayload.commitsFromDate) : undefined;
 	const result = await gitHubClient.getBranchesPage(repository.owner.login, repository.name, perPage, commitSince, cursor as string);
 	const edges = result?.repository?.refs?.edges || [];
 	const branches = edges.map(edge => edge?.node);
 
-	logger.debug("Syncing branches: finished");
-
 	const jiraPayload = transformBranches({ branches, repository }, messagePayload.gitHubAppConfig?.gitHubBaseUrl);
 
+	logger.info({ processingTime: Date.now() - startTime, jiraPayloadLength: jiraPayload?.branches?.length }, "Backfill task complete");
```
gxueatlassian, line 29: Same as above, need `Branch xxx` in the msg. Also not sure we need this line.

src/sync/branches.ts
```diff
@@ -6,25 +6,28 @@ import { BackfillMessagePayload } from "~/src/sqs/sqs.types";
...
+	parentLogger: Logger,
 	gitHubClient: GitHubInstallationClient,
 	_jiraHost: string,
 	repository: Repository,
 	cursor: string | undefined,
 	perPage: number,
 	messagePayload: BackfillMessagePayload) => {
 
-	logger.info("Syncing branches: started");
+	const logger = parentLogger.child({ backfillTask: "Branch" });
```
gxueatlassian, line 17: The app already has in the parent sync/installation.ts. I feel this line is like duplication. 
```
		const logger = parentLogger.child({
				task: nextTask
			});

			logger.info("Starting task");
			nextTask.startTime = Date.now();
```

src/sync/branches.ts
gxueatlassian, line 29: Again, that `{ processingTime: Date.now() - startTime` should be somewhere upper level in sync/installation.ts
However I understand why here, as we need some more logging for `jiraPayload?.branches?.length`. 

I want to understand the purpose of the changes here. If it is for auditing problem, we might need to put more info in it. If it is for performance reason, we probably should use statsD instead.

src/sync/branches.ts
joshkay10: It does, but theres a few steps between that log and the actual execution, the parent log here is to tag all the other logs in the function a bit better, the `task:nextTask` attribute we inherit is an object and its a pain to Splunk with.

Otherwise its pretty much a tidier log :shrug:

src/sync/branches.ts
joshkay10: You are correct! the step after this is to emit that processingTime with StatsD, whether it needs to be in the log or not.... maybe/maybe not but it was floating around :lol: , this log was mainly just to see where it returns, so we can trace the flow if we need to.

Same as before, its an existing log, just aligning them across all the tasks

src/sync/branches.ts
gxueatlassian: Inside the `const taskExecutor = async (nextTask: Task` in the parent ( as in my above comments ), the `logger.info("Starting task");` actually means the start of the task, I don't see any extra steps?

gxueatlassian: I'll mark this as /nit for now to unblock you. But let me know if you can think of anyway to improve.

gxueatlassian (approved):

PR #2174, opened 2023-06-13
ARC-1359 Brahmos Wargames integration

yevhshevchenko:
**What's in this PR?**

**Why**

**Added feature flags**

**Affected issues**  
_Jira Issues_

**How has this been tested?**  

**Whats Next?**


Dockerfile
```diff
@@ -7,6 +7,10 @@ RUN apk add g++ make python3
 RUN apk add --update --upgrade busybox libretls openssl zlib curl
 
 COPY . /app
+
+# Add this the service Dockerfile, at the final stage if multi-stage
```
gxueatlassian, line 11: Hi from this project we can tell this file is not being used during deployment to ddev/stg/prod
https://bitbucket.org/atlassian/github-for-jira-deployment/src/master/

It should the `Dockerfile.prod`

github-for-jira.sd.yml
```diff
@@ -4,6 +4,8 @@ description: Connect app for integrating GitHub into Jira
 organization: Engineering-DevOps
 notifications:
   email: fusion-arc@atlassian.com
+tags:
+  brahmos: enabled
```
gxueatlassian, line 8: /nit
Can you please help me understand, that this tag will not impact the performance on our prod system? Or unless we tell brahmos to run on prod, everything will be as usuall?

gxueatlassian (commented):
nice

github-for-jira.sd.yml
joshkay10: brahmos isn't supported in prod yet so if we could tag it for only stage that would be ideal! 
we have some `environmentOverrides` in there already, but never tried to access that top level values :shrug:

Dockerfile
yevhshevchenko: @gxueatlassian do we have a separate dockerfile for stg and ddev?

gxueatlassian: I think `ddev` `stg` and `prod` shared the same docker file, but you can adjust that in the deployment scripts above.

github-for-jira.sd.yml
yevhshevchenko: @joshkay10 I think that we cannot apply global props on environmentOverrides

yevhshevchenko: @joshkay10 some updates. It works!

gxueatlassian (dismissed):

github-for-jira.sd.yml
gxueatlassian: Hi @yevhshevchenko so this is back to global settings now?

yevhshevchenko: Hi @gxueatlassian, yeah it was a problem when I tried to execute it and faced with an error(

bgvozdev (approved):

Dockerfile.prod
```diff
@@ -32,6 +33,8 @@ RUN apk add curl
 
 USER node
 COPY --chown=node:node --from=build /app /app
+# Add this the service Dockerfile, at the final stage if multi-stage
+COPY --from=docker.atl-paas.net/sox/brahmos-deps/stress-ng:latest /usr/bin/stress-ng /usr/bin/stress-ng
```
bgvozdev, line 37: Was wondering if it is safe to put it to prod Dockerfile though...

yevhshevchenko: @bgvozdev as i understood we do not have a separate file for stg image. I think that it shouldn't be a problem, we just will not have a brahmos on a prod instance

PR #2175, opened 2023-06-13
"Deferred install" flow, no UI

bgvozdev:
**What's in this PR?**
Deferred install flow by GitHub admin

**Why**
In some companes Jira admins and GitHub admins are different people. Having this flow would allow Jira admin to send the link over to GitHub admin to finish the connection.

**Added feature flags**
Yes

**Affected issues**  
too lazy to create one. omg

**How has this been tested?**  
staging

**Whats Next?**
sleeping


src/routes/github/configuration/github-configuration-get.test.ts
```diff
@@ -0,0 +1,164 @@
...
+						githubToken: "token"
+					})
+				);
+			expect(response.status).toStrictEqual(200);
+
+			const url = response.text.match(/data-deferred-install-url="([^"]+)"/)![1];
+			expect(url).toContain(`${envVars.APP_URL}/github/subscription-deferred-install/request`);
+
+			const deferredInstallResponse = await supertest(app).get(url.replace(envVars.APP_URL, ""));
+			expect(deferredInstallResponse.status).toStrictEqual(302);
```
gxueatlassian, line 90: /nit
shall we expect the `Location` in the response headers as well?

src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -97,12 +98,28 @@ const getInstallationsWithAdmin = async (
...
+			deferredInstallUrl = `${envVars.APP_URL}/github/${gitHubAppUuid ? (gitHubAppUuid + "/") : ""}subscription-deferred-install/request/${requestId}`;
+		}
+
 		return {
 			...installation,
 			numberOfRepos,
 			isAdmin,
 			requiresSsoLogin: !!errors.find(err => err instanceof GithubClientSSOLoginError),
-			isIPBlocked: !!errors.find(err => err instanceof GithubClientBlockedIpError)
+			isIPBlocked: !!errors.find(err => err instanceof GithubClientBlockedIpError),
```
gxueatlassian, line 121: /nit
while you are here, shouldn't it be
`isIPBlocked: errors.some(err=>err instanceof GithubClientBlockedIpError)`?

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.test.ts
```diff
@@ -0,0 +1,199 @@
...
+
+		it("should not allow call with corrupted payload", async () => {
+			const result = await supertest(app)
+				.get(`/github/subscription-deferred-install/request/boom`);
+			expect(result.status).toStrictEqual(400);
+			expect(result.body.error).toStrictEqual("Invalid payload");
+		});
+
+		it("should validate UUID and not allow call if belongs to a different GitHub server", async () => {
+			const gitHubServerApp = await DatabaseStateCreator.createServerApp(installation.id);
```
gxueatlassian, line 63: /nit
rename `gitHubServerApp` to `anotherGitHubServerApp`?

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.test.ts
```diff
@@ -0,0 +1,199 @@
...
+			const result = await supertest(app)
+				.get(`/github/${gitHubServerApp.uuid}/subscription-deferred-install/request/${await registerSubscriptionDeferredInstallPayloadRequest(payload)}`);
+			expect(result.status).toStrictEqual(400);
+			expect(result.body.error).toStrictEqual("Invalid payload");
```
gxueatlassian, line 67: /nit
Why `Invalid payload`?

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.ts
```diff
@@ -0,0 +1,53 @@
+import { Request, Response } from "express";
+import { extractSubscriptionDeferredInstallPayload } from "services/subscription-deferred-install-service";
+import { hasAdminAccess } from "services/subscription-installation-service";
+
+const escapeHtml = (unsafe: string) => {
```
gxueatlassian, line 5: lol, can't we use a lib? 
Or isn't handlerbar already escaping?

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.ts
```diff
@@ -0,0 +1,53 @@
...
+		.replace(/'/g, "&#039;");
+};
+
+export const GithubSubscriptionDeferredInstallGet = async (req: Request, res: Response) => {
+	const payload = await extractSubscriptionDeferredInstallPayload(req);
+	const { githubToken, installation } = res.locals;
+
+	if (!await hasAdminAccess(githubToken, installation.jiraHost, payload.gitHubInstallationId, req.log, payload.gitHubServerAppIdPk)) {
+		req.log.warn("Not an admin");
+		res.status(401).json({
```
gxueatlassian, line 20: /nit
User will see a json text right? I am okay for a spike.

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.ts
```diff
@@ -0,0 +1,53 @@
...
+		jiraHost: res.locals.jiraHost,
+		gitHubUrl: res.locals.gitHubAppConfig.hostname,
+		... (res.locals.gitHubAppConfig.gitHubAppId ? {
+			gheClientId: res.locals.gitHubAppConfig.clientId,
+			gheAppId: res.locals.gitHubAppConfig.gheAppId
+		} : { })
+	};
+
+	res.send(
+		`
```
gxueatlassian, line 37: Okay, lol, just see this, you just hate handlerbar don't you?

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-post.ts
```diff
@@ -0,0 +1,23 @@
...
+		githubToken, installation, payload.gitHubServerAppIdPk, payload.gitHubInstallationId, req.log
+	);
+	if (result.error) {
+		res.status(401).json(result);
+		return;
+	}
+
+	await forgetSubscriptionDeferredInstallRequest(req);
+
+	res.status(200).json({ ok: true });
```
gxueatlassian, line 22: What user see? I think a more detail message, even if just json, will be better for `GitHub Admin`, like `Thank you have just connect xxx to xxx successfully blah blah`

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-router.ts
```diff
@@ -0,0 +1,73 @@
...
+	if (parsedPayload.gitHubServerAppIdPk != res.locals.gitHubAppConfig.gitHubAppId) {
+		req.log.warn("Wrong appIdPk");
+		res.status(400).json({ error: INVALID_PAYLOAD_ERR });
+		return;
+	}
+
+	return next();
+};
+
+subRouter.use(validatePayloadAndPopulateJiraHost);
```
gxueatlassian, line 63: /nit
**IMPORTANT**

I just hate how we setup the routing in these nested behaviour. So un-readable. I just would love to see following in ONE SINGLE FILE!!

```
const router = app.router();
const routerConfig = {
   "/github": {
     "/deferInstallUrl": [
        GitHubServerAppMiddleware,
        GitHubAuthMiddlware, 
        {
           "GET": gitHubDeferUrlGet
        }
      ],
      "/configuration: [GitHubServerAppMiddleware, GitHubAuthMiddlware, {
           "GET": gitHubConfigGet
           "POST": gitHubConfigPost
      }"]
  },
  "jira": {
  }
};
router.use(routerConfig);
```

gxueatlassian: **AND**!!
Inside the so-called `gitHubConfigPost`, it is actually just a function, take params return params, and none of the params are http related. There shouldn't be any `res.locals.xxx`. It should be just plain js object params. 

The extractor should be somewhere else.

src/services/subscription-deferred-install-service.ts
```diff
@@ -0,0 +1,66 @@
...
+
+const redis = new IORedis(getRedisInfo("SubscriptionDeferredInstallService"));
+
+const REDIS_CLEANUP_TIMEOUT = 2 * 24 * 3600 * 1000;
+
+const toRedisKey = (requestId: string) => "sub_def_inst_serv_req}" + requestId;
+
+const toRequestId = (req: Request) => req.params["requestId"];
+
+export const extractSubscriptionDeferredInstallPayload = async (req: Request) => {
```
gxueatlassian, line 21: **IMPORTANT**
I don't think we should make the `service` aware of http stuff, such as `Request` here. Would prefer a typed param.

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-router.ts
gxueatlassian: so /nit, just me ranting.

src/services/subscription-deferred-install-service.ts
```diff
@@ -0,0 +1,66 @@
...
+	try {
+		const requestId = toRequestId(req);
+		if (!requestId) {
+			throw new Error("Empty request ID");
```
gxueatlassian, line 25: /nit

This is debatable on how should the code flow works? Exception (like here) or return errorCode?

So is the consumer supposed to catch when calling this api and analyse the text? 

I understand this is spike, so this is just /nit.

Personally I slightly lean towards return an error code instead of throw, but I am not aware what's the consistent behaviour in our code yet, so happy for you to decide.

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-router.ts
```diff
@@ -0,0 +1,73 @@
...
+		res.status(400).json({ error: INVALID_PAYLOAD_ERR });
+		return;
+	}
+
+	if (!await booleanFlag(BooleanFlags.ENABLE_SUBSCRIPTION_DEFERRED_INSTALL, installation.jiraHost)) {
+		res.status(401).json({ error: "Feature is disabled" });
+		return;
+	}
+
+	res.locals.jiraHost = installation.jiraHost;
```
gxueatlassian, line 44: /nit
**IMPORTANT**
So I am a bit concern about the security model here.

ps: lol, nit and important at the same time.

I understand this url with the `requestId` is supposed to be shortLive, however we don't have a timer on the requestId in redis? Or do we? And if any malicious users intercept the `requestId` somehow (Since the jiraAdmin will send this url to gitHubAdmin somehow, the requestId is in the url path), that malicious user can 
1. Hit the url with with its own `gitHubToken`, so can connect his own github org to the victims' jiraHost?
2. Do we have a timer on the requestId on redis the auto destroy?

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-post.ts
rachellerathbone: Nit: success: true (to align with responses in the rest of our middleware and node/express common conventions) + add a message `{ success: true, message: "Something like what Gary said or whatever tickles your fancy" }

src/routes/github/configuration/github-configuration-get.ts
rachellerathbone: My 2 cents on this: both achieve the same result but the latter is more readable and is a little clearer in terms of intent. errors.some conveys the purpose of checking if there exists any error of type GithubClientBlockedIpError and reads as "some errors satisfy the condition." 

At the end of the day, choice between them comes down to preference (I don't think we went this in-depth in coding conventions).

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.ts
```diff
@@ -0,0 +1,53 @@
...
+const escapeHtml = (unsafe: string) => {
+	return unsafe
+		.replace(/&/g, "&amp;")
+		.replace(/</g, "&lt;")
+		.replace(/>/g, "&gt;")
+		.replace(/"/g, "&quot;")
+		.replace(/'/g, "&#039;");
+};
+
+export const GithubSubscriptionDeferredInstallGet = async (req: Request, res: Response) => {
```
rachellerathbone, line 14: Nit: GithubSubscriptionDeferredInstallRouter > GitHubSubscriptionDeferredInstallRouter

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.ts
rachellerathbone: Can we convert this to a template? We're already using Handlebars and it's better to stick with a consistent pattern of using it throughout our codebase.

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-router.ts
rachellerathbone: I agree with Gary here (although maybe disagree with the nit part 😜... where's :wink_: when ya need it?).

To Gary's first point, there's definitely potential for someone to intercept the requestId, hitting the URL with their own GitHub token, and connecting their GitHub organization to the someone else's Jira host. We definitely want to make sure we have some kind of expiration mechanism for the requestId. And to his second point: can we check if we do have an expiration mechanism in place? If we don't we definitely want to set a TTL on the requestId key in Redis to destroy expired requests.

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-router.ts
rachellerathbone: Wow, Gary rage 😜

src/services/subscription-deferred-install-service.ts
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		const request = await redis.get(toRedisKey(requestId));
+		if (!request) {
+			throw new Error("Unknown request");
+		}
+
+		const parsedPayload = JSON.parse(request) as SubscriptionDeferredInstallPayload;
+		if (!parsedPayload.installationIdPk) {
```
rachellerathbone, line 34: I think the first 2 conditional blocks should stay as is but can we convert these 3 checks for parsedPayload to a switch statement just so it's a little more concise and readable? Something like:

```
  switch (true) {
      case !parsedPayload.installationIdPk:
        throw new PayloadExtractionError("No installationIdPk");
      case !parsedPayload.gitHubInstallationId:
        throw new PayloadExtractionError("No gitHubInstallationId");
      case !parsedPayload.orgName:
        throw new PayloadExtractionError("No orgName");
    }
```

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.ts
rachellerathbone: Also, can you provide a screenshot of what this looks like?

src/services/subscription-deferred-install-service.ts
```diff
@@ -0,0 +1,66 @@
...
+		}
+
+		return parsedPayload;
+	} catch (err) {
+		req.log.warn({ err }, "Cannot extract payload");
+		throw err;
+	}
+};
+
+export const registerSubscriptionDeferredInstallPayloadRequest = async (payload: SubscriptionDeferredInstallPayload) => {
```
rachellerathbone, line 53: Add return type: `Promise<string>`

src/services/subscription-deferred-install-service.ts
```diff
@@ -0,0 +1,66 @@
...
+	const key = newUUID();
+	// We don't want to pollute redis, autoexpire after the flag is not being updated
+	await redis.set(toRedisKey(key), JSON.stringify(payload), "px", REDIS_CLEANUP_TIMEOUT);
+	return key;
+};
+
+
+export const forgetSubscriptionDeferredInstallRequest = async (req: Request) => {
```
rachellerathbone, line 61: Add return type: `Promise<void>`

src/services/subscription-deferred-install-service.ts
```diff
@@ -0,0 +1,66 @@
...
+	} catch (err) {
+		req.log.warn({ err }, "Cannot extract payload");
+		throw err;
+	}
+};
+
+export const registerSubscriptionDeferredInstallPayloadRequest = async (payload: SubscriptionDeferredInstallPayload) => {
+	const key = newUUID();
+	// We don't want to pollute redis, autoexpire after the flag is not being updated
+	await redis.set(toRedisKey(key), JSON.stringify(payload), "px", REDIS_CLEANUP_TIMEOUT);
```
rachellerathbone, line 56: Nit: can we break this out into new lines to make it easier to read:

```
await redis.set(
      toRedisKey(key),
      JSON.stringify(payload),
      "px",
      REDIS_CLEANUP_TIMEOUT
    );
```

src/services/subscription-deferred-install-service.test.ts
```diff
@@ -0,0 +1,53 @@
...
+				params: {
+					requestId
+				}
+			} as unknown as Request;
+			await forgetSubscriptionDeferredInstallRequest(request);
+
+			await expect(extractSubscriptionDeferredInstallPayload(request)).toReject();
+		});
+	});
+});
```
rachellerathbone, line 53: Can we add a few extra tests here? Some ideas of things we could also account for:
1. missing/undefined requestId
2. handling an unknown request (requestId does not exist in Redis store)
3. missing properties in the payload

src/services/subscription-deferred-install-service.ts
rachellerathbone, line 21: Missing return type: `: Promise<SubscriptionDeferredInstallPayload>`

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.ts
bgvozdev: given that's a throw-away code, I'd leave it as it is

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-router.ts
bgvozdev: We setup TTL in redis when we put the record, don't we?

1. No, they cannot. The user must be an admin of the github org that is being connected.
2. yes, we have a timer REDIS_CLEANUP_TIMEOUT, currently set up to 2 days
3 . (to Rachelle), they cannot connect to their Jira host because it is inside the payload that they cannot change

image

src/services/subscription-deferred-install-service.ts
bgvozdev: To be frank, I'd prefer to have explicit ifs. This switch looks like a hack to me... Maybe I'm too old though...

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.test.ts
bgvozdev: I opted for returning a common message in case of a error and write details to log, to expose as little info to the outside as possible.

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-router.ts
bgvozdev: I think we should strive for a good test coverage, thus reducing chances that somehting is wrong with this logic. For each endpoint check everything: when it returns 401, when it doesn't, etc. At least this is what I'm doing now :shrug:

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.ts
gxueatlassian: I still think handlebar is easier to achieve since we already had it in place. Just throw in a .hbs file and voila.

src/services/subscription-deferred-install-service.ts
bgvozdev: Well, given that's "extract" (action! do it!), I think exception makes more sense, cause we are not expecting any error... "validate/verify" make more sense to return a error code. But that's my IMHO

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.ts
bgvozdev: One small vs one new file + call render with params, for temp code?

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.ts
bgvozdev: Yeah

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.ts
bgvozdev: That doesn't suppose to have any UI, ok?.....

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-router.ts
bgvozdev: Maybe to clarify it a little bit more:

1. admin prepares a request for connection. Noone can prepare this request for that Jira except of the admin, cause the payload is stored on the backend. Even if requestId is known, that's nothing, just an admin wanting to connect an org to a Jira...
2. to finish the connection, the user must know requestId and to be the GitHub admin of the org that is being requested to connect. Again, if you are a GitHub admin of some other org, or not a GitHub admin at all, you won't be able even to read the details of the request

bgvozdev: feel free to ping me in Slack to discuss if you could see any security threat that I cannot

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.ts
gxueatlassian: If `escapeHtml` is required, then yes, I'd say trust handlebars ( or some libs -- but cons as one more dependancy) rather than hand crafting the escape, as security related stuff, just don't do it yourself.

src/services/subscription-deferred-install-service.ts
gxueatlassian: I agree to @bgvozdev , on both points.

src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -158,34 +171,11 @@ export const GithubConfigurationGet = async (req: Request, res: Response, next:
 
 	const gitHubUserClient = await createUserClient(githubToken, jiraHost, { trigger: "github-configuration-get" }, log, gitHubAppId);
 
-	req.log.debug("found github token");
```
gxueatlassian, line 161: Why aren't these protected by FF?

src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -137,10 +154,6 @@ export const GithubConfigurationGet = async (req: Request, res: Response, next:
 
 	const log = req.log.child({ jiraHost });
 
-	if (!githubToken) {
```
gxueatlassian, line 140: Why aren't these protected by FF?

src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -158,34 +171,11 @@ export const GithubConfigurationGet = async (req: Request, res: Response, next:
 
 	const gitHubUserClient = await createUserClient(githubToken, jiraHost, { trigger: "github-configuration-get" }, log, gitHubAppId);
 
-	req.log.debug("found github token");
-
-	if (!jiraHost) {
```
gxueatlassian, line 163: Why aren't these protected by FF?

src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -97,12 +98,28 @@ const getInstallationsWithAdmin = async (
 		const [isAdmin, numberOfRepos] = await Promise.all([checkAdmin, numberOfReposPromise]);
 		log.info("Number of repos in the org received via GraphQL: " + numberOfRepos);
 
+		let deferredInstallUrl: string | undefined;
+
+		// TODO: we should register the request in a separate POST endpoint. Short-cutting corners now for the spike,
+		// TODO: but must be addressed before rolling out for everyone, to avoid polluting redis
+		if (await booleanFlag(BooleanFlags.ENABLE_SUBSCRIPTION_DEFERRED_INSTALL, jiraHost) && !isAdmin) {
```
gxueatlassian, line 105: Shall we also exclude the `connected` one as well?

src/routes/github/configuration/github-configuration-get.ts
bgvozdev: Because this was some old dead code. We always have gitHubToken here because it is under GitHubOauthMiddelware

src/routes/github/configuration/github-configuration-get.ts
bgvozdev: Because this is some old dead code. We always have jiraHost because it is under jwt-symmetric-middleware

src/routes/github/subscription-deferred-install/github-subscription-deferred-install-get.ts
bgvozdev: Well, that's annoying. We are enabling this code only for 1 customer, and this is such a corner-case

bgvozdev: I'm happy to create some sort of UI with handlebars in my next PR, this is not the purpose of the current one and it is big enough already

src/routes/github/configuration/github-configuration-get.ts
bgvozdev: This shouldn't matter because for the final feature the admin will need to explicitly create the request. We shouldn't be creating it here in the first place 🤷

src/services/subscription-deferred-install-service.ts
rachellerathbone: Definitely agree with the latter :wink_:

krazziekay (approved):

PR #2179, opened 2023-06-18
Maintain environmentMapping order from config.yml

aldenquimby:
**What's in this PR?**
Maintain environmentMapping order from config.yml

**Why**
- [the docs here](https://github.com/atlassian/github-for-jira/blob/3a0f6a47a9ea324f3f5def49dd77ae604eef79f2/SUPPORT.md#deployment-mapping) say "Think top-down if you ever need to override the mapping defined in GitHub for Jira" and suggest the order of environments matters for matching
- [the matching code respects ordering today](https://github.com/atlassian/github-for-jira/blob/e9f93406f398237bd2f4ed0228d70c3c6a1f5686/src/transforms/transform-deployment.ts#L194C1-L197), but the YML -> JS config parser does not, so the docs are inaccurate
- This PR fixes that

**Why - my specific use case**
- My team is trying to use this configuration:
```yml
  environmentMapping:
    development: ["local"]
    staging: ["uat"]
    production: ["production"]
    testing: ["*"]
```
- We have 50+ testing environments and there is no consistent naming convention today, so the idea is "if it's not one of our _special_ environments, then it must be a testing env"
- I have also tried this config, but the negation doesn't work correctly:
```yml
  environmentMapping:
    development: ["local"]
    testing: ["!(local|uat|production)"]
    staging: ["uat"]
    production: ["production"]
```

**Added feature flags**
- No

**Affected issues**  
- No

**How has this been tested?**  
- Added unit test, which failed before this change and passes after

**Whats Next?**
- N/A

### DIFF
--- a/src/services/user-config-service.test.ts
+++ b/src/services/user-config-service.test.ts
@@ -33,10 +33,7 @@ describe("User Config Service", () => {
 		"      - \"test-id-3\"\n" +
 		"      - \"test-id-4\"";
 
-	const configFileContentBase64 = Buffer.from(configFileContent).toString("base64");
-
 	let gitHubClient;
-
 
 	beforeEach(async () => {
 		subscription = await Subscription.create({
@@ -71,11 +68,15 @@ describe("User Config Service", () => {
 
 	});
 
-	const givenGitHubReturnsConfigFile = (repoOwner: string = repoSyncState.repoOwner, repoName = repoSyncState.repoName) => {
+	const givenGitHubReturnsConfigFile = ({
+		repoOwner = repoSyncState.repoOwner,
+		repoName = repoSyncState.repoName,
+		fileContent = configFileContent
+	}: { repoOwner?: string, repoName?: string, fileContent?: string } = {}) => {
 		// see https://docs.github.com/en/rest/repos/contents#get-repository-content
 		githubNock.get(`/repos/${repoOwner}/${repoName}/contents/.jira/config.yml`)
 			.reply(200, {
-				content: configFileContentBase64
+				content: Buffer.from(fileContent).toString("base64")
 			});
 	};
 
@@ -126,10 +127,24 @@ describe("User Config Service", () => {
 		const unknownRepoId = 42;
 
 		githubUserTokenNock(gitHubInstallationId);
-		givenGitHubReturnsConfigFile(unknownRepoOwner, unknownRepoName);
+		givenGitHubReturnsConfigFile({ repoOwner: unknownRepoOwner, repoName: unknownRepoName });
 		const config = await getRepoConfig(subscription, gitHubClient, unknownRepoId, unknownRepoOwner, unknownRepoName);
 		expect(config).toBeTruthy();
 		expect(config?.deployments?.environmentMapping?.development).toHaveLength(4);
 	});
 
+	it("should maintain environment order for top-down matching", async () => {
+		const fileContent = "deployments:\n" +
+			"  environmentMapping:\n" +
+			"    development: [\"DEV-*\"]\n" +
+			"    staging: [\"STG-*\"]\n" +
+			"    production: [\"PRD-*\"]\n" +
+			"    testing: [\"*\"]\n";
+
+		githubUserTokenNock(gitHubInstallationId);
+		givenGitHubReturnsConfigFile({ fileContent });
+		await updateRepoConfig(subscription, repoSyncState.repoId, gitHubClient, ["random.yml", "ignored.yml", ".jira/config.yml"]);
+		const config = await getRepoConfig(subscription, gitHubClient, repoSyncState.repoId, repoSyncState.repoOwner, repoSyncState.repoName);
+		expect(Object.keys(config?.deployments?.environmentMapping ?? {})).toStrictEqual(["development", "staging", "production", "testing"]);
+	});
 });
--- a/src/services/user-config-service.test.ts
+++ b/src/services/user-config-service.test.ts
@@ -147,8 +147,8 @@ describe("User Config Service", () => {
 
 		githubUserTokenNock(gitHubInstallationId);
 		givenGitHubReturnsConfigFile({ fileContent });
-		await updateRepoConfig(subscription, repoSyncState.repoId, gitHubClient, ["random.yml", "ignored.yml", ".jira/config.yml"]);
-		const config = await getRepoConfig(subscription, gitHubClient, repoSyncState.repoId, repoSyncState.repoOwner, repoSyncState.repoName);
+		await updateRepoConfig(subscription, repoSyncState.repoId, gitHubClient, logger, ["random.yml", "ignored.yml", ".jira/config.yml"]);
+		const config = await getRepoConfig(subscription, gitHubClient, repoSyncState.repoId, repoSyncState.repoOwner, repoSyncState.repoName, logger);
 		expect(Object.keys(config?.deployments?.environmentMapping ?? {})).toStrictEqual(["development", "staging", "production", "testing"]);
 	});
 });
--- a/src/services/user-config-service.test.ts
+++ b/src/services/user-config-service.test.ts
@@ -131,7 +131,7 @@ describe("User Config Service", () => {
 		githubUserTokenNock(gitHubInstallationId);
 		givenGitHubReturnsConfigFile({ repoOwner: unknownRepoOwner, repoName: unknownRepoName });
 
-    const config = await getRepoConfig(subscription, gitHubClient, unknownRepoId, unknownRepoOwner, unknownRepoName, logger);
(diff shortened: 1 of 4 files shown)


src/services/user-config-service.ts
```diff
@@ -116,12 +116,7 @@ const convertYamlToUserConfig = (input?: string): Config => {
 	const deployments = {};
 	if (configDeployments != null) {
 		if (configDeployments.environmentMapping) {
-			deployments["environmentMapping"] =  {
-				development: configDeployments.environmentMapping.development,
-				testing: configDeployments.environmentMapping.testing,
-				staging: configDeployments.environmentMapping.staging,
-				production: configDeployments.environmentMapping.production
-			};
+			deployments["environmentMapping"] = configDeployments.environmentMapping;
```
aldenquimby, line 119: If the lack of key filtering is concerning here, I could change this to:
```js
deployments["environmentMapping"] = pickBy(configDeployments.environmentMapping, (_, key) => ['development','testing','staging','production'].includes(key));
```

Lodash `pick` doesn't maintain key order, but `pickBy` does

gxueatlassian: My thinking is, if we filter by key names, and user has a config like
```
{
  prod: ["prod-*"],
  development: ["*"],
}
```
and has a deployment with env name `prod-us-west-1`, the app will map it to `development` as the previous key `prod` will be omitted by the `pickBy`. Which leads to confusion to users.

However, without the filter by key, apps will store `prod` as the key in db, but `prod` is not supported in the upstream  jira api that save the deployment data 
api: https://developer.atlassian.com/cloud/jira/software/rest/api-group-deployments/#api-rest-deployments-0-1-bulk-post
image


I suggest
1. We don't do key filter here. store whatever user supplied in yaml file in db
2. However, during mapping, map it to `unmapped` when that key is not in the supported array found in api doc.

gxueatlassian (commented):
Hey @aldenquimby thank you for raising the PR.
I see you want to achieve here, just some minor comments. Please have a look see whether it make sense.

src/services/user-config-service.ts
gxueatlassian: And I mean checking whether the final mapping is supported and map to `unmapped` if not supported.

here: https://github.com/atlassian/github-for-jira/blob/main/src/transforms/transform-deployment.ts#L222

gxueatlassian (approved):
Thank you @aldenquimby looks good to me 👍 !

src/services/user-config-service.test.ts
```diff
@@ -128,10 +129,26 @@ describe("User Config Service", () => {
 		const unknownRepoId = 42;
 
 		githubUserTokenNock(gitHubInstallationId);
-		givenGitHubReturnsConfigFile(unknownRepoOwner, unknownRepoName);
-		const config = await getRepoConfig(subscription, gitHubClient, unknownRepoId, unknownRepoOwner, unknownRepoName, logger);
+		givenGitHubReturnsConfigFile({ repoOwner: unknownRepoOwner, repoName: unknownRepoName });
+
+    const config = await getRepoConfig(subscription, gitHubClient, unknownRepoId, unknownRepoOwner, unknownRepoName, logger);
```
gxueatlassian, line 134:
```suggestion
		const config = await getRepoConfig(subscription, gitHubClient, unknownRepoId, unknownRepoOwner, unknownRepoName, logger);
```

gxueatlassian (approved):

src/services/user-config-service.test.ts
```diff
@@ -128,10 +129,26 @@ describe("User Config Service", () => {
...
+		const fileContent = "deployments:\n" +
+			"  environmentMapping:\n" +
+			"    development: [\"DEV-*\"]\n" +
+			"    staging: [\"STG-*\"]\n" +
+			"    production: [\"PRD-*\"]\n" +
+			"    testing: [\"*\"]\n";
+
+		githubUserTokenNock(gitHubInstallationId);
+		givenGitHubReturnsConfigFile({ fileContent });
+		await updateRepoConfig(subscription, repoSyncState.repoId, gitHubClient, ["random.yml", "ignored.yml", ".jira/config.yml"]);
```
gxueatlassian, line 150:
```suggestion
		await updateRepoConfig(subscription, repoSyncState.repoId, gitHubClient, logger, ["random.yml", "ignored.yml", ".jira/config.yml"]);
```

src/services/user-config-service.test.ts
```diff
@@ -128,10 +129,26 @@ describe("User Config Service", () => {
...
+		const config = await getRepoConfig(subscription, gitHubClient, repoSyncState.repoId, repoSyncState.repoOwner, repoSyncState.repoName);
```
gxueatlassian, line 151:
```suggestion
		const config = await getRepoConfig(subscription, gitHubClient, repoSyncState.repoId, repoSyncState.repoOwner, repoSyncState.repoName, logger);
```

gxueatlassian:
Hi @aldenquimby this is merged, it will wait for internal CI/CD to deploy it to prod. Will let you know once finished.

gxueatlassian:
> Thank you @gxueatlassian ! Appreciate the help on this, my team will be excited for this change

No worries @aldenquimby , just to follow up the change has been deployed. Please try it out :)

aldenquimby:
@gxueatlassian this still isn't working for my team 🤔. I've add a few commits touching `.jira/config.yml` to our default branch to trigger a resync, but we are still seeing the old behavior from before this PR where the order of the file isn't respected.

Here is my `.jira/config.yml` file:
```yml
deployments:
  environmentMapping:
    development: ["local"]
    staging: ["qa", "uat"]
    production: ["production"]
    testing: ["*"]
```
I am expecting `qa` and `uat` to show as `Staging`, and `production` to show as `Production`, but everything is being mapped to `Testing`:
image

What is the best way for me to debug from here? I have confirmed `transform-deployment.ts` works the way I'd like with my config, but I don't know how to figure out the RepoSyncState for my repo

gxueatlassian:
> @gxueatlassian this still isn't working for my team 🤔. I've added a few commits touching `.jira/config.yml` to our default branch to trigger a resync, but we are still seeing the old behavior from before this PR where the order of the file isn't respected.
> 
> Here is my `.jira/config.yml` file:
> 
> ```yaml
> deployments:
>   environmentMapping:
>     development: ["local"]
>     staging: ["qa", "uat"]
>     production: ["production"]
>     testing: ["*"]
> ```
> 
> I am expecting `qa` and `uat` to show as `Staging`, and `production` to show as `Production`, but everything is still mapped to `Testing` as of today: image
> 
> What is the best way for me to debug from here? I have confirmed `transform-deployment.ts` works the way I'd like with my config, but I don't know how to figure out the RepoSyncState for my repo

Hi @aldenquimby I assume the repo name is `fre*******v2`right? From what I found in our DB the current state of the mapping is 
```
{"deployments":{"environmentMapping":{"development":["local"],"testing":["*"],"staging":["qa","uat"],"production":["production"]}}}
```

It is quite interesting somehow the `testing` bit go above the rest. 
Can you please double check your master/main branch that the file contains the right order? I think there's a "flaw" that currently only master/main branch's config.yml file matter. The file in other branch won't matter. 

The unit test you added looks good so far, as it save into postgres db and read it back using the same read method of transforming deployment
```
	it("should maintain environment order for top-down matching", async () => {
		const fileContent = "deployments:\n" +
			"  environmentMapping:\n" +
			"    development: [\"DEV-*\"]\n" +
			"    staging: [\"STG-*\"]\n" +
			"    production: [\"PRD-*\"]\n" +
			"    testing: [\"*\"]\n";

		githubUserTokenNock(gitHubInstallationId);
		givenGitHubReturnsConfigFile({ fileContent });
		await updateRepoConfig(subscription, repoSyncState.repoId, gitHubClient, logger, ["random.yml", "ignored.yml", ".jira/config.yml"]);
		const config = await getRepoConfig(subscription, gitHubClient, repoSyncState.repoId, repoSyncState.repoOwner, repoSyncState.repoName, logger);
		expect(Object.keys(config?.deployments?.environmentMapping ?? {})).toStrictEqual(["development", "staging", "production", "testing"]);
	});
```

Tbh, I am not sure what goes wrong. As for privacy reason, we can't simply log the actual content of your file in our logging system, unless you agree and we can temporary log it.

gxueatlassian:
> @gxueatlassian yes that is the correct repo name, thank you for the help. My `main` branch contains the file in the correct order. Is there a way for me to force a resync, so the DB pulls a fresh value from the repo? I'd like to rule out the possibility that the DB has a stale value from before my PR

I think if you make a change (on the file) in your main branch, the app should receive it.

aldenquimby:
@gxueatlassian it's working for me now! I Thank you!

PR #2184, opened 2023-06-20
[EAZY kARMA] NONE: add more retries for discovery

bgvozdev:
**What's in this PR?**
Increasing number of retries for repo discovery task.

**Why**
I was digging Db and spotted a backfill that was failing on repo sync. The interesting fact: it had 2 subscriptions for the same gitHubInstallationId and they were failing on different repos, which makes me think increasing number of retries might've solved the issue (see "n" column - this is the number of repos)

```
id   |         createdAt          | numberOfSyncedRepos |       backfillSince        | syncStatus | gitHubInstallationId |   n    |       lastupdatedat
--------+----------------------------+---------------------+----------------------------+------------+----------------------+--------+----------------------------
  70441 | 2021-04-26 13:06:54.071+00 |                   0 |                            | FAILED     |             13083610 | 131604 | 2023-05-22 09:19:26.339+00
  70393 | 2021-04-26 08:01:07.489+00 |                   0 |                            | FAILED     |             13083610 | 110410 | 2023-05-22 09:19:32.752+00
  ```

### DIFF
--- a/github-for-jira.sd.yml
+++ b/github-for-jira.sd.yml
@@ -82,7 +82,7 @@ resources:
   - type: sqs
     name: backfill
     attributes:
-      MaxReceiveCount: 3
+      MaxReceiveCount: 7
       MessageRetentionPeriod: 1209600
       VisibilityTimeout: 602 #Visibility timeout will be overridden by the SQS Listener when we read message from the queue. We set it here anyway in case if that override fails.
       dataType:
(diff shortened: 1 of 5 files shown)


gxueatlassian (approved):
Alternatively while you are here, we can use a ff to fine grain control each task's retry for each jira host.

bgvozdev:
I don't think so @gxueatlassian because that's a config of the SQS queue. Yes, we could've updated the "fail-fast" number, but we didn't have any problem with it so far...

PR #2200, opened 2023-06-26
HOT: loose condition on RepoSyncState

bgvozdev:
**What's in this PR?**
- Check only that the app knows about the org, don't look for exact repo

**Why**
- In case the repo somehow is not in RepoSyncState table (e.g. was created before we enabled create-repo event, or we missed the event whatsoever...)

**Added feature flags**
Nah

**Affected issues**  
HOT

**How has this been tested?**  
tests, locally

### DIFF
--- a/src/routes/github/create-branch/github-create-branch-get.ts
+++ b/src/routes/github/create-branch/github-create-branch-get.ts
@@ -113,9 +113,8 @@ const getReposBySubscriptions = async (subscriptions: Subscription[], logger: Lo
 			const response = await gitHubInstallationClient.getRepositoriesPage(100, undefined, "UPDATED_AT");
 			// The app can be installed in a GitHub org but that org might not be connected to Jira, therefore we must filter them out, or
 			// the next steps (e.g. get repo branches to branch of) will fail
-			const subscriptionRepos = await RepoSyncState.findAllFromSubscription(subscription);
-			const subscriptionRepoIds = new Set(subscriptionRepos.map(repo => repo.repoId));
-			const filteredRepos =  response.viewer.repositories.edges.filter(edge => subscriptionRepoIds.has(edge.node.id));
+			const repoOwners = await RepoSyncState.findAllRepoOwners(subscription);
+			const filteredRepos =  response.viewer.repositories.edges.filter(edge => repoOwners.has(edge.node.owner.login));
 			return filteredRepos.slice(0, MAX_REPOS_RETURNED);
 		} catch (err) {
 			logger.error({ err }, "Create branch - Failed to fetch repos for installation");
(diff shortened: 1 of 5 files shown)


src/routes/github/create-branch/github-create-branch-get.ts
```diff
@@ -113,9 +113,8 @@ const getReposBySubscriptions = async (subscriptions: Subscription[], logger: Lo
 			const response = await gitHubInstallationClient.getRepositoriesPage(100, undefined, "UPDATED_AT");
 			// The app can be installed in a GitHub org but that org might not be connected to Jira, therefore we must filter them out, or
 			// the next steps (e.g. get repo branches to branch of) will fail
-			const subscriptionRepos = await RepoSyncState.findAllFromSubscription(subscription);
-			const subscriptionRepoIds = new Set(subscriptionRepos.map(repo => repo.repoId));
-			const filteredRepos =  response.viewer.repositories.edges.filter(edge => subscriptionRepoIds.has(edge.node.id));
+			const repoOwners = await RepoSyncState.findAllRepoOwners(subscription);
```
gxueatlassian, line 116: We don't need this? As the `createInstallationClient` is already for a particular `org`, no need to further filter

src/routes/github/repository/github-repository-get.ts
```diff
@@ -80,10 +80,9 @@ const getReposBySubscriptions = async (repoName: string, subscriptions: Subscrip
 
 			// We don't want to return repos that are not in Database, otherwise we won't be able to fetch branches to branch from later
 			// The app can be installed in a GitHub org but that org might not be connected to Jira, therefore we must filter them out, or
-			// the next steps (e.g. get repo branches to branch of) will fail
-			const subscriptionRepos = await RepoSyncState.findAllFromSubscription(subscription);
-			const subscriptionRepoIds = new Set(subscriptionRepos.map(repo => repo.repoId));
-			return installationSearch.filter(repo => subscriptionRepoIds.has(repo.id));
+			// the next step (e.g. get repo branches to branch of) will fail
+			const subscriptionOwners = await RepoSyncState.findAllRepoOwners(subscription);
```
gxueatlassian, line 84: Same here actuall.

gxueatlassian (changes requested):
I just realize we may not even need those new sql calls.

src/routes/github/create-branch/github-create-branch-get.ts
bgvozdev: We need because when we fetch branches, we pass the org as a parameter:
- https://github.com/atlassian/github-for-jira/blob/792dab5e01f35ac64a56b73a4228d021f0f82f06/src/routes/github/create-branch/github-branches-get.ts#L19-L19

If the app doesn't know about that org, the call will fail

gxueatlassian (approved):

PR #2216, opened 2023-07-03
ARC 2256 - Enable user to see the basic skeleton of new config experience

krazziekay:
**What's in this PR?**
- Basic skeleton for the new config experience
- Setup for React SPA
  - Local Proxy server setup for hot live reload
  - Production build setup
  - Dummy React component added, using AtlasKit button
  - ViteJs configured for bundling
  - New route added in `atlassian-connect.json` for loading React SPA pages.

**Why**
- For setup of the 5KU first phase.

**How has this been tested?**  
- Locally
  - Simply run the github app, using the command `yarn start`, then the SPA can be viewed at URL `https://ATLASSIAN-INSTANCE.net/spa-index-page`

**Whats Next?**
- Adding test cases


package.json
```diff
@@ -12,10 +12,10 @@
 		"prestart": "run-p setup clean db:dev",
 		"clean": "rimraf coverage build tmp src/**/*.js src/**/*.js.map test/**/*.js test/**/*.js.map",
 		"init": "run-p db",
-		"start": "run-p start:main start:worker",
+		"start": "run-p spa:dev start:main start:worker",
 		"start:main": "tsnd -r tsconfig-paths/register --watch=.env*,db/config.json --inspect=0.0.0.0:9229 --respawn --transpile-only -- src/main.ts",
 		"start:worker": "tsnd -r tsconfig-paths/register --watch=.env*,db/config.json --inspect=0.0.0.0:9230 --respawn --transpile-only -- src/worker.ts",
-		"start:production": "run-p start:main:production start:worker:production",
+		"start:production": "run-p spa:production start:main:production start:worker:production",
```
gxueatlassian, line 18: `spa:production` should not be here, as in production, we don't need it it. We should copy the compiled assets during build time not run time.

spa/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "spa",
```
gxueatlassian, line 2: Maybe not matter, but anyway update the name as well;

spa/tsconfig.json
```diff
@@ -0,0 +1,26 @@
+{
+  "compilerOptions": {
+    "target": "ES2020",
```
gxueatlassian, line 3: /nit
might need put some comments somewhere to justify this.

spa/vite.config.ts
```diff
@@ -0,0 +1,10 @@
+import { defineConfig } from "vite";
+import react from "@vitejs/plugin-react";
+
+/**
+ * Documentation: https://vitejs.dev/config/
+ */
+export default defineConfig({
+	plugins: [react()],
+	base: "/spa"
```
gxueatlassian, line 9: Shall we use something like `/app` instead of `/spa`, as `/spa` is quite technical details, and `/app` is more generic.

package.json
krazziekay: `spa:production` removed and added this step in `build`

spa/package.json
krazziekay: Couldn't think of a better name, so keeping it as it is  for now

gxueatlassian (dismissed):

gxueatlassian (approved):

PR #2220, opened 2023-07-04
None - pass jiraHost to FF

rachellerathbone:
**What's in this PR?**
Passing jiraHost to FF

**Why**
When I initially made the changes to the app descriptor to utilise the FF, there wasn't a way to pass the jiraHost. However, I ended up refactoring my original code and realised it can and should be passed now.

**Added feature flags**
ENABLE_GENERIC_CONTAINERS

**Affected issues**  
none

**How has this been tested?**  
locally, unit tests

**Whats Next?**
Turn on ff for folks that are testing


src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -90,10 +90,10 @@ const getIssuer = (token: string, logger: Logger): string | undefined => {
 	return unverifiedClaims.iss;
 };
 
-export const getTokenType = async (url: string, method: string): Promise<TokenType> => {
-	if (await booleanFlag(BooleanFlags.ENABLE_GENERIC_CONTAINERS)) {
+export const getTokenType = async (url: string, method: string, jiraHost: string): Promise<TokenType> => {
+	if (await booleanFlag(BooleanFlags.ENABLE_GENERIC_CONTAINERS, jiraHost)) {
 		return checkPathValidity(url) && method == "GET"
-		|| await checkGenericContainerActionUrl(`${envVars.APP_URL}${url}`) ? TokenType.normal
+		|| await checkGenericContainerActionUrl(`${envVars.APP_URL}${url}`, jiraHost) ? TokenType.normal
```
gxueatlassian, line 96: Hi, actually in `|| await checkGenericContainerActionUrl(`${envVars.APP_URL}${url}`, jiraHost`, since the FF is already checked in the `if (await booleanFlag(BooleanFlags.ENABLE_GENERIC_CONTAINERS, jiraHost))`, why do we need to check the ff again?

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
```diff
@@ -236,7 +236,8 @@ const	modules = {
 };
 
 export const JiraAtlassianConnectGet = async (_: Request, res: Response): Promise<void> => {
-	modules.jiraDevelopmentTool.actions = await defineJiraDevelopmentToolModuleActions();
+	const { jiraHost } =  res.locals;
+	modules.jiraDevelopmentTool.actions = await defineJiraDevelopmentToolModuleActions(jiraHost);
```
gxueatlassian, line 240: Just to confirm, if we change the ff, user is required to uninstall/re-install the app the make the change effective as this description will only be read once during the app installation?

rachellerathbone: No, the connect team said these changes are automatically detected.

gxueatlassian (approved):

gxueatlassian (approved):

PR #2221, opened 2023-07-05
ARC 2270 - Enable user to view new experience entry page

krazziekay:
**What's in this PR?**
- First page created
- Header component created
- Styled components implemented
- Added Route for assets for both local & prod environment(Assets are all stored under static/ folder)
- Updated eslint rule
- Removed dummy components
- Customized tooltip added

**How has this been tested?**  
- Local

**Screenshot**
Screenshot 2023-07-05 at 12 31 23 pm
Screenshot 2023-07-05 at 12 31 31 pm


spa/src/tests/StartConnection.test.tsx
gxueatlassian, line 1: Can we move this file to the same folder of the `StartConnection.tsx`? I think sitting in same folder is better and consistent with our existing code.

krazziekay: Pages now moved into a folder and the test case moved in the same folder.

PR #2222, opened 2023-07-05
Isoc 3686 add security provider module

mjordan-atlassian:
**What's in this PR?**
- Add the jiraSecurityInfoProviderModule behind a FF
- Add route for the `fetchContainers` action https://developer.atlassian.com/cloud/jira/software/modules/security-information/

**Why**
Part of the work to integrate GitHub with Security in Jira. This module is required in order to use the Security APIs and handle actions from Jira.
Full Spec - https://hello.atlassian.net/wiki/spaces/CDX/pages/2673460294/RFC+GitHub+for+Jira+Security

**Added feature flags**
https://app.launchdarkly.com/github-for-jira/production/features/enable-github-security-in-jira/targeting

**Affected issues**  
https://softwareteams.atlassian.net/browse/ISOC-3686

**How has this been tested?**  
- Test for connect descriptor
- Test for new route

**Whats Next?**
- Implement fetchWorkspaces and searchContainers actions


src/middleware/jira-symmetric-jwt-middleware.ts
```diff
@@ -5,7 +5,8 @@ import { getJWTRequest, TokenType, validateQsh } from "~/src/jira/util/jwt";
 import { Installation } from "~/src/models/installation";
 import {
 	getGenericContainerUrls,
-	moduleUrls
+	moduleUrls,
+	securityContainerActionUrls
```
rachellerathbone, line 9: Nit: for consistency, maybe we could call this getSecurityContainerActionUrls

src/models/reposyncstate.ts
```diff
@@ -310,6 +310,24 @@ export class RepoSyncState extends Model implements RepoSyncStateProperties {
 		return repositories as RepoSyncState[];
 	}
 
+	static async findOneForRepoUrlAndRepoId(repoUrl: string, repoId: string) {
```
rachellerathbone, line 313: I didn't get a chance to do this during inno week but we should probably add a test/s for this.

src/routes/jira/security/workspaces/containers/jira-security-workspaces-containers-post.ts
```diff
@@ -0,0 +1,90 @@
...
+};
+
+export const JiraSecurityWorkspacesContainersPost = async (req: Request, res: Response): Promise<void> => {
+	req.log.info({ method: req.method, requestUrl: req.originalUrl }, "Request started for security POST repositories");
+
+	const { ids: repoIds } = req.body;
+
+	if (!repoIds) {
+		// TODO: Return fetchContainers error handling spec once implemented
+		// https://hello.atlassian.net/wiki/spaces/CDX/pages/2639314030/RFC+Handle+fetchContainers+errors
```
rachellerathbone, line 75: Oh nice. Thanks for linking to his.

src/routes/jira/security/workspaces/containers/jira-security-workspaces-containers-post.ts
```diff
@@ -0,0 +1,90 @@
...
+
+export const transformRepositories = async (
+	repos: RepoSyncState[]
+): Promise<SecurityContainer[]> => {
+	const transformedSubscriptions = repos.map((repo) => {
+		const { repoId, repoName, repoUrl, updatedAt } = repo;
+		const baseUrl = new URL(repoUrl).origin;
+		return {
+			id: transformRepositoryId(repoId, baseUrl),
+			name: repoName,
```
gxueatlassian, line 58: So looking at the PR, my question is:
1. Is the `repoName`, `avatarUrl` ( and `lastUpdatedDate` ) the only info needed by the security module? Which is the whole point of this PR? 
2. Since security module already have `repoId` and `repoUrl`, if `1` is true, and assuming we can modify our app to push those info into DataDepot at the first places, wouldn't that means we don't need this PR at all? As all those information is already available?

I mean, this doesn't seem heaps of (critical) information are missing. Can't we just save the app from calling this endpoint?

src/routes/jira/security/workspaces/containers/jira-security-workspaces-containers-post.ts
```diff
@@ -0,0 +1,90 @@
...
+	if (!repoIds) {
+		// TODO: Return fetchContainers error handling spec once implemented
+		// https://hello.atlassian.net/wiki/spaces/CDX/pages/2639314030/RFC+Handle+fetchContainers+errors
+		const errMessage = Errors.MISSING_SECURITY_CONTAINER_IDS;
+		req.log.warn(errMessage);
+		res.status(400).send(errMessage);
+		return;
+	}
+
+	const repos = await getRepos(repoIds);
```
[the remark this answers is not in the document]
rachellerathbone: Good callout. I overlooked this last week in my rush to throw stuff together for inno week last week. @mjordan-atlassian do you think you could update this? Otherwise, I can do it but probably won't be able to tackle until next week.

src/models/reposyncstate.ts
mjordan-atlassian: Will do 👍

src/routes/jira/security/workspaces/containers/jira-security-workspaces-containers-post.ts
mjordan-atlassian: @gxueatlassian `fetchContainers` and `searchContainers` (to be implemented) both only need to return `id, name, url, avatarUrl, lastUpdatedDate` https://developer.atlassian.com/cloud/jira/software/modules/security-information/

> Can't we just save the app from calling this endpoint?
In an ideal situation, we would like to get to this point eventually. We are looking into how we can store Security Container information in DD but this is not planned for this quarter and needs more investigation.

We already have existing patterns in place on both the Security in Jira page and Toolchain page which rely on calling these actions specified by the security module. This pattern is already being used by 5 different security providers and would like to avoid creating specific rules just for GitHub.

src/routes/jira/security/workspaces/containers/jira-security-workspaces-containers-post.ts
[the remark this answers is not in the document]
mjordan-atlassian: Yeah, good call. I can look into this a bit later today and will reach out if I run into any issues.

src/models/reposyncstate.ts
```diff
@@ -310,6 +310,32 @@ export class RepoSyncState extends Model implements RepoSyncStateProperties {
 		return repositories as RepoSyncState[];
 	}
 
+	static async findOneForRepoUrlAndRepoIdAndJiraHost(repoUrl: string, repoId: number, jiraHost: string):Promise<RepoSyncState | null> {
+		const results = await this.sequelize!.query(
+			`SELECT rss.* 
+			FROM "RepoSyncStates" rss
+			JOIN "Subscriptions" s ON rss."subscriptionId" = s."id"
```
gxueatlassian, line 317: I have hesitation on the performance here:
1. Why do we need to use `JOIN`, can't we grab the subscriptionId first, and just search on RepoSyncStates table?
2. Why do we need to use LIKE? Can't we just do "repoUrl" = :repoUrl?

Not that it doesn't work, it is more like a performance concern.

src/routes/jira/security/workspaces/containers/jira-security-workspaces-containers-post.ts
gxueatlassian: Okay got cha, I am good with this now. 👍

gxueatlassian (approved):

PR #2228, opened 2023-07-06
NONE: get rid of maybe jwt something something

bgvozdev:
**What's in this PR?**
Removing some remnants of jiraHost/jwt query params

**Why**
Cause they are not secure

**Added feature flags**
Nah

**How has this been tested?**  
Locally

### DIFF
--- a/static/js/github-create-branch-options.js
+++ b/static/js/github-create-branch-options.js
@@ -45,11 +45,9 @@ $(document).ready(() => {
 });
 
 const createUrlForGH = (uuid, multiGHInstance) => {
-	const jiraHost = $("#jiraHost").val();
-	const jwt = $("#jwt").val();
   return uuid ?
-    `github/${uuid}/create-branch?issueKey=${issueKey}&issueSummary=${issueSummary}&jiraHost=${jiraHost}&jwt=${jwt}&ghRedirect=to&multiGHInstance=${multiGHInstance}` :
-    `github/create-branch?issueKey=${issueKey}&issueSummary=${issueSummary}&jiraHost=${jiraHost}&jwt=${jwt}&multiGHInstance=${multiGHInstance}`;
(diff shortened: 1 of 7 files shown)


joshkay10 (approved):

static/js/github-create-branch-options.js
```diff
@@ -45,11 +45,9 @@ $(document).ready(() => {
 });
 
 const createUrlForGH = (uuid, multiGHInstance) => {
-	const jiraHost = $("#jiraHost").val();
-	const jwt = $("#jwt").val();
   return uuid ?
-    `github/${uuid}/create-branch?issueKey=${issueKey}&issueSummary=${issueSummary}&jiraHost=${jiraHost}&jwt=${jwt}&ghRedirect=to&multiGHInstance=${multiGHInstance}` :
-    `github/create-branch?issueKey=${issueKey}&issueSummary=${issueSummary}&jiraHost=${jiraHost}&jwt=${jwt}&multiGHInstance=${multiGHInstance}`;
+    `github/${uuid}/create-branch?issueKey=${issueKey}&issueSummary=${issueSummary}&ghRedirect=to&multiGHInstance=${multiGHInstance}` :
```
gxueatlassian, line 49: Did you tested it that it still works without `jwt=${jwt}`

joshkay10, line 49: im not sure if JWT still needs to be passed through here, will have a check

bgvozdev: Yep. jiraHost is put to session by jwtSymmetricMiddleware and is used from there afterwards

gxueatlassian (approved):

static/js/github-create-branch-options.js
joshkay10: nope, looks like used locals values now, should be ok

bgvozdev: I don't think so: once JWT is verified it is put to session and used from there after that https://github.com/atlassian/github-for-jira/blob/e698f8241b16668bd2ec659737061a4ce1b552fd/src/middleware/jira-symmetric-jwt-middleware.ts#L45-L45

PR #2229, opened 2023-07-06
ARC-2253 delete backfill msg if there's another job running it

gxueatlassian:
**What's in this PR?**
Delete the backfill message if it is found working on another worker

**Why**

1. Our backfill msg count climb. up very high
2. Alert fire off
3. msg count doesn't comes down
4. Rate limit slooooow the process
5. I think there're a lots of same backfill msg for one customer
6. Previously there's no retry on backfill message in some cases. Now we cover all cases to retry for backfill, so it is safe to delete the message it self.j

**Added feature flags**
enable-dedup-backfill-sqs-msg

**Affected issues**
ARC-2253

**How has this been tested?**
N/A

**Whats Next?**
Turn on ff and test.

### DIFF
--- a/src/config/feature-flags.ts
+++ b/src/config/feature-flags.ts
@@ -28,7 +28,7 @@ export enum BooleanFlags {
 	EARLY_EXIT_ON_VALIDATION_FAILED = "early-exit-on-validation-failed",
 	USE_REST_API_FOR_DISCOVERY = "use-rest-api-for-discovery-again",
 	ENABLE_GENERIC_CONTAINERS = "enable-generic-containers",
-	DELETE_MESSAGE_ON_BACKFILL_WHEN_OTHERS_WORKING_ON_IT = "delete_message_on_backfill_when_others_working_on_it",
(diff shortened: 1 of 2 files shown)


src/config/feature-flags.ts
```diff
@@ -28,6 +28,7 @@ export enum BooleanFlags {
 	EARLY_EXIT_ON_VALIDATION_FAILED = "early-exit-on-validation-failed",
 	USE_REST_API_FOR_DISCOVERY = "use-rest-api-for-discovery-again",
 	ENABLE_GENERIC_CONTAINERS = "enable-generic-containers",
+	DELETE_MESSAGE_ON_BACKFILL_WHEN_OTHERS_WORKING_ON_IT = "delete_message_on_backfill_when_others_working_on_it",
```
bgvozdev, line 31: nit: I think our convention is to use hyphen, not underscore

bgvozdev (approved):

src/config/feature-flags.ts
gxueatlassian: Damn it, I knew that, somehow mis typed it. Thanks

joshkay10 (approved):

PR #2234, opened 2023-07-07
draft/test

gxueatlassian:


spa/src/api/rest-api.ts
```diff
@@ -0,0 +1,75 @@
+export function createRestApi() {
+
+	let gitHubAccessToken: string | undefined;
```
gxueatlassian, line 3: My idea is that, assume there's some malicious code running in user browser, those malicious scripts cannot read the value of those token, like `window.gitHubToken` or `window.react.__context.PageConnect.__imortedAuthManager.gitHubAccessToken`

gxueatlassian: Above `gitHubAccessToken` is in the "closure" of this function, so it wont' be disclosed.

spa/src/api/rest-api.ts
```diff
@@ -0,0 +1,75 @@
...
+	let gitHubRefreshToken: string | undefined;
+
+	return {
+		storeGitHubToken,
+		isTokenStillValid,
+		generateOAuthUrl,
+		fetchInstalledOrganisations
+	};
+
+	function storeGitHubToken(tokens: { gitHubAccessToken: string; githubRefreshToken?: string }) {
```
gxueatlassian, line 13: A write only function.

spa/src/api/rest-api.ts
```diff
@@ -0,0 +1,75 @@
...
+
+	async function isTokenStillValid() {
+
+		try {
+			const resp = await fetch("https://api.github.com/user", {
+				headers: {
+					Authorization: `Bearer ${gitHubAccessToken}`
+				}
+			});
+			if (!resp.ok) {
```
gxueatlassian, line 26: This is a catch of `fetch`, not sure whether you awared of it, `fetch` doesn't throw on 400/500, it only throw on network problem. Need to check `resp.ok` to determine whether it is 200

spa/src/api/rest-api.ts
```diff
@@ -0,0 +1,75 @@
...
+
+	async function fetchInstalledOrganisations() {
+
+		try {
+
+			const jiraJwt = await getNewJiraToken();
+
+			const resp = await fetch("/rest/app/cloud/organizations", {
+				headers: {
+					Authorization: `Bearer ${jiraJwt}###${gitHubAccessToken}`
```
gxueatlassian, line 51: We need both `gitHubToken` and `JiraJwtToken` in api request now. And I don't trust any other header than `Authorization`. <--- I think some switch/router in the middle of the network will all be aware of NOT to log the value of `Authorization`, but if we use `Some-random-header-key`, those router/switch might log them in the log.

gxueatlassian: But I am not too sure, I am open for suggestions.

spa/src/services/auth-manager.ts
```diff
@@ -0,0 +1,15 @@
+import { createRestApi } from "../api/rest-api";
+
+const restApi = createRestApi();
+
+export {
+};
+
+async function startOAuth(onDone: () => void) {
```
gxueatlassian, line 8: This is the bit I want in a `manager` layer. It does two things, and I don't want it to pollute our beautiful react code.

PR #2235, opened 2023-07-07
CES-15622 add more logs (curlv) for customer to investigate

gxueatlassian:
**What's in this PR?**
Add more curl logs to a customer for listing the org.

**Why**
They claim the org is not showing. Need to log curl for github to investigate.

**Added feature flags**
N/A

**Affected issues**  
CES-15622

**How has this been tested?**  

**Whats Next?**


gxueatlassian:
case closed already.

PR #2239, opened 2023-07-10
ARC-2271 share same ts typing and add jwt handling

gxueatlassian:
**What's in this PR?**
1. Add JWT handling
2. Add ts types sharing between frontend and backend.

**Why**
1. The existing JWT handling relying too much on the session cookie, which I'd like to remove in the new 5ku experience as it is all frontend SPA in local js memory now.

**Added feature flags**
N/A

**Affected issues**  
ARC-2271

**How has this been tested?**  
Unit test

**Whats Next?**
Add oauth dance.


spa/src/rest-interfaces
```diff
@@ -0,0 +1 @@
+../../src/rest-interfaces
```
gxueatlassian, line 1: This is a linux symbolic link. I hope the production pipeline (linux) will recognise this as well as other developer's machine.

src/routes/rest/oauth/oauth-router.ts
```diff
@@ -0,0 +1,11 @@
+import { Router, Request, Response } from "express";
+import { getRedirectUrl } from "./oauth-service";
+
+export const OAuthRouter = Router({ mergeParams: true });
+
+OAuthRouter.post("/redirecturl", async (_req: Request, res: Response) => {
+	const { jiraHost } = res.locals;
```
gxueatlassian, line 7: A dummy endpoint.

PR #2244, opened 2023-07-11
ARC 2271 - [BE] Enable user to authorised with GitHub

krazziekay:
**What's in this PR?**
- Implemented the backend logic for logging into GitHub
- Added services & components in frontend for handling the backend logic

**How has this been tested?**  
- Locally

**Whats Next?**
- Test cases and probably cleanup, cause the changes here are growing massive.

**Video**

https://github.com/atlassian/github-for-jira/assets/13076255/c6281f92-4bfb-47b1-a71e-ea3e53d62d89


spa/src/oauth-manager/index.tsx
```diff
@@ -0,0 +1,55 @@
...
+	let username: string | undefined;
+	let email: string | undefined;
+
+	async function checkValidity() {
+		if (!accessToken) return;
+		const res = await ApiRequest.token.getUserDetails(accessToken);
+		username = res.data.login;
+		email = res.data.email;
+
+		return res.status === 200;
```
gxueatlassian, line 17: If res.status !== 200, will this return `false` or will axios throw an Error?

spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -55,31 +56,82 @@ const InlineDialog = styled(TooltipPrimitive)`
...
+	const [expandStep1, setExpandStep1] = useState(!isAuthenticated);
+	const [expandStep2, setExpandStep2] = useState(isAuthenticated);
 
-	const authorize = () => {
+	const [isLoggedIn, setIsLoggedIn] = useState(isAuthenticated);
+	const [loggedInUser, setLoggedInUser] = useState<string>(username);
+	const [loaderForLogin, setLoaderForLogin] = useState(false);
+
+	useEffect(() => {
+		window.addEventListener("message", (event) => {
```
gxueatlassian, line 85: Do you think we should clear the eventListener as well?

spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -55,31 +56,82 @@ const InlineDialog = styled(TooltipPrimitive)`
...
+			setIsLoggedIn(true);
+			setCompletedStep1(true);
+			setExpandStep1(false);
+			setExpandStep2(true);
+			setCanViewContentForStep2(true);
+		});
+	}, []);
+
+	useEffect(() => {
+		OAuthManagerInstance.checkValidity().then((status: boolean | undefined) => {
```
gxueatlassian, line 97: Write the function as async and do normal async code here? Instead of Promise chain.

spa/src/utils/axiosInstance.ts
```diff
@@ -0,0 +1,19 @@
+import axios from "axios";
```
gxueatlassian, line 1: I assume only places use this `axiosInstance` is the code within `src/api`? So maybe we should move this file to under the folder `src/api/`?

src/rest/routes/github-callback/service.ts
```diff
@@ -0,0 +1,34 @@
+import { Request } from "express";
+import { envVars } from "config/env";
+import axios from "axios";
+
+const GITHUB_CLOUD_LOGIN_URL = "https://github.com/login/oauth/access_token";
+export const getAccessToken = async (req: Request) => {
```
gxueatlassian, line 6: This is a pure business service file, so let's try not to put any http-related classes here, like `Request`. Just pass in the param as normal args.

src/rest/routes/github-callback/service.ts
```diff
@@ -0,0 +1,34 @@
...
+	const { code } = req.query;
+	if (!code) {
+		req.log.error("No code provided!");
```
gxueatlassian, line 9: Throw error if node code?

src/rest/routes/github-callback/service.ts
gxueatlassian, line 6: Put a proper typing on the return value please.

spa/src/api/auth/index.ts
```diff
@@ -0,0 +1,7 @@
+import AxiosInstance from "../../utils/axiosInstance";
+
+const GitHubAuth = {
+	authenticate: () => AxiosInstance.get("/rest/app/cloud/oauth/redirectUrl")
```
gxueatlassian, line 4: Add proper typing on response type.

spa/src/api/token/index.ts
```diff
@@ -0,0 +1,9 @@
+import axios from "axios";
+
+const Token = {
+	getUserDetails: (token: string) => axios.get("https://api.github.com/user", {
```
gxueatlassian, line 4: Add proper typing on response type.

spa/src/oauth-manager/index.tsx
krazziekay: Added a try-catch to catch errors, in case of errors this will only return false.

spa/src/pages/ConfigSteps/index.tsx
krazziekay: Removed the eventListener once the message has been received.

spa/src/pages/ConfigSteps/index.tsx
krazziekay: Using `async-await` inside useEffect causes [memory leakage](https://ultimatecourses.com/blog/using-async-await-inside-react-use-effect-hook). That's why I'm using a then chain here.

spa/src/utils/axiosInstance.ts
krazziekay: Updated

src/rest/routes/github-callback/service.ts
krazziekay: Updated

src/rest/routes/github-callback/service.ts
krazziekay: Updated

src/rest/routes/github-callback/service.ts
krazziekay: Updated

spa/src/api/auth/index.ts
krazziekay: Updated

spa/src/api/token/index.ts
krazziekay: Updated

spa/src/api/auth/index.ts
```diff
@@ -0,0 +1,8 @@
+import AxiosInstance from "../axiosInstance";
+import { AxiosResponse } from "axios";
+
+const GitHubAuth = {
+	authenticate: (): Promise<AxiosResponse> => AxiosInstance.get("/rest/app/cloud/oauth/redirectUrl")
```
gxueatlassian, line 5: Can you type like this? Coz that's the whole point of typing the rest api

```typescript
import { GetRedirectUrlResponse } from "../../rest-interfaces/oauth-types";

const GitHubAuth = {
       authenticate: (): Promise<AxiosResponse> => AxiosInstance.get("/rest/app/cloud/oauth/redirectUrl")
       authenticate: (): Promise<AxiosResponse<GetRedirectUrlResponse>> => AxiosInstance.get("/rest/app/cloud/oauth/redirectUrl")
};
```

spa/src/api/token/index.ts
```diff
@@ -0,0 +1,9 @@
+import axios, { AxiosResponse } from "axios";
+
+const Token = {
+	getUserDetails: (token: string): Promise<AxiosResponse> => axios.get("https://api.github.com/user", {
```
gxueatlassian, line 4: Same typing as above please.

spa/src/oauth-manager/index.tsx
gxueatlassian: Actually where's the try/catch? I didn't see it?

gxueatlassian (approved):

PR #2249, opened 2023-07-12
Arc 2271 secure state in oauth

gxueatlassian:
**What's in this PR?**
Add usage of `state` to verify the oauth flow

**Why**
Best practice to protected the app 
https://auth0.com/docs/secure/attack-protection/state-parameters

**Added feature flags**
N/A

**Affected issues**  
ARC-2271

**How has this been tested?**  
Unit test, local

**Whats Next?**
N/A


spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -85,7 +85,8 @@ const ConfigSteps = () => {
 		const handler = async (event: any) => {
 			if (event.origin !== originalUrl) return;
 			if (event.data?.code) {
-				await OAuthManagerInstance.finishOAuthFlow(event.data?.code, event.data?.state);
+				const success = await OAuthManagerInstance.finishOAuthFlow(event.data?.code, event.data?.state);
+				if (!success) return;
```
gxueatlassian, line 89: @krazziekay maybe another PR, again. I don't know where to show the error message (like which UI component). 
For now, I just `NOT` call the following steps (setIsLoggin(true)) if the oauth fail.

src/rest/routes/oauth/service.ts
```diff
@@ -41,13 +62,32 @@ export const finishOAuthFlow = async (
...
+		const redisState = await redis.get(state) || "";
+
+		if (!redisState) {
+			log.warn({ state }, "state is missing in redis in oauth exchange token");
+			return null;
+		}
+
+		const parsedState = JSON.parse(redisState);
+
+		if (jiraHost !== parsedState.jiraHost) {
```
gxueatlassian, line 86: This is the whole purpose of the check (2 of 2)

spa/src/oauth-manager/index.tsx
```diff
@@ -19,16 +21,28 @@ const OauthManager = () => {
...
 	}
 
 	async function finishOAuthFlow(code: string, state: string) {
+
+		if (!code) return false;
+		if (!state) return false;
+
+		const prevState = window.localStorage.getItem(STATE_KEY);
+		window.localStorage.removeItem(STATE_KEY);
+		if (state !== prevState) return false;
```
gxueatlassian, line 37: This is the whole purpose of the check (1 of 2) to prevent CSRF attack.

src/rest-interfaces/oauth-types.ts
```diff
@@ -8,41 +9,6 @@ export type ExchangeTokenResponse = {
 };
 
 export type UsersGetAuthenticatedResponse = {
-	avatar_url: string;
```
gxueatlassian, line 11: I think we should type only what we need.

krazziekay (approved):

PR #2250, opened 2023-07-12
ISOC-3717 - Adding fetchWorkspace api to github for jira app. 

Blastoplex:
**What's in this PR?**
Adding the fetchWorkspaces action to the security module. Didn't feature flag as this was already implemented in the previous PR
Added the route, read path, and new attribute to subscription entity (avatar url) and appropriate migration script.

Updated the workspace id to be a subscription rather than a hybrid githubinstallationId as subscriptions better represent the workspace entity 

**Why**
This is necessary for jira to be able to fetch workspace data from the github app and render things like the name. 

**Added feature flags**
n/a

**Affected issues**  
ISOC-3717

**How has this been tested?**  
Added integration tests to `src/routes/jira/security/workspaces/jira-security-workspaces-post.test.ts`

**Whats Next?**


src/services/subscription-installation-service.ts
```diff
@@ -117,19 +117,13 @@ export const verifyAdminPermsAndFinishInstallation =
...
 
 	try {
-		const gitHubAppClient = await createAppClient(logger, installation.jiraHost, gitHubServerAppIdPk, { trigger: "github-configuration-post" });
-
-		logger.info("Fetching info about GitHub installation");
-		const { data: ghInstallation } = await gitHubAppClient.getInstallation(gitHubInstallationId);
-
 		const jiraClient = await JiraClient.getNewClient(installation, logger);
-		await jiraClient.linkedWorkspace(ghInstallation.account.id);
+		await jiraClient.linkedWorkspace(subscription.id);
```
Blastoplex, line 126: Updated linked workspaces to be subscription ids

src/util/not-empty.ts
```diff
@@ -0,0 +1,7 @@
+const notEmpty = <TValue>(
```
Blastoplex, line 1: this is just super useful for Typscript to filter out potentially null items from your array

gxueatlassian (changes requested):
Just blocking it temporary until db migrated.

src/routes/jira/security/workspaces/jira-security-workspaces-post.test.ts
```diff
@@ -0,0 +1,324 @@
...
+				method: "POST",
+				pathname: "/jira/security/workspaces",
+				query
+			}, false),
+			iss: installation.plainClientKey
+		}, await installation.decrypt("encryptedSharedSecret", getLogger("test")));
+	};
+
+	it("Should return a 400 status if no IDs are passed in the body", async () => {
+		app = express();
```
gxueatlassian, line 100: existing sampel *.test.ts may not be good, I think just use following should be good
```
const app = getFrontendApp();
await supertest(app).post()...
`

Comment (gxueatlassian) [reply src/routes/jira/security/workspaces/jira-security-workspaces-post.test.ts:100]: And remove 
```
app.use((req, _, next) => {
			req.log = getLogger("test");
			next();
		});
```

File: src/services/subscription-installation-service.ts (hunk 1 above)
Comment (gxueatlassian) [reply src/services/subscription-installation-service.ts:126]: Just to make sure, if user disconnect and reconnect a subscription (but without re-install the app in github org), the security data will be different. 
Is there a place to remove the data upon disconnect?
I think Harmindar have that here https://github.com/atlassian/github-for-jira/pull/2242/files#diff-087bf7a763d243dfd6ae90017a14d41c979f3e27171e77523c9e3561cf0dccb1R85
But that one still use the orgId.

File: src/models/subscription.ts [hunk 4]
```diff
@@ -188,7 +189,8 @@ export class Subscription extends Model {
 				gitHubInstallationId: payload.installationId,
 				jiraHost: payload.host,
 				jiraClientKey: payload.hashedClientKey,
-				gitHubAppId: payload.gitHubAppId || null
+				gitHubAppId: payload.gitHubAppId || null,
+				avatarUrl: payload.avatarUrl || null
```
gxueatlassian, line 193: Just to double check the install of `avatarUrl` is always `null` in the PR right? We can do that in a separate PR is that the plan?

src/services/subscription-installation-service.ts
Blastoplex: I missed that! ill update that to subscription also.

src/routes/jira/security/workspaces/jira-security-workspaces-post.test.ts
Blastoplex: Ill give that a try! thanks

src/models/subscription.ts
Blastoplex: Yep leaving it as null, and we have a default avatar that we return when it is null. I dont think jira actually uses the avatar url atm, But at some point we should start storing it in another pr

gxueatlassian (approved):

gxueatlassian (approved):

src/models/subscription.ts
Blastoplex: Update added the setting of the url in this pr.

PR #2255, opened 2023-07-12
Arc 2271 fix up oauth bug

gxueatlassian:
**What's in this PR?**
fix minor bugs on the poco.


spa/src/oauth-manager/index.tsx
```diff
@@ -37,7 +37,7 @@ const OauthManager = () => {
 		if (state !== prevState) return false;
 
 		const token = await ApiRequest.githubAuth.exchangeToken(code, state);
-		if (token.data.accessToken && token.data.refreshToken) {
+		if (token.data.accessToken) {
```
gxueatlassian, line 40: On stage our app is not configured to return refresh token. Hence it will always be empty.

krazziekay (approved):

PR #2256, opened 2023-07-12
None resolve 504

joshkay10:
tested in stage

**What's in this PR?**
Apparently 301/302 redirects can strip header data usually **Authorization**, in the logs we have some calls getting 401s at the create-branch route, but successfully making through the create-branch-options route, which means that they were authorized for the first call but not the second, we have that backed up by some error messaging saying header data is incomplete... Hopefully this resolves that!

### DIFF
--- a/src/routes/github/create-branch/github-create-branch-options-get.ts
+++ b/src/routes/github/create-branch/github-create-branch-options-get.ts
@@ -5,6 +5,7 @@ import { GitHubServerApp } from "~/src/models/github-server-app";
 import { sendAnalytics } from "utils/analytics-client";
 import { AnalyticsEventTypes, AnalyticsScreenEventsEnum } from "interfaces/common";
 import { envVars } from "config/env";
+import { getLogger } from "config/logger";
 
 // TODO - this entire route could be abstracted out into a generic get instance route on github/instance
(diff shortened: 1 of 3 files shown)


src/routes/github/create-branch/github-create-branch-options-get.ts
```diff
@@ -36,12 +44,16 @@ export const GithubCreateBranchOptionsGet = async (req: Request, res: Response,
 	const url = new URL(`${req.protocol}://${req.get("host")}${req.originalUrl}`);
 	// Only has cloud instance
 	if (servers.hasCloudServer && servers.gheServerInfos.length == 0) {
-		res.redirect(`/github/create-branch${url.search}`);
+		logger.info("redirecting to cloud.");
+		res.set("Authorization", req.headers.authorization);
```
gxueatlassian, line 48: Sorry what is `res.set("Authorization"` for? Will the client js pick up the Auth value from the header?

joshkay10: ahh good catch that was was first attempt to force it there. 1 moment

krazziekay (approved):

PR #2260, opened 2023-07-13
Arc 2272 allow installation GitHub app

gxueatlassian:
**What's in this PR?**
Add an installation button for customer to install cloud github app

**Why**
Users need to install app into their github acc

**Added feature flags**
N/A

**Affected issues**  
ARC-2272

**How has this been tested?**  
Unit test, local

**Whats Next?**
Some design/ux review?


spa/src/oauth-manager/index.tsx
```diff
@@ -45,6 +47,23 @@ const OauthManager = () => {
 		return false;
 	}
 
+	async function installNewApp(onFinish: () => void) {
+		const app = await ApiRequest.gitHubApp.getAppNewInstallationUrl();
+		const exp = new Date(new Date().getTime() + FIFTEEN_MINUTES_IN_MS);
+		document.cookie = `is-spa=true; expires=${exp.toUTCString()}; path=/; SameSite=None; Secure`;
+		const winInstall = window.open(app.data.appInstallationUrl, "_blank");
+		const hdlWinInstall = setInterval(() => {
```
krazziekay, line 55: I think its better to use the `postMessage` similar to the how its used in `ConfigSteps.tsx`?
Simply pass a message after closing the child window and grab that message here and implement this method.

spa/src/oauth-manager/index.tsx
```diff
@@ -45,6 +47,23 @@ const OauthManager = () => {
...
+			if (winInstall?.closed) {
+				try {
+					document.cookie = "is-spa=;expires=Thu, 01 Jan 1970 00:00:00 GMT; path=/; SameSite=None; Secure";
```
krazziekay, line 58: Can't we just clear the cookie?

spa/src/api/apps/index.ts
```diff
@@ -0,0 +1,9 @@
+import { GetGitHubAppsUrlResponse } from "../../rest-interfaces/oauth-types";
+import AxiosInstance from "../axiosInstance";
+import { AxiosResponse } from "axios";
+
+const GitHubApps = {
+	getAppNewInstallationUrl: async (): Promise<AxiosResponse<GetGitHubAppsUrlResponse>> => AxiosInstance.get("/rest/app/cloud/installation/new"),
```
krazziekay, line 6: Do we need the `async` here? The async is already implied by the Promise definition.

spa/src/oauth-manager/index.tsx
gxueatlassian: Yup, will try to fix that.

spa/src/oauth-manager/index.tsx
gxueatlassian: Hmm, I don't know how to "clearCookie", it seems this is the way to "remove" cookie in js?

spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -183,6 +183,12 @@ const ConfigSteps = () => {
 		navigate("/spa/connected");
 	};
 
+	const installNewOrg = async () => {
+		await OAuthManagerInstance.installNewApp(() => {
```
krazziekay, line 187: /nit
```suggestion
		await OAuthManagerInstance.installNewApp(getOrganizations)
```

spa/src/api/apps/index.ts
gxueatlassian: Probably don't, can fix that.

spa/src/pages/ConfigSteps/index.tsx
gxueatlassian: Actually I leave this as it was, feels it could be more flexible in future.

spa/src/api/apps/index.ts
gxueatlassian: fixed.

spa/src/oauth-manager/index.tsx
gxueatlassian: Can we merged this one for now? I need to think about how to structure the code. Don't want to add duplicate events listener for the clearing.

src/rest/rest-router.ts
```diff
@@ -21,6 +22,8 @@ subRouter.use(JwtHandler);
 
 subRouter.use("/oauth", OAuthRouter);
 
+subRouter.use("/installation", GitHubAppsRoute);
```
krazziekay, line 25: \nit 
Can we name this GitHubAppsRouter, just to make it similar to the other routes?

krazziekay (approved):

PR #2267, opened 2023-07-20
ARC-1913 and SPOT-197: Remove admins permissions to call /deleteInstallation

atrigueiro:
**What's in this PR?**
Just removing the permissions for us to call the `deleteInstallation` endpoint manually via the cli. It's currently being called by Pollinator only, so they will continue to do so (policies are already defined in the `extras-stg` and `extras-prod` policies).

**Why**
[Spotlight](https://hello.atlassian.net/wiki/spaces/JST/pages/2101826308/Spotlight+Agile+and+Open+Toolchain+dangerous+operations+audit) project (response to last year's sev 0).
Removing operations that can be considered dangerous.

**Added feature flags**
None.

**Affected issues**  
ARC-1913
SPOT-197

**How has this been tested?**  
Added a new test on `mais-test.json`

### DIFF
--- a/etc/poco/bundle/main.json
+++ b/etc/poco/bundle/main.json
@@ -51,7 +51,8 @@
         "/api/**"
       ],
       "methods": [
-        "*"
+        "GET",
+				"POST"
       ],
       "principals": {
         "staff": {
--- a/etc/poco/bundle/main-test.json
+++ b/etc/poco/bundle/main-test.json
@@ -152,7 +152,6 @@
 			"principals": [],
 			"allowed": true
 		},
-
 		{
 			"name": "Allow external users to disconnect github account",
 			"path": "/jira/configuration",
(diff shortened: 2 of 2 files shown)


etc/poco/bundle/main.json
```diff
@@ -51,7 +51,8 @@
         "/api/**"
       ],
       "methods": [
-        "*"
+        "GET",
```
atrigueiro, line 54: I had a look at all routers we have that do a DELETE call and the only endpoints are:
/deleteInstallation 
/:jiraHost
/repo-sync-states

Also had a look in Splunk and those other endpoints haven't been called in a month so I think it's ok to restrict for now and if we do come across the need to allow one or both of them again, we can do so.

joshkay10 (approved):
Yeah its pretty rare that we call them from CLI

PR #2274, opened 2023-07-24
Arc 2274 analytics for spa

gxueatlassian:
**What's in this PR?**
1. Add optional analytics events
2. Replace the build scripts with react-scripts

**Why**
1. Because we want to send analytics events
2. Because the previous build scripts `vite` is not well thought. It is using native esm module, and dynamic import on the optional dependency will always trigger a complain on it. So I update the use the react-scripts, which transform the `require` to return `undefined` on missing dep. Also this means the bundled out is determined at compile time, instead of runtime (at users' browser')

**Added feature flags**
N/A

**Affected issues**  
ARC-2274

**How has this been tested?**  
Unit
Local

**Whats Next?**
Add more events


spa/.yarnrc
```diff
@@ -0,0 +1,2 @@
+--install.ignore-engines true
```
gxueatlassian, line 1: One of the sub deps in  latest react-scripts requires node16. 
We are using node14.
So ignore it for now, will upgrade later.

spa/package.json
```diff
@@ -2,14 +2,27 @@
   "name": "spa",
   "private": true,
   "version": "0.0.0",
-  "type": "module",
+  "homepage": "/spa",
```
gxueatlassian, line 5: `module` is used by `vite`
`homepage` is used by `react-scripts`

src/main.ts
```diff
@@ -5,21 +5,14 @@ import throng from "throng";
...
-	const server = app.listen(port, () => {
+	app.listen(port, () => {
 		getLogger("frontend-app").info(`started at port ${port}`);
 	});
-
-	/**
-	 * Running Proxy for Web sockets for running SPA locally,
-	 * Only for Dev environments for hot reload
-	 */
-	proxyLocalWSForDev(server);
```
gxueatlassian, line 22: This is removed as the react-scripts can directly direct user browser to connect websocker on provided port

package.json
```diff
@@ -43,7 +43,7 @@
 		"db:dev": "run-s db:create:dev db:migrate:dev",
 		"db:test": "run-s db:create:test db:migrate:test",
 		"prepare": "husky install",
-		"precommit": "run-p lint build",
+		"precommit": "run-p lint",
```
gxueatlassian, line 46: it is too slow, let's remove it.
ps: app should not rely on this to ensure code in main branch is compilable anway, so it should be safe to remove it.
ps2: the github onpush.yml will do a build as well.

spa/.yarnrc
```diff
@@ -0,0 +1 @@
+--ignore-engines true
```
gxueatlassian, line 1: This is because one of the sub component in the latest react-script replies on node16. 
But the base node version on root folder relies on node 14 (dockerfile). 
I think we need to upgrade to node 16 later.

spa/src/index.tsx
```diff
@@ -1,39 +1,7 @@
 import React from "react";
 import ReactDOM from "react-dom/client";
 import "@atlaskit/css-reset";
-import { setGlobalTheme } from "@atlaskit/tokens";
```
gxueatlassian, line 4: This is due to switch to react-script, it emits many more warnings.
This is to extract the separate file so that the hot reload works.

spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -135,17 +137,17 @@ const ConfigSteps = () => {
 		setLoaderForOrgFetching(true);
 		const response = await AppManager.fetchOrgs();
 		if (response) {
-			setOrganizations(response?.orgs.map((org: any) => ({
+			setOrganizations(response?.orgs.map((org) => ({
 				label: org.account.login,
-				value: org.id,
+				value: String(org.id),
```
gxueatlassian, line 142: According to the typing, this has to be string.

spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -316,7 +319,12 @@ const ConfigSteps = () => {
 								isLoading={loaderForOrgFetching}
 								onChange={(value) => {
 									setOrgConnectionDisabled(false);
-									setSelectedOrg(value);
+									if(value) {
+										setSelectedOrg({
+											label: value.label,
+											value: parseInt(value.value)
```
gxueatlassian, line 325: And this need to be parsed back to Int

src/spa-proxy.ts
```diff
@@ -8,10 +7,10 @@ const SPA_PATH = "/spa";
 const proxy = httpProxy.createProxyServer({
 	target: {
 		host: "localhost",
-		port: 5173,
+		port: 3000,
```
gxueatlassian, line 10: react-script port is 3000 instead of 5173 from vite

spa/config-overrides.js
```diff
@@ -0,0 +1,8 @@
+//https://stackoverflow.com/questions/70591567/module-not-found-error-cant-resolve-fs-in-react
+module.exports = function override(config, env) {
+  console.log("React app rewired works!")
+  config.resolve.fallback = {
```
gxueatlassian, line 4: So that on github when building and package not found, it will return something instead. 
ps: funny it doesn't return `false`, it return some empty object instead.

spa/package.json
```diff
@@ -2,14 +2,25 @@
...
-    "dev": "vite",
-    "build": "tsc && vite build",
+    "dev": "BROWSER=none WDS_SOCKET_HOST=localhost WDS_SOCKET_PORT=3000 react-app-rewired start",
+    "build": "tsc && react-app-rewired build",
     "lint": "eslint src --ext ts,tsx --report-unused-disable-directives --max-warnings 0",
-    "preview": "vite preview",
     "test": "jest"
   },
+  "browserslist": {
+    "production": [
```
gxueatlassian, line 13: Definitely need some discussion later. It is always a trade off of bundle size vs browser compatibility

spa/public/index.html
```diff
@@ -0,0 +1,14 @@
+<!DOCTYPE html>
+<html lang="en">
```
krazziekay, line 2: Do we need this html file? If we run a `yarn build`, doesn't that create the html file?

spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -183,6 +185,7 @@ const ConfigSteps = () => {
 			case 1: {
 				setLoaderForLogin(true);
 				try {
+					analyticsClient.sendUIEvent({ actionSubject: "authorizeButton", action: "clicked" });
```
krazziekay, line 188: I think the name of the action should be more specific here.
```suggestion
					analyticsClient.sendUIEvent({ actionSubject: "authorizeToGitHubCloud", action: "clicked" });
```

krazziekay (commented):
There are some conflicts with the latest merge, please resolve them too.

spa/public/index.html
gxueatlassian: Yes we need this, as the reactscripts relies on this.

krazziekay (approved):

PR #2276, opened 2023-07-25
ARC-2316 add jirahost to analytics client

bgvozdev:
**What's in this PR?**
- Passing through jiraHost to analytics client
- Add logging

**Why**
- We will need to enrich the payload with a special properties based on jiraHost (next PR), see the description of the ticket
- To figure out the volume of events we are sending, per node.js node. That should help us to decide how to implement the enrichment


src/util/analytics-client.ts
```diff
@@ -10,12 +10,12 @@ const logger = getLogger("analytics");
...
-	(eventType: "track", attributes: { name: string, source: string } & Record<string, unknown>);
-	(eventType: "ui" | "operational", attributes: Record<string, unknown>);
-} = (eventType: string, attributes: Record<string, unknown> = {}): void => {
+	(jiraHost: string, eventType: "screen", attributes: { name: string } & Record<string, unknown>);
+	(jiraHost: string, eventType: "track", attributes: { name: string, source: string } & Record<string, unknown>);
+	(jiraHost: string, eventType: "ui" | "operational", attributes: Record<string, unknown>);
+} = (jiraHost: string, eventType: string, attributes: Record<string, unknown> = {}): void => {
 
-	logger.debug(analyticsClient ? "Found analytics client." : `No analytics client found.`);
+	logger.info({ jiraHost }, analyticsClient ? "Found analytics client." : `No analytics client found.`);
```
gxueatlassian, line 18: /nit 
Are you sure to turn this to info? As there will be a lot of logs?

gxueatlassian (approved):
Is there a reason why the `jiraHost` is not actually being sent in this PR?

src/util/analytics-client.ts
bgvozdev: That's exactly what I want to check, how much events do we send per jiraHost :) I'll remove it once we have this data

PR #2280, opened 2023-07-26
ARC-2322 adding logs to help debug a couple of dos tickets

rachellerathbone:
**What's in this PR?**
Additional, temporary logs.

**Why**
To try and pin down why some customers are missing data.

**Added feature flags**
temp-logs-for-dos-tickets

**Affected issues**  
ARC-2322

**How has this been tested?**  
Locally

**Whats Next?**
Find the reason why customers aren't seeing data :blob_pray: then circle back and clean these logs up.


src/config/feature-flags.ts
```diff
@@ -29,6 +29,7 @@ export enum BooleanFlags {
 	ENABLE_GENERIC_CONTAINERS = "enable-generic-containers",
 	ENABLE_GITHUB_SECURITY_IN_JIRA = "enable-github-security-in-jira",
 	DELETE_MESSAGE_ON_BACKFILL_WHEN_OTHERS_WORKING_ON_IT = "delete-message-on-backfill-when-others-working-on-it",
+	TEMP_LOGS_FOR_DOS_TICKETS = "temp-logs-for-dos-tickets",
```
gxueatlassian, line 32: Can you use `VERBOSE_LOGGING` instead?

gxueatlassian: Or does it make sense to use `VERBOSE_LOGGING`?

rachellerathbone: I can. Just wondering if it's going to add any extra value though?

rachellerathbone: Actually, I just checked that feature flag and it looks like it's already on for several customers. I really want to limit this to the 2 customers who I'm helping out. Starting with one that has given permission for me to log this additional data (waiting for the other one to get back to me).

src/sqs/backfill-error-handler.ts
```diff
@@ -19,10 +20,16 @@ const handleTaskError = async (sendSQSBackfillMessage: (message, delaySec, logge
 		receiveCount: context.receiveCount,
 		lastAttempt: context.lastAttempt
 	});
-	log.info("Handling error task");
+
+	const logAdditionalData = await booleanFlag(BooleanFlags.VERBOSE_LOGGING, context.payload.installationId.toString());
```
rachellerathbone, line 24: I don't think there's any way to access the JH here so thought I'd give this a crack instead?

gxueatlassian (approved):

PR #2283, opened 2023-07-27
ISOC-3690 Secret scanning webhook handler

Harminder84:
**What's in this PR?**
Adding support to listen to `secret_scanning_alert` webhook event, transform the payload to Jira vulnerability and submit to Jira.
Updated `@octokit/webhooks-types` to latest version as support for secret scanning added in version `7.x`

**Why**
Part of the work to integrate GitHub with Security in Jira

**Added feature flags**

**Affected issues**  
[ISOC-3690]

**How has this been tested?**  
Unit test
Local testing

[ISOC-3690]: https://softwareteams.atlassian.net/browse/ISOC-3690

### DIFF
--- a/src/routes/github/webhook/webhook-receiver-post.ts
+++ b/src/routes/github/webhook/webhook-receiver-post.ts
@@ -21,6 +21,7 @@ import { dependabotAlertWebhookHandler } from "~/src/github/dependabot-alert";
 import { BooleanFlags, booleanFlag } from "~/src/config/feature-flags";
 import { Subscription } from "~/src/models/subscription";
 import { extraLoggerInfo } from "./webhook-logging-extra";
+import { secretScanningAlertWebhookHandler } from "~/src/github/secret-scanning-alert";
 
 export const WebhookReceiverPost = async (request: Request, response: Response): Promise<void> => {
 	const eventName = request.headers["x-github-event"] as string;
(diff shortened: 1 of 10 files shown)


src/routes/github/webhook/webhook-receiver-post.ts
```diff
@@ -135,11 +136,16 @@ const webhookRouter = async (context: WebhookContext) => {
...
 			break;
 		case "dependabot_alert":
-			subscriptions = await Subscription.findOneForGitHubInstallationId(context.payload.installation.id, undefined);
+			subscriptions = await Subscription.findOneForGitHubInstallationId(context.payload.installation.id, context.gitHubAppConfig.gitHubAppId);
 			if (subscriptions && await booleanFlag(BooleanFlags.ENABLE_GITHUB_SECURITY_IN_JIRA, subscriptions.jiraHost)) {
 				await GithubWebhookMiddleware(dependabotAlertWebhookHandler)(context);
 			}
 			break;
+		case "secret_scanning_alert":
+			subscriptions = await Subscription.findOneForGitHubInstallationId(context.payload.installation.id, context.gitHubAppConfig.gitHubAppId);
```
gxueatlassian, line 145: /nit 
I have a gut feeling that the availability of `subscription` will be checked inside the `GithubWebhookMiddleware`, so not really need to check here. 

I understand we need to jiraHost, but we can put the FF check inside the method of `secretScanningAlertWebhookHandler` where jiraHost is available

gxueatlassian (dismissed):

src/routes/github/webhook/webhook-receiver-post.ts
rachellerathbone: Good call out. We do check for subscriptions in GithubWebhookMiddleware. I vote we rip this check out (alogn with the one or dependabot_alert.

src/transforms/transform-secret-scanning-alert.ts
```diff
@@ -0,0 +1,57 @@
...
+	return {
+		vulnerabilities: [{
+			schemaVersion: "1.0",
+			id: `d-${transformRepositoryId(repository.id, githubClientConfig.baseUrl)}-${alert.number}`,
+			updateSequenceNumber: Date.now(),
+			containerId: transformRepositoryId(repository.id, githubClientConfig.baseUrl),
+			displayName: alert.secret_type_display_name || `${alert.secret_type} secret exposed`,
+			description:  "Secret scanning alert",
+			url: alert.html_url,
+			type: "sca",
```
mjordan-atlassian, line 23: I guess this is actually `sast` since this is scanning your source code for exposed secrets rather than your dependencies.

src/transforms/transform-secret-scanning-alert.ts
```diff
@@ -0,0 +1,57 @@
...
+			introducedDate: alert.created_at,
+			lastUpdated: alert.updated_at || alert.created_at,
+			severity: {
+				level: JiraVulnerabilitySeverityEnum.CRITICAL
+			},
+			identifiers: [],
+			status: transformGitHubStateToJiraStatus(context.action, context.log),
```
mjordan-atlassian, line 30: Won't we fetch the `state` here? Should use that instead and then the `context.action` as a fallback?

rachellerathbone (approved):

PR #2297, opened 2023-07-31
ARC-2275 add ff and add drop button

gxueatlassian:
**What's in this PR?**
1. Add ff
2. Add drop experience button

**Added feature flags**
enable-5ku-experience--cloud-connect

**Affected issues**  
ARC-2275

**How has this been tested?**  
Local

**Whats Next?**

### DIFF
--- a/spa/src/pages/ConfigSteps/index.tsx
+++ b/spa/src/pages/ConfigSteps/index.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from "react";
+import { useEffect, useState, useCallback } from "react";
 import Button, { LoadingButton } from "@atlaskit/button";
 import styled from "@emotion/styled";
 import SyncHeader from "../../components/SyncHeader";
@@ -138,7 +138,9 @@ const ConfigSteps = () => {
 		});
 	};
 
-	const getOrganizations = async () => {
+	const lastTokenGeneratedTime = OAuthManager.getLastTokenGeneratedTime();
+	const getOrganizations = useCallback(async () => {
+		console.log("fetch orgs", { lastTokenGeneratedTime });
 		setLoaderForOrgFetching(true);
 		const response = await AppManager.fetchOrgs();
 		setLoaderForOrgFetching(false);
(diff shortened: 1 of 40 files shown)


spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -167,6 +165,7 @@ const ConfigSteps = () => {
 		}
 	}, []);
 
+	/*eslint-disable react-hooks/exhaustive-deps*/
```
krazziekay, line 168: Do we need this disabling comment anymore?

src/config/feature-flags.ts
```diff
@@ -75,9 +76,11 @@ const getLaunchDarklyValue = async <T = boolean | string | number>(flag: Boolean
 };
 
 // Include jiraHost for any FF that needs to be rolled out in stages
-export const booleanFlag = async (flag: BooleanFlags, key?: string): Promise<boolean> =>
+export const booleanFlag = async (flag: BooleanFlags, key?: string): Promise<boolean> => {
+	if (flag === BooleanFlags.USE_NEW_5KU_SPA_EXPERIENCE) return true;
```
krazziekay, line 80: Why are we returning true for this flag check? Can't we just set it from Launchdarkly?

src/routes/jira/jira-get.ts
```diff
@@ -160,13 +161,15 @@ const renderJiraCloudAndEnterpriseServer = async (res: Response, req: Request):
 
 	const hasConnections =  !!(installations.total || gheServers?.length);
 
+	const useNewSPAExperience = await booleanFlag(BooleanFlags.USE_NEW_5KU_SPA_EXPERIENCE, jiraHost);
```
krazziekay, line 164: What if we change the URL in the `atlassian-connect.json` instead of this?
Cause this firstly loads all the info for the current GitHub for Jira app, then does a redirect. We don't need to load anything from the current app.

spa/src/pages/ConfigSteps/index.tsx
gxueatlassian: haha no, must be from previous

src/config/feature-flags.ts
gxueatlassian: good catch!

src/routes/jira/jira-get.ts
gxueatlassian: Actually it doesn't `then does a redirect`
This page is the home page, if the user click the left sidebar, it will goes to home page. So I still want user able to go to home page at some point.

krazziekay (approved):

PR #2306, opened 2023-08-02
Arc 2324 fix popup in the middle

gxueatlassian:
**What's in this PR?**
1. Fix the popup to be middle of screen ( and same monitor in case of dual monitor, but I can't screenshot multi monitor though)
2. Remove the loading effect of the authorise button, as user might not finish the authorise flow but directly close
image

**Why**

**Added feature flags**

**Affected issues**  
ARC-2324

**How has this been tested?**  
Local

**Whats Next?**
N/A

### DIFF
--- a/spa/src/pages/ConfigSteps/index.tsx
+++ b/spa/src/pages/ConfigSteps/index.tsx
@@ -17,6 +17,7 @@ import OAuthManager from "../../services/oauth-manager";
 import analyticsClient from "../../analytics";
 import { AxiosError } from "axios";
 import { ErrorObjType, modifyError } from "../../utils/modifyError";
+import { popup } from "../../utils";
 
 type GitHubOptionType = {
 	selectedOption: number;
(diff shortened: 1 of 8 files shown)


spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -325,14 +321,7 @@ const ConfigSteps = () => {
 					{
 						isLoggedIn ? <>
 							{
-								loaderForLogin ? <>
```
krazziekay, line 328: Why is this loader being removed? Without this loader, there will be a sudden jumpy kinda effect in case of slow networks.

gxueatlassian: Yeah, happy to chat about this.
So right now there's a problem when I clicked the auth button. If I close the popup window directly, this auth button stays in "loading" state, I can't retry my login.

So we can fix either:
1. Use an `setInterval` to monitor close of sub window and remove the "loading" state if close.
2. Like in this PR, just remove the "loading" state. I think there won't be "jumpy" for that screen is because in order to have "jumpy", you have to have `isLoggedIn` and `loaderForLogin` change in some states. I still don't see when will that skeleton shown, as if `isLoggedIn` is true, `loaderForLogin` won't be true right?

krazziekay: We can remove the loading state for the content inside login step, but I still think we need the loader for the Authorize button. If not with loader, then something else to at least disable the button, otherwise the user can just spam click the Authorize button sending multiple requests.

gxueatlassian: Yeah thanks @krazziekay , let's chat this in another PR. 
Revert some of the changes.

krazziekay (approved):

PR #2307, opened 2023-08-02
ISOC-3776 Installation webhook handler

Harminder84:
**What's in this PR?**
Implemented `installation` webhook handler to check whether security permissions has been accepted or not. If yes, then
 - update`isSecurityPermissionsAccepted` column in `Subscriptions` table to true
 -  Submit request to Jira to link security workspace

**Why**
We are introducing new permissions to access `dependabot_alerts` and `secret_scanning`. Once these permissions are made available in the GH app, the GH Admin will need to accept them. We will use this data to track customers whose permission requests are still pending. Backfill for security VULNS will also be triggered on permission acceptance 

**Added feature flags**

**Affected issues**  
[ISOC-3776]

**How has this been tested?**  
Unit test
Local

[ISOC-3776]: https://softwareteams.atlassian.net/browse/ISOC-3776

### DIFF
--- a/src/github/installation.ts
+++ b/src/github/installation.ts
@@ -1,0 +1,101 @@
+import { WebhookContext } from "routes/github/webhook/webhook-context";
+import { JiraClient } from "../jira/client/jira-client";
+import { BooleanFlags, booleanFlag } from "../config/feature-flags";
+import { InstallationEvent } from "@octokit/webhooks-types";
+import { Subscription } from "../models/subscription";
+import { submitSecurityWorkspaceToLink } from "../services/subscription-installation-service";
+import { Installation } from "../models/installation";
+import { emitWebhookProcessedMetrics } from "../util/webhook-utils";
+import Logger from "bunyan";
+
+const SECURITY_PERMISSIONS = ["secret_scanning_alerts", "security_events", "vulnerability_alerts"];
+
+export const installationWebhookHandler = async (
+	context: WebhookContext<InstallationEvent>,
+	jiraClient: JiraClient,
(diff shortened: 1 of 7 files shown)


src/github/installation.ts
```diff
@@ -0,0 +1,101 @@
...
+	}
+};
+
+const hasSecurityPermissions = (permissions: InstallationEvent["installation"]["permissions"]) => {
+	return SECURITY_PERMISSIONS.every(securityPermission => securityPermission in permissions);
+};
+
+const setSecurityPermissionAccepted = async (subscription: Subscription, logger: Logger) => {
+	try {
+		if (subscription) {
```
gxueatlassian, line 95: no need for `if`?

src/github/installation.ts
```diff
@@ -0,0 +1,101 @@
...
+			}
+		},
+		gitHubAppConfig: {
+			gitHubAppId
+		}
+	} = context;
+
+	const jiraHost = jiraClient.baseURL;
+
+	const logger = context.log.child({
```
gxueatlassian, line 37: I don't think we need those
```
const logger = context.log.child({
		gitHubInstallationId,
		jiraHost
	});
```
Should be already set.

gxueatlassian (changes requested):
Looks good ✅  to me, let me know once db scripts are ran for all env, then I can unblock this one.

Harminder84:
@gxueatlassian, I ran the DB scripts for all three environments.

gxueatlassian (approved):

PR #2308, opened 2023-08-03
Arc 2327 send ad hoc sentry error

gxueatlassian:
**What's in this PR?**
Manually capture error

**Why**
Sentry won't capture those as they are caught and default value is used.

**Added feature flags**
N/A

**Affected issues**  
ARC-2327

**How has this been tested?**  
Local

**Whats Next?**

### DIFF
--- a/spa/src/utils/index.ts
+++ b/spa/src/utils/index.ts
@@ -10,14 +10,13 @@ export function popup (url: string, opts: { width: number, height: number }) {
 	return window.open(url, "_blank", `popup,${popup_params(opts.width, opts.height)}`);
 }
 
-export function reportError(err: Error) {
+export function reportError(err: unknown) {
 	try {
 		Sentry.captureException(err);
 	} catch (_) {
 		//donothing
 	}
 }
-
 
 //https://stackoverflow.com/a/4682246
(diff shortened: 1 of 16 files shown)


spa/src/utils/index.ts
```diff
@@ -8,6 +10,14 @@ export function popup (url: string, opts: { width: number, height: number }) {
 	return window.open(url, "_blank", `popup,${popup_params(opts.width, opts.height)}`);
 }
 
+export function reportError(err: unknown) {
+	try {
+		Sentry.captureException(err);
+	} catch (_) {
+		//do nothing
```
joshkay10, line 17: /nit Do we want to log this out for us still, in case Sentry has a cry about something

```
catch (err) {
 console.error("Sentry failed blah blah blah", error)
}
```

joshkay10 (approved):

spa/src/utils/index.ts
gxueatlassian: Not sure, in fact, in local, since sentry will not be set, nothing will happen ( and no error thrown anyway )
So unless you grab the (prod) sentry token in set it in your local env, you won't see a thing as sentry will NOT be inited locally.

PR #2313, opened 2023-08-04
NONE fix PRs with deleted branches

bgvozdev:
**What's in this PR?**
Fixing backfilling of PRs which feature branches were deleted

**Why**
Because the customers are complaining that they cannot find their PRs in Jira

**Added feature flags**
Existing one (we turned of GraphQL calls for now and opting for REST)

**Affected issues**  
DISTURBED

**How has this been tested?**  
Locally, tests

**Whats Next?**
!!! Currently blocked because PR GraphQL query is not working and throwing exceptions !!! @joshkay10 is preparing a fix


src/transforms/transform-pull-request.ts
```diff
@@ -198,15 +200,21 @@ export const transformPullRequest = (_jiraHost: string, pullRequest: pullRequest
...
 		issueKeys,
 		lastUpdate: pullRequest.updatedAt,
 		reviewers: mapReviews(pullRequest.reviews?.nodes, pullRequest.reviewRequests?.nodes),
-		sourceBranch: pullRequest.headRef?.name || "",
-		sourceBranchUrl: `https://github.com/${pullRequest.headRef?.repository?.owner?.login}/${pullRequest.headRef?.repository?.name}/tree/${pullRequest.headRef?.name}`,
+		sourceBranch: pullRequest.headRefName,
+		...(
+			pullRequest.headRef
+				? {
+					sourceBranchUrl: `https://github.com/${pullRequest.headRef?.repository?.owner?.login}/${pullRequest.headRef?.repository?.name}/tree/${pullRequest.headRef?.name}`
```
gxueatlassian, line 214: Need a bit of thinking for GHE as it won't be github.com

gxueatlassian: I think we can use headRef.repository.url to replace the need for `github.com`

bgvozdev: Nice find @gxueatlassian , but I think that exactly the same logic that was before. We probably should raise this as a bug and not consider it as a blocker, wdyt?

joshkay10 (approved):
LGTM

PR #2314, opened 2023-08-04
ISOC 3689 add git hub code scanning webhook

auzwang:
**What's in this PR?**
Adds Security in Jira `code_scanning_alert` webhook support. GitHub [Security Code Scanning Alert webhooks](https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#code_scanning_alert) will be transformed and submitted as [Security Vulnerabilities in Jira](https://developer.atlassian.com/cloud/jira/software/rest/api-group-security-information/#api-rest-security-1-0-linkedworkspaces-bulk-post).

Vulnerabilities are only submitted from the default or protected branch. This is done by checking
```TypeScript
alert.most_recent_instance?.ref?.startsWith("refs/heads")
```
When a Code Scan Alert is found in a PR, then merged into the default or protected branch, a webhook with the action `appeared_in_branch` is triggered with the same alert number, which is subsequently submitted to Jira.

Remote Links will continue be handled in the same way.

Fixes an error where Code Scanning Alert Remote Link Webhook metrics were not being processed.

```diff
- context.payload.webhookReceived
+ context.webhookReceived
```

**Why**
Integrate GitHub with Security in Jira

**Added feature flags**

**Affected issues**
[ISOC-3689](https://softwareteams.atlassian.net/browse/ISOC-3689)

**How has this been tested?**  
Locally
Unit Tests

**Questions**
We are emitting a new webhook metric for Code Scanning Alert Security Vulnerabilities (`code_scanning_alert_security`) in addition to the Remote Links handler. Does this make sense?

[ISOC-3689]: https://softwareteams.atlassian.net/browse/ISOC-3689


src/transforms/transform-code-scanning-alert.ts
```diff
@@ -101,3 +109,43 @@ export const transformCodeScanningAlert = async (context: WebhookContext, github
...
+			displayName: alert.rule.name,
+			description: alert.rule.full_description || alert.rule.description,
+			url: alert.html_url,
+			type: "sast",
+			introducedDate: alert.created_at,
+			lastUpdated: alert.dismissed_at || alert.fixed_at || alert.updated_at || alert.created_at,
+			severity: {
+				level: transformGitHubSeverityToJiraSeverity(alert.rule.security_severity_level, handleUnmapped)
+			},
+			identifiers: [],
```
mjordan-atlassian, line 144: This field is actually required. I didn't specify it in the RFC since I wasn't sure yet what value to use here. 

We will need to explore the payload to see what values we can use here, there is an `alert.rule.tags` but doesn't seem to have the URL so we might need to build this ourselves. There is a URL in `alert.rule.help` but I'm not sure if that is available for all action types and backfilling so will need to check.

src/transforms/transform-code-scanning-alert.ts
```diff
@@ -101,3 +109,43 @@ export const transformCodeScanningAlert = async (context: WebhookContext, github
...
+		return null;
+	}
+
+	const metrics = {
+		trigger: "webhook",
+		subTrigger: "code_scanning_alert"
+	};
+	const gitHubInstallationClient = await createInstallationClient(githubInstallationId, jiraHost, metrics, context.log, context.gitHubAppConfig?.gitHubAppId);
+
+	const handleUnmapped = (state) => context.log.info(`Received unmapped state from code_scanning_alert webhook: ${state}`);
```
mjordan-atlassian, line 127: Can we also add a `field` param here so we can tell if it was the `severity` or `status` that received the unmapped value.

CONTRIBUTING.md
```diff
@@ -76,6 +76,7 @@ Once you've set up your GitHub app and cloned this repo, copy the file `.env.dev
 + `PRIVATE_KEY_PATH`: You'll also need to generate a new private key on your GitHub app page, download it, move it to the source root of this repo, and set `PRIVATE_KEY_PATH=<your-private-key-name>.pem`
 + `ATLASSIAN_URL`: The URL for the Jira instance you're testing on. If you don't have one now, [please set the value of this variable from the steps mentioned here](#create-your-jira-instance).
 + `APP_KEY`: Your Jira app key - need to be unique for your development app
++ `WEBHOOK_SECRETS`: `["development"]` as previously set up in the GitHub app page.
```
gxueatlassian, line 79: 👍

gxueatlassian:
One small thing to double check, for

```
Vulnerabilities are only submitted from the default or protected branch. This is done by checking
alert.most_recent_instance?.ref?.startsWith("refs/heads")
```

I think `refs/heads` can also means any branches? We are accepting alerts from any branches, is that correct? But I couldn't be wrong.

gxueatlassian (approved):

auzwang:
> One small thing to double check, for
> 
> ```
> Vulnerabilities are only submitted from the default or protected branch. This is done by checking
> alert.most_recent_instance?.ref?.startsWith("refs/heads")
> ```
> 
> I think `refs/heads` can also means any branches? We are accepting alerts from any branches, is that correct? But I could be wrong.

Code scanning can be configured to trigger on push and pull requests to projected branches. So pushes to unprotected branches such as `feature/abc-123` won't trigger code scan alerts until a PR is made targeting the default branch or a protected branch.

src/transforms/transform-code-scanning-alert.ts
auzwang: Added the transformation of `alert.rule.tags` to vulnerability `identifiers`. When no CWE tags are found, `identifiers` will be omitted.

src/transforms/transform-code-scanning-alert.ts
```diff
@@ -101,3 +109,66 @@ export const transformCodeScanningAlert = async (context: WebhookContext, github
...
+		return null;
+	}
+	// CWE tags from GitHub take the format 'external/cwe/cwe-259'
+	const cwePrefix = "external/cwe/cwe-";
+	const identifiers = tags.filter(tag => tag.startsWith(cwePrefix)).map(tag => {
+		// Remove starting 0s as GitHub can provide 'cwe-079'
+		const cweId = tag.split(cwePrefix)[1].replace(/^0+/, "");
+		return {
+			displayName: `CWE-${cweId}`,
+			url: `https://cwe.mitre.org/cgi-bin/jumpmenu.cgi?id=${cweId}`
```
mjordan-atlassian, line 124: What's the difference between using this URL and `https://cwe.mitre.org/data/definitions/${cweId}` ? 
The one you are using redirects to the `data/definitions` page anyway?

auzwang: I used the URL https://cwe.mitre.org/ uses in their ID Lookup input. The difference is the error page when supplying an invalid ID. I suspect there is future-proof credence using the `cgi` URL as it would be out in the wild and reference in PDFs, hence the redirect.

Navigating using the definitions URL we get:
Screenshot 2023-08-15 at 1 27 31 pm

Whereas with the `id` param, we get:
Screenshot 2023-08-15 at 1 27 37 pm

gxueatlassian (approved):

PR #2318, opened 2023-08-07
ARC 2391 - Refactor UI to meet new design

krazziekay:
**What's in this PR?**
- UI changes 
- Minor cleanup
- Test cases updated

**Why**
- Better UX

**How has this been tested?**  
- Test cases
- Locally

**Video**
- When there are no orgs installed / all orgs are installed

https://github.com/atlassian/github-for-jira/assets/13076255/f0b81fdd-4e9e-43fc-828d-d84699175074

- When orgs are installed, but not connected (This should not happen with this new UI, but just in case)

https://github.com/atlassian/github-for-jira/assets/13076255/21ec0195-7928-4f61-97cf-e0c0e5b6c651

**Whats Next?**
- Test in staging & merge


spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -137,76 +132,21 @@ const ConfigSteps = () => {
 		});
 	};
 
-	const getOrganizations = useCallback(async () => {
+	const getOrganizations = async (autoRedirectToInstall = true) => {
```
krazziekay, line 135: Switched this back to just normal method instead of using callback and using `// eslint-disable`, otherwise there are just too many eslint errors which is happening because of the multiple usage of methods within each other.

gxueatlassian (approved):

PR #2319, opened 2023-08-08
Fix connect button still loading after error

gxueatlassian:
**What's in this PR?**
1. Restore the UI connect button after error.
2. "temporary" set an error message for lack of admin permission.

**Why**

**Added feature flags**

**Affected issues**  

**How has this been tested?**  
Local

**Whats Next?**

### DIFF
--- a/spa/src/pages/ConfigSteps/index.tsx
+++ b/spa/src/pages/ConfigSteps/index.tsx
@@ -187,13 +187,13 @@ const ConfigSteps = () => {
 		analyticsClient.sendUIEvent({ actionSubject: "switchGitHubAccount", action: "clicked" });
 	};
 
-	const doCreateConnection = async (gitHubInstallationId: number, mode: "auto" | "manual") => {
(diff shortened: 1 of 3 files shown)


spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -307,7 +307,11 @@ const ConfigSteps = () => {
 																		onClick={() => {
 																			setLoaderForOrgClicked(true);
 																			setClickedOrg(org.id);
-																			doCreateConnection(org.id, "manual");
+																			try {
```
krazziekay, line 310: Do we need the try catch here? `doCreateConnection` already has a try-catch. Can we just add the `finally` inside the `doCreateConnection`?

gxueatlassian: ?? It is the `finally` I added here.

gxueatlassian: Although I might further refactor the button into sub componnent with loading states as well.

krazziekay (approved):

PR #2322, opened 2023-08-08
None use state in new app install

gxueatlassian:
**What's in this PR?**
Use state instead of cookie to distinguish spa vs old exp.

**Why**
3rd party cookie is blocked by safari

**Added feature flags**

**Affected issues**  

**How has this been tested?**  
Local
Unit test

**Whats Next?**


krazziekay (dismissed):
🎉

krazziekay (approved):

PR #2324, opened 2023-08-08
Fix potential duplication of Subscriptions bug

gxueatlassian:
**What's in this PR?**
Move the avatarUrl from `where` to `default`

**Why**
During my local development, I from times to times find out there're duplicate Subscriptions record. I am not 100% sure why, but this `avatarUrl` looks really suspicious. 

**Added feature flags**
N/A

**Affected issues**  
N/A

**How has this been tested?**  
Local
AvatarUrl is available after change

image

**Whats Next?**


src/models/subscription.ts
```diff
@@ -198,10 +198,10 @@ export class Subscription extends Model {
 				gitHubInstallationId: payload.installationId,
 				jiraHost: payload.host,
 				jiraClientKey: payload.hashedClientKey,
-				gitHubAppId: payload.gitHubAppId || null,
-				avatarUrl: payload.avatarUrl || null
+				gitHubAppId: payload.gitHubAppId || null
 			},
 			defaults: {
+				avatarUrl: payload.avatarUrl || null,
```
bgvozdev, line 204: Do we even use it?

bgvozdev: If not, should we rather drop the column?

bgvozdev (approved):

src/models/subscription.ts
gxueatlassian: It is a new column, added by xxxx team. :) I forgot the name.

gxueatlassian: https://github.com/atlassian/github-for-jira/pull/2250/files#r1260508317

PR #2328, opened 2023-08-09
None fix multi org install popup

gxueatlassian:
**What's in this PR?**
1. Remove the cycle reference on getOrgs <--> installNewOrgs
2. Add lock to forbid open tabs multi times. 

**Why**
1. Cycle reference is always bad
2. From some reference page https://upmostly.com/tutorials/why-is-my-useeffect-hook-running-twice-in-react , it is better to model our api (in useEffect) to be idempotent (as much as we can)

**Added feature flags**

**Affected issues**  

**How has this been tested?**  
Local

**Whats Next?**

### DIFF
--- a/spa/src/pages/ConfigSteps/index.tsx
+++ b/spa/src/pages/ConfigSteps/index.tsx
@@ -255,7 +255,7 @@ const ConfigSteps = () => {
 			setLoggedInUser(OAuthManager.getUserDetails().username);
 			setLoaderForLogin(false);
 			setOrganizations([]);
-			if (status === true) {
+			if (status) {
 				const result = await getOrganizations();
 				if (result.success) {
 					if (result.orgs.length === 0) {
--- a/spa/src/pages/ConfigSteps/index.tsx
+++ b/spa/src/pages/ConfigSteps/index.tsx
@@ -255,7 +255,7 @@ const ConfigSteps = () => {
 			setLoggedInUser(OAuthManager.getUserDetails().username);
 			setLoaderForLogin(false);
 			setOrganizations([]);
-			if (status) {
+			if (status === true) {
 				const result = await getOrganizations();
 				if (result.success) {
 					if (result.orgs.length === 0) {
--- a/spa/src/pages/ConfigSteps/index.tsx
+++ b/spa/src/pages/ConfigSteps/index.tsx
@@ -132,19 +132,16 @@ const ConfigSteps = () => {
 		});
 	};
 
-	const getOrganizations = async (autoRedirectToInstall = true) => {
+	const getOrganizations = async () => {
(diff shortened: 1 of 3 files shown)


gxueatlassian:
This was still happening after the fix if we goback-goforward in the page. `useEffect` is very likely to fire off twice.

spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -257,7 +254,15 @@ const ConfigSteps = () => {
 			}
 			setLoggedInUser(OAuthManager.getUserDetails().username);
 			setLoaderForLogin(false);
-			await getOrganizations();
+			setOrganizations([]);
+			if (status === true) {
```
krazziekay, line 258:
```suggestion
			if (status) {
```

spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -257,7 +254,15 @@ const ConfigSteps = () => {
...
+				const result = await getOrganizations();
+				if (result.success) {
```
krazziekay, line 260: Can combine them instead of the nested ifs
```suggestion
				if (result.success && result.orgs.length === 0) {
```

spa/src/pages/ConfigSteps/index.tsx
gxueatlassian: Sorry it won't work, if status is instanace of AxiosError, `if(status)` will evaluate to true. 
Generally speaking, explicit coding is better, so `xxx === true` will always be clear and less error prone.

gxueatlassian: Hmm, actually it is shortcutted via 
```
if (status instanceof AxiosError) {
				setError(modifyError(status, {}, { onClearGitHubToken: clearGitHubToken }));
				return;
			}
```

So in this case will be fine. But still, I prefer to make it clear, instead of relying in some previous check and early return.

krazziekay:
> This was still happening after the fix if we goback-goforward in the page. `useEffect` is very likely to fire off twice.


The double fire of useEffect only happens in local environment. It shouldn't happen in prod. This is because of the use of `React.strict`, which makes the render happen twice but just while developing.([source](https://react.dev/reference/react/StrictMode#strictmode))

krazziekay (approved):

gxueatlassian:
> > This was still happening after the fix if we goback-goforward in the page. `useEffect` is very likely to fire off twice.
> 
> The double fire of useEffect only happens in local environment. It shouldn't happen in prod. This is because of the use of `React.strict`, which makes the render happen twice but just while developing.([source](https://react.dev/reference/react/StrictMode#strictmode))

Nice! Good to know that.

PR #2332, opened 2023-08-10
None check server cloud create branch

joshkay10:
**What's in this PR?**
Make sure that we get the cloud or server subscription correctly

Tested on stage and works a treat

**Why**
There is a bug on create-branch when it doesn't work because we return the wrong subscription

**BEFORE FIX**
Screenshot 2023-08-10 at 3 00 33 PM

**AFTER FIX**
Screenshot 2023-08-10 at 3 47 07 PM


krazziekay (approved):

iurynogueira:
This was deployed?

PR #2335, opened 2023-08-10
ISOC-3693 Secret scanning alerts backfill

Harminder84:
**What's in this PR?**
Adding support to backfill the secret scanning alerts along with the other entities. The backfill behaviour is in sync with the backfill for existing entities.

https://hello.atlassian.net/wiki/spaces/CDX/pages/2748121866/LDR+How+do+we+backfill+vulnerabilities+for+existing+GitHub+for+Jira+customers

**Why**
Part of the work to integrate GitHub with Security in Jira

**Added feature flags**

**Affected issues**  
[ISOC-3693]

**How has this been tested?**  
Unit Test

[ISOC-3693]: https://softwareteams.atlassian.net/browse/ISOC-3693


src/sync/secret-scanning-alerts.ts
```diff
@@ -0,0 +1,113 @@
...
+	logger: Logger,
+	gitHubAppId: number | undefined
+): Promise<JiraVulnerabilityBulkSubmitData> => {
+
+	const gitHubClientConfig = await getGitHubClientConfigFromAppId(gitHubAppId, jiraHost);
+
+	const vulnerabilities = alerts.map((alert) => {
+		return {
+			schemaVersion: "1.0",
+			id: `d-${transformRepositoryId(repository.id, gitHubClientConfig.baseUrl)}-${alert.number}`,
```
mjordan-atlassian, line 89: Secret scanning vulnerability IDs should start with `s` to differentiate them from dependabot and code scanning

src/sync/secret-scanning-alerts.ts
```diff
@@ -0,0 +1,113 @@
...
+			severity: {
+				level: JiraVulnerabilitySeverityEnum.CRITICAL
+			},
+			identifiers: [{
+				displayName: alert.secret_type,
+				url: alert.html_url
+			}],
+			status: transformGitHubStateToJiraStatus(alert.state, logger),
+			additionalInfo: {
+				content: alert.secret_type
```
mjordan-atlassian, line 107: We might as well add the `alert.html_url` here to the `url` property so that we can link back to this alert in the vulnerability table. We should also add that here as well https://github.com/atlassian/github-for-jira/blob/main/src/transforms/transform-secret-scanning-alert.ts#L35

mjordan-atlassian: Let's actually just remove `additionalInfo` entirely from all secret scanning alerts since we have this info as the identifier

gxueatlassian (approved):

PR #2343, opened 2023-08-14
ARC-2316: start sending analytic events to SQS

bgvozdev:
**What's in this PR?**
Sending analytics event to SQS instead of via analytics client

**Why**

**Added feature flags**
Yes

**Affected issues**  
in title

**How has this been tested?**  
in ddev

image
image

**Whats Next?**

### DIFF
--- a/src/util/analytics-client.ts
+++ b/src/util/analytics-client.ts
@@ -6,6 +6,9 @@ import { createHashWithoutSharedSecret } from "utils/encryption";
 import _, { omit } from "lodash";
 import { MicrosEnvTypeEnum } from "interfaces/common";
 import { isTestJiraHost } from "config/jira-test-site-check";
+import { booleanFlag, BooleanFlags } from "config/feature-flags";
+import AWS from "aws-sdk";
+import { SendMessageRequest } from "aws-sdk/clients/sqs";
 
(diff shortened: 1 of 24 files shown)


src/util/analytics-client.ts
```diff
@@ -34,6 +37,50 @@ const calculateClientEnv = () => {
...
+			eventAttributes: attributes
+		}
+	};
+	const sqs = new SQS({ apiVersion: "2012-11-05", region: envVars.SQS_INCOMINGANALYTICEVENTS_QUEUE_REGION });
+	const params: SendMessageRequest = {
+		MessageBody: JSON.stringify(payload),
+		QueueUrl: envVars.SQS_INCOMINGANALYTICEVENTS_QUEUE_URL,
+		DelaySeconds: 0
+	};
+	try {
```
gxueatlassian, line 75: /nit
Is there a reason the `try{` doesn't move to the very top of this method?

gxueatlassian (approved):

src/util/analytics-client.ts
bgvozdev: While "sendMessage" is anticipated to result in a error from time to time, and we are OK to ignore it (this is analytics), if there's a bug above - that's not anticipated and we should uncover this ASAP.

PR #2357, opened 2023-08-21
Arc 968 view repos addons

kamaksheeAtl:
**What's in this PR?**
add-ons and fixes:
1. make search-bar not case sensitive
2. re-render repo list on search clear
3. add debouncing to search
4. add FINISHED option in filter

**Why**
Make the feature better

**Added feature flags**
NA

**Affected issues**  
NA

**How has this been tested?**  
Jest unit test cases and locally via atlas tunnel

**Whats Next?**

### DIFF
--- a/src/routes/jira/jira-get-connected-repos.ts
+++ b/src/routes/jira/jira-get-connected-repos.ts
@@ -54,7 +54,7 @@ export const JiraGetConnectedRepos = async (
 				[Op.and]: [
 					{
 						repoName: {
-							[Op.like]: `%${repoName}%`
+							[Op.iLike]: `%${repoName}%`
 						}
 					},
 					{
@@ -72,7 +72,7 @@ export const JiraGetConnectedRepos = async (
 				[Op.and]: [
 					{
 						repoName: {
-							[Op.like]: `%${repoName}%`
+							[Op.iLike]: `%${repoName}%`
 						}
 					},
 					{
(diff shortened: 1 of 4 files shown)


static/js/jira-connected-repos.js
```diff
@@ -21,6 +23,18 @@ $(document).ready(() => {
 	$("#status-filter").on("change", () => {
 		loadRepos(1);
 	});
+	let repoSearchTimeoutId;
+	$('#repo-search').on('input', function() {
+		// re-render the original list after clearing the search bar
+		if($(this).val().length === 0) {
+			loadRepos(1)
```
krazziekay, line 30: \nit
```suggestion
			loadRepos(1);
```

static/js/jira-connected-repos.js
```diff
@@ -1,4 +1,6 @@
 $(document).ready(() => {
+	// to get the focus back on seach bar after comp reload
```
krazziekay, line 2: \nit Typo
```suggestion
	// to get the focus back on search bar after comp reload
```

krazziekay (approved):

src/routes/jira/jira-get-connected-repos.ts
```diff
@@ -54,7 +54,7 @@ export const JiraGetConnectedRepos = async (
 				[Op.and]: [
 					{
 						repoName: {
-							[Op.like]: `%${repoName}%`
```
gxueatlassian, line 57: Hi while you are here, can you also have a look at is this replacement vulnerable for the SQL injection attack? I suspect we should use something else instead of ``[Op.like]: `...${}...` ``

PR #2366, opened 2023-08-23
add attr on 5ku for orgs lack admin status

gxueatlassian:
Send fail reason for orgs to analytic events


krazziekay (dismissed):

spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -221,10 +221,12 @@ const ConfigSteps = () => {
 		try {
 			analyticsClient.sendUIEvent({ actionSubject: "connectOrganisation", action: "clicked" }, { mode });
 			const connected = await AppManager.connectOrg(gitHubInstallationId);
-			analyticsClient.sendTrackEvent({ actionSubject: "organisationConnectResponse", action: connected ? "success" : "fail"}, { mode });
 			if (connected instanceof AxiosError) {
-				showError(modifyError(connected, { orgLogin }, { onClearGitHubToken: clearGitHubToken, onRelogin: reLogin }));
+				const errorObj = modifyError(connected, { orgLogin }, { onClearGitHubToken: clearGitHubToken, onRelogin: reLogin });
```
krazziekay, line 225: \nit
We can already grab the errorCode from `connected`, as it is an instance of `AxiosError`. 
We can avoid returning the `errorCode` from `modifyError`.

gxueatlassian: Hmm to get the errorCode, we need to do this `errorCode = connected?.response?.data?.errorCode || "UNKNOWN";` .  I think I'd rather use the errorObj instead.

krazziekay (approved):

PR #2367, opened 2023-08-23
Add userAccountId to existing events

gxueatlassian:
**What's in this PR?**
Add userAccountId to existing events

**Why**
We can build a better funnel from existing evens to new spa events
Also helpful for analysing user behaviour on existing events.

**How has this been tested?**  
None.

**Whats Next?**

### DIFF
--- a/src/middleware/jira-admin-permission-middleware.ts
+++ b/src/middleware/jira-admin-permission-middleware.ts
@@ -19,7 +19,6 @@ export const fetchAndSaveUserJiraAdminStatus = async (req: Request, res: Respons
 			res.locals.accountId = userAccountId;
 		}
 
-		res.locals.accountId = userAccountId;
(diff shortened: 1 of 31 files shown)


src/middleware/jira-admin-permission-middleware.ts
```diff
@@ -15,7 +15,10 @@ export const fetchAndSaveUserJiraAdminStatus = async (req: Request, claims: Reco
 		// Can't check permissions without userAccountId
 		if (!userAccountId) {
 			return;
```
bgvozdev, line 17: Curious why are we not throwing an exception here?

bgvozdev (approved):

gxueatlassian:
It won't work. The whole point of the session cookie in existing page is because in some routes there's no JWT token. Hence unless we save that userAccountId in the session ( which incurr complexity ), we won't be to have userAccountId in all routes.

PR #2368, opened 2023-08-23
Isoc 3949 remove security provider module FF

mjordan-atlassian:
**What's in this PR?**
This PR moves the GitHub Security Feature Flag check out of the module descriptor and into the individual security routes. 

**Why**
This will allow us to easily enable GitHub security without having to wait for users to update their connect app or for the 33-hour automatic update due to the connect app descriptor version changing.

We already have a feature flag in place in Jira frontend that disables GitHub Security from showing in the Security empty states due to the presence of the security provider module.

https://app.launchdarkly.com/jira/production/features/core-x-isotopes-security-page-in-jira-multivendor-github/targeting

**How has this been tested?**  
Tests have been added for every security route to return a 403 when trying to access the route with the feature flag off.


src/routes/jira/security/workspaces/containers/jira-security-workspaces-containers-post.test.ts
```diff
@@ -103,6 +137,10 @@ describe("Repositories Post", () => {
 	});
 
 	it("Should return an empty array if no matching repositories are found", async () => {
+		when(booleanFlag).calledWith(
+			BooleanFlags.ENABLE_GITHUB_SECURITY_IN_JIRA, jiraHost
+		).mockResolvedValue(true);
```
Harminder84, line 142: This seems to be requirement for each unit test. I think, it would be nice move this piece of code to `beforeEach`

gxueatlassian (approved):

Harminder84 (approved):

PR #2369, opened 2023-08-24
ARC-2426 fix error on parsing deployment env when config is null

gxueatlassian:
**What's in this PR?**
Fix a bug in the env mapping that the RepoSyncState config mapping is null

image

**Why**
Lots of deployment DLQ msg due to 

When the `globPatterns` is null, it won't trigger javascript to use a default value `[]`. (Seems only `undefined` will be replaced with a `[]`

**Added feature flags**

**Affected issues**  
ARC-2426

**How has this been tested?**  
Unit test

**Whats Next?**


src/interfaces/common.ts
```diff
@@ -94,12 +94,7 @@ export interface Config {
 		 * globs that are used in the `mapEnvironment()` function to match a given environment with one
 		 * of the valid Jira environment types.
 		 */
-		environmentMapping?: {
-			development?: string[];
```
gxueatlassian, line 98: This mapping is supplied by user during runtime, we don't know for sure

src/services/user-config-service.ts
```diff
@@ -100,7 +100,7 @@ const hasTooManyPatternsPerEnvironment = (config: Config): boolean => {
 		return false;
 	}
 	return Object.keys(environmentMapping).some(key => {
-		return environmentMapping[key]?.length > MAX_PATTERNS_PER_ENVIRONMENT;
```
gxueatlassian, line 103: TS warning due to above type changed.

src/transforms/transform-deployment.ts
```diff
@@ -203,8 +203,8 @@ export const mapState = (state: string | undefined): string => {
 	}
 };
 
-const matchesEnvironment = (environment: string, globPatterns: string[] = []): boolean => {
```
gxueatlassian, line 206: In existing code, if `null` is passed in, the `globPatterns` will not be replaced by `[]`. 
ps: I am a bit worry about the rest of our code (where default value is used) as well, but will leave it for next time.

src/interfaces/common.ts
atrigueiro: We don't know it but that's exactly what the mapping does. Jira only accepts these 4, so if we can't match any of them, then Jira will reject the payload.

gxueatlassian: On push webhook, the config of the mappings is saved to RepoSyncState [from here](https://github.com/atlassian/github-for-jira/blob/main/src/services/user-config-service.ts#L123), which takes a free form. 

On deployment webhook, this config is read. But [further map here](https://github.com/atlassian/github-for-jira/blob/main/src/transforms/transform-deployment.ts#L247).

So in short, the current behaviour is on saving the config, we take anything user provides and keep the ordering. When use it, we map it according to what user provide/order and set to "unmapped" if the result is not a valid jira env state.

---
Let me know if above make sense. And btw, the typings here doesn't reflect the "free form" part during push webhook, so I want to change it to match it.

---
Alternatively we can restrict from "free form" and only accept "development/prod/test/etc" on push webhook, but then we need to deal with existing data and backfill them, too much trouble.

bgvozdev (approved):

PR #2370, opened 2023-08-25
[ENHANCEMENT] - Clearing the configured app property when there are no connections

krazziekay:
**What's in this PR?**
- Clearing the configured app property when there are no connections

### DIFF
--- a/src/routes/jira/jira-get.ts
+++ b/src/routes/jira/jira-get.ts
@@ -13,6 +13,7 @@ import { AnalyticsEventTypes, AnalyticsScreenEventsEnum } from "interfaces/commo
 import { getCloudOrServerFromGitHubAppId } from "utils/get-cloud-or-server";
 import { Errors } from "config/errors";
 import { booleanFlag, BooleanFlags } from "config/feature-flags";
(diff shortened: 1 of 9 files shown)


src/routes/jira/jira-get.ts
```diff
@@ -161,6 +162,11 @@ const renderJiraCloudAndEnterpriseServer = async (res: Response, req: Request):
 
 	const hasConnections =  !!(installations.total || gheServers?.length);
 
+	if (!hasConnections) {
+		const jiraClient = await JiraClient.getNewClient(installation, req.log);
+		await jiraClient.appPropertiesDelete();
```
mjordan-atlassian, line 167: Is there a reason to delete instead of set to false? This would still work for our use case just curious :)

mjordan-atlassian (approved):

src/routes/jira/jira-get.ts
krazziekay: The configure message shows up on either undefined or false, so doesn't really matter for this case. No specific reason for using delete here.

PR #2371, opened 2023-08-25
None add user account id to existing attr

gxueatlassian:
**What's in this PR?**
Add user account id to analytics event

**Why**
Better tracking in amplitude and databrick

**Added feature flags**
N/A

**Affected issues**  
N/A

**How has this been tested?**  
N/A

**Whats Next?**
N/A


gxueatlassian:
Nah. More thinking needed.

PR #2372, opened 2023-08-25
Fix some analytics events

gxueatlassian:
**What's in this PR?**
Fix some bugs on analytics events sending.

### DIFF
--- a/spa/src/pages/ConfigSteps/index.tsx
+++ b/spa/src/pages/ConfigSteps/index.tsx
@@ -220,7 +220,7 @@ const ConfigSteps = () => {
 	const doCreateConnection = async (gitHubInstallationId: number, mode: "auto" | "manual", orgLogin?: string) => {
 		try {
 			analyticsClient.sendUIEvent({ actionSubject: "connectOrganisation", action: "clicked" }, { mode });
-			const connected = await AppManager.connectOrg(gitHubInstallationId);
+			const connected: boolean | AxiosError = await AppManager.connectOrg(gitHubInstallationId);
 			if (connected instanceof AxiosError) {
 				const errorObj = modifyError(connected, { orgLogin }, { onClearGitHubToken: clearGitHubToken, onRelogin: reLogin });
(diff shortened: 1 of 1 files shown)


spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -264,14 +264,14 @@ const ConfigSteps = () => {
...
-				const response = await OAuthManager.finishOAuthFlow(event.data?.code, event.data?.state);
+				const response: boolean | AxiosError = await OAuthManager.finishOAuthFlow(event.data?.code, event.data?.state);
 				setLoaderForLogin(false);
 				if (response instanceof AxiosError) {
 					showError(modifyError(response, {}, { onClearGitHubToken: clearGitHubToken, onRelogin: reLogin }));
 					analyticsClient.sendTrackEvent({ actionSubject: "finishOAuthFlow", action: "fail" });
 					return;
 				} else {
-					analyticsClient.sendTrackEvent({ actionSubject: "finishOAuthFlow", action: "success" });
+					analyticsClient.sendTrackEvent({ actionSubject: "finishOAuthFlow", action: response === true ?  "success" : "fail" });
```
krazziekay, line 274:
```suggestion
					analyticsClient.sendTrackEvent({ actionSubject: "finishOAuthFlow", action: response ?  "success" : "fail" });
```

krazziekay (approved):

spa/src/pages/ConfigSteps/index.tsx
gxueatlassian: Nah, this is intentional, I know ts checked it, but just to make sure we are only sending "success" when the response is exactly "true"

gxueatlassian: Coz otherwise, if the `resposne` is an `string: "something wrong"`, the TS may failed to checked that since the inner code does
```
return e as AxiosError;
```
In that case, `response ?  "success" : "fail"` will return `success`

PR #2374, opened 2023-08-26
[ENHANCEMENT] - Clearing the configured app property when there are no connections

krazziekay:
**What's in this PR?**
- Set the property to false when there are no connections.


src/util/is-connected.ts
```diff
@@ -0,0 +1,17 @@
+import { Subscription } from "models/subscription";
+import { GitHubServerApp } from "models/github-server-app";
+import { Request } from "express";
+import { getConnectionsAndInstallations } from "routes/jira/jira-get";
+
+
+export const isConnected = async (req: Request, jiraHost: string, installationId: number): Promise<boolean> => {
+	const subscriptions = await Subscription.getAllForHost(jiraHost) || [];
```
gxueatlassian, line 8: Is there a reason we can't just do `return subscriptions.length >0`?

atrigueiro (approved):

src/util/is-connected.ts
```diff
@@ -0,0 +1,17 @@
...
+	const gheServers: GitHubServerApp[] = await GitHubServerApp.findForInstallationId(installationId) || [];
```
gxueatlassian, line 9: Tbh, I don't think `GitHubServerApp.findForInstallationId(installationId)` is safe, as the `installationId` can be overlap between different customers

gxueatlassian (changes requested):
Just some comments.

src/util/is-connected.ts
```diff
@@ -0,0 +1,17 @@
+import { Subscription } from "models/subscription";
+import { GitHubServerApp } from "models/github-server-app";
+import { Request } from "express";
+import { getConnectionsAndInstallations } from "routes/jira/jira-get";
```
gxueatlassian, line 4: This is a bit strange in terms of importing. `jira-delete` --> import `utils/is-connecrted` --> import `jira-get`.

src/util/is-connected.ts
krazziekay: That will only work if there were cloud apps. The enterprise server works a bit differently. Don't exactly remember the relation right now but I've been testing and it seems that the GitHub server apps that have been deleted will only delete the entry in the table `GitHubServerApps` and their corresponding `Subscription` is not deleted.

krazziekay: Ok, when we're deleting a GitHub server app, its corresponding subscription is not deleted. It just stays there for now. That needs to be done first if we wanna use `subscriptions.length`. 
Adding a TODO for that now and have created a ticket for that. That needs to be tackled separately.

gxueatlassian (approved):

PR #2381, opened 2023-08-30
ARC-2449 add debug logging for pinging a site for cert issue

gxueatlassian:
**What's in this PR?**
1. Add more ping in the ping route

**Why**
1. Need investigate more about the cert issue

**Added feature flags**
N/A

**Affected issues**  
ARC-2449

**How has this been tested?**  
Tested locally

**Whats Next?**
Run the curl in server and check output

### DIFF
--- a/src/routes/api/api-ping-post.ts
+++ b/src/routes/api/api-ping-post.ts
@@ -27,6 +27,8 @@ export const ApiPingPost = async (req: Request, res: Response): Promise<void> =>
 			authorization: "",
 			proxy: data.useProxy === true ? envVars.PROXY : undefined
 		});
+		req.log.info(`Api ping post output: body`, { body: output.body });
+		req.log.info(`Api ping post output: meta`, { meta: output.meta });
 		logCurlOutputInChunks(output, req.log);
 	} catch (e) {
 		req.log.warn("Fail initiate ping url using curl", { err: e, url: data.url });
--- a/src/routes/api/api-ping-post.ts
+++ b/src/routes/api/api-ping-post.ts
@@ -24,11 +24,9 @@ export const ApiPingPost = async (req: Request, res: Response): Promise<void> =>
 		const output = await runCurl({
 			fullUrl: data.url,
 			method: "GET",
-			authorization: "",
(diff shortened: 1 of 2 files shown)


src/routes/api/api-ping-post.ts
```diff
@@ -16,6 +19,34 @@ export const ApiPingPost = async (req: Request, res: Response): Promise<void> =>
 		return;
 	}
 
+	try {
+		const output = await runCurl({
+			fullUrl: data.url,
+			method: "GET",
+			authorization: ""
+		});
+		logCurlOutputInChunks(output, req.log);
```
gxueatlassian, line 28: I am hoping this will give us output on `ssl verification: OK` or `ssl verification: fail` on the `curl` command.

src/routes/api/api-ping-post.ts
```diff
@@ -16,6 +19,34 @@ export const ApiPingPost = async (req: Request, res: Response): Promise<void> =>
...
+	try {
+		if (data.jiraHost) {
+			const gitHubClient = await createAnonymousClientByGitHubAppId(data.gitHubAppId, jiraHost, { trigger: "api-ping-post" }, req.log);
+			try {
+				await gitHubClient.checkGitHubToken("invalid-token");
+			} catch (e) {
+				if (e instanceof GithubClientError) {
+					req.log.warn("Anonymous client failed at GitHubClientError", { err: e, url: data.url, gitHubAppId: data.gitHubAppId, jiraHost: data.jiraHost });
+				} else {
+					throw e;
```
gxueatlassian, line 42: And this line should throw on fail on ssl connection.

src/routes/api/api-ping-post.ts
bgvozdev: btw do we use outbound proxy for curl commands? I think this customer is using ip whitelisting, which means it will timeout if not

gxueatlassian: Good point. I don't think the curl use outbound. Let me see how to do it

gxueatlassian: fixed.

bgvozdev (approved):

PR #2388, opened 2023-09-01
SAI-13825 Adds feature flags to control whether entities should be sent even without associations

tavomaciel:
**What's in this PR?**
- Adds feature flag to send entities even without associations;
- Change transformers to always send entities if flag is on

**Added feature flags**

- StringFlag: `send-all`, value is a JSON list, with the possible values inside of it:
```
"branches-backfill" | "builds-backfill" | "commits-backfill" | "deployments-backfill" | "prs-backfill" |
"branches" | "builds" | "commits" | "deployments" | "prs"
```

Why so many? Well, it should be safer to roll it out separately, easier to keep track if things are on or off with many booleanFlags than with a stringFlag acting as a "set"!

**Affected issues**  
- SAI-13825

**How has this been tested?**  
Create a commit/branch/pr/deployment/build without an issue key -> it should still be sent


src/config/feature-flags.ts
```diff
@@ -88,6 +89,23 @@ export const onFlagChange = (flag: BooleanFlags | StringFlags | NumberFlags, lis
 	launchdarklyClient.on(`update:${flag}`, listener);
 };
 
+type ShouldSendAllStringTypes =
+	"branches-backfill" | "builds-backfill" | "commits-backfill" | "deployments-backfill" | "prs-backfill" |
+	"branches" | "builds" | "commits" | "deployments" | "prs";
+
+export const shouldSendAll = async (type: ShouldSendAllStringTypes, jiraHost?: string, logger?: Logger): Promise<boolean> => {
```
bgvozdev, line 96: nit: curious why jiraHost and logger are optional?

tavomaciel: Cause some of the callers don't have it and lifting the values to that point would be quite hard from what I saw :|

I could be wrong tho! happy to be corrected here

bgvozdev: I quickly checked the callers: looks like there's only one that doesn't have logger around, but I couldn't spot any missing jiraHosts... :susp:

tavomaciel: This is the one I found: https://github.com/atlassian/github-for-jira/pull/2388/files#diff-e2e6b06c457708ee48dd963356ef9816859ade5936ff382cbb65a13c59639cb8L95

```
export const transformPullRequestRest = async (
	gitHubInstallationClient: GitHubInstallationClient,
	pullRequest: Octokit.PullsGetResponse,
	reviews?: Array<{ state?: string, user: Octokit.PullsUpdateResponseRequestedReviewersItem }>,
	log?: Logger,
	jiraHost?: string // <<<< ----------
)
```

bgvozdev: @tavomaciel https://github.com/atlassian/github-for-jira/pull/2393/files - that should fix your problem, please feel free to merge in the branch :)

bgvozdev (dismissed):

bgvozdev (approved):

PR #2389, opened 2023-09-01
Enhance react integration test

gxueatlassian:
**What's in this PR?**
Enhance the react integration test.

**Why**
1. Instead of mocking the API, use lib to mock backend (http) directly. So that we are testing the api client as well.
2. Usecase base testing, give more authentic testing.

**Added feature flags**

**Affected issues**  
N/A

**How has this been tested?**  
Integration Test

**Whats Next?**


krazziekay (dismissed):

spa/src/pages/ConfigSteps/empty-state-flow.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+import { act, render, screen, fireEvent, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import ConfigSteps from "./index";
+import Connected from "../Connected";
+import { MemoryRouter, Route, Routes } from 'react-router-dom'
+import * as nock from "nock";
+
+describe("Empty state user flow", () => {
+
+	let mockWinOpenFunc: jest.Mock;
```
yvonnevictoria, line 11: /nit As someone looking at this with no context, the function name isn't super clear. Does it mean `mockWindowOpenFunc`?

spa/src/pages/ConfigSteps/empty-state-flow.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+			"Allow": "OPTIONS, HEDA, GET, POST, DELETE, PUT, PATCH"
+		});
+		gitHubScope.get("/user").reply(200, {
+			email: "some@email.com",
+			login: "some-login"
+		});
+	}
+
+	async function clickButton(buttonText: string) {
+		await act(async() => {
```
yvonnevictoria, line 97: It would be good to move these to a different file to allow them to be reused more globally or moved directly into the test case if they won't be reused at all

spa/src/pages/ConfigSteps/empty-state-flow.test.tsx
```diff
@@ -0,0 +1,141 @@
...
+		});
+	}
+
+	function expectUserNotLogin() {
+		expect(screen.queryByText("Select your GitHub product", { exact: false })).toBeInTheDocument();
+		expect(screen.queryByText("Logged in as", { exact: false })).not.toBeInTheDocument();
+	}
+
+	function expectUserLoginSuccess(login: string) {
+		expect(screen.queryByText("Select your GitHub product")).not.toBeInTheDocument();
```
yvonnevictoria, line 108: I have a personal preference for not adding assertions in a function, because it may be confusing to someone writing test cases if a "hidden" assertion fails. Would love to hear more about your opinion on this @gxueatlassian

spa/css.stub.js
```diff
@@ -0,0 +1 @@
+module.exports = {};
```
gxueatlassian, line 1: In some react component we use `import xxx.css`. 
Seems the `OrgsContainer` we import ScrollBar which import SimpleBar which import `xx.css` and ts-jest is failing on recongise it. 
Search web give me answer that to put a stub here just to pase the ts check.

spa/jest.config.json
```diff
@@ -3,6 +3,9 @@
 	"transform": {
 		"^.+\\.tsx?$": "ts-jest"
 	},
+	"moduleNameMapper": {
+		"^.+\\.css$": "<rootDir>/css.stub.js"
```
gxueatlassian, line 7: Same as above.

spa/src/components/Error/test.tsx
```diff
@@ -1,4 +1,3 @@
-import "@testing-library/jest-dom";
```
gxueatlassian, line 1: To speed up the test coz we don't need it in each file, just the setup.ts file.

spa/tsconfig.json
```diff
@@ -22,5 +23,5 @@
 			"rest-interfaces": ["./src/rest-interfaces"]
 		}
 	},
-  "include": ["src"]
+  "includes": [ "./setup.ts", "src" ]
```
gxueatlassian, line 26: Otherwise doesn't looks like the ts-jest is including the tesing-lib/js-dom thing during the ts compile phase.

gxueatlassian: I mean after I remove `import "@testing-library/jest-dom";` tests failed, turns out that `import "@testing-library/jest-dom";` is not called by ts-jest during compile.

ps: is has nothing to do with jest at compile time.

spa/src/pages/ConfigSteps/empty-state-flow.test.tsx
gxueatlassian: I've abstracted this away.

spa/src/pages/ConfigSteps/empty-state-flow.test.tsx
gxueatlassian: Yup have moved it in a utils file.

spa/src/pages/ConfigSteps/empty-state-flow.test.tsx
gxueatlassian: have abstracted it away

spa/src/pages/ConfigSteps/empty-state-flow.test.tsx
gxueatlassian: My idea is the treat `expectUserLoginSuccess` as any other `expectXXX`.
If a "hidden" assertion fail, dev shouldn't be confused that something fail inside `expectUserLoginSuccess`, although dev still need to look inside to see what's failing.

I am keen to keep this as it is, but I can change it if you strongly suggest should use plain expect.I am also okay with that.

PR #2399, opened 2023-09-04
NONE Update eslint

atraversatlassian:
**What's in this PR?**
Updating the Typescript eslint version and rules to be more strict.  Most file edits come from running npx eslint --fix --ext .ts,.tsx .

Eslint was not bumped by a major version because the current node version is incompatible with eslint 6.X.X

To control the size of PRs lint errors that are producing lots of errors are currently disabled.  These should be enabled with fixed styling later.

**Why**
Moving towards stricter linting

**Added feature flags**
No, there is not new functionality

**Affected issues**  
NONE

**How has this been tested?**  
There should be no functional changes

**Whats Next?**
There should be separate PRs to remove the disabling of:
"@typescript-eslint/no-non-null-assertion",
"@typescript-eslint/no-explicit-any",
"@typescript-eslint/prefer-nullish-coalescing",
"@typescript-eslint/restrict-template-expressions",
"@typescript-eslint/no-unsafe-assignment",
"@typescript-eslint/no-unsafe-argument",
"@typescript-eslint/no-unsafe-member-access",
"@typescript-eslint/no-unsafe-return",
"@typescript-eslint/no-unsafe-call",
"@typescript-eslint/unbound-method",
"@typescript-eslint/require-await",
"@typescript-eslint/await-thenable",
"@typescript-eslint/restrict-plus-operands",
"@typescript-eslint/no-misused-promises"


src/backfill/backfill.types.ts
```diff
@@ -1,6 +1,6 @@
 import { RepoSyncState } from "models/reposyncstate";
 
-export type JobId = {
+export interface JobId {
```
gxueatlassian, line 3: Hi @atraversatlassian , do you know why's the recommendation of interface over type?

src/github/branch.ts
```diff
@@ -85,14 +85,14 @@ export const deleteBranchWebhookHandler = async (context: WebhookContext<DeleteE
 		return;
 	}
 
-	context.log.info({ prRef: context.payload.ref }, `Deleting branch for repo ${context.payload.repository?.id}`);
+	context.log.info({ prRef: context.payload.ref }, `Deleting branch for repo ${context.payload.repository.id}`);
```
gxueatlassian, line 88: /nit <-- can ignore this comment.

Since the payload comes from external (and possible comes from multiple version of github cloud/server), and this use case is just a log, I'd say leave it as `repository?.id` for safety reason. 

Ideally, I'd suggest we should validate all external input with json schema (either directly download from github or generated from typings) before processing it.

src/github/client/github-client-errors.ts
```diff
@@ -121,7 +121,7 @@ export class GithubClientGraphQLError extends GithubClientError {
 		this.isRetryable = !!errors.find(
 			(error) =>
 				"MAX_NODE_LIMIT_EXCEEDED" == error.type ||
-				error.message?.startsWith("Something went wrong while executing your query")
+				error.message.startsWith("Something went wrong while executing your query")
```
gxueatlassian, line 124: The consumer of this constructor is like
```
		const graphqlErrors = response.data?.errors;
		if (graphqlErrors?.length) {
			const err = new GithubClientGraphQLError(response, graphqlErrors);
```

I'd say a bit of defensive programming using `error.message?` is better. Cause the `errors` comes from GitHub graphql response, which isn't really validated to any format.

src/backfill/backfill.types.ts
atraversatlassian: This was changed from this rule https://typescript-eslint.io/rules/consistent-type-definitions/

src/github/branch.ts
atraversatlassian: This is auto fixed by https://typescript-eslint.io/rules/no-unnecessary-condition/

The payload is a DeleteEvent that has Repository as a required member.  Repository should never be null here

src/github/client/github-client-errors.ts
atraversatlassian: I might misunderstand but this should map to GraphQLError object type where this member should always be set

gxueatlassian: `this member should always be set` <-- I think this is the point where we weren't on the same page yet.

I am not sure about the percentage, maybe 90% of the cases above `this member should always be set` is what is happening in the real world. 

So for the rest 10% case ( or even less ), when that GraphQL errors doesn't contains the `message` property, and error happen:
1. The current behaviour is the process is treated as not `isRetryable`, and the error handling flow continue.
2. Whereas in the new behaviour, the process will throw an Error in this error handling flow, and leads to some unexpected/coded behaviour. <-- this is what I concern, and not sure what's the impact to users.

------

While majority of the errors above PR is okay, I wonder do we need to consider some edge cases? Tbh I am not fuzzy about it, we can say **"Hey the errors itself are just edges cases, errors without `message` in it are just edge cases of edge cases, so we don't need to care about it"**. <-- as long as we are aware of it and agree on that, I am fine with the PR.

src/backfill/backfill.types.ts
gxueatlassian: Thanks. From the link I see the reason is "Keep declaration consistent", but not mentioning why `interface` is better than `type`. So I guess the rule is to help being consistent and sort of pick `interface`?

gxueatlassian: And this is me out of curious, I am fine with changing from type to interface.

PR #2401, opened 2023-09-05
None enhance sentry error info

gxueatlassian:
**What's in this PR?**
1. Enhance the error reporting

**Why**
1. So Sentry can better group those errors.

Screenshot 2023-09-05 at 12 34 37 pm

### DIFF
--- a/spa/src/pages/ConfigSteps/index.tsx
+++ b/spa/src/pages/ConfigSteps/index.tsx
@@ -9,7 +9,7 @@ import LoggedinInfo from "../../common/LoggedinInfo";
 import Tooltip, { TooltipPrimitive } from "@atlaskit/tooltip";
 import { token } from "@atlaskit/tokens";
 import { useNavigate } from "react-router-dom";
-import Error from "../../components/Error";
+import ErrorUI from "../../components/Error";
 import AppManager from "../../services/app-manager";
 import OAuthManager from "../../services/oauth-manager";
 import analyticsClient from "../../analytics";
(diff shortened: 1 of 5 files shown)


spa/tsconfig.json
```diff
@@ -2,10 +2,9 @@
   "compilerOptions": {
     "target": "ES2020",
     "useDefineForClassFields": true,
-    "lib": ["ES2020", "DOM", "DOM.Iterable"],
```
gxueatlassian, line 5: I think adding a ES2022.Error is pretty safe. 
https://caniuse.com/mdn-javascript_builtins_error_cause

spa/src/pages/ConfigSteps/index.tsx
```diff
@@ -182,7 +182,7 @@ const ConfigSteps = () => {
 				} catch (e) {
 					setLoaderForLogin(false);
 					showError(modifyError(e as AxiosError, {}, { onClearGitHubToken: clearGitHubToken, onRelogin: reLogin }));
-					reportError(e, {
```
gxueatlassian, line 185: We have to call the `new Error(xxx)` here instead of inside the `reportError` because the stacktrace will only be caught during the calling of `new Error()`

gxueatlassian: And wrap it with a `new Error` will help Sentry group the error into different issues.

spa/tsconfig.json
gxueatlassian: Have to add this to support the `{ cause }` in Error constructor.

krazziekay (approved):

PR #2402, opened 2023-09-05
ARC-2459 use app client to check user permission

gxueatlassian:
**What's in this PR?**
1. Use github installation client to check permission

**Why**
1. The user client is impacted by SSO error. Installation client should not.

**Added feature flags**
use-installation-client-to-check-permission

**Affected issues**  
ARC-2459

**How has this been tested?**  
Unit test
Local

**Whats Next?**

### DIFF
--- a/src/util/github-utils.ts
+++ b/src/util/github-utils.ts
@@ -2,6 +2,7 @@ import { GitHubUserClient } from "~/src/github/client/github-user-client";
 import { GitHubAppClient } from "~/src/github/client/github-app-client";
 import Logger from "bunyan";
 import { booleanFlag, BooleanFlags } from "config/feature-flags";
+import { GithubClientNotFoundError, GithubClientInvalidPermissionsError } from "~/src/github/client/github-client-errors";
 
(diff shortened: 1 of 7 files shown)


src/util/github-utils.ts
```diff
@@ -1,12 +1,25 @@
...
 		logger.info("isUserAdminOfOrganization: orgType is a user");
 		return orgName === username;
 	}
 
+	try {
+		if (await booleanFlag(BooleanFlags.USE_INSTALLATION_CLIENT_CHECK_PERMISSION, jiraHost)) {
+			logger.info("isUserAdminOfOrganization: orgType is an org, checking membership (app client)");
+			const { data: { state, role } } = await gitHubInstallationClient.getUserMembershipForOrg(username, orgName);
+			logger.info({ orgName, username, state, role }, `isUserAdminOfOrganization: User has role for org`);
+			return role === "admin" && state === "active";
```
bgvozdev, line 17: curious why do we need the second check? it wasn't in the version with the user client...

bgvozdev (approved):

src/util/github-utils.ts
gxueatlassian: I don't know. Maybe we didn't do it right at the first place? Like inactive admin can still do sh**?

PR #2409, opened 2023-09-06
NONE: add eventloop-lag

bgvozdev:
**What's in this PR?**
Emitting metrics about evenloop lag

**Why**
Trying to understand wtf is going on with web nodes and why the latency became a problem

Tested in stg

**Added feature flags**
🤠

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -5,6 +5,9 @@ import throng from "throng";
 import { initializeSentry } from "config/sentry";
 import { isNodeProd } from "utils/is-node-env";
 import { getFrontendApp } from "./app";
+import createLag from "event-loop-lag";
+import { statsd } from "config/statsd";
+import { metricLag } from "config/metric-names";
 
 const start = async () => {
 	initializeSentry();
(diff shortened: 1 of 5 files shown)


src/main.ts
```diff
@@ -13,6 +16,10 @@ const start = async () => {
 	app.listen(port, () => {
 		getLogger("frontend-app").info(`started at port ${port}`);
 	});
+	const lag = createLag(1000);
+	setInterval(() => {
+		statsd.histogram(metricLag.lagHist, lag(), { }, { });
```
krazziekay, line 21: Do we have any attributes to identify this 1 sec delay? Maybe pass some extra params just in case.

bgvozdev: Sorry @krazziekay not sure I understand... 1 sec is just a frequency how often we want to send this metric to sfx...

krazziekay (approved):

src/main.ts
krazziekay: Approved then, its just for testing anyways right?

bgvozdev: Yeah, it doesn't affect prod. Just to make sure node.js' event loop is not lagging behind (or to identify that this is the problem...)

PR #2412, opened 2023-09-07
ARC-1449 redis reconnect

joshkay10:
**What's in this PR?**
When we hit a readonly error, its most likely that we are connected to the read-only replica.

In this case we ask ioredis to reconnect to the redis instance and it should point back to master on reconnect!!

**Why**
During failover testing we had an incident when we didn't reconnect to the master instance and kept trying to write to the replica which is readonly 😢 

**Affected issues**  
ARC-1449

**How has this been tested?**  
Testing in stage with a micros failover event for REDIS

I also was able to force my local docker to create a slave instance and connect to that move it to read only and see the logs keep attempting to reconnect until i restored the master.

**Whats Next?**
Nothing!! hopefully


src/config/redis-info.ts
```diff
@@ -6,7 +6,14 @@ export const getRedisInfo = (connectionName: string): IORedis.RedisOptions => ({
...
 	db: 0,
+	reconnectOnError(err) {
+		const targetError = "READONLY";
 
+		if (err.message.includes(targetError)) {
+			return 1;
+		}
+		return false;
+	},
 	// TODO find out if we still need these options
```
joshkay10, line 17: They put a fix in on that github link, but people are still complaining, will investigate further outside this PR

src/config/redis-info.ts
```diff
@@ -6,7 +6,14 @@ export const getRedisInfo = (connectionName: string): IORedis.RedisOptions => ({
 	port: Number(envVars.REDISX_CACHE_PORT) || 6379,
 	host: envVars.REDISX_CACHE_HOST || "127.0.0.1",
 	db: 0,
+	reconnectOnError(err) {
+		const targetError = "READONLY";
```
joshkay10, line 10: considered making `const targetErrors = ["READONLY", "BLAH BLAH"];` work, but the reconnection logic that is handled by `ioredis` was working for the examples that I could find.... If we run into problems in the future will be pretty easy to slap it in.

src/config/redis-info.ts
```diff
@@ -6,7 +6,14 @@ export const getRedisInfo = (connectionName: string): IORedis.RedisOptions => ({
 	port: Number(envVars.REDISX_CACHE_PORT) || 6379,
 	host: envVars.REDISX_CACHE_HOST || "127.0.0.1",
 	db: 0,
+	reconnectOnError(err) {
```
joshkay10, line 9: read PR main comments for more details

src/config/redis-info.ts
```diff
@@ -6,7 +6,14 @@ export const getRedisInfo = (connectionName: string): IORedis.RedisOptions => ({
...
+		const targetError = "READONLY";
 
+		if (err.message.includes(targetError)) {
+			return 1;
```
bgvozdev, line 13: curious why 1 and not true?

bgvozdev: if true is acceptible, could be `return err.messages.includes(targetError)`

bgvozdev (approved):

src/config/redis-info.ts
joshkay10: `ioredis` library expects a number, its milliseconds until reconnect, and it throws a tantrum if you pass true.

PR #2413, opened 2023-09-07
NONE Enabling more eslint rules and fixing errors

atraversatlassian:
**What's in this PR?**
Enables the following eslint rules
"@typescript-eslint/no-non-null-assertion",
"@typescript-eslint/no-explicit-any",
"@typescript-eslint/prefer-nullish-coalescing",
"@typescript-eslint/restrict-template-expressions",
"@typescript-eslint/restrict-plus-operands"

**Why**
Empowering the linter longer term

**Added feature flags**
No

**Affected issues**  
NONE

**How has this been tested?**  

**Whats Next?**

### DIFF
--- a/src/config/logger.ts
+++ b/src/config/logger.ts
@@ -40,8 +40,8 @@ const GIT_REF_URL_REGEX = /^(.*)\/git\/ref\/([^/]+)$/;
 const isGitRefUrl = (url: string) => url.match(GIT_REF_URL_REGEX);
 
 const removeGitRefFromUrl = (url: string) =>
-	url.replace(GIT_REF_URL_REGEX, (_, prefix, gitRef) =>
-		`${prefix as string}/git/ref/${createHashWithSharedSecret(decodeURIComponent(gitRef))}`
+	url.replace(GIT_REF_URL_REGEX, (_, prefix: string, gitRef) =>
+		`${prefix}/git/ref/${createHashWithSharedSecret(decodeURIComponent(gitRef))}`
 	);
 
 const USERS_URL_REGEX = /^(\/api\/v3)?\/users\/([^/]+)$/;
@@ -62,14 +62,14 @@ const REST_DEVINFO_BRANCH_URL_REGEX = /^\/rest\/devinfo\/([^/]+)\/repository\/([
 const isDevInfoBranchUrl = (url: string) => url.match(REST_DEVINFO_BRANCH_URL_REGEX);
 
 const removeBranchFromDevInfoUrl = (url: string) =>
-	url.replace(REST_DEVINFO_BRANCH_URL_REGEX, (_, version, repoNo, branchaName, updateSequenceId) =>
+	url.replace(REST_DEVINFO_BRANCH_URL_REGEX, (_, version, repoNo, branchaName, updateSequenceId: string) =>
 		[
 			"/rest/devinfo",
 			version,
 			"repository",
 			repoNo,
 			"branch",
-			`${createHashWithSharedSecret(decodeURIComponent(branchaName))}?_updateSequenceId=${updateSequenceId as string}`
+			`${createHashWithSharedSecret(decodeURIComponent(branchaName))}?_updateSequenceId=${updateSequenceId}`
 		].join("/")
 	);
 
--- a/src/config/logger.ts
+++ b/src/config/logger.ts
@@ -41,7 +41,7 @@ const isGitRefUrl = (url: string) => url.match(GIT_REF_URL_REGEX);
 
 const removeGitRefFromUrl = (url: string) =>
 	url.replace(GIT_REF_URL_REGEX, (_, prefix, gitRef) =>
-		`${prefix}/git/ref/${createHashWithSharedSecret(decodeURIComponent(gitRef))}`
+		`${prefix as string}/git/ref/${createHashWithSharedSecret(decodeURIComponent(gitRef))}`
 	);
 
 const USERS_URL_REGEX = /^(\/api\/v3)?\/users\/([^/]+)$/;
@@ -69,7 +69,7 @@ const removeBranchFromDevInfoUrl = (url: string) =>
 			"repository",
 			repoNo,
 			"branch",
-			`${createHashWithSharedSecret(decodeURIComponent(branchaName))}?_updateSequenceId=${updateSequenceId}`
+			`${createHashWithSharedSecret(decodeURIComponent(branchaName))}?_updateSequenceId=${updateSequenceId as string}`
 		].join("/")
 	);
 
@@ -101,7 +101,7 @@ const removeOwnersAndReposFromUrl = (url: string) =>
 			"repos",
 			createHashWithSharedSecret(decodeURIComponent(repos)),
 			"branches"
-		].join("/") + suffix
+		].join("/") + (suffix as string)
 	);
 
 const censorUrl = (url) => {
--- a/src/github/client/github-client-errors.ts
+++ b/src/github/client/github-client-errors.ts
@@ -17,7 +17,7 @@ export class GithubClientError extends Error {
 		this.uiErrorCode = "UNKNOWN";
 
 		this.cause = { ...cause };
-		this.stack = this.stack?.split("\n").slice(0, 2).join("\n") + "\n" + cause.stack;
+		this.stack = (this.stack?.split("\n").slice(0, 2).join("\n") ?? "Missing Stack") + "\n" + (cause.stack ?? "Missing Stack");
 	}
 }
 
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -28,7 +28,7 @@ import { getLogger } from "config/logger";
 
 export interface GitHubClientApiKeyConfig {
 	headerName: string;
-	apiKeyGenerator: () => Promise<string | undefined>;
+	apiKeyGenerator: () => Promise<string>;
 }
 
 export interface GitHubConfig {
@@ -142,7 +142,7 @@ export class GitHubClient {
 				if (!config.headers) {
 					config.headers = {};
 				}
-				config.headers[apiKeyConfig.headerName] = await apiKeyConfig.apiKeyGenerator() ?? "undefined";
+				config.headers[apiKeyConfig.headerName] = await apiKeyConfig.apiKeyGenerator();
 				return config;
 			});
 		}
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -28,7 +28,7 @@ import { getLogger } from "config/logger";
 
 export interface GitHubClientApiKeyConfig {
 	headerName: string;
-	apiKeyGenerator: () => Promise<string>;
+	apiKeyGenerator: () => Promise<string | undefined>;
 }
 
 export interface GitHubConfig {
@@ -131,7 +131,7 @@ export class GitHubClient {
 				...this.metrics
 			}),
 			instrumentFailedRequest(metricHttpRequest.github, this.restApiUrl, jiraHost, {
-				withApiKey: (!!gitHubConfig.apiKeyConfig)?.toString() || "undefined",
+				withApiKey: (!!gitHubConfig.apiKeyConfig).toString(),
 				...this.metrics
 			})
 		);
@@ -142,7 +142,7 @@ export class GitHubClient {
 				if (!config.headers) {
 					config.headers = {};
 				}
-				config.headers[apiKeyConfig.headerName] = await apiKeyConfig.apiKeyGenerator();
+				config.headers[apiKeyConfig.headerName] = await apiKeyConfig.apiKeyGenerator() ?? "undefined";
 				return config;
 			});
 		}
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -127,11 +127,11 @@ export class GitHubClient {
 		);
 		this.axios.interceptors.response.use(
 			instrumentRequest(metricHttpRequest.github, this.restApiUrl, jiraHost, {
-				withApiKey: (!!gitHubConfig.apiKeyConfig).toString(),
+				withApiKey: (!!gitHubConfig.apiKeyConfig)?.toString() || "undefined",
 				...this.metrics
 			}),
 			instrumentFailedRequest(metricHttpRequest.github, this.restApiUrl, jiraHost, {
-				withApiKey: (!!gitHubConfig.apiKeyConfig).toString(),
+				withApiKey: (!!gitHubConfig.apiKeyConfig)?.toString() || "undefined",
 				...this.metrics
 			})
 		);
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -127,7 +127,7 @@ export class GitHubClient {
 		);
 		this.axios.interceptors.response.use(
 			instrumentRequest(metricHttpRequest.github, this.restApiUrl, jiraHost, {
-				withApiKey: (!!gitHubConfig.apiKeyConfig)?.toString() || "undefined",
+				withApiKey: (!!gitHubConfig.apiKeyConfig).toString(),
 				...this.metrics
 			}),
 			instrumentFailedRequest(metricHttpRequest.github, this.restApiUrl, jiraHost, {
--- a/src/github/client/github-client.ts
+++ b/src/github/client/github-client.ts
@@ -127,11 +127,11 @@ export class GitHubClient {
 		);
 		this.axios.interceptors.response.use(
 			instrumentRequest(metricHttpRequest.github, this.restApiUrl, jiraHost, {
-				withApiKey: "" + (!!gitHubConfig.apiKeyConfig),
+				withApiKey: (!!gitHubConfig.apiKeyConfig).toString(),
 				...this.metrics
 			}),
 			instrumentFailedRequest(metricHttpRequest.github, this.restApiUrl, jiraHost, {
-				withApiKey: "" + (!!gitHubConfig.apiKeyConfig),
+				withApiKey: (!!gitHubConfig.apiKeyConfig).toString(),
 				...this.metrics
 			})
 		);
--- a/src/github/code-scanning-alert.test.ts
+++ b/src/github/code-scanning-alert.test.ts
@@ -10,6 +10,11 @@ import { mocked } from "jest-mock";
 import { emitWebhookProcessedMetrics } from "utils/webhook-utils";
 import Logger from "bunyan";
 import { booleanFlag } from "config/feature-flags";
+import { DeploymentsResult, JiraClient } from "../jira/client/jira-client";
+import { AxiosResponse } from "axios";
+import { JiraIssue, JiraSubmitOptions, JiraBuildBulkSubmitData, JiraDeploymentBulkSubmitData } from "../interfaces/jira";
+import { TransformedRepositoryId } from "../transforms/transform-repository-id";
+
 jest.mock("config/feature-flags");
 jest.mock("utils/webhook-utils");
 jest.mock("../transforms/transform-code-scanning-alert");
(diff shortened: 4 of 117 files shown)


src/config/logger.ts
```diff
@@ -41,7 +41,7 @@ const isGitRefUrl = (url: string) => url.match(GIT_REF_URL_REGEX);
 
 const removeGitRefFromUrl = (url: string) =>
 	url.replace(GIT_REF_URL_REGEX, (_, prefix, gitRef) =>
-		`${prefix}/git/ref/${createHashWithSharedSecret(decodeURIComponent(gitRef))}`
+		`${prefix as string}/git/ref/${createHashWithSharedSecret(decodeURIComponent(gitRef))}`
```
bgvozdev, line 44: nit: curious if it is possible to add the type to the argument instead?

src/github/client/github-client-errors.ts
```diff
@@ -17,7 +17,7 @@ export class GithubClientError extends Error {
 		this.uiErrorCode = "UNKNOWN";
 
 		this.cause = { ...cause };
-		this.stack = this.stack?.split("\n").slice(0, 2).join("\n") + "\n" + cause.stack;
+		this.stack = (this.stack?.split("\n").slice(0, 2).join("\n") ?? "Missing Stack") + "\n" + (cause.stack ?? "Missing Stack");
```
bgvozdev, line 20: nice!

src/github/code-scanning-alert.test.ts
```diff
@@ -52,10 +57,83 @@ describe("Code Scanning Alert Webhook Handler", () => {
 
 	const mockJiraRemoteLinkSubmit = jest.fn();
 	const mockJiraSecuritySubmit = jest.fn();
-	const jiraClient = {
+	const jiraClient: JiraClient = {
 		baseURL: JIRA_BASE_URL,
 		remoteLink: { submit: mockJiraRemoteLinkSubmit },
-		security: { submitVulnerabilities: mockJiraSecuritySubmit }
+		security: { submitVulnerabilities: mockJiraSecuritySubmit },
+		issues: {
```
bgvozdev, line 64: Wow, that's pretty cumbersome, but guess we don't have much options here, do we?

bgvozdev: Maybe to have some factory that would generate a deep mock and then just grab the functions/methods we are interested in? E.g.
```
const jiraClient: JiraClient = generateDeeplyMockedJiraClient();
mockJiraRemoteLinkSumit=jiraClient.remoteLink.submit;
```
Also not sure if it is suitable here though :)

src/github/code-scanning-alert.ts
```diff
@@ -5,18 +5,19 @@ import {
...
 
-export const handleRemoteLinks = async (context: WebhookContext, jiraClient, gitHubInstallationId: number): Promise<void> => {
+export const handleRemoteLinks = async (context: WebhookContext, jiraClient: JiraClient, gitHubInstallationId: number): Promise<void> => {
 	context.log = context.log.child({
 		gitHubInstallationId,
 		jiraHost: jiraClient.baseURL
 	});
 
-	const jiraPayload = await transformCodeScanningAlert(context, gitHubInstallationId, jiraClient.baseUrl);
+	const jiraPayload = await transformCodeScanningAlert(context, gitHubInstallationId, jiraClient.baseURL);
```
bgvozdev, line 20: did we have a bug here?

src/github/issue-comment.ts
```diff
@@ -106,7 +106,7 @@ const syncIssueCommentsToJira = async (jiraHost: string, context: WebhookContext
 				"content": [
 					{
 						"type": "text",
-						"text": `${comment.user.login}`,
+						"text": `${comment.user.login as string}`,
```
bgvozdev, line 109: I have hard time parsing this 🤣  Do we really need `s?

src/github/code-scanning-alert.ts
atraversatlassian: Probably

src/middleware/github-webhook-middleware.ts
```diff
@@ -116,7 +116,7 @@ export const GithubWebhookMiddleware = (
 			fields: {
 				webhookId,
 				gitHubInstallationId,
-				gitHubServerAppIdPk: "" + gitHubAppId,
+				gitHubServerAppIdPk: gitHubAppId?.toString(),
```
bgvozdev, line 119: Hmm... isn't this changes the behaviour for undefined? Before the change it would be "undefined" as a string, after the change it would be undefined as a value... If so then wondering which one of those is the behaviour we want?

src/models/subscription.ts
```diff
@@ -25,6 +25,28 @@ export interface Repository {
 	updated_at: string; // TODO: is this a date object or a timestamp?  Different places uses different things
 }
 
+export interface Branch {
```
bgvozdev, line 28: Just curious where the schema is it coming from, and where it is used?

atraversatlassian: Yes this is generated based on its usage

src/routes/api/api-router.ts
```diff
@@ -130,7 +130,7 @@ const cryptorDebugEndpoint = async (_req: Request, resp: Response) => {
 	try {
 		let data = "";
 		for (let i = 0; i < 10; i++) {
-			data = data + "-" + Math.floor((Math.random() * 10));
+			data = `${data}-${Math.floor((Math.random() * 10))}`;
```
bgvozdev, line 133: nit: pretty sure we can just kill the endpoint

src/routes/api/commits-from-date/recover-commits-from-dates.ts
```diff
@@ -52,12 +52,12 @@ export const RecoverCommitsFromDatePost = async (req: Request, res: Response): P
...
 		}
-		info(`${count} subscriptions updated, last subscription id is ${lastId})`);
+		info(`${count} subscriptions updated, last subscription id is ${lastId ? lastId.toString() : "undefined"})`);
 
 		res.end();
 
 	} catch (e) {
 		log.error({ err: e }, "Error happen when recovering commits from date");
-		res.end(`Error happen when recovering commits from date: ${safeJsonStringify(e)}`);
+		res.end(`Error happen when recovering commits from date: ${safeJsonStringify(e) as string}`);
```
bgvozdev, line 61: Should we rather `yarn add @types/safe-json-stringify`?

src/routes/api/db-migrations/db-migration-down.ts
```diff
@@ -53,7 +53,7 @@ const validateScriptAgainstDB = async (targetScript: string) => {
 	if (lastScript.length < 1) {
 		throw {
 			statusCode: 500,
-			message: `There're no scripts in db to migration down, stop proceeding. \n ${lastScript}`
+			message: `There're no scripts in db to migration down, stop proceeding. \n ${lastScript.toString()}`
```
bgvozdev, line 56: nit: Hmm.... if lastScript.length < 1.... does it make sense to include it into the message at all?

src/routes/api/db-migrations/db-migration-up.ts
```diff
@@ -11,7 +11,7 @@ export const DBMigrationUp = async (req: Request, res: Response): Promise<void>
 
 	try {
 
-		const targetScript = getTargetScript(req);
+		const targetScript: string = getTargetScript(req);
```
bgvozdev, line 14: nit: shouldn't we rather make `getTargetScript` to return a string?

src/routes/jira/connect/enterprise/jira-connect-enterprise-post.ts
```diff
@@ -69,9 +69,9 @@ export const JiraConnectEnterprisePost = async (
 	// inside the handler
 	const TIMEOUT_PERIOD_MS = parseInt(process.env.JIRA_CONNECT_ENTERPRISE_POST_TIMEOUT_MSEC || "30000");
 
-	const gheServerURL = req.body.gheServerURL?.trim();
-	const apiKeyHeaderName = req.body.apiKeyHeaderName?.trim();
-	const apiKeyValue = req.body.apiKeyValue?.trim();
+	const gheServerURL: string = req.body.gheServerURL?.trim();
```
bgvozdev, line 72: Hmm, curious why TS is not complaining... `req.body.gheServerURL?.trim()` could be `undefined`, right?

atraversatlassian: Correct updated to address.  TS can't work it out because it is an any object.  It will complain here once we enable unsafe call to any rules later.

src/github/code-scanning-alert.ts
bgvozdev: @joshkay10 @atrigueiro  - Looks like the feature had been broken for GHE users all the time (I see the jiraHost is used to decrypt the GHE's secrets)...

src/routes/jira/connect/enterprise/app/jira-connect-enterprise-app-put.ts
```diff
@@ -58,6 +58,6 @@ export const JiraConnectEnterpriseAppPut = async (
 		});
 
 		res.status(404).send({ message: "Failed to update GitHub App." });
-		return next(new Error(`Failed to update GitHub app: ${error}`));
+		return next(new Error(`Failed to update GitHub app: ${error instanceof Error ? error.toString() : "unkown"}`));
```
bgvozdev, line 61: I see this construction over and over... wondering if we need some `toError(cause: unknown, message: string)` utility function?

src/routes/jira/connect/enterprise/jira-connect-enterprise-post.ts
```diff
@@ -176,7 +176,7 @@ export const JiraConnectEnterprisePost = async (
 			return;
 		}
 
-		const codeOrStatus = "" + (axiosError.code || axiosError.response?.status);
+		const codeOrStatus = "" + (axiosError.code || axiosError.response?.status?.toString() || "undefined");
```
bgvozdev, line 179: nit: do we still need `"" +` prefix?

src/models/subscription.ts
bgvozdev: I'm wondering if that's the right place for this interface? Don't think we ever store branch data in our Db...

bgvozdev:
Just a general concern @atraversatlassian (not sure how justifiable it is, but flashing out anyway 😄 ): if foo is undefined then

```
log(`${foo}` )
```
won't throw a error, while

```
log(foo.toString())
```

I see in the majority of places you have added `foo ? foo.toString() : "undefined"` , but not everywhere, I believe because TS didn't complain about that... I assume still there could be some corner-cases (e.g. the schema of some REST response in our code assumes the value is there, but in reality it is not). Wondering how confident we are that's not the case?

src/routes/api/api-router.ts
atraversatlassian: This PR is already big.  I will leave it to you to decide to do that clean up in a separate PR if you are confident it can die

src/models/subscription.ts
atraversatlassian: Happy to move it if you have a better location for it

src/github/client/github-client.ts
```diff
@@ -127,11 +127,11 @@ export class GitHubClient {
 		);
 		this.axios.interceptors.response.use(
 			instrumentRequest(metricHttpRequest.github, this.restApiUrl, jiraHost, {
-				withApiKey: "" + (!!gitHubConfig.apiKeyConfig),
+				withApiKey: (!!gitHubConfig.apiKeyConfig)?.toString() || "undefined",
```
bgvozdev, line 130: nit: pretty sure !! operation will be always defined, therefore no need in "?"

src/github/issue-comment.ts
```diff
@@ -83,7 +83,10 @@ export const issueCommentWebhookHandler = async (
 
 const syncIssueCommentsToJira = async (jiraHost: string, context: WebhookContext, gitHubInstallationClient: GitHubInstallationClient) => {
 	const { comment, repository, issue } = context.payload;
-	const { body: gitHubMessage, id: gitHubId, html_url: gitHubCommentUrl } = comment;
+	const { id: gitHubId } = comment;
+	const gitHubLogin: string = comment.user?.login || "undefined";
+	const gitHubMessage: string = comment.body || "undefined";
```
bgvozdev, line 88: probably better to use ?? , otherwise an empty string will also be mapped to undefined

src/github/issue-comment.ts
```diff
@@ -106,7 +109,7 @@ const syncIssueCommentsToJira = async (jiraHost: string, context: WebhookContext
 				"content": [
 					{
 						"type": "text",
-						"text": `${comment.user.login}`,
+						"text": `${gitHubLogin}`,
```
bgvozdev, line 112: nit probably don't need `

src/jira/util/jwt.ts
```diff
@@ -139,8 +140,8 @@ const validateSymmetricJwt = (secret: string, request: JWTRequest, tokenType: To
 	let unverifiedClaims: any;
 	try {
 		unverifiedClaims = decodeSymmetric(token, "", algorithm, true); // decode without verification;
-	} catch (e) {
-		throw `Invalid JWT: ${e.message}`;
+	} catch (e: unknown) {
+		throw `Invalid JWT: ${errorStringFromUnknown(e)}`;
```
bgvozdev, line 144: wrap in Error?

src/jira/util/jwt.ts
```diff
@@ -151,8 +152,8 @@ const validateSymmetricJwt = (secret: string, request: JWTRequest, tokenType: To
 	let verifiedClaims: any; //due to decodeSymmetric return any
 	try {
 		verifiedClaims = decodeSymmetric(token, secret, algorithm, false);
-	} catch (error) {
-		throw `Unable to decode JWT token: ${error}`;
+	} catch (e: unknown) {
+		throw `Unable to decode JWT token: ${errorStringFromUnknown(e)}`;
```
bgvozdev, line 156: wrap in Error?

src/models/subscription.ts
bgvozdev: maybe to `src/github/client/github-client.types.ts` ?

src/routes/api/client-key/recover-client-key.ts
```diff
@@ -94,7 +94,7 @@ const getAndVerifyplainClientKey = async ({ id, jiraHost, clientKey: hashedClien
 		//https://hello.atlassian.net/wiki/spaces/EDGE/pages/315794033/Whitelist+Proxy+-+Usage
 		//use whitelist proxy so that we can hit xxx.jira-dev.com
 		//Unless we don't care about those records in db?
-		const proxy = `http://${process.env.WHITELIST_PROXY_HOST}:${process.env.WHITELIST_PROXY_PORT}`;
+		const proxy = `http://${process.env.WHITELIST_PROXY_HOST ?? ""}:${process.env.WHITELIST_PROXY_PORT ?? ""}`;
```
bgvozdev, line 97: supernit: why not "undefined"? 🤣  doesn't matter as I think it doesn't matter, but in other places it is "undefined"

src/routes/github/configuration/github-configuration-get.ts
```diff
@@ -138,8 +139,8 @@ const removeFailedConnectionsFromDb = async (logger: Logger, installations: Inst
 					host: jiraHost,
 					gitHubAppId
 				});
-			} catch (err) {
-				const deleteSubscriptionError = `Failed to delete subscription: ${err}`;
+			} catch (err: unknown) {
+				const deleteSubscriptionError = `Failed to delete subscription: ${errorStringFromUnknown(err)}`;
 				logger.error(deleteSubscriptionError);
```
bgvozdev, line 144: since you are touching this, would you mind also to change this line to 

```
logger.error({ err }, deleteSubscriptionError);
```
🙏  We usually pass { err } to logger in case of an exception

src/github/client/github-client.ts
```diff
@@ -127,11 +127,11 @@ export class GitHubClient {
 		);
 		this.axios.interceptors.response.use(
 			instrumentRequest(metricHttpRequest.github, this.restApiUrl, jiraHost, {
-				withApiKey: "" + (!!gitHubConfig.apiKeyConfig),
+				withApiKey: (!!gitHubConfig.apiKeyConfig).toString(),
 				...this.metrics
 			}),
 			instrumentFailedRequest(metricHttpRequest.github, this.restApiUrl, jiraHost, {
-				withApiKey: "" + (!!gitHubConfig.apiKeyConfig),
+				withApiKey: (!!gitHubConfig.apiKeyConfig)?.toString() || "undefined",
```
bgvozdev, line 134: nit: pretty sure !! is always defined :) should've duplicated my prev. comment, feel free to ignore!

src/routes/jira/connect/enterprise/jira-connect-enterprise-post.ts
```diff
@@ -127,7 +127,7 @@ export const JiraConnectEnterprisePost = async (
 			apiKeyHeaderName
 				? {
 					headerName: apiKeyHeaderName,
-					apiKeyGenerator: () => Promise.resolve(apiKeyValue)
+					apiKeyGenerator: () => Promise.resolve(apiKeyValue || "undefined")
```
bgvozdev, line 130: actually I think this must be undefined, otherwise "undefined' string will be used as an apiKeyValue (undefined means no value here)

src/models/subscription.ts
bgvozdev: I see you have copied it to github-client.types... shall we remove this one then?

src/sqs/error-handlers.ts
```diff
@@ -78,9 +75,9 @@ const maybeHandleNonRetryableResponseCode = <MessagePayload extends BaseMessageP
 	//Unfortunately we can't check if error is instance of Octokit.HookError because it is not a class, so we'll just rely on status
 	//New GitHub Client error (GithubClientError) also has status parameter, so it will be covered by the following check too
 	//TODO When we get rid of Octokit completely add check if (error instanceof GithubClientError) before the following code
-	const maybeErrorWithStatus: any = error;
-	if (maybeErrorWithStatus.status && UNRETRYABLE_STATUS_CODES.includes(maybeErrorWithStatus.status)) {
-		context.log.warn({ err: maybeErrorWithStatus }, `Received error with ${maybeErrorWithStatus.status} status. Unretryable. Discarding the message`);
+	const status: number = error["status"];
```
bgvozdev, line 78: Shouldn't it be `number | undefined` ?

src/sync/deployment.test.ts
```diff
@@ -206,10 +206,24 @@ describe("sync/deployments", () => {
 			commitsFromDate: "2023-01-01T00:00:00Z"
 		});
 
+		interface DeploymentEntity {
```
bgvozdev, line 209: I think we already have these types here: https://github.com/atlassian/github-for-jira/blob/NONE-add-mem-stats-and-improve-logging/src/github/client/github-queries.ts#L533-L533

src/util/workers-health-monitor.ts
```diff
@@ -32,7 +32,7 @@ export const startMonitorOnWorker = (parentLogger: Logger, iAmAliveInervalMsec:
 const logRunningProcesses = (logger: Logger) => {
 	exec("ps aux", (err, stdout) => {
 		if (err) {
-			logger.error({ err }, `exec error: ${err}`);
+			logger.error({ err }, `exec error: ${err?.toString()}`);
```
bgvozdev, line 35: err is always defined (`if (err)` above)

bgvozdev: I think TS should understand that?

src/github/client/github-client.ts
```diff
@@ -28,7 +28,7 @@ import { getLogger } from "config/logger";
 
 export interface GitHubClientApiKeyConfig {
 	headerName: string;
-	apiKeyGenerator: () => Promise<string>;
+	apiKeyGenerator: () => Promise<string | undefined>;
```
bgvozdev, line 31: Not quite sure I understand this... API key config consists of 2 items: header and value. Either both should be absent (then the object that holds a reference to the config should contain null/undefined), or present. In both cases, I think apiKeyGenerator should always return a string

src/github/client/github-client.ts
```diff
@@ -142,7 +142,7 @@ export class GitHubClient {
 				if (!config.headers) {
 					config.headers = {};
 				}
-				config.headers[apiKeyConfig.headerName] = await apiKeyConfig.apiKeyGenerator();
+				config.headers[apiKeyConfig.headerName] = await apiKeyConfig.apiKeyGenerator() ?? "undefined";
```
bgvozdev, line 145: We have already checked on line 139 that the apiKeyConfig is defined, which should mean that both headerName and the value are defined.

bgvozdev (approved):

PR #2422, opened 2023-09-13
NONE: trying to pin memory problem, improve logging

bgvozdev:
**What's in this PR?**
More logs, better logs

**Why**
Still no idea what's going on with memory


bgvozdev:
actually no

PR #2423, opened 2023-09-13
[INNOVATION] ARC-2444 - Add Feature Flag for backfill status page

krazziekay:
**What's in this PR?**
- Added new env variable in node app - `LAUNCHDARKLY_CLIENT_KEY`
  - This env variable along with the `HASHED_JIRAHOST` is sent from the node to the react app for production builds.
- Added FF in react /spa. 

**Why**
- Part of the Innovation!

**Tests**
- Tested in staging


bgvozdev (changes requested):
We cannot use proprietary code in open-source project 😢

PR #2424, opened 2023-09-13
[INNOVATION - PLAN B] ARC-2444 - Add Feature Flag for backfill status page

krazziekay:
**What's in this PR?**
- Pass the feature flag value from the node to the react app for production builds.

**Why**
- Part of the Innovation!

**Tests**
- Tested in staging


bgvozdev (approved):

spa/src/feature-flags.ts
```diff
@@ -0,0 +1,3 @@
+const featureFlags = () => FRONTEND_FEATURE_FLAGS ?
+	JSON.parse(JSON.stringify(FRONTEND_FEATURE_FLAGS)) : null;
```
bgvozdev, line 2: Do we have lodash in spa? If so, then _.cloneDeep

bgvozdev (approved):

spa/src/feature-flags.ts
krazziekay: Whats the reason for cloning? 
The value for `FRONTEND_FEATURE_FLAGS` never changes.
Locally, its always gonna be `###FRONTEND_FEATURE_FLAGS###` and in production/staging, it will be set when the react app is built and never after, so don't really see the need for cloning it.

PR #2427, opened 2023-09-13
'Cannot define if should send all' - fix bug throwing this error

joshkay10:
**What's in this PR?**
We have had about 350k error logs in the last 7 days in prod since the flag is returning false and treating that like an array.
Have checked its an array before trying to use .includes on it


bgvozdev (approved):

bgvozdev:
@tavomaciel FYI

PR #2431, opened 2023-09-18
NONE Enable eslint for worker directory

atraversatlassian:
**What's in this PR?**
Enable eslint for worker directory

**Why**
Improve lint coverage

**Added feature flags**
None

**Affected issues**  
NONE

**How has this been tested?**  
unit tests

**Whats Next?**
Enable more directories


src/sqs/backfill-discovery.test.ts
```diff
@@ -67,7 +67,7 @@ describe("Discovery Queue Test - GitHub Client", () => {
 			gitHubAppId: gitHubServerApp.id
 		});
 
-		await sqsQueues.backfill.start();
+		sqsQueues.backfill.start();
```
bgvozdev, line 70: TBH this change looks quite bizarre. In one place we are adding await, in another we are removing it, exactly for the same method...

bgvozdev (approved):

src/sqs/backfill-discovery.test.ts
atraversatlassian: This is clean up.  This function had previously been changed to not be async.  When that change was made these calls were not updated to remove the awaits here that no longer do anything.  They are awaiting a void instead the previous Promise<void>

PR #2436, opened 2023-09-19
NONE eslint services directory

atraversatlassian:
**What's in this PR?**
Enable eslint in services directory

**Why**
More lint

**Added feature flags**
None

**Affected issues**  
NONE

**How has this been tested?**  
Unit tests

**Whats Next?**
more lint


src/services/micros/lifecycle.ts
```diff
@@ -66,13 +66,13 @@ type Callback = (data: LifecycleMessageData) => void;
 // We're interested in type "Notification" and Subject "Micros Lifecycle Notification"
 // There are also messages on this queue used internally
 interface LifecycleData {
-	Type: string;
-	Subject: string;
-	Message: string;
+	Type?: string;
+	Subject?: string;
+	Message?: string;
```
bgvozdev, line 71: Curious about this change, should we really make our typing looser?

atraversatlassian: Where these are currently used it is doing a direct assignment from an any coming from parsing json or yaml.  We could keep these tighter in their definition but would need more validation when assigning objects to these interfaces

bgvozdev: I think we should do that IMHO, otherwise we will need to check if it is defined or not everywhere

atraversatlassian: I have updated LifecycleData back to having them as set properties and added undefined checks after assignment.

I left LifecycleMessageData with optionals on its fields for now as it is only used in one place of the code right after being converted from json

bgvozdev (approved):

PR #2442, opened 2023-09-19
NONE eslint sqs

atraversatlassian:
**What's in this PR?**
Enable eslint for sqs

**Why**
LINT

**Added feature flags**
None

**Affected issues**  
NONE

**How has this been tested?**  
unit tests

**Whats Next?**
More lint


src/sqs/backfill-error-handler.test.ts
```diff
@@ -208,9 +210,11 @@ describe("backfillErrorHandler", () => {
 		expect(result).toEqual({
 			isFailure: false
 		});
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
```
joshkay10, line 213: /nit ... can avoid the error suppression with something like
	```
	const messageMock = sendMessageMock.mock.calls[0] as number[];

		expect(messageMock[0]).toEqual(
			{ installationId: DatabaseStateCreator.GITHUB_INSTALLATION_ID, jiraHost }
		);
```

File: src/models/axios-error-event-decorator.ts [hunk 2]
```diff
@@ -27,7 +27,7 @@ export class AxiosErrorEventDecorator {
 		return this.response?.request;
 	}
 
-	static decorate(event: any, hint: any): any {
+	static decorate(this: void, event: any, hint: any): any {
```
joshkay10, line 30: /nit /opinion-piece lol - From what I understand, the `this: void` is just another way of saying this is a static method? Typescript will already enforce rules around calling this function with the `static` declaration so seems just like extra effort having  `this: void`

src/sqs/branch.test.ts
```diff
@@ -31,7 +31,7 @@ describe("Branch Webhook", () => {
 			jiraHost,
 			jiraClientKey: clientKey
 		});
-		await sqsQueues.branch.start();
```
joshkay10, line 34: ha, good catch!

src/sqs/branch.test.ts
```diff
@@ -120,9 +121,10 @@ describe("Branch Webhook", () => {
 		it("should not update Jira issue if there are no issue Keys in the branch name", async () => {
 			const getLastCommit = jest.fn();
 
-			await expect(app.receive(branchNoIssues as any)).toResolve();
+			await expect(app.receive(branchNoIssues)).toResolve();
 			expect(getLastCommit).not.toBeCalled();
 
+			// eslint-disable-next-line @typescript-eslint/require-await
 			await waitUntil(async () => {
```
joshkay10, line 128: /nit ughh waitUntil :shakesfist: we can get around this suppression with the old promise.all but I'm not sure if thats worse or better lol. Looks like eslint gets in a tiz when callbacks come into play... so it feels like a valid suppress?



```
			await waitUntil(async () => {
				await Promise.all([
					expect(githubNock).toBeDone(),
					expect(jiraNock).toBeDone()
				]);
```

src/models/axios-error-event-decorator.ts
atraversatlassian: This is related to this eslint rule https://typescript-eslint.io/rules/unbound-method/

My understanding of it is when passing this function into a method you don't get the class binding.  Adding this to the function ensures you don't try to access anything internally on the class in the method call.

joshkay10 (approved):

PR #2446, opened 2023-09-20
NONE: fix OOM

bgvozdev:
**What's in this PR?**
Fixing OOM

**Why**
Cause OOMs are bad, m'kay.

**Added feature flags**
Relying on tests and typescript

**How has this been tested?**  
tests and typescript

**Whats Next?**
Celerbrate! 🍾

### DIFF
--- a/src/models/reposyncstate.ts
+++ b/src/models/reposyncstate.ts
@@ -12,7 +12,7 @@ import {
 	STRING,
 	UpdateOptions,
 	JSON,
-	QueryTypes
+	QueryTypes, Order
 } from "sequelize";
 import { Subscription, TaskStatus } from "./subscription";
 import { merge } from "lodash";
@@ -215,13 +215,24 @@ export class RepoSyncState extends Model implements RepoSyncStateProperties {
 		return results[0] as RepoSyncState & Subscription;
 	}
 
-	static async findAllFromSubscription(subscription: Subscription, options: FindOptions = {}): Promise<RepoSyncState[]> {
+	static async findAllFromSubscription(subscription: Subscription, limit: number, offset: number, order: Order, options: FindOptions = {}): Promise<RepoSyncState[]> {
 		const result = await RepoSyncState.findAll(merge(options, {
+			where: {
+				subscriptionId: subscription.id
+			},
+			limit,
+			offset,
+			order
+		}));
+		return result || [];
+	}
+
+	static async countAllFromSubscription(subscription: Subscription, options: FindOptions = {}): Promise<number> {
+		return await RepoSyncState.count(merge(options, {
 			where: {
 				subscriptionId: subscription.id
 			}
 		}));
-		return result || [];
 	}
 
 	// TODO: move repoOwner to Subscription table and get rid of this.
--- a/src/models/reposyncstate.ts
+++ b/src/models/reposyncstate.ts
@@ -227,14 +227,6 @@ export class RepoSyncState extends Model implements RepoSyncStateProperties {
 		return result || [];
 	}
 
-	static async countAllFromSubscription(subscription: Subscription, options: FindOptions = {}): Promise<number> {
-		return await RepoSyncState.count(merge(options, {
-			where: {
-				subscriptionId: subscription.id
-			}
-		}));
-	}
-
 	// TODO: move repoOwner to Subscription table and get rid of this.
 	// The current schema implies a subscription might have multiple
 	// "repoOwner"s associated with it, while that's impossible
--- a/src/routes/api/api-reset-subscription-failed-tasks.ts
+++ b/src/routes/api/api-reset-subscription-failed-tasks.ts
@@ -28,10 +28,12 @@ export const ApiResetSubscriptionFailedTasks = async (req: Request, res: Respons
 
 	let offset = 0;
 	let hasNextPage = true;
+	const PAGE_SIZE = 100;
+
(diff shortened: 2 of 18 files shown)


src/models/reposyncstate.ts
```diff
@@ -215,15 +215,26 @@ export class RepoSyncState extends Model implements RepoSyncStateProperties {
...
-			}
+			},
+			limit,
+			offset,
+			order
 		}));
 		return result || [];
 	}
 
+	static async countAllFromSubscription(subscription: Subscription, options: FindOptions = {}): Promise<number> {
```
atraversatlassian, line 230: This function looks to do the same thing as countSubscriptionRepos on line 182

src/routes/api/api-reset-subscription-failed-tasks.ts
```diff
@@ -26,20 +26,27 @@ export const ApiResetSubscriptionFailedTasks = async (req: Request, res: Respons
...
+	let hasNextPage = true;
+	do {
+		const repoSyncStates = await RepoSyncState.findAllFromSubscription(subscription, 100, offset, [["repoFullName", "ASC"]]);
+		offset += 100;
+		hasNextPage = repoSyncStates.length === 100;
+
+		await Promise.all(repoSyncStates.map(async (repoSyncState) => {
+			let updated = false;
+			targetTasks.forEach(targetTask => {
+				if (repoSyncState[`${targetTask}Status`] === "failed") {
```
atraversatlassian, line 39: Could we pass this as filter to the RepoSyncState query and have it only return items that are failed instead?

src/routes/api/api-reset-subscription-failed-tasks.ts
```diff
@@ -26,20 +26,27 @@ export const ApiResetSubscriptionFailedTasks = async (req: Request, res: Respons
...
-		let updated = false;
-		targetTasks.forEach(targetTask => {
-			if (repoSyncState[`${targetTask}Status`] === "failed") {
-				repoSyncState[`${targetTask}Status`] = null;
-				logs.push({ repoSyncStateId: repoSyncState.id, targetTask });
-				updated = true;
+	let offset = 0;
+	let hasNextPage = true;
+	do {
+		const repoSyncStates = await RepoSyncState.findAllFromSubscription(subscription, 100, offset, [["repoFullName", "ASC"]]);
```
atraversatlassian, line 32: nit would be better if the 100 value could come from config

src/routes/api/installation/api-installation-syncstate-get.ts
```diff
@@ -4,9 +4,32 @@ import { Subscription } from "models/subscription";
...
+		return;
+	}
+
 	const githubInstallationId = Number(req.params.installationId);
 	const gitHubAppId = Number(req.params.gitHubAppId) || undefined;
 	const jiraHost = req.params.jiraHost;
+	const limit = Number(req.query.limit);
+	const offset = Number(req.query.offset);
+
+	if (limit > 10000) {
```
atraversatlassian, line 27: nit 10000 coming from config better.  At least make it a constant value

atraversatlassian (dismissed):

src/routes/api/api-reset-subscription-failed-tasks.ts
bgvozdev: that's some API endpoint that we probably never use; I wouldn't bother too much about it

src/models/reposyncstate.ts
bgvozdev: Awesome find! :killwithfire:

src/routes/api/api-reset-subscription-failed-tasks.ts
bgvozdev: That was an existing code, don't want to touch it. I only wrapped it in a loop for pagination (changed how we call `findAllFromSubsscription`), the rest is left as it was.

src/routes/api/installation/api-installation-syncstate-get.ts
bgvozdev: just an admin endpoint we probably never used, I woudn't bother much

src/routes/api/api-reset-subscription-failed-tasks.ts
bgvozdev: turned into a constant though :) thanks!

src/routes/api/installation/api-installation-syncstate-get.ts
bgvozdev: turned into a constant though :) thanks!

src/routes/api/installation/api-installation-syncstate-get.ts
```diff
@@ -4,9 +4,34 @@ import { Subscription } from "models/subscription";
 import { pick } from "lodash";
 
 export const ApiInstallationSyncstateGet = async (req: Request, res: Response): Promise<void> => {
+	if (req.query.limit === undefined) {
```
atraversatlassian, line 7: could query be undefined? and need a null propagator

atraversatlassian (approved):

src/routes/api/installation/api-installation-syncstate-get.ts
bgvozdev: No, I think it is always defined, at least according to express typing

PR #2455, opened 2023-09-22
NONE eslint rest

atraversatlassian:
**What's in this PR?**
eslint for rest directory

**Why**
lint

**Added feature flags**
none
**Affected issues**  
None

**How has this been tested?**  
test

**Whats Next?**
more lint


src/models/jira-client.ts
```diff
@@ -9,7 +9,7 @@ import { envVars } from "config/env";
 export class JiraClient {
 	axios: AxiosInstance;
 
-	static async getNewClient(installation: Installation, log: Logger) {
+	static async getNewClient(this: void, installation: Installation, log: Logger) {
```
bgvozdev, line 12: Curious how does this work? What will happen with the current calls (which passes 2 args only)?

src/rest/routes/oauth/index.ts
```diff
@@ -9,18 +9,25 @@ export const OAuthRouter = Router({ mergeParams: true });
 OAuthRouter.get("/redirectUrl", errorWrapper("OAuthRedirectUrl", async function OAuthRedirectUrl(req: Request, res: Response<GetRedirectUrlResponse>) {
 
 	const cloudOrUUID = req.params.cloudOrUUID;
-	const { jiraHost } = res.locals;
+	const jiraHost = res.locals["jiraHost"] as string | undefined;
```
bgvozdev, line 12: `||`

bgvozdev: ah, sorry, ignore me

src/routes/jira/workspaces/jira-workspaces-get.ts
```diff
@@ -58,8 +58,8 @@ export const JiraWorkspacesGet = async (req: Request, res: Response): Promise<vo
 	}
 
 	const orgName = sanitizeHtml(req.query?.searchQuery as string);
-	const page = Number(sanitizeHtml(req.query?.page)) || DEFAULT_PAGE_NUMBER;
-	const limit = Number(sanitizeHtml(req.query?.limit)) || DEFAULT_LIMIT;
+	const page = Number(sanitizeHtml(req.query?.page as string ?? "undefined")) || DEFAULT_PAGE_NUMBER;
```
bgvozdev, line 61: nit could be simplified with a ternary, I think

```
req.query?.page ? Number(sanitizeHtml(req.query.page as string)) : DEFAULT_PAGE_NUMBER
```

bgvozdev: (here and below in this file and in the next one)

src/models/jira-client.ts
atraversatlassian: You don't need to update the calls.  You are just telling the class that 'this' needs to be treated as void in the scope of the function

bgvozdev (approved):

PR #2457, opened 2023-09-25
5KU-ENHANCEMENT - Customized scrollbars added

krazziekay:
**What's in this PR?**
- Added component for showing custom scrollbars.
- Removed the gradient logic

**Why**
- Better UI for users with long list of orgs

**Test**
- Added test case for error states
- Tested in staging

**Video**

https://github.com/atlassian/github-for-jira/assets/13076255/0d06238b-8c01-402d-a27f-edebb17f935a


spa/src/pages/ConfigSteps/OrgsContainer/index.tsx
```diff
@@ -103,66 +81,68 @@ const OrganizationsList = ({
 		}
 	};
 	return (
-		<>
-			<div css={orgsWrapperStyle} ref={containerRef}>
+		<Scrollbars
+			style={{
+			maxHeight: 250,
```
rachellerathbone, line 86: Nit: can we avoid using magic numbers here and opt for constants? Will just make the code a little more readable.

spa/src/pages/ConfigSteps/OrgsContainer/index.tsx
```diff
@@ -103,66 +81,68 @@ const OrganizationsList = ({
...
-											}}
-										>
-											Connect
-										</Button>
-									)}
-								</>
-							) : (
-								<>
-									<div>
+						const hasError = !canConnect(org);
```
rachellerathbone, line 93: Can we add some tests for this component? Don't think we need anything too crazy but we should at least have some for canConnect with different inputs to ensure it returns the correct result based on the organization's properties (requiresSsoLogin, isIPBlocked, isAdmin).

krazziekay: Added test cases for the error conditions.

spa/src/pages/ConfigSteps/OrgsContainer/index.tsx
krazziekay: Done

rachellerathbone (approved):

PR #2460, opened 2023-09-26
fix: whow browser popup block alter

kamaksheeAtl:
What's in this PR?
This is to fix an error:
- Sentry link: [link](https://atlassian-2y.sentry.io/issues/4451313518/?project=4505604384292864&query=replayId%3Aa389024472974f6d942051975ac2f96c&referrer=issue-stream&stream_index=0)
- Jira link: [link](https://softwareteams.atlassian.net/jira/software/projects/ARC/boards/385?selectedIssue=ARC-2457)
image

Why?
As there are some blocked popups in the flow of GH4J config 5ku exp, we tried to enhance the user design to address this issue. The goal is to detect blocked popups on the browser setting and show an alert message to encourage users to unblock the popups for GH4J 5ku exp.

Added feature flags:
- Not applicable

Affected issues:
- Sentry link: [link](https://atlassian-2y.sentry.io/issues/4451313518/?project=4505604384292864&query=replayId%3Aa389024472974f6d942051975ac2f96c&referrer=issue-stream&stream_index=0)

How has this been tested?
Manually on local and stg env

What's Next?

### DIFF
--- a/spa/src/app.tsx
+++ b/spa/src/app.tsx
@@ -18,7 +18,7 @@ const SentryRoutes = Sentry.withSentryReactRouterV6Routing(Routes);
 
 const App = () => {
 	// to reset the local storage value on app reload
-	localStorage.setItem("isPopupBlocked", "false");
+	localStorage.removeItem("isPopupBlocked");
 	setGlobalTheme({
 		light: "light",
 		dark: "dark",
--- a/spa/src/app.tsx
+++ b/spa/src/app.tsx
@@ -17,7 +17,8 @@ initSentry();
 const SentryRoutes = Sentry.withSentryReactRouterV6Routing(Routes);
 
 const App = () => {
-
+	// to reset the local storage value on app reload
+	localStorage.setItem("isPopupBlocked", "false");
 	setGlobalTheme({
 		light: "light",
 		dark: "dark",
--- a/spa/src/app.tsx
+++ b/spa/src/app.tsx
@@ -18,7 +18,9 @@ const SentryRoutes = Sentry.withSentryReactRouterV6Routing(Routes);
 
 const App = () => {
 	// to reset the local storage value on app reload
-	localStorage.removeItem("isPopupBlocked");
+	if(localStorage.getItem("isPopupBlocked")){
+		localStorage.removeItem("isPopupBlocked");
+	}
 	setGlobalTheme({
 		light: "light",
 		dark: "dark",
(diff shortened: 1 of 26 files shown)


spa/src/services/oauth-manager/index.ts
```diff
@@ -53,6 +55,13 @@ async function authenticateInGitHub(onWinClosed: () => void): Promise<void> {
 				}
 			}, 1000);
 		}
+		else{
```
krazziekay, line 58: This is just one method that uses popup, there are other methods too. So, instead of adding else to just one method here, its better to throw this error in the `popup` method.

krazziekay (changes requested):

spa/src/services/oauth-manager/index.ts
kamaksheeAtl: I think its better to keep it as null 
as if we are throwing the error here and using the popup method into onClick event 
If in case we forget to handle it for some cases 
It will break the UI

krazziekay: With this way we would need to add this same else check in all the places where `popup` has been used.
Instead let's just use check inside `popup` method. If the popup is not opened, then set a `popupNotEnabled` value within a constant or localStorage. And simply watch this value in the top level component, maybe `header` or `wrapper`, which will display the error banner when it is set.
Can talk over in details in call...

spa/src/utils/index.ts
```diff
@@ -10,9 +10,10 @@ export function popup (url: string) {
 	const openedPopup = window.open(url, "_blank");
 	if (!openedPopup || openedPopup.closed || typeof openedPopup.closed === "undefined") {
 		// TODO: Meaningful UI when browser pop-up is blocked
```
krazziekay, line 12: Can this remove this todo now.

spa/src/app.tsx
```diff
@@ -17,7 +17,8 @@ initSentry();
 const SentryRoutes = Sentry.withSentryReactRouterV6Routing(Routes);
 
 const App = () => {
-
+	// to reset the local storage value on app reload
+	localStorage.removeItem("isPopupBlocked");
```
krazziekay, line 21: Please add a check first before removing it. 
```suggestion
	if (localStorage.getItem("isPopupBlocked") === null) {
	  // to reset the local storage value on app reload
	  localStorage.removeItem("isPopupBlocked");
}
```

krazziekay (dismissed):

krazziekay (approved):

PR #2462, opened 2023-09-27
ARC-2290 new jira client

joshkay10:
**What's in this PR?**
Created a new JiraClient class and merged all the requests from the other two JiraClients - this is not the final product... but it brings everything to one and we can enhance from here.

**Why**
We have wanted to do it for a while, the existing JiraClients are in two seperate directories and needs to be tidied up converted to a more developer friendly format... and Typed, being two seperate clients wasn't an issue as they do serve slighty different functions one being Jira requests for GitHub installation data the other more  general Jira requests eg getCloudId......  

also, I need to introduce a new TCS(tenantContextService) sidecar to run a Axios interceptor before trying to update suspended Jira accounts, so now was the time.

**Added feature flags**
Not required for this change, no code is being used in Production, next steps will start consuming with FFs.

**Affected issues**  
_Jira Issues_

**How has this been tested?**  

**Whats Next?**
Tests for the requests, there is a nbunch there that are fine, but a lot are missing, this is intentional as the PR was already too big. next PR will be just tests.

Also have a few helper functions i will pull out and import back in.

Will start consuming one endpoint at a time, behind feature flags, expect many small PRs

As mentioned in the WHY , they did originally serve slightly different purposes, we will likely split up this new client into smaller classes much like GHClients.


src/jira/client/jira-api-client.test.ts
```diff
@@ -0,0 +1,108 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { getLogger } from "config/logger";
+import { JiraClient } from "./jira-api-client";
+import { DatabaseStateCreator } from "test/utils/database-state-creator";
+
+
```
joshkay10, line 6: Pulled the models/jiraclient tests across and just tweaked the before each to fit the new create style, otherwise same tests

src/jira/client/jira-api-client.ts
```diff
@@ -0,0 +1,494 @@
+import Logger from "bunyan";
```
joshkay10, line 1: Sorry its massive, mostly just brought things over from old clients as is, will highlight with comments changes

src/jira/client/jira-api-client.ts
```diff
@@ -0,0 +1,494 @@
...
+import { Installation } from "models/installation";
+import { Subscription } from "models/subscription";
+import { envVars } from "config/env";
+import { uniq } from "lodash";
+import { jiraIssueKeyParser } from "utils/jira-utils";
+import { TransformedRepositoryId, transformRepositoryId } from "~/src/transforms/transform-repository-id";
+import { getJiraId } from "../util/id";
+import { getCloudOrServerFromGitHubAppId } from "utils/get-cloud-or-server";
+import { getDeploymentDebugInfo, extractDeploymentDataForLoggingPurpose } from "./jira-client-deployment-helper";
+import {
```
joshkay10, line 13: all these functions used to exists  here and there around the outside of the classes, i have bundled them into a helper file... more on that when we get to it

src/jira/client/jira-api-client.ts
```diff
@@ -0,0 +1,494 @@
...
+		await jiraClient.initialize(installation);
+		return jiraClient;
+	}
+
+	private async initialize(installation: Installation): Promise<void> {
+		const secret = await installation.decrypt("encryptedSharedSecret", this.logger);
+		this.axios = getAxiosInstance(installation.jiraHost, secret, this.logger);
+	}
+
+	private constructor(jiraHost: string, gitHubInstallationId: number, gitHubAppId: number | undefined, logger: Logger) {
```
joshkay10, line 59: **private constructor** - we need to make a async call to get a secret and since
constructors are a synchronous call as a rule, we get around it with the create function calling the constructor, its private so noone accidentally tries.

its similar to one of the old JCs but its just cleaned up and works for all now.

src/jira/client/jira-api-client.ts
```diff
@@ -0,0 +1,494 @@
...
+		const payload = {
+			accountId,
+			globalPermissions: [
+				"ADMINISTER"
+			]
+		};
+		return await this.axios.post("/rest/api/latest/permissions/check", payload);
+	}
+
+	// TODO TEST
```
joshkay10, line 119: Lots of TODO TEST comments, left there intentionally, after this PR I will add a test block for each of these functions....

src/jira/client/jira-api-client.ts
```diff
@@ -0,0 +1,494 @@
...
+	const uniqueCommits = commits.reduce((accumulator: JiraCommit[], currentCommit: JiraCommit) => {
+		if (!accumulator.some((commit) => commit.id === currentCommit.id)) {
+			accumulator.push(currentCommit);
+		}
+		return accumulator;
+	}, []);
+	return uniqueCommits;
+};
+
+// TODO MOVE TO new jira-client-commit-helper.ts
```
joshkay10, line 449: More TODO i was getting concious of the PR size, deduplicateCommits and batchedBulkUpdate will be moved to commit-helper file

src/jira/client/jira-api-client.ts
```diff
@@ -0,0 +1,494 @@
...
+			},
+			operationType: options?.operationType || "NORMAL"
+		};
+		this.logger.info("Sending vulnerabilities payload to jira.");
+		return await this.axios.post("/rest/security/1.0/bulk", payload);
+	}
+}
+
+// TODO MOVE TO new jira-client-commit-helper.ts
+const deduplicateCommits = (commits: JiraCommit[] = []): JiraCommit[] => {
```
joshkay10, line 439: This is a re-write of dedupCommits in one of the JCs, reason being I found the other one unclear. Have just simplified it a bit but is still concise

src/jira/client/jira-client-deployment-helper.ts
```diff
@@ -16,3 +18,21 @@ export const getDeploymentDebugInfo = (jiraPayload: JiraDeploymentBulkSubmitData
 		commitCount: associations.filter(a => a.associationType === "commit").map(a => a.values?.length || 0).reduce((a, b) => a + b, 0)
 	};
 };
+
+export const extractDeploymentDataForLoggingPurpose = (data: JiraDeploymentBulkSubmitData, logger: Logger): Record<string, any> => {
```
joshkay10, line 22: Just pulled from JC and placed here so we can import it in. code is same same. Will add a TODO for tests!

src/jira/client/jira-client-issue-key-helper.ts
```diff
@@ -0,0 +1,172 @@
+import { uniq } from "lodash";
```
joshkay10, line 1: ooooo again a lot of these are just pulled from JC and dumped in here, I've done a bunch of house keeping, more typings but still not quite done.

will highlight any major changes

src/jira/client/jira-client-issue-key-helper.ts
```diff
@@ -0,0 +1,172 @@
...
+ * Truncates branches, commits and PRs to their first ISSUE_KEY_API_LIMIT issue keys
+ */
+export const truncateIssueKeys = (repositoryObj) => {
+	updateRepositoryIssueKeys(repositoryObj, truncate);
+};
+
+/**
+ * Get truncated issue keys and associations based on the ISSUE_KEY_API_LIMIT.
+ */
+export const getTruncatedIssueKeys = (data: IssueKeyObject[] = []): IssueKeyObject[] =>
```
joshkay10, line 35: I had to rewrite this function, old version wasn't truncating... picked it up when writing tests.

src/jira/client/jira-client-issue-key-helper.test.ts
```diff
@@ -0,0 +1,667 @@
+import {
```
joshkay10, line 1: These are all fresh tests, has super high coverage....
I have broken each function into its own describe block..... since the are all isolated its pretty easy to follow i think. I will have a re-review over these as well.

joshkay10:
Sigh, moving to https://github.com/atlassian/github-for-jira/pull/2465

PR #2469, opened 2023-09-28
Arc 2496 update 5ku success screen

kamaksheeAtl:
**What's in this PR?**
Update success screen of 5KU config experience
Part of GH4J: Post Phase-1 Improvements
image

**Why**
It was not clear where we want users to go next
https://hello.atlassian.net/wiki/spaces/OTFS/pages/2877686448/GH4J+Post+Phase-1+Improvements#Update-success-page

**Added feature flags**
NA

**Affected issues**  
[_Jira Issues_](https://softwareteams.atlassian.net/browse/ARC-2496)

**How has this been tested?**  
Locally and added unit test-cases 

**Whats Next?**

### DIFF
--- a/spa/src/pages/Connected/index.tsx
+++ b/spa/src/pages/Connected/index.tsx
@@ -23,7 +23,7 @@ const titleStyle = css`
 const paragraphStyle = css`
 	color: ${token("color.text.subtle")};
 	margin: ${token("space.300")} ${token("space.0")};
-	padding: 0px ${token("space.500")};
+	padding: 0px ${token("space.800")};
 `;
 const flexWrapperStyle = css`
 	padding: ${token("space.400")} ${token("space.0")};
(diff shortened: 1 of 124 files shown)


spa/src/pages/Connected/index.tsx
```diff
@@ -47,18 +43,16 @@ const sectionStyle = css`
...
-`;
 
 const buttonStyle = css`
-	padding-left: 0;
-	height: 14px;
-	line-height: 14px;
-	display: inline-flex;
+	margin: 0px 10px;
+`;
+const subtleBtnStyle = css`
```
kamaksheeAtl, line 50: have applied !important to the color prop of the Atlas Toolkit button, because According to the UX provided to us, the button with the appearance of 'subtle' cannot be used as is, as its color looks a bit darker than what is shown in Figma.
and the original Button comp has css color prop with !important, so cant override it with normal css specificity!

Please suggest a solution.

krazziekay (approved):

PR #2475, opened 2023-10-04
fix: org with zero repo count display

kamaksheeAtl:
**What's in this PR?**
Currently we are still not showing repo count as zero for the org with zero repo 
Issue:
Screenshot 2023-10-04 at 9 12 19 am
Applying fix for the same.
Fix:
Screenshot 2023-10-04 at 11 24 39 am

**Why**
https://softwareteams.atlassian.net/browse/ARC-316

**Added feature flags**
NA

**Affected issues**  
[_Jira Issues_](https://softwareteams.atlassian.net/browse/ARC-316)

**How has this been tested?**  
Manually in local env and via unit test cases

**Whats Next?**

### DIFF
--- a/src/util/handlebars/handlebar-helpers.ts
+++ b/src/util/handlebars/handlebar-helpers.ts
@@ -27,7 +27,7 @@ export const registerHandlebarsHelpers = () => {
 				: `${numberOfSyncedRepos} / ${totalNumberOfRepos}`
 	);
 
-	hbs.registerHelper("checkRepoCount", (totalNumberOfRepos: number) => (totalNumberOfRepos >= 0));
+	hbs.registerHelper("checkRepoCount", (totalNumberOfRepos: unknown | undefined) => (typeof totalNumberOfRepos === "number" &&  totalNumberOfRepos >= 0));
 
 	hbs.registerHelper("repoAccessType", (repository_selection: string) =>
 		repository_selection === "all" ? "All repos" : "Only select repos"
--- a/src/util/handlebars/handlebar-helpers.ts
+++ b/src/util/handlebars/handlebar-helpers.ts
@@ -27,8 +27,7 @@ export const registerHandlebarsHelpers = () => {
 				: `${numberOfSyncedRepos} / ${totalNumberOfRepos}`
 	);
 
-	//check to display repo count as zero in case of org with zero repos
-	hbs.registerHelper("hasTotalNumberOfRepos", (totalNumberOfRepos: number) => (totalNumberOfRepos || totalNumberOfRepos === 0));
+	hbs.registerHelper("checkRepos", (totalNumberOfRepos: number) => (totalNumberOfRepos >= 0));
 
 	hbs.registerHelper("repoAccessType", (repository_selection: string) =>
 		repository_selection === "all" ? "All repos" : "Only select repos"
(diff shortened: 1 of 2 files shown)


src/util/handlebars/handlebar-helpers.ts
```diff
@@ -27,6 +27,9 @@ export const registerHandlebarsHelpers = () => {
 				: `${numberOfSyncedRepos} / ${totalNumberOfRepos}`
 	);
 
+	//check to display repo count as zero in case of org with zero repos
```
bgvozdev, line 30: nit comment duplicating code

bgvozdev: Either we should remove it or change the code to be expressive enough so it won't need comments describing what it does ;)

bgvozdev (dismissed):

src/util/handlebars/handlebar-helpers.ts
```diff
@@ -27,6 +27,9 @@ export const registerHandlebarsHelpers = () => {
...
+	hbs.registerHelper("hasTotalNumberOfRepos", (totalNumberOfRepos: number) => (totalNumberOfRepos || totalNumberOfRepos === 0));
```
fusion-arc-e2e-test, line 31: This helper name is a bit confusing, can we change it to something like `checkRepos` or something.
Also, I think its better to do a check to `totalNumberOfRepos >= 0` than `OR` check cause this `OR` check will fail for unlikely scenarios like negative numbers.

src/util/handlebars/handlebar-helpers.ts
```diff
@@ -27,8 +27,7 @@ export const registerHandlebarsHelpers = () => {
 				: `${numberOfSyncedRepos} / ${totalNumberOfRepos}`
 	);
 
-	//check to display repo count as zero in case of org with zero repos
-	hbs.registerHelper("hasTotalNumberOfRepos", (totalNumberOfRepos: number) => (totalNumberOfRepos || totalNumberOfRepos === 0));
+	hbs.registerHelper("checkRepoCount", (totalNumberOfRepos: number) => (totalNumberOfRepos >= 0));
```
bgvozdev, line 30: Wondering if  `typeof totalNumberOfRepos === 'number'` would make it more clear?...

kamaksheeAtl: yes it would !

kamaksheeAtl: no i thought applying that as an additional check type check and no to be equal or grater than zero

bgvozdev: `typeof totalNumberOfRepos === 'number' && totalNumberOfRepos >=0` looks good to my taste :)

krazziekay: We already have typescript check for that, so feels redundant

krazziekay (dismissed):

bgvozdev (approved):

src/util/handlebars/handlebar-helpers.ts
```diff
@@ -27,6 +27,8 @@ export const registerHandlebarsHelpers = () => {
 				: `${numberOfSyncedRepos} / ${totalNumberOfRepos}`
 	);
 
+	hbs.registerHelper("checkRepoCount", (totalNumberOfRepos: unknown | undefined) => (typeof totalNumberOfRepos === "number" &&  totalNumberOfRepos >= 0));
```
joshkay10, line 30: /nit can we use `number` instead of `unknown`? There may be handlebarsy reasons you did that so ok to ignore if thats the case

kamaksheeAtl: yes @joshkay10 
this has been done because of handlebar reason

PR #2476, opened 2023-10-04
golden image usage for node

joshkay10:
**What's in this PR?**
Use the new Golden image from 

**Whats not in this PR**
Local development changes, once we have settled on this we can update the local dockerfile to work


Dockerfile.prod
```diff
@@ -1,21 +1,25 @@
-FROM node:14.21-alpine3.16 as build
+FROM docker.atl-paas.net/sox/micros-node-14:1.0.5 as build
+
+# root user for installs
+USER root
 
 # adding python for node-gyp
-RUN apk add g++ make python3
```
joshkay10, line 4: no longer Alpine distribution so apk gets replaces with apt-get for Debian

Dockerfile.prod
```diff
@@ -1,21 +1,25 @@
...
-RUN apk add --update --upgrade openssl
-RUN apk add --update --upgrade zlib
+COPY . /opt/service
 
-COPY . /app
+WORKDIR /opt/service
 
-WORKDIR /app
+# Change ownership of the /opt/service directory to appuser
+USER root
```
joshkay10, line 20: need to change ownership to allow these next steps to work as appuser...
using root the whole way through would likely work but its a big no no security wise. its slowwww so might see if we can do it a bit cleverer

Dockerfile.prod
```diff
@@ -25,31 +29,29 @@ RUN yarn list
...
 
 # For coredumps
-RUN apk add gdb
-RUN apk add bash
+RUN apt-get install -y gdb
 
-USER node
-COPY --chown=node:node --from=build /app /app
+# non-root user
+USER appuser
```
joshkay10, line 48: appuser is a predefined user on the image, node user does not exist and appuser is same-same-but-dfferent

Dockerfile.prod
gxueatlassian: I don't get why the whole change to root/user back and forth. 
Does the `/opt/service` contains other stuff? If so, can we `COPY /opt/service/app` instead?

joshkay10: This was exactly the comment i was looking for!! maybe? the chown call takes sooooo long.

joshkay10:
Not moving to this image.

PR #2482, opened 2023-10-05
Arc 1192 update backfilling status

kamaksheeAtl:
**What's in this PR?**

**Why**

**Added feature flags**

**Affected issues**  
_Jira Issues_

**How has this been tested?**  

**Whats Next?**


src/routes/jira/jira-get-connections-backfill-status.ts
```diff
@@ -0,0 +1,113 @@
...
+		return next(new Error(`Failed to render connected repos`));
+	}
+};
+
+const getBackfillCompletionStatus = (backfillStatus: BackFillType): boolean =>
+	Object.values(backfillStatus).every(
+		(backFill: SubscriptionBackfillState): boolean => backFill?.isSyncComplete
+	);
+
+const getBackfillStatus = async (subscriptionsById): Promise<BackFillType> => {
```
kamaksheeAtl, line 91: I have created a new PR as i made loads of changes in this file to keep this backfill status polling insync with the way we are providing data to jira-configuration-table.hbs via getInstallation

src/routes/jira/jira-get-connections-backfill-status.ts
```diff
@@ -0,0 +1,113 @@
...
+			.map(Number)
+			.filter(Boolean);
+
+		if (subscriptionIds.length === 0) {
+			req.log.warn("Missing Subscription IDs");
+			res.status(400).send("Missing Subscription IDs");
+			return;
+		}
+
+		const subscriptions = await Subscription.findAll({
```
joshkay10, line 43: theres a `findAllForSubscriptionIds` on Subscriptions you could use here if ya wanna

KamaksheeSamant: yeah this works fine as per the tests 
making the change

src/routes/jira/jira-get-connections-backfill-status.ts
```diff
@@ -0,0 +1,110 @@
...
+			req.log.error("Missing Subscription");
+			res.status(400).send("Missing Subscription");
+			return;
+		}
+
+		const jiraHosts = subscriptions.map(
+			(subscription) => subscription?.jiraHost
+		);
+
+		const jiraHostsMatched = jiraHosts.every(
```
joshkay10, line 59: /nit 🤔 - i think you could drop the map and just use the every like this

const jiraHostsMatched = subscriptions.every(
    (subscription) => subscription?.jiraHost === localJiraHost
);

src/routes/jira/jira-get-connections-backfill-status.ts
```diff
@@ -0,0 +1,110 @@
...
+		const isBackfillComplete = getBackfillCompletionStatus(backfillStatus);
+		res.status(200).send({
+			data: {
+				subscriptions: backfillStatus,
+				isBackfillComplete,
+				subscriptionIds: resultSubscriptionIds
+			}
+		});
+	} catch (error) {
+		return next(new Error(`Failed to render connected repos`));
```
joshkay10, line 79: Can we log the error plz
	`logger.error({ error }, "oh nooooooooooooooooooooo");`

src/routes/jira/jira-get-connections-backfill-status.ts
```diff
@@ -0,0 +1,110 @@
...
+		const isSyncComplete =
+			subscription?.syncStatus === SyncStatus.COMPLETE ||
+			subscription?.syncStatus === SyncStatus.FAILED;
+		const failedSyncErrors = await getRetryableFailedSyncErrors(subscription);
+
+		backfillStatus[subscriptionId] = {
+			isSyncComplete,
+			syncStatus: mapSyncStatus(subscription?.syncStatus),
+			totalRepos: subscription?.totalNumberOfRepos,
+			syncedRepos: await RepoSyncState.countFullySyncedReposForSubscription(
```
joshkay10, line 101: if either of the awaits fail inside the loop it will short circuit so might wanna try catch around it so it keeps going, unless you want it to quit then ignore more lol

```
const failedSyncErrors = await getRetryableFailedSyncErrors(subscription);
syncedRepos: await RepoSyncState.countFullySyncedReposForSubscription(
```

src/routes/jira/jira-get-connections-backfill-status.ts
kamaksheeAtl: yup thats r8 short and sweet! done

src/routes/jira/jira-get-connections-backfill-status.ts
kamaksheeAtl: Done

src/routes/jira/jira-get-connections-backfill-status.ts
kamaksheeAtl: well, this is a good suggestion!
Yes in this case till the time we dont get a separate UX to show the error when we fail backfill status polling fails for a particular subscription
it will be better to exclude that result from final result and continue with others

src/routes/jira/jira-get-connections-backfill-status.ts
```diff
@@ -0,0 +1,130 @@
...
+		const subscriptions = await Subscription.findAllForSubscriptionIds(
+			subscriptionIds
+		);
+
+		const resultSubscriptionIds = subscriptions.map(
+			(subscription) => subscription.id
+		);
+
+		if (!subscriptions || subscriptions.length === 0) {
+			req.log.error("Missing Subscription");
```
rachellerathbone, line 56: We have an enum set up for this in error.ts. Just import `MISSING_SUBSCRIPTION`

src/routes/jira/jira-get-connections-backfill-status.ts
```diff
@@ -0,0 +1,130 @@
...
+				subscriptions: backfillStatus,
+				isBackfillComplete,
+				subscriptionIds: resultSubscriptionIds,
+				errors
+			}
+		});
+	} catch (error) {
+		req.log.error(
+			{ error },
+			"Failed to poll repo backfill status for provided subscription ID"
```
rachellerathbone, line 86: nit: pull this out into a const instead of repeating the same string.

gxueatlassian (changes requested):
Sorry, code looks good. Just quite a bit of complexity in the `.js` file, and this is not cover in our unit test / e2e test. Also I'd like to push towards using spa/react for the backfill page, then the update of status will be just a `useEffect`, should be much easier. 

Let me know if you object this, happy to discuss.

PR #2484, opened 2023-10-05
attacking rate limit with exp back off and priority by count

joshkay10:
**What's in this PR?**
A two pronged assault on ratelimit retries.

We gets bursts of traffic and it overwhelms the rate limit, we try again on ratelimit reset, but the data is so large that it just keeps getting rate limited....

Prong 1
sooo an attempt to semi-normalize the incoming request over a greater spread of time... so we take the reset time and add additional waits receivecount x Github reset time

Prong 2
We have a buffer so we dont accidently put it back just before the reset / grace for Github to be updated....
i have reused this and given priority to messages that have been seen more often, so a receive count of 5 will be visible before a receive count of 1.... in a ultra busy environment this may not make a huge difference since our queue arent FIFO but might be just enough breathing room!!!?!?!!?


src/sqs/error-handlers.ts
```diff
@@ -86,9 +87,12 @@ const maybeHandleNonRetryableResponseCode = <MessagePayload extends BaseMessageP
 const maybeHandleRateLimitingError = <MessagePayload extends BaseMessagePayload>(error: Error, context: SQSMessageContext<MessagePayload>): ErrorHandlingResult | undefined => {
 	if (error instanceof GithubClientRateLimitingError) {
 		context.log.warn({ error }, `Rate limiting error, retrying`);
-		const delaySec = error.rateLimitReset + RATE_LIMITING_DELAY_BUFFER_SEC - (Date.now() / 1000);
-		return { retryable: true, retryDelaySec: delaySec, isFailure: true };
+		const buffer = BASE_RATE_LIMITING_DELAY_BUFFER_SEC - context.receiveCount * 10;
```
bgvozdev, line 90: Hmm... 1. could we please move 10 to a constant with a descriptive name? 2. could it be negative, should we handle it accordingly?

joshkay10: good news, i just pushed that exact change when you made this comment.

will look at negative option now

bgvozdev: 3. why do we need this sophisticated logic? really we are just calculating some buffer which is from 0 to 60 seconds. I think a clarifying comment is needed

src/sqs/error-handlers.ts
```diff
@@ -86,9 +87,12 @@ const maybeHandleNonRetryableResponseCode = <MessagePayload extends BaseMessageP
...
+		const rateLimitReset = error.rateLimitReset + buffer - Date.now() / 1000;
+		const retryDelaySec = rateLimitReset + ONE_HOUR_IN_SECONDS * (context.receiveCount - 1);
```
bgvozdev, line 93: That's quite aggressive. I think we need to put a comment clarifying this also

joshkay10: Commenting.

src/sqs/error-handlers.ts
```diff
@@ -85,10 +86,16 @@ const maybeHandleNonRetryableResponseCode = <MessagePayload extends BaseMessageP
 
 const maybeHandleRateLimitingError = <MessagePayload extends BaseMessagePayload>(error: Error, context: SQSMessageContext<MessagePayload>): ErrorHandlingResult | undefined => {
 	if (error instanceof GithubClientRateLimitingError) {
+		const DELAY_BUFFER_STEP = 10;
```
bgvozdev, line 89: shouldn't it be among all the constants? And probably prefixed with `RATE_LIMIT_`?

src/sqs/error-handlers.ts
```diff
@@ -85,10 +86,16 @@ const maybeHandleNonRetryableResponseCode = <MessagePayload extends BaseMessageP
...
 		context.log.warn({ error }, `Rate limiting error, retrying`);
-		const delaySec = error.rateLimitReset + RATE_LIMITING_DELAY_BUFFER_SEC - (Date.now() / 1000);
-		return { retryable: true, retryDelaySec: delaySec, isFailure: true };
+		const buffer = Math.max(DELAY_BUFFER_STEP, BASE_RATE_LIMITING_DELAY_BUFFER_SEC - context.receiveCount * DELAY_BUFFER_STEP);
```
bgvozdev, line 91: Sill would like to see a comment clarifying this line, too ;)

joshkay10: Clarified! please let me know if it makes sense though!

src/sqs/error-handlers.ts
joshkay10: haha, i got stuck between two worlds, a lot of me wanted to bring the other ones inside here. moved with the rest!

bgvozdev (approved):

PR #2492, opened 2023-10-11
FIX - FF value not being reflected in SPA

krazziekay:
**What's in this PR?**
- Separated the logic to add the FF value into the build HTML pages from the `if` check.

**Why**
- The FF value from LaunchDarkly was not being picked up in SPA.

**How has this been tested?**  
- Staging


src/routes/spa/spa-router.ts
```diff
@@ -17,16 +17,19 @@ let indexHtmlContent: string = "";
 
 SpaRouter.use(JwtHandlerWithoutQsh);
 SpaRouter.use("/*", async (_, res) => {
+	const { jiraHost } = res.locals;
+	const featureFlags = {
+		ENABLE_5KU_BACKFILL_PAGE: await booleanFlag(BooleanFlags.ENABLE_5KU_BACKFILL_PAGE, jiraHost)
```
gxueatlassian, line 22: The `jiraHost` is not available at this stage, as this is the design principle of SPA:
1. the static html/js/css is always static, no authentication token will be passed in the AUTHORIZATION header, hence the router won't have jiraHost in the res.locals

---------

Having say that, we can do either of bellow:

1. Read the query param (`jwt`) here directly, and try to decode it (by reusing some logic the JwtHandler) and get the jiraHost. But don't fail the page if token not availabe, as the defer install will need this page for GH admin (which dosen't have jira token)
2. Or alternatively, in the spa Connected component, always do `AP.navigator.go("addonmodule", { moduleKey: "gh-addon-admin" });` and now do that navigation in the backend routing.

src/routes/spa/spa-router.ts
```diff
@@ -17,16 +17,19 @@ let indexHtmlContent: string = "";
...
-		};
-
-		indexHtmlContent = (await fs.readFile(path.join(process.cwd(), "spa/build/index.html"), "utf-8"))
-			.replace("##SPA_APP_ENV##", envVars.MICROS_ENVTYPE || "")
-			.replace("##SENTRY_SPA_DSN##", envVars.SENTRY_SPA_DSN || "")
-			.replace("\"##FRONTEND_FEATURE_FLAGS##\"", JSON.stringify(featureFlags));
+		indexHtmlContent = await fs.readFile(path.join(process.cwd(), "spa/build/index.html"), "utf-8");
 	}
+
+	indexHtmlContent = indexHtmlContent
```
gxueatlassian, line 29: This will override the global `indexHtmlContent`. I'd suggest use a local variable.

krazziekay: Yep we are overriding these values every time when the request is called, so that we get the latest value.

gxueatlassian (changes requested):
Since the FF is always ON now, and I believe there might be a concurrency override issue here. Let's chat if you still need this PR approved.

gxueatlassian (approved):

PR #2495, opened 2023-10-11
Fix localstorage permission problem by remove it.

gxueatlassian:
**What's in this PR?**
1. Remove localStorage usage for popup detection.
2. Add a "dismiss" link ( so that we can rest the error )

image

**Why**
1. Some device has permission problem accessing the local storage, we don't want that permission stop the whole app.

**Added feature flags**
N/A

**Affected issues**  
N/A

**How has this been tested?**  
Local, Unit

**Whats Next?**
N/A

### DIFF
--- a/spa/src/pages/InstallationRequested/index.tsx
+++ b/spa/src/pages/InstallationRequested/index.tsx
@@ -42,9 +42,7 @@ const InstallationRequested = () => {
 			<LoggedinInfo
 				username={username}
 				logout={navigateBackToSteps}
-				onPopupBlocked={() => {
-					//TODO: show proper UI
-				}}
+				onPopupBlocked={() => { /* do nothing as ui flow will redirect to entry page anyway */ }}
 			/>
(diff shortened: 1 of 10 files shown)


spa/src/pages/InstallationRequested/index.tsx
```diff
@@ -42,6 +42,9 @@ const InstallationRequested = () => {
 			<LoggedinInfo
 				username={username}
 				logout={navigateBackToSteps}
+				onPopupBlocked={() => {
+					//TODO: show proper UI
```
kamaksheeAtl, line 46: is this suppose to be todo?
should we not show popup block alert of this page ?

gxueatlassian: Hi I've changed the comments and leave this out. When popup is blocked, the ui will still be redirected to home page, so actually we don't need to do anything.

kamaksheeAtl (approved):

PR #2502, opened 2023-10-16
None splice bug

joshkay10:
**What's in this PR?**
simplfying the code so we can handle trash better.

The little tweaks werent working so have done a more proper clean... hopefully!

### DIFF
--- a/src/transforms/push.ts
+++ b/src/transforms/push.ts
@@ -186,7 +186,7 @@ export const processPush = async (github: GitHubInstallationClient, payload: Pus
 		// break the array up into chunks of 400
 		const chunks: JiraCommit[][] = [];
 
-		while (commits.length) {
+		while (commits?.length) {
 			chunks.push(commits.splice(0, 400));
 		}
 
--- a/src/transforms/push.ts
+++ b/src/transforms/push.ts
@@ -138,54 +138,60 @@ export const processPush = async (github: GitHubInstallationClient, payload: Pus
 		}
 
 		const recentShas = shas.slice(0, MAX_COMMIT_HISTORY);
-		const commits: JiraCommit[] = await Promise.all(
-			recentShas.map(async (sha): Promise<JiraCommit> => {
-				log.info("Calling GitHub to fetch commit info " + sha.id);
-				try {
-					const commitResponse = await github.getCommit(owner.login, repo, sha.id);
-					const {
-						files,
-						author: author = {},
-						parents,
-						sha: commitSha,
-						html_url
-					} = commitResponse.data;
 
-					const githubCommitAuthor = commitResponse.data.commit?.author;
-					const message = commitResponse.data.commit?.message;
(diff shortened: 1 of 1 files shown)


src/transforms/push.ts
```diff
@@ -138,54 +138,59 @@ export const processPush = async (github: GitHubInstallationClient, payload: Pus
...
+					files: compact(filesToSend.map((file) => mapFile(file, repo, sha.id, owner.name))),
+					id: commitSha,
+					issueKeys: sha.issueKeys,
+					url: html_url,
+					updateSequenceId: Date.now(),
+					flags: isMergeCommit ? ["MERGE_COMMIT"] : undefined
+				};
+			} catch (err: unknown) {
+				log.warn({ err }, "Failed to fetch data from GitHub");
+				throw err;
```
atraversatlassian, line 183: You haven't changed the existing behaviour but I do have a concern that this throw will be collected with all other promises.  This would mean if we get one transient error from github we would reject the entire batch and need to reprocess.  Is that the correct behaviour to have here or should we be more resilient and be able to collect those for retry separately?

atraversatlassian: This is probably scope creep on this ticket so if we think this is something to address it is probably best to create a new ticket for it and do it with a separate PR

atraversatlassian (approved):

src/transforms/push.ts
joshkay10: So yeah, for the most part its exactly as is, 
breaking it into 3 seperate actions instead of one big chain... to try simplify, but!!! there is one extra piece of logic

```
if (!commitSha) {
					return null;
}
```

Then we filter out the trash
`.filter(commit => commit !== null) as JiraCommit[];`

So if we get a unprocessable 200 result from GitHub we continue instead of doing exactly what you said, quitting for everything...

But your point still stands that we are still throwing a tantrum across the board for one failed event....
I was on the edge with this, 
we could replace within the catch something like.....
```
log.warn({ err }, "Failed to fetch data from GitHub");
return null
```
But I feel like there are enough scenarios when a singular failure should cause us to stop then try again later. At least for 5XXs. Things get a little less clear when we throw an error and bubble all the way up to a error handler to decide whether to try again or not, but there is absolutely chance of failure here if we get a 404 for 1 out of 5 commits we probably wouldn't retry the task...

haha after that massive story, I agree, we probably need to be slightly smarter when we handle these

joshkay10: Yeah, main change was splitting it up so its easy to read since I added a .filter as well

also added returning null for unprocessable data from github to filter out....


I left the error handling alone at the moment because id agree its a bit bigger scope, we could probably be smarter in the catch and `return null` instead of `throw err` for ones that we are ok with skipping / never expect a success response from, its all a bit of a tangled web, we propagate the errors up to some should/should-not retry logic but we treat the whole task for the same error so definite hole I think 🤔

PR #2507, opened 2023-10-20
ARC 2495 [BE] - Deferred Installation 

krazziekay:
**What's in this PR?**
- Added new endpoints for the deferred installation flow.
- Added test cases.
- Updated `atlassian-connect.json` for the new deferred route in `spa`.

**Why**
- To make things super easy for the customers

**How has this been tested?**  
- Test cases
- Tested in staging - https://www.loom.com/share/5c8da6cd93e84beba69bc13c83ad6da9

**Whats Next?**
- Merge with the FE and test.


src/rest/routes/deferred/deferred-request-parse.ts
```diff
@@ -0,0 +1,19 @@
...
+import { InvalidArgumentError } from "config/errors";
+
+export const DeferredRequestParseRoute = errorWrapper("ParseRequestId", async function DeferredRequestParseRoute(req: Request, res: Response<SubscriptionDeferredInstallPayload>) {
+	const requestId = req.params.requestId;
+	if (!requestId) {
+		req.log.warn("Missing requestId in query");
+		throw new InvalidArgumentError("Missing requestId in query");
+	}
+
+	const deferredInstallPayload = await extractSubscriptionDeferredInstallPayload(requestId);
```
gxueatlassian, line 16: Please cross check that jiraHost from extracted match the one from JwtHandler. Otherwise it could be a security leak that anyone with a requestId (supposed to be public over the network in url) can request information from other jira host.

src/rest/routes/github-callback/index.ts
```diff
@@ -49,8 +49,10 @@ const getPostMessageScript = function(opts: Record<string, unknown>) {
 	return `<html>
 			<body></body>
 			<script>
+			if (window.opener) {
```
gxueatlassian, line 52: Doesn't it always have that `window.opener`?
And in your change, what if it doesn't have `window.opener`, what UI error will bring user to?

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
```diff
@@ -218,7 +218,17 @@ const	modules = {
 			url: "/spa?from={ac.from}",
 			location: "none",
 			conditions: adminCondition
-		}, {
+		},
+		{
+			key: "spa-deferred-page",
```
gxueatlassian, line 223: I don't think we need this one, as the defer page will NOT be rendered in iframe at all from Jira.

src/rest/routes/deferred/deferred-check-ownership-and-connect.ts
```diff
@@ -0,0 +1,42 @@
...
+		githubToken as string,
+		installation,
+		undefined,// TODO: Need to pass this value later for GHE apps
+		gitHubInstallationId,
+		true,
+		req.log
+	);
+	if (result.errorCode === "NOT_ADMIN") {
+		throw new InsufficientPermissionError(result.error || "Not admin of org");
+	} else {
```
joshkay10, line 38: /nit - dont need the else here, the throw will return if prcessess so can simplify a smidge..

```
if (result.errorCode === "NOT_ADMIN") {
		throw new InsufficientPermissionError(result.error || "Not admin of org");
}
res.sendStatus(200);
```

src/rest/routes/deferred/deferred-check-ownership-and-connect.ts
```diff
@@ -0,0 +1,42 @@
...
+	if (!jiraHost) {
+		req.log.warn("Missing jiraHost from the requestId");
+		throw new InvalidArgumentError("Missing jiraHost from the requestId");
+	}
+	const installation = await Installation.findByPk(installationIdPk);
+	if (!installation) {
+		req.log.warn("Installation not found for this requestId");
+		throw new InvalidArgumentError("Installation not found for this requestId");
+	}
+	const result = await verifyAdminPermsAndFinishInstallation(
```
joshkay10, line 28: /nit slightly more descriptive variable name would be a + instead of just result

src/rest/routes/deferred/deferred-check-ownership-and-connect.ts
```diff
@@ -0,0 +1,42 @@
...
+		githubToken as string,
+		installation,
+		undefined,// TODO: Need to pass this value later for GHE apps
+		gitHubInstallationId,
+		true,
+		req.log
+	);
+	if (result.errorCode === "NOT_ADMIN") {
+		throw new InsufficientPermissionError(result.error || "Not admin of org");
```
joshkay10, line 37: include a req.log.warn for this sucker too..

on a side note.. does InsufficientPermissionError log out the message we pass it? then we can drop the additional warn logs each error since its the same msg?

src/rest/routes/deferred/deferred-installation-url.ts
```diff
@@ -0,0 +1,40 @@
...
+	if (!gitHubInstallationId) {
+		req.log.warn("Missing gitHubInstallationId in query");
+		throw new InvalidArgumentError("Missing gitHubInstallationId in query");
+	}
+
+	if (!gitHubOrgName) {
+		req.log.warn("Missing gitHubOrgName in query");
+		throw new InvalidArgumentError("Missing gitHubOrgName in query");
+	}
+
```
joshkay10, line 25: is it possible for res.locals.installation to not be available at this point?

src/rest/routes/deferred/deferred-installation-url.ts
```diff
@@ -0,0 +1,40 @@
...
+	const payload: SubscriptionDeferredInstallPayload = {
+		installationIdPk: installation.id,
+		jiraHost: installation.jiraHost,
+		gitHubInstallationId: parseInt(gitHubInstallationId.toString()),
+		orgName: gitHubOrgName.toString(),
+		gitHubServerAppIdPk: undefined // TODO: This only works for cloud, Add this value for GHE servers
+	};
+	const requestId = await registerSubscriptionDeferredInstallPayloadRequest(payload);
```
joshkay10, line 33: Is it worth wrapping up the payload gen and the async call in a try/catch?

krazziekay: We can, but don't think that'll be necessary. 
The only time that this would fail is if redis is not running. And redis is running in docker, so we would already know if it fails from the docker. 
Plus, I haven't seen any try catch being used in the other places where this method is being used.

src/rest/routes/deferred/deferred-installation-url.ts
krazziekay: Nop, `res.locals.installation` should always be here because this route passes through `JwtHandler` middleware, which parses the JWT token and saves the `installation` to `res.locals`.

src/rest/routes/deferred/deferred-check-ownership-and-connect.ts
krazziekay: `Req.log` added.
The frontend does catch the `INSUFFICIENT_PERMISSION` error and shows the corresponding error banner.

src/rest/routes/deferred/deferred-check-ownership-and-connect.ts
krazziekay: Updated the variable name.

src/rest/routes/deferred/deferred-check-ownership-and-connect.ts
krazziekay: Else removed

src/routes/jira/atlassian-connect/jira-atlassian-connect-get.ts
krazziekay: Removed

src/rest/routes/github-callback/index.ts
krazziekay: When we are requesting to install an org from Github, then the admin receives an email with the link from Github. When dealing with that link, then at that time this `window.opener` is undefined, which causes the response screen to be a JS error screen, but the installation is done. That's why I added this check.
We do not have any UI for showing the error at the moment, we can do that in another task. The objective here is to not show any JS error.

src/rest/routes/deferred/deferred-request-parse.ts
krazziekay: Check added

src/rest/routes/deferred/deferred-request-parse.ts
```diff
@@ -0,0 +1,24 @@
...
+
+	if (deferredInstallPayload.jiraHost !== res.locals.jiraHost) {
+		req.log.warn("Jirahost mismatch");
+		throw new RestApiError(500, "JIRAHOST_MISMATCH", "Jirahost mismatch");
+	}
+
+	res.status(200).send(deferredInstallPayload);
```
gxueatlassian, line 23: send the info only we need

src/rest/routes/deferred/deferred-request-parse.ts
krazziekay: This check will not be required as this path does not go through JwtHandler middleware. This change is done in the FE PR.

src/rest/routes/deferred/deferred-request-parse.ts
krazziekay: Updated

src/rest/routes/deferred/index.ts
```diff
@@ -0,0 +1,14 @@
...
+import { DeferredRequestParseRoute } from "./deferred-request-parse";
+import { DeferredCheckOwnershipAndConnectRoute } from "./deferred-check-ownership-and-connect";
+import { JwtHandler } from "../../middleware/jwt/jwt-handler";
+import { GitHubTokenHandler } from "~/src/rest/middleware/jwt/github-token";
+
+export const DeferredRouter = Router({ mergeParams: true });
+
+DeferredRouter.get("/installation-url", JwtHandler, DeferredInstallationUrlRoute);
+
+DeferredRouter.get("/parse/:requestId", JwtHandler, DeferredRequestParseRoute);
```
gxueatlassian, line 12: Should we remove JwtHandler?

krazziekay: This has been removed in the [FE PR](https://github.com/atlassian/github-for-jira/pull/2508).

krazziekay: I've just merged in the FE pr to this one, so this middleware has now been removed

src/rest/routes/deferred/deferred-check-ownership-and-connect.ts
```diff
@@ -0,0 +1,43 @@
...
+		undefined,// TODO: Need to pass this value later for GHE apps
+		gitHubInstallationId,
+		true,
+		req.log
+	);
+	if (isAdminResponse.errorCode === "NOT_ADMIN") {
+		req.log.warn("User is not an admin of the org");
+		throw new InsufficientPermissionError(isAdminResponse.error || "Not admin of org");
+	} else {
+		res.sendStatus(200);
```
gxueatlassian, line 40: if `isAdminResponse.errorCode === "NOT_ADMIN"`, it may still be some other errors. Maybe add a catch all error here.

gxueatlassian (approved):

src/rest/routes/deferred/deferred-check-ownership-and-connect.ts
krazziekay: Updated the checks.

gxueatlassian (approved):

gxueatlassian (approved):

PR #2508, opened 2023-10-20
ARC 2495 [FE] - Deferred Installation

krazziekay:
**What's in this PR?**
- New page for the deferred installation flow.
- Added new endpoints to request deferred requests.

**Why**
- To make things super easy for the customers

**How has this been tested?**  
- Local
- Test in staging - https://www.loom.com/share/5c8da6cd93e84beba69bc13c83ad6da9

**Whats Next?**
- Merge with the BE and test.


spa/src/components/Error/KnownErrors/index.tsx
```diff
@@ -39,14 +59,103 @@ export const ErrorForSSO = ({ orgName, accessUrl, resetCallback, onPopupBlocked
...
+			You’re not an owner for this organization. To connect:
+			<ol css={olStyle}>
+				<li>
+					<a css={linkStyle} onClick={getOrgOwnerUrl}>
+						Find an organization owner.
+					</a>
+				</li>
+				<li>
+					<a css={linkStyle} onClick={getDeferredInstallationUrl}>
+						Send them a link and ask them to connect.
```
gxueatlassian, line 115: Can you double check with design? I feel the word "Send" is a bit too strong, user might hestigate to click it.
A "lighter" word will be like "Generate" or "Prepare"

spa/src/pages/DeferredInstallation/index.tsx
```diff
@@ -0,0 +1,211 @@
...
+	}, [ isLoggedIn ]);
+
+	// Authenticate if no token/username is set
+	const authenticate = async () => {
+		setIsLoading(true);
+		try {
+			await OAuthManager.authenticateInGitHub({
+				onWinClosed: () => {
+					setIsLoading(false);
+					console.log("Successfully authenticated", username);
```
gxueatlassian, line 136: Shouldn't we call `connectDeferredOrgOrg()` here? So that we don't have to do another `useEffect`?

spa/src/services/deferral-manager/index.ts
```diff
@@ -0,0 +1,46 @@
...
+		if(!ret) {
+			reportError(
+				{ message: "Response status for parsing deferred request id is not 200", status: response.status },
+				{ path: "parseDeferredRequestId" }
+			);
+		}
+
+		return response.data;
+	} catch (e: unknown) {
+		console.log("What is hts");
```
gxueatlassian, line 18: Maybe more clear msg.

spa/src/pages/DeferredInstallation/index.tsx
krazziekay: The `connectDeferredOrg` should happen after the whole OAuth flow is finished, which happens on the other useEffect with `OAuthManager.finishOAuthFlow`.
And I've kept it in separate useEffect for better readability and avoiding massive code in a single useEffect.

spa/src/services/deferral-manager/index.ts
krazziekay: Removed.

gxueatlassian (approved):

PR #2509, opened 2023-10-23
Node js Upgrade to 18

kamaksheeAtl:
**What's in this PR?**

**Why**

**Added feature flags**

**Affected issues**  
_Jira Issues_

**How has this been tested?**  

**Whats Next?**


src/transforms/transform-code-scanning-alert.test.ts
```diff
@@ -182,41 +181,40 @@ describe("code_scanning_alert transform", () => {
 				gitHubInstallationId,
 				jiraHost
 			);
-			expect(result?.vulnerabilities[0]).toMatchObject(
-				{
-					"additionalInfo": {
-						"content": "CodeQL"
+			expect(result?.vulnerabilities[0]).toMatchObject({
```
gxueatlassian, line 184: /nit
What's changed in this file? Can you make it dosen't shown that much change here?

src/util/workers-health-monitor.test.ts
```diff
@@ -159,26 +159,26 @@ describe("workers-health-monitor", () => {
 			}));
 
 			const workerSendingImAliveToMasterInterval = setInterval(() => {
-				if ((cluster.workers[1]!.on as jest.Mock).mock.calls.length > 0) {
-					(cluster.workers[1]!.on as jest.Mock).mock.calls[0][1]();
-					(cluster.workers[2]!.on as jest.Mock).mock.calls[0][1]();
+				if ((cluster.workers?.[1]?.on as jest.Mock).mock.calls.length > 0) {
```
gxueatlassian, line 162: Why is this?

gxueatlassian: I mean why use `?` instead of the `!`, as the `!` still make it a "valid" test

krazziekay: Using `!` was throwing in typescript errors, cause `cluster.workers` can be `undefined`.

src/transforms/transform-code-scanning-alert.test.ts
krazziekay: The only change is the removal of the double quotes for the keys within the object.

src/util/workers-health-monitor.test.ts
gxueatlassian: I'd suggest using `!` when possible, because for test code, we want a more strict behaviour, like we always want to test to fail if the value is `undefined`. Using `?`, while okay in these couple lines, may leave the test in a less strict behaviour in some other cases. 

I've tried the following should remove the complained from ts

```
                               if ((cluster.workers![1]!.on as jest.Mock).mock.calls.length > 0) {
                                       (cluster.workers![1]!.on as jest.Mock).mock.calls[0][1]();
                                       (cluster.workers![2]!.on as jest.Mock).mock.calls[0][1]();
                                }
```

yvonnevictoria (approved):

yvonnevictoria:
Happy to approve once the comments @gxueatlassian wrote about the tests are addressed. 

In future, please fill in the description template above :) It's always helpful to have some context on why a change is being made, and the major changes that have occurred. It'd also be good if you added a descriptive branch name e.g. `UA-9615-node-18-upgrade`.

PR #2513, opened 2023-10-24
remove-stale-msg-flag

joshkay10:
**What's in this PR?**
removing the flag... and the code!!!!!!!!!!!!!!!!!!!!!!!!!!

This isn't as extreme as is appears, the flag has been returning false for months, it was useful to clean up the deployment queue when we had troubles, but now we are actively fixing things in the queue so we wont need it.

MOSTLY... i don't want code left in there thats going to be skipped indefinitely.


yvonnevictoria (approved):

gxueatlassian (approved):

yvonnevictoria:
@joshkay10 Do we need to close this PR?

PR #2549, opened 2023-11-13
NONE: eslint re-enable rules

atraversatlassian:
**What's in this PR?**
Re-enable eslint rules

**Why**
More lint

**Added feature flags**
None

**Affected issues**  
NONE

**How has this been tested?**  
Tests

**Whats Next?**
More lint

### DIFF
--- a/src/models/sentry-scope-proxy.ts
+++ b/src/models/sentry-scope-proxy.ts
@@ -1,4 +1,5 @@
 /* eslint-disable @typescript-eslint/no-explicit-any */
+import { Event, EventHint } from "@sentry/types";
 
 /*
  * Acts like a Sentry scope. See https://docs.sentry.io/enriching-error-data/scopes/?platform=node#configuring-the-scope
@@ -8,6 +9,11 @@
  *
  * Learn more at https://docs.sentry.io/platforms/node/#eventprocessors
  */
+
+export interface SentryScopeProxyEvent extends Event {
+	extra?: Record<string, any>;
+}
+
 export class SentryScopeProxy {
 	event: any;
 	hint: any;
@@ -18,9 +24,9 @@ export class SentryScopeProxy {
 		this.extra = {};
 	}
 
-	static processEvent(this: void, event: any, hint: any) {
-		if (hint.originalException.sentryScope) {
-			hint.originalException.sentryScope.addTo(event);
+	static processEvent(event: Event, hint?: EventHint | undefined): SentryScopeProxyEvent {
+		if (hint?.originalException && hint?.originalException["sentryScope"]) {
+			hint?.originalException["sentryScope"].addTo(event);
 		}
 
 		return event;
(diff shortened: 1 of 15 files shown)


src/util/webhook-utils.ts
```diff
@@ -71,6 +71,7 @@ export const emitWebhookProcessedMetrics = (
 			{ err },
 			"Failed to send webhook processing time metrics."
 		);
+		return undefined;
```
joshkay10, line 74: /nit - it implicitly returns undefined anyway so dont need to add it?

src/routes/api/api-reset-subscription-failed-tasks.ts
```diff
@@ -5,7 +5,7 @@ import { TaskType } from "../../sync/sync.types";
 
 interface OutputLogRecord {
 	repoSyncStateId: number,
-	targetTask: "pull" | "branch" | "commit" | "build" | "deployment" | "dependabotAlert" | string
+	targetTask:  string // "pull" | "branch" | "commit" | "build" | "deployment" | "dependabotAlert"
```
joshkay10, line 8: There is a taskTypes type in `sync.types.ts` that 'should' match this.... but with all the security and code scanning changes it might be out of sync :ohnoes:

src/models/sentry-scope-proxy.ts
```diff
@@ -18,9 +24,9 @@ export class SentryScopeProxy {
 		this.extra = {};
 	}
 
-	static processEvent(this: void, event: any, hint: any) {
-		if (hint.originalException.sentryScope) {
-			hint.originalException.sentryScope.addTo(event);
+	static processEvent(event: Event, hint?: EventHint | undefined): SentryScopeProxyEvent {
```
joshkay10, line 27: /nit i feel like ? or | undefined is enough, is there any reason to include both other than being super explicit?

joshkay10 (approved):

yvonnevictoria (approved):

src/routes/api/api-reset-subscription-failed-tasks.ts
atraversatlassian: Yeah maybe this should get updated to keep the stronger typing.  Since it has string in the list it is the only typing used as it overwrites everything else.  I would be happy for us to go to the strong typing but this was function as a comment right now so I just made it into one

src/util/webhook-utils.ts
atraversatlassian: I am indifferent between return; and return undefined;  I opted for the more explicit....maybe there should be a lint rule for it ;)

src/models/sentry-scope-proxy.ts
atraversatlassian: Ah I should have cleaned this up, I will get it with my next PR.  Personally I prefer the ? on the param name than undefined.  I don't have a strong preference

PR #2550, opened 2023-11-13
NONE: re-enable eslint rules

atraversatlassian:
**What's in this PR?**
Enable eslint rules

**Why**
lint

**Added feature flags**
None

**Affected issues**  
NONE

**How has this been tested?**  
Tests

**Whats Next?**
More lint

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -145,8 +145,7 @@
 			"@typescript-eslint/no-unnecessary-condition": "off",
 			"@typescript-eslint/no-explicit-any": "off",
 			"@typescript-eslint/prefer-ts-expect-error": "off",
-			"@typescript-eslint/no-unnecessary-type-arguments": "off",
-			"@typescript-eslint/no-non-null-assertion": "off"
+			"@typescript-eslint/no-unnecessary-type-arguments": "off"
 		}
 	}
 ]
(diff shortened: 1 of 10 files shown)


.eslintrc.json
```diff
@@ -145,8 +145,7 @@
 			"@typescript-eslint/no-unnecessary-condition": "off",
 			"@typescript-eslint/no-explicit-any": "off",
 			"@typescript-eslint/prefer-ts-expect-error": "off",
-			"@typescript-eslint/no-unnecessary-type-arguments": "off",
-			"@typescript-eslint/no-non-null-assertion": "off"
```
gxueatlassian, line 149: I am mind blown by trying to understand what (removing this line) this mean now?

Does it mean `"@typescript-eslint/no-non-null-assertion": "on"` (by removing the line) and --> means in code you can NOT have `non-null` assertion? --> means you can't do `this.sequelize!` --> which is contracditory to bellow code ???>

gxueatlassian: Okay, so the ts-lint now allow
`await this.sequelize!.query`
but forbid
`const installation: Installation = (await Installation.findByPk(gitHubServerApp.installationId))!`

hmm, still don't get it.

atraversatlassian: It doesn't allow either.  For the first line you mentioned if you look I have left a comment plus an explicit eslint override for it.  So both break the rule because it is now "on"

gxueatlassian (approved):

PR #2556, opened 2023-11-16
NONE: Re-enable eslint rules

atraversatlassian:
**What's in this PR?**
Lint

**Why**
Lint

**Added feature flags**
None

**Affected issues**  
NONE

**How has this been tested?**  
Tests

**Whats Next?**
Lint

### DIFF
--- a/src/middleware/github-server-app-middleware.ts
+++ b/src/middleware/github-server-app-middleware.ts
@@ -5,7 +5,7 @@ import { envVars } from "config/env";
 import { GITHUB_CLOUD_BASEURL } from "~/src/github/client/github-client-constants";
 import { BaseLocals } from "../rest/routes";
 
-export const GithubServerAppMiddleware = async (req: Request, res: Response<any, BaseLocals>, next: NextFunction): Promise<void> => {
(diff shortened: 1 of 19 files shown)


src/middleware/github-server-app-middleware.ts
```diff
@@ -40,7 +40,7 @@ export const GithubServerAppMiddleware = async (req: Request, res: Response<any,
 		res.locals.gitHubAppId = gitHubServerApp.id;
 		res.locals.gitHubAppConfig = {
 			gitHubAppId: gitHubServerApp.id,
-			appId: gitHubServerApp.appId,
+			appId: String(gitHubServerApp.appId),
```
bgvozdev, line 43: Just wondering about the justification of this change :)

bgvozdev (approved):

src/middleware/github-server-app-middleware.ts
atraversatlassian: The parent object is now more strongly typed.  In other usage it is used as a String this makes this invocation consistent with that typing expectation

PR #2566, opened 2023-11-24
Added routes for adding/removing a dummy subscription

krazziekay:
**What's in this PR?**
- Updated api-configuration-post to set the configured state.

**Why**
- In pollinator tests, when logging in to GitHub, it has a code verification step or a M2FA step, which becomes a blocker in testing adding/removing subscriptions.
- So this endpoint is an alternative to the original setting the configured state without having any token check and logs.

**How has this been tested?**  
- Test cases

**Whats Next?**
- Merge this and use these endpoints in the pollinator tests.


src/routes/api/subscription/api-dummy-subscription-post.ts
```diff
@@ -0,0 +1,34 @@
...
+ * @param req
+ * @param res
+ * @constructor
+ */
+export const ApiDummySubscriptionPost = async (req, res) => {
+	const jiraHost = req.params.jiraHost as string;
+
+	try {
+		// Create a dummy Subscription
+		await Subscription.install({
```
gxueatlassian, line 21: Actually, do we even need this line? Maybe what we need is just to use api to `await saveConfiguredAppProperties(jiraHost, req.log, true);`

krazziekay: True the main thing here is the `saveConfiguredAppProperties`. 
So, I've updated the `ApiConfigurationPost` route.

gxueatlassian (approved):

PR #2579, opened 2023-11-28
ARC-2643 allow multi workflow entries in audit log during backfill

gxueatlassian:
**What's in this PR?**
Allow multi builds in audit log during backfill
Some typings.

**Why**
Polli test needs it

**Added feature flags**

**Affected issues**  
[arc-2643]

**How has this been tested?**  
Unit test

**Whats Next?**

### DIFF
--- a/src/github/branch.ts
+++ b/src/github/branch.ts
@@ -74,7 +74,7 @@ export const processBranch = async (
 		operationType: "NORMAL",
 		auditLogsource: "WEBHOOK",
 		entityAction: "BRANCH_CREATE",
-		subscriptionId: subscription?.id
+		subscriptionId: subscription?.id || 0
 	});
 
 	emitWebhookProcessedMetrics(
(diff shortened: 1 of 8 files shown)


src/github/branch.ts
```diff
@@ -74,7 +74,7 @@ export const processBranch = async (
 		operationType: "NORMAL",
 		auditLogsource: "WEBHOOK",
 		entityAction: "BRANCH_CREATE",
-		subscriptionId: subscription?.id
+		subscriptionId: subscription?.id || 0
```
gxueatlassian, line 77: This shouldn't happen, but we didn't `return` above if the `subscription` is empty, so just allow safe.

src/jira/client/jira-client.ts
```diff
@@ -411,9 +412,9 @@ export const getJiraClient = async (
 					status: response.status,
 					data:response.data
 				};
-				const reqBuildData = data?.builds[0];
```
gxueatlassian, line 414: This is where some builds are missing during backfill.

kamaksheeAtl (approved):

PR #2582, opened 2023-11-29
NONE: Reenable lint rules

atraversatlassian:
**What's in this PR?**
Lint

**Why**
Lint

**Added feature flags**
Nope

**Affected issues**  
NONE

**How has this been tested?**  
Tests

**Whats Next?**
Lint

### DIFF
--- a/src/routes/github/create-branch/github-create-branch-options-get.ts
+++ b/src/routes/github/create-branch/github-create-branch-options-get.ts
@@ -13,7 +13,7 @@ export const GithubCreateBranchOptionsGet = async (req: Request, res: Response, 
 	const { issueKey } = req.query;
 
 	if (!issueKey) {
-		return next(new Error(Errors.MISSING_ISSUE_KEY));
+		next(new Error(Errors.MISSING_ISSUE_KEY)); return;
 	}
 
 	const jiraHost: string = res.locals.jiraHost;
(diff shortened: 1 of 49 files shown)


src/routes/github/create-branch/github-create-branch-options-get.ts
```diff
@@ -13,7 +13,7 @@ export const GithubCreateBranchOptionsGet = async (req: Request, res: Response,
 	const { issueKey } = req.query;
 
 	if (!issueKey) {
-		return next(new Error(Errors.MISSING_ISSUE_KEY));
+		next(new Error(Errors.MISSING_ISSUE_KEY)); return;
```
bgvozdev, line 16: /nit my personal preference is to keep different statements on different lines

src/routes/jira/connect/enterprise/jira-connect-enterprise-post.ts
```diff
@@ -39,7 +39,7 @@ const sendErrorMetricAndAnalytics = (jiraHost: string, errorCode: ErrorResponseC
 	}, {
 		jiraHost,
 		...errorCodeAndStatusObj
-	});
+	}).catch(() => { /* ignore */ });
```
bgvozdev, line 42: Curious why are we are swallowing a error here?

atraversatlassian: I believe we were previously swallowing this error because the promise was never awaited.  This is a little more explicit about it.  I should probably have logging....but the underlying functions are already logging on throws and silently passing up undefined.

bgvozdev (approved):

PR #2587, opened 2023-11-30
NONE: Re-enable lint rules

atraversatlassian:
**What's in this PR?**
Lint

**Why**
Lint

**Added feature flags**
Nope

**Affected issues**  
NONE

**How has this been tested?**  
tests

**Whats Next?**
lint

### DIFF
--- a/src/config/env.ts
+++ b/src/config/env.ts
@@ -23,7 +23,7 @@ type Transforms<T, K extends keyof T = keyof T> = {
 };
 
 const transforms: Transforms<EnvVars> = {
-	MICROS_ENV: (value?: string) =>  EnvironmentEnum[value || ""] as EnvironmentEnum || EnvironmentEnum.development,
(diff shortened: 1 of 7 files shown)


src/config/env.ts
```diff
@@ -23,7 +23,7 @@ type Transforms<T, K extends keyof T = keyof T> = {
 };
 
 const transforms: Transforms<EnvVars> = {
-	MICROS_ENV: (value?: string) =>  EnvironmentEnum[value || ""] as EnvironmentEnum || EnvironmentEnum.development,
+	MICROS_ENV: (value?: string) =>  EnvironmentEnum[value || ""] as EnvironmentEnum,
```
bgvozdev, line 26: :susp: Curious why we are removing the fallback value?

atraversatlassian: ah good catch I think this is bad resolution because of the casting.  I will fix it

PR #2594, opened 2023-12-04
NONE: Re-enable more lint rules

atraversatlassian:
**What's in this PR?**
lint

**Why**
lint

**Added feature flags**
Nope

**Affected issues**  
NONE

**How has this been tested?**  
tests

**Whats Next?**
lint

### DIFF
--- a/src/sqs/sqs.ts
+++ b/src/sqs/sqs.ts
@@ -48,7 +48,7 @@ export class SqsQueue<MessagePayload extends BaseMessagePayload> {
 	/**
 	 * Context of the currently active listener, or the last active if the queue stopped
 	 */
-	listenerContext: SQSContext;
+	listenerContext: SQSContext | undefined;
 
 	public constructor(settings: QueueSettings, messageHandler: MessageHandler<MessagePayload>, errorHandler: ErrorHandler<MessagePayload>) {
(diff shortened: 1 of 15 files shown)


src/sqs/sqs.ts
```diff
@@ -306,10 +306,10 @@ export class SqsQueue<MessagePayload extends BaseMessagePayload> {
...
 				executionId: uuidv4(),
 				queue: this.queueName,
-				jiraHost: payload?.jiraHost,
-				installationId: payload?.installationId,
-				gitHubAppId: payload?.gitHubAppConfig?.gitHubAppId,
-				webhookId: payload?.webhookId,
+				jiraHost: payload.jiraHost,
+				installationId: payload.installationId,
+				gitHubAppId: payload.gitHubAppConfig?.gitHubAppId,
+				webhookId: payload.webhookId,
```
bgvozdev, line 312: Are we sure these are always there? :susp: Backfilling, for instance, doesn't have webhookId, and this file looks like some common one...

bgvozdev: Ah wait, probably that would be the same outcome anyway, which is "undefined"

atraversatlassian: That is right these properties could be undefined but payload will never be undefined here

PR #2601, opened 2023-12-06
chore: add new endpoint for backfill

kamaksheeAtl:
**What's in this PR?**

**Why**

**Added feature flags**

**Affected issues**  
_Jira Issues_

**How has this been tested?**  

**Whats Next?**

### DIFF
--- a/spa/src/api/subscriptions/index.ts
+++ b/spa/src/api/subscriptions/index.ts
@@ -1,5 +1,12 @@
 import { axiosRest } from "../axiosInstance";
 
 export default {
-	getSubscriptions: () => axiosRest.get("/rest/subscriptions")
+	getSubscriptions: () => axiosRest.get("/rest/subscriptions"),
+	syncSubscriptions: (data: {
+		installationId: number;
+		commitsFromDate: string;
+		appId?: number;
+		source: string;
+		syncType: string;
+	}) => axiosRest.post("/rest/sync", data),
 };
--- a/src/rest/rest-router.ts
+++ b/src/rest/rest-router.ts
@@ -1,5 +1,6 @@
 import { Router } from "express";
 import { JwtHandler } from "./middleware/jwt/jwt-handler";
+import { body } from "express-validator";
 import { OAuthRouter } from "./routes/oauth";
 import { OAuthCallbackHandler, OrgsInstalledHandler, OrgsInstallRequestedHandler } from "./routes/github-callback";
 import { GitHubOrgsRouter } from "./routes/github-orgs";
@@ -10,6 +11,7 @@ import { RestErrorHandler } from "./middleware/error";
 import { JiraAdminEnforceMiddleware } from "./middleware/jira-admin/jira-admin-check";
 import { AnalyticsProxyHandler } from "./routes/analytics-proxy";
 import { SubscriptionsRouter } from "./routes/subscriptions";
+import { SyncRouter } from "./routes/sync";
 import { DeferredRouter } from "./routes/deferred";
 
 export const RestRouter = Router({ mergeParams: true });
@@ -20,6 +22,17 @@ const subRouter = Router({ mergeParams: true });
  * Separate route which returns the list of both cloud and server subscriptions
  */
 RestRouter.use("/subscriptions", JwtHandler, JiraAdminEnforceMiddleware, SubscriptionsRouter);
+
+/**
+ * Separate route for SPA to start backfill of both cloud and server subscriptions
+ */
+RestRouter.use(
+	"/sync",
+	body("commitsFromDate").optional().isISO8601(),
+	JwtHandler,
+	JiraAdminEnforceMiddleware,
+	SyncRouter
+);
 
 /**
  * For cloud flow, the path will be `/rest/app/cloud/XXX`,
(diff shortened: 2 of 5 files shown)


src/rest/routes/sync/index.ts
```diff
@@ -0,0 +1,92 @@
...
+			req.log,
+			syncType,
+			commitsFromDate || subscription.backfillSince,
+			targetTasks,
+			{ source }
+		);
+
+		res.sendStatus(202);
+	} catch (error: unknown) {
+		next(new Error("Unauthorized"));
```
krazziekay, line 88: Please use `RestApiError` for this with the appropriate status and error codes.

spa/src/api/subscriptions/index.ts
```diff
@@ -1,5 +1,7 @@
+import { RestSyncReqBody } from "~/src/rest-interfaces";
 import { axiosRest } from "../axiosInstance";
 
 export default {
-	getSubscriptions: () => axiosRest.get("/rest/subscriptions")
+	getSubscriptions: () => axiosRest.get("/rest/subscriptions"),
+	syncSubscriptions: (data: RestSyncReqBody) => axiosRest.post(`/rest/app/cloud/sync`, data),
```
gxueatlassian, line 6: It is similar to a comment I left for @krazziekay , I think we should follow the rest api semantic, so it would be `/rest/app/cloud/subscription/${subscriptionId}/sync`

gxueatlassian: Using the installationId is wrong (in existing page) in my opinion, coz each entity IS a subscription, so we should natuarally deal with subscriptionId. Otherwise, when it comes to GHE, we will need to combine "installationId" and "gheAppId" together to avoid conflicts.

src/rest/rest-router.ts
```diff
@@ -27,6 +29,14 @@ RestRouter.use("/subscriptions", JwtHandler, JiraAdminEnforceMiddleware, Subscri
  */
 RestRouter.use("/app/:cloudOrUUID", subRouter);
 
+subRouter.post(
```
gxueatlassian, line 32: We should move this bellow to be after the JWT handler, Admin middlware, etc.

src/rest/routes/sync/index.ts
```diff
@@ -0,0 +1,74 @@
...
+const restSyncPost = async (
+	req: Request<ParamsDictionary, unknown, RestSyncReqBody>,
+	res: Response<unknown, BaseLocals>
+) => {
+	//TODO: We are yet to handle enterprise backfill
+	const cloudOrUUID = req.params.cloudOrUUID;
+	const gheUUID = cloudOrUUID === "cloud" ? undefined : req.params.cloudOrUUID;
+	let gitHubAppId : number | undefined = undefined;
+	if (gheUUID) {
+		gitHubAppId = parseFloat(gheUUID);
```
krazziekay, line 21: `UUID` and `githubAppId` are 2 completely different things.
We have `UUID` in the frontend, which we use to identify the exact GHE app, so in order to get the `githubAppId`, you will get that from the table `GitHubServerApp`
` await GitHubServerApp.getForUuidAndInstallationId(UUID, res.locals.installation.id)`

src/rest/routes/sync/index.ts
```diff
@@ -0,0 +1,74 @@
...
+import { findOrStartSync } from "~/src/sync/sync-utils";
+import { determineSyncTypeAndTargetTasks } from "~/src/util/github-sync-helper";
+import { BaseLocals } from "..";
+import { RestApiError } from "~/src/config/errors";
+import { RestSyncReqBody } from "~/src/rest-interfaces";
+
+
+const restSyncPost = async (
+	req: Request<ParamsDictionary, unknown, RestSyncReqBody>,
+	res: Response<unknown, BaseLocals>
```
gxueatlassian, line 14: Shall we type the `unknown` --> `ResponseBody` as well?

src/rest/routes/sync/index.ts
```diff
@@ -0,0 +1,74 @@
...
+			req.log,
+			syncType,
+			commitsFromDate || subscription.backfillSince,
+			targetTasks,
+			{ source }
+		);
+
+		res.sendStatus(202);
+	} catch (error: unknown) {
+		throw new RestApiError(500, "UNKNOWN", "Something went wrong");
```
gxueatlassian, line 70: Let's not do try catch here. As now everything become 500. Should let the rest-router to figure out by the error type. Otherwise everything is 500 in this route.

src/rest/routes/sync/index.ts
kamaksheeAtl: Not sure what how it would look like

src/rest/routes/sync/index.ts
kamaksheeAtl: I moved it around main logic and not other errors are now out of try cach

src/rest/routes/sync/index.ts
gxueatlassian: just use void or string

gxueatlassian (dismissed):

src/routes/jira/sync/jira-sync-post.ts
krazziekay, line 1: The logic for `restSyncPost` and this endpoint is the same. Would be better to pull this common logic out into another method and reuse it in both places.
As for the analytics, can add an extra attribute `spa: true | false`.

gxueatlassian (approved):

PR #2602, opened 2023-12-06
[INNO] - New endpoint for deleting subscriptions

krazziekay:
**What's in this PR?**
- New endpoint for removing subscriptions.
- Only handles cloud subscriptions.
- Added an API for hitting this new endpoint.

**Why**
- Innovation Week.

**How has this been tested?**  
- Test cases


src/rest/rest-router.ts
```diff
@@ -50,4 +51,6 @@ subRouter.use(GitHubTokenHandler);
 
 subRouter.use("/org", GitHubOrgsRouter);
 
+subRouter.delete("/subscription/:installationId", SubscriptionDeleteRoute);
```
gxueatlassian, line 54: I'd use a `subRouter.use` to be consistent with others. The routes can be `/subscription/:subscriptionId`

src/rest/routes/subscriptions/delete.ts
```diff
@@ -0,0 +1,54 @@
...
+		installation.target_type,
+		req.log
+	)) {
+		throw new InsufficientPermissionError("Unauthorized to delete subscription");
+	}
+
+	try {
+		const subscription = await Subscription.getSingleInstallation(jiraHost, gitHubInstallationId, gitHubAppId);
+		if (!subscription) {
+			throw new RestApiError(500, "RESOURCE_NOT_FOUND", "Subscription not found");
```
gxueatlassian, line 46: Should be `404` to be consistent with rest api ?

src/rest/routes/subscriptions/delete.ts
```diff
@@ -0,0 +1,54 @@
...
+		throw new RestApiError(400, "RESOURCE_NOT_FOUND", "Jirahost not found");
+	}
+
+	const gitHubAppClient = await createAppClient(req.log, jiraHost, gitHubAppId, metrics);
+	const gitHubInstallationClient = await createInstallationClient(gitHubInstallationId, jiraHost, metrics, req.log, gitHubAppId);
+	const gitHubUserClient = await createUserClient(githubToken, jiraHost, metrics, req.log, gitHubAppId);
+	const { data: installation } = await gitHubAppClient.getInstallation(gitHubInstallationId);
+	const { data: { login } } = await gitHubUserClient.getUser();
+
+	if (!await isUserAdminOfOrganization(
```
gxueatlassian, line 32: So this is more a production question, do we need user to be GitHub admin to disconnect a subscription? If so, this is a bit contracdictory to the purpose of "defer install"

gxueatlassian:
Tbh, I'd feel more comfortable if you:
1. Extract the current logic of `jira-delete.ts` into a separate method in a separate service-class,
2. Call that method in this route.

So that we reuse some of the logic, (which are missing here), like the remove devinfo and send analytics events.

src/rest/routes/subscriptions/delete.ts
krazziekay: Removed this admin usage, I was using the wrong logic before. Have corrected it now.

src/rest/routes/subscriptions/delete.ts
krazziekay: Updated

src/rest/rest-router.ts
krazziekay: Updated

src/rest/rest-router.ts
```diff
@@ -46,6 +47,8 @@ subRouter.use("/installation", GitHubAppsRoute);
 
 subRouter.use("/jira/cloudid", JiraCloudIDRouter);
 
+subRouter.use("/subscription/:subscriptionId", SubscriptionDeleteRouter);
```
gxueatlassian, line 50:
```suggestion
subRouter.use("/subscriptions/:subscriptionId", SubscriptionRouter);
```

gxueatlassian: I think we should use a one-level higher router here, coz: later if we want to do other operation on that subscription, we can't
```
subRouter.use("/subscription/:subscriptionId", SubscriptionDeleteRouter);
subRouter.use("/subscription/:subscriptionId", SubscriptionOtherOpsRouter);
```
Coz the route conflcts.

src/util/jira-utils.ts
```diff
@@ -114,3 +127,73 @@ export const hasJiraIssueKey = (str: string): boolean => !isEmpty(jiraIssueKeyPa
...
+	} catch (err: unknown) {
+		logger.warn({ err }, "Failed to delete security workspace or vulnerabilities from Jira");
+	}
+};
+
+export const removeSubscription = async (
+	installation: Installation,
+	ghInstallationId: number,
+	gitHubAppId: number | undefined,
+	req: Request,
```
gxueatlassian, line 154: Please try not to take `req` and `res`.
A rule of thumb will be `require as less as you can` when writing function args.
For `req`, I guess we just need a `log`, and for `res`, I guess we don't need anything, the function can just return a `result: succes or not` (And don't return `202`, as this is HTTP concept, should not be related to non-http service/utils code

src/rest/routes/subscriptions/delete.ts
```diff
@@ -0,0 +1,22 @@
...
+/**
+ * This delete endpoint only handles Github cloud subscriptions
+ */
+SubscriptionDeleteRouter.delete("/", errorWrapper("SubscriptionDelete", async (req: Request, res: Response) => {
+	// TODO: Get the github app id for GHE subscriptions
+	const gitHubAppId = undefined;
+	const subscriptionId: number = Number(req.params.subscriptionId);
+	const { installation } = res.locals as { installation: Installation; };
+
+	// Passing 0 as the `ghInstallationId` cause its not gonna be used anyways
```
gxueatlassian, line 19: `Passing 0 as the `ghInstallationId` cause its not gonna be used anyways` <-- if so, why not take `undefined` for now?

Or better, write three functions
```javascript
export function removeSubscriptionById(installation, subscriptionId) {
  return doRemoveSubscription(sub: Subscription);
}
export function removeSubscriptionGHInstallationIDAndAppId(installation, ghInstId, ghAppId) {
  return doRemoveSubscription(sub: Subscription);
}
//__do_not_export__this
function doRemoveSubscription(sub: Subscription) {
  //do real stuff
}
```

src/rest/rest-router.ts
krazziekay: Re-structured the routes.

src/util/jira-utils.ts
krazziekay: - Removed the `res` and updated `req` to `req.log`
- Updated this method to return boolean value, which is then used to return the response.

src/rest/routes/subscriptions/delete.ts
krazziekay: Updated the type and passed `undefined`.

src/rest/routes/subscriptions/index.ts
```diff
@@ -0,0 +1,42 @@
...
+	const cloudOrUUID = req.params.cloudOrUUID;
+	if (!cloudOrUUID) {
+		throw new InvalidArgumentError("Invalid route, couldn't determine if its cloud or enterprise!");
+	}
+
+	// TODO: Check and add test cases for GHE later
+	const gitHubAppId = cloudOrUUID === "cloud" ? undefined :
+		(await GitHubServerApp.getForUuidAndInstallationId(cloudOrUUID, installation.id))?.appId; //TODO: validate the uuid regex
+
+	if (await removeSubscription(installation, undefined, gitHubAppId, req.log, subscriptionId)) {
```
gxueatlassian, line 37: I don't see when `removeSubscription` return `false`? 
Maybe we can just remove this `if else`?

krazziekay: Removed the if-else.

src/util/jira-utils.ts
```diff
@@ -166,6 +166,10 @@ export const removeSubscription = async (
 		throw new RestApiError(404, "RESOURCE_NOT_FOUND", "Can not find subscription");
 	}
 
+	if (subscription.jiraHost !== jiraHost) {
+		throw new RestApiError(500, "JIRAHOST_MISMATCH", "Jirahosts do not match for this subscription");
```
gxueatlassian, line 170: Well, it should be 403 or 401 anyway, cause it not the server's fault that client submit a wrong subscription Id. 
But /nit

gxueatlassian (approved):

src/util/jira-utils.ts
krazziekay: 403 and 401 might give the wrong idea here, coz they indicate that the user has no permission or not authenticated. I think 500 is better suited, so leaving it as it is for now.

PR #2610, opened 2023-12-12
reduce calls to jira issue status

gxueatlassian:
**What's in this PR?**
1. Reduce times to called to jira issue status
2. Add redis cache timeout ff

**Why**

**Added feature flags**
skip-process-queue-when-issue-not-exists-timeout

**Affected issues**  
None

**How has this been tested?**  
Unit test.

**Whats Next?**

### DIFF
--- a/src/transforms/push.ts
+++ b/src/transforms/push.ts
@@ -1,4 +1,5 @@
 import Logger from "bunyan";
+import { uniq } from "lodash";
 import { Subscription } from "models/subscription";
 import { getJiraClient, JiraClient } from "../jira/client/jira-client";
 import { JiraClientError } from "../jira/client/axios";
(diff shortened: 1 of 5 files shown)


src/transforms/push.ts
```diff
@@ -140,11 +141,14 @@ export const processPush = async (github: GitHubInstallationClient, payload: Pus
...
 
 		const recentShas = shas.slice(0, MAX_COMMIT_HISTORY);
+
+		const invalidIssueKeys = await tryGetInvalidIssueKeys(recentShas, subscription, jiraClient, log);
+
 		const commitPromises: Promise<JiraCommit | null>[] = recentShas.map(async (sha): Promise<JiraCommit | null> => {
 			try {
 
 				if (await booleanFlag(BooleanFlags.SKIP_PROCESS_QUEUE_IF_ISSUE_NOT_FOUND, jiraHost)) {
-					if (!await someIssueKeysExistsOnJira(subscription.jiraHost, sha.issueKeys, jiraClient, log)) {
```
gxueatlassian, line 147: The current problem is, this api calls are fired in parallel because we are using 
```
await Promise.all(recentShas.map(sha=>Promise()))
```

So even for same jira issue, there's no time for redis to cache the result.

joshkay10 (approved):

PR #2613, opened 2023-12-13
Arc 2745 fix exp time

gxueatlassian:
What's in this PR?
Add ff to reduce the app token exp time

Why
TEN MINUTES is the max allowed, prone to error

Added feature flags
app-token-exp-in-milli-sec

Affected issues
[ARC-2745]

How has this been tested?
Unit test

Whats Next?

[ARC-2745]: https://softwareteams.atlassian.net/browse/ARC-2745

### DIFF
--- a/src/routes/api/ghes-app-verification/ghes-app-verify-get-apps.ts
+++ b/src/routes/api/ghes-app-verification/ghes-app-verify-get-apps.ts
@@ -5,6 +5,7 @@ import { Installation } from "models/installation";
 import { runCurl } from "utils/curl/curl-utils";
 import { AppTokenHolder } from "~/src/github/client/app-token-holder";
 
+const FIVE_MINUTES_IN_MILLI_SEC = 5 * 60 * 1000;
 export const GHESVerifyGetApps = async (req: Request, res: Response): Promise<void> => {
 
 	const gitHubAppId = parseInt(req.params["gitHubAppId"] || "");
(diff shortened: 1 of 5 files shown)


bgvozdev (approved):

yvonnevictoria (approved):

src/routes/api/ghes-app-verification/ghes-app-verify-get-apps.ts
```diff
@@ -36,7 +37,7 @@ export const GHESVerifyGetApps = async (req: Request, res: Response): Promise<vo
 			const output = await runCurl({
 				fullUrl: `${gitHubAppClient.restApiUrl}/app`,
 				method: "GET",
-				authorization: `Bearer ${AppTokenHolder.createAppJwt(await app.getDecryptedPrivateKey(installation.jiraHost), String(app.appId)).token}`
+				authorization: `Bearer ${AppTokenHolder.createAppJwt(await app.getDecryptedPrivateKey(installation.jiraHost), String(app.appId), FIVE_MINUTES_IN_MILLI_SEC).token}`
```
krazziekay, line 40: Why are we not feature flagging this too?

krazziekay: I realize that this is for the API endpoints that we use, but we can't test the exp time set in FF from this endpoint , can we? 
So if the customer is still facing this issue, we can't test the value in the FF through this endpoint cause its set to 5 minutes.

gxueatlassian:
I don't like my PR.

PR #2617, opened 2023-12-14
add dlq-service during BB pipeline

gxueatlassian:
**What's in this PR?**
add dlq-service during BB pipeline

**Why**
1. The public yarn.lock doesn't contains the optional deps
2. But during BB pipeline, it will install the optional, but now the fronzenfile is wrong, so will error.

**Added feature flags**

**Affected issues**  

**How has this been tested?**  

https://bitbucket.org/atlassian/github-for-jira-deployment/pipelines/results/8402

image

**Whats Next?**

### DIFF
--- a/.yarnrc
+++ b/.yarnrc
@@ -1,2 +1,0 @@
---install.ignore-optional true
---add.ignore-optional true
--- a/Dockerfile.prod
+++ b/Dockerfile.prod
@@ -18,8 +18,7 @@ COPY . /app
 WORKDIR /app
 
 # Installing packages
-RUN rm .yarnrc
-RUN yarn add @atlassian/sqs-queue-dlq-service --frozen-file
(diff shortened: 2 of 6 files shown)


.yarnrc
gxueatlassian, line 1: No point of having this file now.

Dockerfile.prod
```diff
@@ -18,7 +18,7 @@ COPY . /app
 WORKDIR /app
 
 # Installing packages
-RUN rm .yarnrc
+RUN yarn add @atlassian/sqs-queue-dlq-service
```
gxueatlassian, line 21: The only drawback is we are not `frozen` this dep to a specific version, but I guess that's okay, since it is only used as our queue replay api, won't affect customer (in a disaster way)

bgvozdev (approved):

PR #2618, opened 2023-12-18
ARC-2759 skip the error "Commit not found by sha..." if on last try

gxueatlassian:
**What's in this PR?**
Add code to skip the commit if error is "422 not found by sha" when on the last try. Only apply for push queue.

**Why**
I think on the last re-try, if we still couldn't find the commit by sha from either github.com or ghes, it could means either the server is wrong or the commit is gone. We should skip the message instead of putting them in DLQ.

**Added feature flags**
skip-commit-if-sha-not-found-on-last-try

**Affected issues**  
[ARC-2759]

**How has this been tested?**  
Unit test.

**Whats Next?**

[ARC-2759]: https://softwareteams.atlassian.net/browse/ARC-2759

### DIFF
--- a/src/config/feature-flags.ts
+++ b/src/config/feature-flags.ts
@@ -26,7 +26,8 @@ export enum BooleanFlags {
 	USE_CUSTOM_ROOT_CA_BUNDLE = "use-custom-root-ca-bundle",
 	GENERATE_CORE_HEAP_DUMPS_ON_LOW_MEM = "generate-core-heap-dumps-on-low-mem",
 	USE_RATELIMIT_ON_JIRA_CLIENT = "use-ratelimit-on-jira-client",
-	SKIP_PROCESS_QUEUE_IF_ISSUE_NOT_FOUND = "skip-process-queue-when-issue-not-exists"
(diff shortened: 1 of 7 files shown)


src/config/feature-flags.ts
```diff
@@ -26,7 +26,8 @@ export enum BooleanFlags {
 	USE_CUSTOM_ROOT_CA_BUNDLE = "use-custom-root-ca-bundle",
 	GENERATE_CORE_HEAP_DUMPS_ON_LOW_MEM = "generate-core-heap-dumps-on-low-mem",
 	USE_RATELIMIT_ON_JIRA_CLIENT = "use-ratelimit-on-jira-client",
-	SKIP_PROCESS_QUEUE_IF_ISSUE_NOT_FOUND = "skip-process-queue-when-issue-not-exists"
+	SKIP_PROCESS_QUEUE_IF_ISSUE_NOT_FOUND = "skip-process-queue-when-issue-not-exists",
+	SKIP_COMMIT_IF_SHA_NOT_FOUND_ON_LAST_TRY = "skip-commit-if-sha-not-found-on-last-try"
```
bgvozdev, line 30: Out of curiousity, what "last-try" means here? Could we change it to something like `-on-webhook-push` ? Cause I had impression we are talking about SQS processing...

bgvozdev (approved):

